Issue #10, opened 2020-06-05
Command: check the integration


mtojek:
The command is used by the CI to verify correctness of the integration. Ultimately it can also be used by a developer to make sure that they will push the correct code to the Git repository.

Suggested command: `elastic-package check`

PR #31, opened 2020-07-30
Implement check command

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/10

Note: some commands are placeholders, they will be implemented in another issues.


mtojek:
We had an offline discussion, agreed on the following approach:

* `check` - aggregates multiple commands (format, lint, build)
* `test` - keep it aside
* `validate` becomes `lint`

cmd/check.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func setupCheckCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "check",
+		Short: "Check the integration",
+		Long:  "Use check command to verify if the integration is correct in terms of formatting, validation, building, tests.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := composeCommandActions(cmd, args,
+				formatCommandAction,
```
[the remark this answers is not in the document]
exekias: It would be good to have a parameter in this one to fail if there are changes. We have normally done this using `git diff-index`

mtojek:
I think Github is facing an outage right now. Missing commits (already pushed) here: https://github.com/mtojek/elastic-package/commits/10-check-integration

cmd/check.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func setupCheckCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "check",
+		Short: "Check the integration",
```
masci, line 11: Sorry for hijacking this PR but I think we could make a more consistent use of the terms `package` and `integration`. 

I understand that at this point in time `integration`s represent the only available `type` of package but this will hopefully change in the future.

So we should decide whether this tool is somehow dedicated to the development of specific packages of `type: integration`, and in this case I would change the name of the tool and maybe move the source code into the `elastic/integrations` repo; or we say this is a general purpose tool to support the development of packages, and in this case I'd consolidate the naming scheme around source code, docs, etc. avoiding to refer to `integration`s when the command can actually work with any package type.

mtojek: Hey @masci, thanks for looking into this! I believe we have a discussion about this in the past.

We decided to move on with packages, but as long as there are no other types even defined (or in plans), I sticked to integrations. I can try to get rid of words referring to "integrations" in as many places as possible, but internally there will be support only for integrations.

I prefer to keep this in a separate repository as the original plan is to keep integration source in the integrations repo and nothing else (no logic, tooling, side scripting).

I'd like to listen to other opinions on this. @exekias @ruflin @ycombinator

exekias: I'm +1 on having this tool as the reference for package development, where integrations will be only a subset of that. In that sense I would refrain from mentioning "integrations" here.

In that sense I think it also makes sense to have the tool in a different repo, not only because of this "generic" soul, but also because I expect that once things get stable, `elastic-package` users won't necessarily be go developers (with the go toolchain installed for instance). This project should have its own lifecycle.

mtojek: If you don't mind I would go over this in another PR. It's not directly in the scope of this PR.

mtojek:
Resolving in favor of https://github.com/elastic/elastic-package/pull/34

cmd/check.go
mtojek: https://github.com/elastic/elastic-package/issues/35

Issue #35, opened 2020-07-30
Adjust wording to "package" rathen than "integration"


mtojek:
The tool should be oriented on package development, not only integrations. @masci suggested to adjust all references to integrations and replaces with packages (where possible).

ycombinator:
IIRC the original intent was for this tool to be just for integrations packages. Just want to clarify that going forward we want it to be used for _any_ packages (e.g. endpoint as well), yes?

mtojek:
I understand this :

The tool is intended to be used for development of different packages, but currently we support "integrations" only.

ycombinator:
Gotcha, limiting the scope that way for now sounds good. We should make sure none of the interfaces (commands, arguments to commands, etc.) are integrations specific so it's easy to grow the scope in the future. I think this is already the case but probably good to keep checking anytime we introduce a new interface that we're not backing ourselves into any corners that'd be hard to get out of later when we broaden the scope.

masci:
Thanks for filing this issue!

@ycombinator  I think it's ok if the tool has a few integrations-specific subcommands (with relative flags and args) if needed, as integrations are indeed packaged. My intent was to consolidate the message we send through docs, source code and CLI about the scope of the tool, which might be (as @mtojek perfectly condensed in the comment above) _The tool is intended to be used for development of packages_.

PR #36, opened 2020-07-30
Refer to packages instead of integrations

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/35

README preview: https://github.com/mtojek/elastic-package/blob/35-adjust-wording/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The elastic-package is a command line tool, written in Go, used for developing E
 
 TODO
 
-## Supported packages
+## Supported package types
 
 * [Elastic Integrations](https://github.com/elastic/integrations)
 
--- a/README.md
+++ b/README.md
@@ -1,10 +1,16 @@
 # elastic-package
 
-The elastic-package is a command line tool, written in Go, used for developing Elastic Integrations.
+The elastic-package is a command line tool, written in Go, used for developing Elastic packages.
+
+*For experimental use only*
 
 ## Features
 
 TODO
+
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1,11 +1,17 @@
...
-The elastic-package is a command line tool, written in Go, used for developing Elastic Integrations.
+The elastic-package is a command line tool, written in Go, used for developing Elastic packages.
+
+*For experimental use only*
 
 ## Features
 
 TODO
 
+## Supported packages
```
masci, line 11: Let's do "package types" so the `type` key will sound more familiar if an user peeks into a manifest

cmd/testrunner.go
```diff
@@ -9,8 +9,8 @@ import (
 func setupTestCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "test",
-		Short: "Run test suite for the integration",
-		Long:  "Use test runners to verify if the integration collect logs are metrics properly.",
+		Short: "Run test suite for the package",
+		Long:  "Use test runners to verify if the package collects logs are metrics properly.",
```
masci, line 13: typo: s/`logs are metrics`/`logs and metrics`

README.md
mtojek: Agree.

Changed to "Supported package types".

cmd/testrunner.go
mtojek: Fixed.

PR #38, opened 2020-07-31
Add built binary to .gitignore

ruflin:
I assume this binary should never be checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@
 # Test binary, built with `go test -c`
 *.test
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -8,6 +8,9 @@
 # Test binary, built with `go test -c`
 *.test
 
+# elastic-package binary built with go build
```
mtojek, line 11: If you use "go build", then most likely you can produce the binary in the repo folder.

PR #39, opened 2020-07-31
Fix: create build dir if it's not present

mtojek:
This PR adjusts the elastic-package to create build directory for integrations if it's missing.


mtojek:
I need this fix, I will merge it.

PR #42, opened 2020-08-06
Implement lint subcommand

### LINKED ISSUE #5: Command: lint the integration
The command runs a basic validation of the package content, it doesn’t necessarily need to build the content. YAML manifests, JSON files, mandatory files, etc.  It also verifies if dashboard file are properly encoded.

The command exposes a flag that fails when any file modification is required (for "check" purposes). It must be part of the implementation.

Suggested command: `elastic-package lint`

ycombinator:
This PR implements the `elastic-package lint` command.

Depends on https://github.com/elastic/package-spec/pull/12.
Resolves #5.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -1,10 +1,11 @@
 package cmd
 
 import (
-	"os"
-
 	"github.com/elastic/package-spec/code/go/pkg/validator"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
+	"github.com/elastic/elastic-package/internal/packages"
 )
 
 func setupLintCommand() *cobra.Command {
@@ -18,9 +19,12 @@ func setupLintCommand() *cobra.Command {
 }
 
 func lintCommandAction(cmd *cobra.Command, args []string) error {
-	packageRootPath, err := os.Getwd()
+	packageRootPath, found, err := packages.FindPackageRoot()
+	if !found {
+		return errors.New("package root not found")
+	}
 	if err != nil {
-		return err
+		return errors.Wrap(err, "locating package root failed")
 	}
 
 	return validator.ValidateFromPath(packageRootPath)
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -10,8 +10,8 @@ import (
 func setupLintCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "lint",
-		Short: "Lint the integration",
-		Long:  "Use lint command to lint the integration files.",
+		Short: "Lint the package",
+		Long:  "Use lint command to lint the package files.",
 		RunE:  lintCommandAction,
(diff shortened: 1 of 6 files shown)


cmd/lint.go
```diff
@@ -1,9 +1,9 @@
 package cmd
 
 import (
-	"fmt"
-
+	"github.com/elastic/package-spec/code/go/pkg/validator"
 	"github.com/spf13/cobra"
+	"os"
```
mtojek, line 6: nit: please sort imports

cmd/lint.go
```diff
@@ -12,11 +12,16 @@ func setupLintCommand() *cobra.Command {
 		Short: "Lint the integration",
 		Long:  "Use lint command to lint the integration files.",
 		RunE:  lintCommandAction,
+		SilenceUsage: true,
```
mtojek, line 15: I wonder if there is global switch to fix these cases. Otherwise we need to mute usage for every subcommand.

cmd/lint.go
```diff
@@ -12,11 +12,16 @@ func setupLintCommand() *cobra.Command {
...
 	}
 	return cmd
 }
 
 func lintCommandAction(cmd *cobra.Command, args []string) error {
-	fmt.Println("lint is not implemented yet.")
-	return nil
+	packageRootPath, err := os.Getwd()
```
mtojek, line 21: There is a method that will help you: https://github.com/elastic/elastic-package/blob/master/internal/builder/packages.go#L62 You can expose it or even extract to `internal/packages`.

mtojek:
There was an issue spotted by CI:

```
[2020-08-06T21:45:00.276Z] go: github.com/elastic/package-spec/code/go@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

cmd/lint.go
ycombinator: It might be possible to set SilenceUsage at the root command level, but I'm 100% sure based on this thread: https://github.com/spf13/cobra/issues/340. I will try it out and update this PR accordingly.

ycombinator:
> There was an issue spotted by CI...

This is expected to happen until https://github.com/elastic/package-spec/pull/12 gets merged.

cmd/lint.go
ycombinator: Fixed in 5ffd82d.

cmd/lint.go
ycombinator: Fixed in 008f7e4.

cmd/lint.go
ycombinator: Done in 75eb0d2.

ycombinator:
@mtojek I think I've addressed all your feedback on this PR. It's ready for your 👀 again, when you get a chance please. Thanks!

PR #43, opened 2020-08-11
Implement "Promote" subcommand

mtojek:
Fixes: https://github.com/elastic/package-storage/issues/123

This PR implements "promote" subcommand.

Please play on your own :) 

`elastic-package promote`

Sample PRs created:
https://github.com/elastic/package-storage/pull/341
https://github.com/elastic/package-storage/pull/342
(equal number of files 👍 )


internal/github/auth.go
```diff
@@ -0,0 +1,46 @@
...
+			envAuth, authTokenFile)
+	}
+	return nil
+}
+
+// AuthToken method finds and returns the GitHub authorization token.
+func AuthToken() (string, error) {
+	githubTokenVar := os.Getenv(envAuth)
+	if githubTokenVar != "" {
+		fmt.Println("Using GitHub token from environment variable.")
```
ycombinator, line 31: Not for this PR but at some point later we should consider introducing proper logging into this tool.

internal/promote/storage.go
```diff
@@ -0,0 +1,410 @@
...
+		},
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "creating remote failed")
+	}
+	return r, nil
+}
+
+// ListPackages method lists available packages in the package-storage.
+// It skips technical packages (snapshot, staging).
```
ycombinator, line 99: What do you mean by "technical" packages?

internal/promote/pull_requests.go
```diff
@@ -0,0 +1,176 @@
...
+
+func buildPullRequestRemoveDescription(sourceStage string, revisions PackageRevisions) string {
+	var builder strings.Builder
+	builder.WriteString(fmt.Sprintf("This PR removes packages from `%s`.\n", sourceStage))
+	builder.WriteString("\n")
+	builder.WriteString("Removed packages:\n")
+	for _, revision := range revisions {
+		builder.WriteString(fmt.Sprintf("* `%s-%s`\n", revision.Name, revision.Version))
+	}
+	return builder.String()
```
ycombinator, line 136: Would be nice to include the promotion PR that is associated with this removal PR, especially since the removal PR should probably not be merged before the promotion PR is merged?

internal/promote/pull_requests.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return nil
+}
+
+func buildPullRequestRemoveDescription(sourceStage string, revisions PackageRevisions) string {
+	var builder strings.Builder
+	builder.WriteString(fmt.Sprintf("This PR removes packages from `%s`.\n", sourceStage))
+	builder.WriteString("\n")
+	builder.WriteString("Removed packages:\n")
+	for _, revision := range revisions {
```
ycombinator, line 133: Nit: I think you could loop over `revisions.Strings()` here. Then you only need to prefix the string with `* ` inside the loop.

internal/promote/pull_requests.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return builder.String()
+}
+
+func buildPullRequestPromoteDescription(sourceStage, destinationStage string, revisions PackageRevisions) string {
+	var builder strings.Builder
+	builder.WriteString(fmt.Sprintf("This PR promotes packages from `%s` to `%s`.\n", sourceStage, destinationStage))
+	builder.WriteString("\n")
+	builder.WriteString("Promoted packages:\n")
+	for _, revision := range revisions {
```
ycombinator, line 144: Nit: I think you could loop over `revisions.Strings()` here. Then you only need to prefix the string with `* ` inside the loop.

internal/promote/storage.go
```diff
@@ -0,0 +1,410 @@
...
+// User method returns the Git user.
+func User(r *git.Repository) (string, error) {
+	c, err := r.ConfigScoped(config.SystemScope)
+	if err != nil {
+		return "", errors.Wrap(err, "reading config failed")
+	}
+	return c.User.Name, nil
+}
+
+func sortPackageRevisions(revisions PackageRevisions) PackageRevisions {
```
ycombinator, line 402: Might be personal preference but when I see a function whose name contains a type in it and also takes an argument of the same type, it feels like the function should really be a method with the type as the receiver. So something like:

```go
func (p PackageRevisions) sort() PackageRevisions {
   ...
}
```

internal/github/auth.go
ruflin: ECS logging lib ;-)

ruflin:
@mtojek Will this also promote packages like `snapshot` and `staging`?

mtojek:
> @mtojek Will this also promote packages like `snapshot` and `staging`?

I assume it's shouldn't touch them as they are technical packages.

internal/github/auth.go
mtojek: https://github.com/elastic/elastic-package/issues/45

internal/promote/storage.go
mtojek: It's a bit domain driven design. Packages like `snapshot` and `staging` don't below to the domain, but are strictly technical (we use them to highlight the currently used stage). 

If it's not straightforward I can reword it to something more clear.

mtojek:
> Might be overkill for now but if we start adding more GitHub API interactions in the `elastic-package` tool it might be worth replacing the "bare" HTTP calls and boilerplate with a GitHub library like https://github.com/google/go-github.

I checked that it's BSD, so it's allowed. I'll try to include it here.

internal/promote/pull_requests.go
mtojek: Fixed

internal/promote/pull_requests.go
mtojek: Fixed

internal/promote/pull_requests.go
mtojek: Fixed

internal/promote/storage.go
mtojek: I like this idea, it's better than simple producedure. Fixed.

mtojek:
> I checked that it's BSD, so it's allowed. I'll try to include it here.

Done

internal/promote/storage.go
```diff
@@ -0,0 +1,410 @@
...
+				Version: versionDir.Name(),
+				semver:  *packageVersion,
+			})
+		}
+	}
+	return revisions.sort(), nil
+}
+
+// FilterPackages method filters package revisions based on the "newest revision only" policy.
+func FilterPackages(allPackages PackageRevisions, newestOnly bool) PackageRevisions {
```
ycombinator, line 157: This function looks like another candidate to become a method with a `PackageRevisions` receiver.

ycombinator:
When I run `elastic-package promote`, the second question that comes up is:
> ? Would you like to promote newest revisions only and remove older ones? (Y/n)

Maybe I'm misunderstanding the process but this question is actually asking about two independent choices, no — whether to promote newest revisions only and whether to remove old packages? In other words, isn't it possible to say yes to one of those choices but no to the other?

Also, related: what does it mean _not_ to promote newest revisions only? My understanding from the code is that it would copy (promote) all packages from the source branch to the destination branch at that point. But what is the use case for saying no to this question?

internal/promote/storage.go
mtojek: Fixed

ycombinator:
> ? Which packages would you like to promote  [Use arrows to move, space to select, <right> to all, <left> to none, type to filter]

UX: It takes a few seconds (at least for me) for the list of packages to load up. I imagine this is because network I/O is involved in order to generate the list of packages. Might be good to show some sort of message like "creating list of packages..." or something before the network I/O is started so the user knows that the waiting is normal and not some sort of error state.

ycombinator:
At the very last step of running this tool I got this error:

```
Error: pushing changes failed: pushing branch failed: unexpected client error: unexpected requesting "https://github.com/Shaunak%20Kashyap/package-storage/info/refs?service=git-receive-pack" status code: 400
```

Here is the complete interaction:

```
? Which promotion would you like to run snapshot - staging
? Would you like to promote newest revisions only and remove older ones? Yes
X Sorry, your reply was invalid: Value is required
? Which packages would you like to promote checkpoint-0.1.0
Promote packages from snapshot to staging...
Remove packages from snapshot...
Error: pushing changes failed: pushing branch failed: unexpected client error: unexpected requesting "https://github.com/Shaunak%20Kashyap/package-storage/info/refs?service=git-receive-pack" status code: 400
```
I assume it has something to do with the `Shaunak%20Kashyap` there in the URL instead of `ycombinator`. How would one go about fixing this issue?

internal/promote/storage.go
ycombinator: Yeah, "technical" is still unclear to me personally but that might be just me. Let's get another opinion on it. @ruflin WDYT?

mtojek:
> Maybe I'm misunderstanding the process but this question is actually asking about two independent choices, no — whether to promote newest revisions only and whether to remove old packages? In other words, isn't it possible to say yes to one of those choices but no to the other?
> 
> Also, related: what does it mean _not_ to promote newest revisions only? My understanding from the code is that it would copy (promote) all packages from the source branch to the destination branch at that point. But what is the use case for saying no to this question?

Consider the current state in the package-storage (snapshot -> staging) and the endpoint package. The endpoint package is available in two version 0.14 and 0.15
You can either:
1. Promote newest version only, so  endpoint-0.15 and remove both 0.14 and 0.15
2. Promote the one you'd like, e.g. only endpoint-0.14

It might be bad wording, but this is actually our use case.

mtojek:
> I assume it has something to do with the Shaunak%20Kashyap there in the URL instead of ycombinator. How would one go about fixing this issue?

Hm.. this is a matter of Git config (System scope). In my case this is `mtojek`. The question is: which property should be used...

internal/promote/storage.go
mtojek: I can rephrase it:

`It skips packages: snapshot, staging`.

mtojek: Ok, I changed it to the above form.

Issue #15, opened 2020-06-05
Feature: run pipeline tests


mtojek:
Master issue: https://github.com/elastic/elastic-package/issues/14

Pipeline tests are quick, small tests that can verify the ingest pipeline without a need for booting up the entire cluster (only the Elasticsearch instance).

Input:
`test-a-events.json`: defines a set of JSON events, which could be produced by the input. These input data are sent to the pipeline.
`test-a-expected.json`: contains a document expected to be produced by the tested pipeline

Execution:

The test runner (elastic-package) installs the ingest pipeline in Elasticsearch, sends mocked events to the pipeline and compares stored documents with the expected output.

ycombinator:
@mtojek I started working on a PR for this but didn't get too far. Please go ahead and work on it if you have time! I've removed my assignment.

mtojek:
I'm going with this now.

PR #44, opened 2020-08-11
Implement system test command stub

### LINKED ISSUE #16: Feature: run system tests
Master issue: https://github.com/elastic/elastic-package/issues/14

System tests are end-to-end tests that involve multiple Elastic products and the real service (under tests) or collected log data (which could be mocked). They support multiple versions of the tested service (e.g. selected with environment variables).
The current design assumes that system tests only consist of configuration files and there are no test cases in Go language.

Input:
`test-a-agent.yml`: (possibly optional?) configuration file used to spawn a standalone instance of the agent
`Dockerfile`: Dockerfile definition that defines an instance of the tested service (might be single for entire integration)
`docker-compose.yml`: (optional) the manifest defines other properties of the tested service, e.g. scaling factor, volumes, exposed ports (might be single for entire integration)
`supported-variants.yml`: contains environment variables defining multiple versions of the product (might be single for entire integration)
`test-a-input-file.log`: (optional) other resources referenced in the configuration, e.g. log files to be harvested.
`test-a-expected.json`: contains a document expected to be present in Elasticsearch. The document format allows for defining basic matchers to verify non-exact values.

Execution:

The test runner (elastic-package) performs full end-to-end testing. It performs the following actions:
1. Start the next variant of the tested service.
2. Tear up the testing cluster (Elasticsearch, Kibana, Package Registry).
3. Install the package using the Kibana API.
4. Start the agent using the provided configuration file.
5. Wait for logs/metrics and compare results with the document in the database.
6. Verify that all fields in the document are documented.
7. Tear down the testing cluster.

ycombinator:
This PR implements the following command:
```
elastic-package test system [flags]
```

Where the optional `flags` are:
- `--fail-on-missing/-m`: Fail if system tests are missing. Default: ignore missing system tests.
- `--dataset <dataset>`: Only run system tests for specified `<dataset>`s (comma-separated). Default: run system tests for all datasets.

Running this command from anywhere inside a package's folder will run the system tests for that package.

**NOTE:**  at the moment this command is implemented as a stub. It does not _actually_ perform system tests. That functionality — the actual system test runner — will be implemented in follow up PR(s).
  
Resolves partially #16.


cmd/testrunner.go
```diff
@@ -1,21 +1,60 @@
...
+		RunE:  testSystemCommandAction,
+	}
+
+	cmd := &cobra.Command{
 		Use:   "test",
 		Short: "Run test suite for the package",
 		Long:  "Use test runners to verify if the package collects logs and metrics properly.",
 		RunE:  testCommandAction,
 	}
+	cmd.PersistentFlags().BoolP("ignoreMissing", "i", false, "ignore missing tests")
```
mtojek, line 25: Hm... I wonder if the command should ignore the fact that there are no tests by default. Similarly to:

```bash
➜  elastic-package git:(promote-subcommand) go test
?   	github.com/elastic/elastic-package	[no test files]
➜  elastic-package git:(promote-subcommand) echo $?
0
```

internal/testrunner/system/system.go
```diff
@@ -0,0 +1,44 @@
...
+
+	if err := runner.run(); err != nil {
+		return errors.Wrap(err, "system tests failed")
+	}
+
+	return nil
+}
+
+func findSystemTestsPath(packageRootPath string) (string, error) {
+	systemTestsPath := path.Join(packageRootPath, "_dev", "test", "system")
```
mtojek, line 30: I'm not quite sure why is there `_dev` folder here. Is it temporarily for development purposes?

internal/testrunner/system/system.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"os"
+	"path"
+
+	"github.com/pkg/errors"
+)
+
+// Run runs the system tests for the given package.
+func Run(packageRootPath string) error {
+	systemTestsPath, err := findSystemTestsPath(packageRootPath)
```
mtojek, line 12: Just to confirm - there is one single directory for system test across entire package, not per every dataset?

EDIT:

I looked into the design paper and I think it mentions `test` directories in datasets instead. The reason why is it like that, is because you can enable different datasets in Elastic Agent and don't collect same logs in all test cases.

mtojek, line 12: I'm thinking about the best place for filtering tests:

`elastic-package test --dataset ingest_controller`

The `--dataset` parameter filters test files based on the dataset criteria, which may be valid for both kinds of tests (pipeline, system). Do you think we can put this in some common place, before executing runners?

cmd/testrunner.go
```diff
@@ -1,21 +1,60 @@
...
+	}
+
+	if err != nil {
+		return errors.Wrap(err, "error running package system tests")
+	}
+
 	return nil
 }
+
+func testCommandAction(cmd *cobra.Command, args []string) error {
```
mtojek, line 57: Follow up on another other comment:

This could be a good place to run test filtering (selected dataset) and then pass this directory to composed runners (which also can be selected with command args).

mtojek:
I think you introduced some common logic valid also for pipeline runner. To prevent potential merge conflicts I kindly suggest to merge this PR after basic polishing, so I can focus on https://github.com/elastic/elastic-package/issues/15

cmd/testrunner.go
ycombinator: Yeah, I went back and forth on this. I'm okay ignoring missing tests as the default behavior. To reflect this better, should I change the flag to be something like `--failOnMissingTests/-m`, defaulting to `false`?

internal/testrunner/system/system.go
ycombinator: My goal with the `_dev` folder is to create a space in the package structure where "development time" files can be kept, e.g. test-related files. This folder would not be included in the package build (i.e. the result of `elastic-package build`) and therefore would never make it into the package registry, etc.

In some ways the `_dev` folder is similar to the `_meta` folders we see in Beats, except I wanted to make it quite explicit via the name that "this space was for development only purposes". That said, I'm open to other names or ways of accomplishing the same goal.

I'm experimenting with this idea for now (hence the draft status of this PR) but if it turns out to be useful I plan to create  an issue in the elastic-spec repo to discuss the proposal of such a folder.

internal/testrunner/system/system.go
ycombinator: Yes, you're right. It should be per dataset, not per package. Will fix, thanks for pointing this out!

ycombinator: I like this idea. Just to confirm, if no `--dataset` is specified we iterate over all datasets, yes?

> Do you think we can put this in some common place, before executing runners?

Yup, by specifying this flag as a persistent flag on the root `test` command it's value will be available to all sub-commands of `test`: `test system`, `test pipeline`, etc.

ycombinator:
> I think you introduced some common logic valid also for pipeline runner. To prevent potential merge conflicts I kindly suggest to merge this PR after basic polishing, so I can focus on #15

Do you mean I should pull out the common logic into a separate PR that we merge before this one?

cmd/testrunner.go
```diff
@@ -1,21 +1,60 @@
...
+		return errors.Wrap(err, "locating package root failed")
+	}
+
+	ignoreMissing, err := cmd.Flags().GetBool("ignoreMissing")
+	if err != nil {
+		return errors.Wrap(err, "error parsing ignoreMissing flag")
+	}
+
+	err = system.Run(packageRootPath)
+	if err != nil && ignoreMissing && err == system.ErrNoSystemTests {
```
ycombinator, line 46: On further thought I think the the `test*Action` functions here should try to be ignorant of any testing-specific flags such as `--ignoreMissing`. I think they should parse such flags' values into variables and then pass the variables to the test runner for further interpretation. So I'm thinking of moving this `if` condition here inside the  implementation of `system.Run`. 

LMK if you disagree @mtojek.

internal/testrunner/system/system.go
ycombinator: Couple more questions about the `--dataset` flag:
- should we accept multiple occurrences of `--dataset` as a way for a user to specify multiple datasets?
- should we make it behave like the `-run` flag of `go test` in that it can take a regular expression and execute tests for matching datasets?

I'm also okay deferring one or both of the above features to follow up PRs. WDYT?

cmd/testrunner.go
mtojek: Works for me!

It might be personal preference, but maybe let's stick to dashes: `--fail-on-missing`.

internal/testrunner/system/system.go
mtojek: > My goal with the _dev folder is to create a space in the package structure where "development time" files can be kept, e.g. test-related files. This folder would not be included in the package build (i.e. the result of elastic-package build) and therefore would never make it into the package registry, etc.

I think I already pushed some nginx test files to the package-storage. I assume we'll to clean them up.

I like your idea, although I thought about simply skipping such files/dirs (test/, Dockerfile, etc.), but there can be plenty of them. 

Do you think such directory should be also on the dataset level?

>I'm experimenting with this idea for now (hence the draft status of this PR) but if it turns out to be useful I plan to create an issue in the elastic-spec repo to discuss the proposal of such a folder.

Yes, please do.

internal/testrunner/system/system.go
mtojek: > should we accept multiple occurrences of --dataset as a way for a user to specify multiple datasets?

I think the proposal defines multiple values for single flag: `--dataset=a,b,c`.

> should we make --dataset behave like the -run flag of go test in that it can take a regular expression and execute tests for matching datasets?

I admit I didn't think about it.

For flexibility we can add such feature, but I'm not sure about potential use cases. Usually dataset names differ much from each other. I don't reckon any names that have a common prefix. Another use case - run tests for all datasets except X.

Maybe just delay such feature until we have a real requirement.

> I'm also okay deferring one or both of the above features to follow up PRs. WDYT?

Absolutely right, it's not a first need feature.

mtojek:
> Do you mean I should pull out the common logic into a separate PR that we merge before this one?

If you're close to finalize works on the PR, I can wait for it ;) I had plans to start working on the pipeline tests and would be nice if I can fetch your changes in `cmd/testrunner.go`.

But as I said, don't need to hurry up. I have stuff to look into ;)

cmd/testrunner.go
mtojek: Let's consider the use case:

There is an integration that has system tests, but doesn't have pipeline tests. Pipeline test runner shouldn't fail the `elastic-package test` command.

That's why I was thinking about detecting test directories (package scope, dataset scope) in the `cmd/testrunner.go` and for each of them run all appropriate runners.

Sample flow:

Find test directories:
1. Found `nginx/test`.
2. Found `nginx/access/test`.
3. Found `nginx/error/test`.

Use pipeline test runner:
4. Run pipeline tests for `nginx/access/test`.
5. Run pipeline tests for `nginx/error/test`.

Use system test runner:
6. Run system tests for `nginx/test`.
7. Run system tests for `nginx/access/test`.
8. Run system tests for `nginx/error/test`.

We can sync up/discuss this later on if you prefer.

cmd/testrunner.go
ycombinator: Hmmm, I prefer the kebab case too but I'm not sure this is supported. Will try though. Are you also suggesting we remove the short option (`-m`) or just that you prefer the kebab case over the camel case for the long option?

mtojek: Sorry, I missed it. The short option can stay there.

ycombinator: Changed `--failOnMissing` to `--fail-on-missing` in 72a3ded.

ycombinator:
@mtojek I decided to

internal/testrunner/system/system.go
ycombinator: Made a proposal issue for the `_dev` folder in the package-spec repo for discussion: https://github.com/elastic/package-spec/issues/22.

internal/testrunner/system/system.go
ycombinator: Fixed in 8275c27e0a25bb9e55cace984657391fa74fa272.

ycombinator: Implemented `--dataset` as an optional flag, allowing it to take a comma-separated values, in 165e30b.

cmd/testrunner.go
ycombinator: I've implemented this now.

ycombinator:
@mtojek I decided to scope this PR to just the command interface and a bit of logic around test folder selection. That way you can use the code in this PR to make progress on the pipeline testing feature. I will implement the actual system test runner in a follow up PR. This also keeps the size of this PR here smaller and hopefully makes it easier to review. Please re-review when you get a chance. Thanks!

cmd/check.go
```diff
@@ -11,7 +13,7 @@ func setupCheckCommand() *cobra.Command {
 		Short: "Check the package",
 		Long:  "Use check command to verify if the package is correct in terms of formatting, validation and building.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := composeCommandActions(cmd, args,
+			err := internalCmd.ComposeCommandActions(cmd, args,
```
mtojek, line 16: nit: is it possible not to expose this method? or maybe put it in the `cmd/utils`?

ycombinator: Sure, originally I was thinking of putting it in `cmd/utils.go` but I thought `cmd` was reserved (by convention) for the external API. That's why I put it under `internal/cmd/utils.go` instead.

mtojek: Hmm... I don't think it can shared with some other cmd packages. I think we don't need to artificially extract it to a separate package. It should be fine just not exposed in cmd.

ycombinator: Okay, relocated in 7dba5c5. Please take a look to see if this is what you were expecting. Thanks!

Issue #24, opened 2020-07-28
Command: version


ycombinator:
The command would output the version of `elastic-package` currently installed. 

As `elastic-package` is rapidly under development it would be good to know which version is being run for debugging issues from users, etc.

exekias:
This would be nice, I'm wondering what will be the release cycle here? It sounds like for now we may want to stay up to date will any change to master? In that case the version would be the commit hash 😈 

I created this related issue: https://github.com/elastic/elastic-package/issues/32

PR #46, opened 2020-08-12
Show the application version

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/24


cmd/version.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	// BuildTime is the build time of the binary (set externally with ldflags).
+	BuildTime = "unknown"
+
+	// CommitHash is the Git hash of the branch, used for version purposes (set externally with ldflags).
```
ycombinator, line 11: Thank you for these comments, especially the "set externally..." part. I remember it took me some time to figure out how this was working in Beats 🙂.

Makefile
```diff
@@ -1,5 +1,6 @@
 build:
-	go get github.com/elastic/elastic-package
+	go get -ldflags "-X github.com/elastic/elastic-package/cmd.CommitHash=`git describe --always --long --dirty` -X github.com/elastic/elastic-package/cmd.BuildTime=`date +%FT%T%z`" \
```
ycombinator, line 2: This is a nitpick so feel free to ignore.

The one issue I see with using `date +%FT%T%z` is that it generates the date in the timezone of the system where the build is made. So, for example, in my case that's `PDT` right now, corresponding to an offset of `-0700`. This is not a problem for accuracy because of the `%z` at the end of the format string — the offset gets included as well.

However, it might be a bit inconvenient if the user of `elastic-package version` was not in the same timezone as where it was built. WDYT about instead using `date +%s` here to store the number of seconds since the Epoch, UTC. Then in `versionCommandAction` you can convert this to the user's local timezone and display that for the user's convenience.

mtojek: This is good idea :) Fixed.

PR #50, opened 2020-08-13
Use 7.9 as default environment

ruflin:
As for the snapshot registry currently 7.9 is used as default environment, I think this should also be the default environment spinned up here. Long term it would be nice if the version could be passed in as param.

### DIFF
--- a/internal/install/static_snapshot_yml.go
+++ b/internal/install/static_snapshot_yml.go
@@ -3,7 +3,7 @@ package install
 const snapshotYml = `version: '2.3'
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.9.0-SNAPSHOT
     healthcheck:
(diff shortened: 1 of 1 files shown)


ruflin:
@mtojek This did not have the effect I hoped. When I run ` go run . cluster up` it still boots up a 8.0 cluster so I assume I also need to change things somewhere else too.

mtojek:
Check what is inside of the `~/.elastic-package`. Unfortunately this is required until I implement a version check for this.

BTW this is very unusual way of running the application :) Use `make build` to build the app and then simply `elastic-package`.

ruflin:
Cleaning the .elastic-package dir works. Perhaps we need a `elastic-package clean` command? :-)

I run all the go apps I'm developing in this way as like this I never have to deal with outdated binaries. It "should" not make a difference.

mtojek:
We can say it's a bug. I created an issue for it: https://github.com/elastic/elastic-package/issues/52 Not sure about "clean", I would rather focus on keep the application stateless. Apart from the issue you faced, we're following this rule.

PR #51, opened 2020-08-13
Add "cluster pull" command

ruflin:
Often it is useful to get the most recent version of the images. For this the most recent images must be pulled. I initially added it to the bootUp command directly but realised this could become annoying as sometimes starting a cluster could take much longer. Instead I made it a manual command.

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -32,6 +32,18 @@ func setupClusterCommand() *cobra.Command {
 		},
 	}
 
+	pullCommand := &cobra.Command{
+		Use:   "pull",
+		Short: "Pulls down the most recent version of the images",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := cluster.DockerComposePull()
+			if err != nil {
+				return errors.Wrap(err, "pulling down the images failed")
+			}
+			return nil
+		},
+	}
+
 	shellInitCommand := &cobra.Command{
 		Use:   "shellinit",
 		Short: "Initiate environment variables",
@@ -53,6 +65,7 @@ func setupClusterCommand() *cobra.Command {
 	cmd.AddCommand(
 		upCommand,
 		downCommand,
+		pullCommand,
 		shellInitCommand)
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


cmd/cluster.go
```diff
@@ -32,6 +32,18 @@ func setupClusterCommand() *cobra.Command {
 		},
 	}
 
+	pullCommand := &cobra.Command{
+		Use:   "pull",
+		Short: "Pulls down the most recent version of the images",
```
mtojek, line 37: nit: the versions

cmd/cluster.go
```diff
@@ -32,6 +32,18 @@ func setupClusterCommand() *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := cluster.DockerComposePull()
```
mtojek, line 39: I prefer not to leak the Docker internals from the package (package). How about `cluster.Update()`?

internal/cluster/boot.go
```diff
@@ -84,6 +84,27 @@ func dockerComposeBuild() error {
 	return nil
 }
 
+// DockerComposePull pulls down the most recent versions of the Docker images
+func DockerComposePull() error {
```
mtojek, line 88: This method could be private and be called by `cluster.Update()`. Consistently with `TearDown()`.

ruflin:
@mtojek I adjusted it to use `update`

internal/cluster/boot.go
```diff
@@ -64,6 +64,11 @@ func TearDown() error {
 	return nil
 }
 
+// Update pulls down the most recent versions of the Docker images
+func Update() error {
+	return dockerComposePull()
```
mtojek, line 69: Could you wrap the error here? "updating docker images failed" or sth similar (see `TearDown`).

ruflin: It seems like duplication but can do.

mtojek: not duplication, but you're encapsulating errors and forbidding leaking internals up.

Issue #33, opened 2020-07-30
format, lint: fail fast, but don't modify integration source


mtojek:
The goal of this issue is to modify commands (format, lint) to fail fast when a specific flag is passed (e.g. `--fail-on-modification`). "Check" command will set and pass this flag to subcommands to fail in case of any inconsistency in the integration (e.g. not formatted source code).

mtojek:
@ycombinator I assume that `lint` doesn't need this flag at all? This is mainly to prevent formatting on the CI side?

ycombinator:
My understanding of this flag is something like this: "if modifications are detected, fail immediately". If so, then I agree — it doesn't make sense for `lint` to accept this flag.

Issue #4, opened 2020-06-05
Command: format the integration


mtojek:
The command format the structure for an existing integration. It's for cleaning up potential whitespaces, etc.

The command exposes a flag that fails when any file modification is required (for "check" purposes). It must be part of the implementation.

Suggested command: `elastic-package format`

mtojek:
I'm going with the MVP (formatting JSON and YAML files).

PR #53, opened 2020-08-13
Implement "format" subcommand

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/4
Fixes: https://github.com/elastic/elastic-package/issues/33

This PR implements "format" subcommand (JSON and YAML files).

As the follow-up I will reformat all integrations.


cmd/check.go
```diff
@@ -22,6 +22,7 @@ func setupCheckCommand() *cobra.Command {
 			return nil
 		},
 	}
+	cmd.PersistentFlags().BoolP(failFastFlagName, "f", true, "fail if any file requires updates")
```
ycombinator, line 25: Nit: s/fail if any/fail immediately if any/?

ycombinator, line 25: Nit: to reflect the "fast" part of fail fast:

```suggestion
	cmd.PersistentFlags().BoolP(failFastFlagName, "f", true, "fail immediately if any file requires updates")
```

cmd/format.go
```diff
@@ -1,22 +1,43 @@
...
-	"fmt"
-
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
+	"github.com/elastic/elastic-package/internal/formatter"
+	"github.com/elastic/elastic-package/internal/packages"
 )
 
+const failFastFlagName = "fail-fast"
```
ycombinator, line 11: Not for this PR but in https://github.com/elastic/elastic-package/pull/44 I introduced an `internal/cmd` package for various common utilities/consts/etc. we need across multiple command implementations. We could consider moving this const there later.

cmd/format.go
```diff
@@ -1,22 +1,43 @@
...
+
 func setupFormatCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "format",
 		Short: "Format the package",
 		Long:  "Use format command to format the package files.",
 		RunE:  formatCommandAction,
 	}
+	cmd.Flags().BoolP(failFastFlagName, "f", false, "fail if any file requires formatting")
```
ycombinator, line 20: Nit: to reflect the "fast" part of fail fast:

```suggestion
	cmd.Flags().BoolP(failFastFlagName, "f", false, "fail immediately if any file requires formatting")
```

cmd/format.go
ycombinator: Ref: https://github.com/elastic/elastic-package/pull/44/files#diff-e343a6b946d6d699b7b341b437c58e4b

cmd/format.go
```diff
@@ -1,22 +1,43 @@
...
+	if err != nil {
+		return errors.Wrap(err, "locating package root failed")
+	}
+	if !ok {
+		return errors.New("package root not found")
+	}
+
+	ff, err := cmd.Flags().GetBool(failFastFlagName)
+	if err != nil {
+		return errors.Wrapf(err, "flag not found (flag: %s)", failFastFlagName)
```
ycombinator, line 35: Not for this PR but as we start parsing flags in various command implementations we should probably standardize on the error message when the flag is not found. Maybe create a common function to generate the error or something like that in `internal/cmd/util.go` (related: https://github.com/elastic/elastic-package/pull/53/files#r470018706).

cmd/format.go
```diff
@@ -1,22 +1,43 @@
...
+	}
+
+	err = formatter.Format(packageRoot, ff)
+	if err != nil {
+		return errors.Wrapf(err, "formatting the integration failed (path: %s, failFast: %t)", packageRoot, ff)
```
ycombinator, line 40: Since the user will be familiar with `--fail-fast` not `failFast`.

```suggestion
		return errors.Wrapf(err, "formatting the integration failed (path: %s, --fail-fast: %t)", packageRoot, ff)
```

internal/formatter/formatter.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func formatFile(path string, failFast bool) error {
+	file := filepath.Base(path)
+	ext := filepath.Ext(file)
+
+	var c interface{}
+	var unmarshal func([]byte, interface{}) error
+	var marshal func(interface{}, string, string) ([]byte, error)
+	switch ext {
```
ycombinator, line 44: I wonder if we should use https://golang.org/pkg/net/http/#DetectContentType instead of relying on the extension, especially since you are opening the file further down in this function anyway.

mtojek: Take a look at the internals of this method. It depends on mime types definition of the operating system, not really the Go code. I think it last detection of JSON and YAML content type.

cmd/format.go
mtojek: In this case I'm not convinced, it's just logging an error.

internal/formatter/formatter.go
ycombinator: Hmmm, looks like `http.DetectContentType` cannot detect JSON or YAML after all. So ++ to relying on the extension instead.

cmd/check.go
mtojek: Fixed

cmd/format.go
mtojek: I looked into IDE and saw the problem you spotted. If we follow your idea with moving the Compose.. to `internal`, the `cmd` package would be cleaner.

As this is strongly related with https://github.com/spf13/cobra , I suggest to name the package `cobraext`. WDYT?

cmd/format.go
mtojek: Fixed.

ruflin:
I'm curious to see what things change when you apply it to existing packages.

mtojek:
> I'm curious to see what things change when you apply it to existing packages.

In most cases, fields are sorted in alpha num order - that's the first impression. If you dive into diffs deeper, mostly redundant whitespaces.

ruflin:
The sorting is probably what I'm worried about. I think it is important that a user can create a package and use his own logic on how he sorts the yaml keys. Imagine `go fmt` would sort all your methods alphabetically instead of the order you select to have related methods closely together.

cmd/format.go
mtojek: Good idea, Shaunak. I prepared such method, also applied in couple of places.

mtojek:
I upgraded the library to `gopkg.in/yaml.v3` which provides `yaml.Node` that preserves comments and other meta information. We don't need to modify the JSON marshalers as the only JSON files we've are Kibana files and the formatter is inline with them.

As the consequence of having comments, the `yaml.Node` preserves order of fields, so no alpha-num sorting.

cmd/format.go
```diff
@@ -1,9 +1,18 @@
...
+
+	"github.com/elastic/elastic-package/internal/cobraext"
+	"github.com/elastic/elastic-package/internal/formatter"
+	"github.com/elastic/elastic-package/internal/packages"
+)
+
+// Flag "fail-fast" is used by commands: check, format
+const (
+	failFastFlagName        = "fail-fast"
+	failFastFlagDescription = "fail immediately if any file requires updates"
```
ycombinator, line 15: Since these flags are used by multiple commands, should we move them to `internal/cobraext`, maybe in a `const.go` file or something?

cmd/format.go
```diff
@@ -13,10 +22,27 @@ func setupFormatCommand() *cobra.Command {
 		Long:  "Use format command to format the package files.",
 		RunE:  formatCommandAction,
 	}
+	cmd.Flags().BoolP(failFastFlagName, "f", false, failFastFlagDescription)
 	return cmd
 }
 
 func formatCommandAction(cmd *cobra.Command, args []string) error {
-	fmt.Println("format is not implemented yet.")
+	packageRoot, ok, err := packages.FindPackageRoot()
```
ycombinator, line 30: nit: s/`ok`/`found`/

cmd/testrunner.go
```diff
@@ -7,26 +7,35 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
+	"github.com/elastic/elastic-package/internal/cobraext"
 	"github.com/elastic/elastic-package/internal/packages"
 	"github.com/elastic/elastic-package/internal/testrunner"
 	"github.com/elastic/elastic-package/internal/testrunner/system"
 )
 
+const (
```
ycombinator, line 16: I can see these flags (and their descriptions) being used in multiple commands, especially the `dataset` flag. In any case it might be good to keep all possible flags and their descriptions in a single file to make it easier to achieve consistency in naming and language. WDYT about moving them to something like `internal/cobraext/const.go`?

internal/formatter/formatter.go
```diff
@@ -0,0 +1,94 @@
...
+
+func formatFile(path string, failFast bool) error {
+	file := filepath.Base(path)
+	ext := filepath.Ext(file)
+
+	switch ext {
+	case ".json":
+	case ".yml":
+	default:
+		return nil
```
ycombinator, line 45: Should we be returning an error here about the format/extension not being supported?

ycombinator: Or, if not return an error, then log a message if we are skipping formatting of a file type?

ycombinator: BTW, I just checked all integrations packages and this is the distribution of extensions, in case we want to add support for more formats in the future:

```
$ find packages -type f | awk -F\. '{print $NF}' | sort | uniq -c | sort -nr
 750 json
 561 yml
 149 hbs
  54 png
  27 md
  26 svg
   4 log
   2 jpg
   1 js
```

ycombinator: The only ones in that list that look like worth supporting in the future (not this PR) are `yml.hbs` and _maybe_ `md`.

internal/formatter/formatter.go
```diff
@@ -0,0 +1,94 @@
...
+		err = yaml.Unmarshal(content, &node)
+		if err != nil {
+			return errors.Wrap(err, "unmarshalling YAML file failed")
+		}
+
+		formatted, err = yaml.Marshal(&node)
+		if err != nil {
+			return errors.Wrap(err, "marshalling JSON raw message failed")
+		}
+	}
```
ycombinator, line 79: So we end up having two switch-case constructs in this function that are switching on the same cases. I don't think this is a big deal right now as we only have two cases so feel free to keep things as-is. 

However, just as a suggestion, if you like you could consider doing something like:

```go
type formatter func(path string) ([]byte, bool, error)

var formatters = map[string]formatter{
  "json": jsonFormatter,
  "yml": yamlFormatter,
  "yaml": yamlFormatter,
}

func jsonFormatter(filePath []byte) ([]byte, bool, error) {
   // return formatted bytes, whether file is already in good shape, error
}

// more formatter functions
```

and then use the map inside this function instead of the two `switch-case` constructs.

mtojek: I like this design idea! I've adjusted the code.

cmd/format.go
mtojek: Fixed

cmd/format.go
mtojek: Good idea, fixed

cmd/testrunner.go
mtojek: Fixed

internal/formatter/formatter.go
mtojek: The problem with `yml.hbs` files is that they're templates, so you can't easily assume which spaces/tabs you can adjust/remove and which ones are meaningful because of some conditional rendering.

Regarding `md` I believe it's doable, but this should be coordinated with tech writers. AFAIK they prefer to have a single web service you can use to format your docs, but our use case is a bit different (I assume we won't allow community users to use such web service).

ycombinator: Makes sense to keep this as-is. We don't want to to return errors for un-formattable files or log messages either as it might get too spammy.

internal/formatter/formatter.go
```diff
@@ -0,0 +1,75 @@
...
+	file := filepath.Base(path)
+	ext := filepath.Ext(file)
+
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return errors.Wrap(err, "reading file content failed")
+	}
+
+	format, undefined := formatters[ext]
+	if !undefined {
```
ycombinator, line 53: Shouldn't the variable name be `defined` so the `if` condition will read as, "if not defined"? Right it reads as "if not undefined", which seems inaccurate?

mtojek: Of course, it's bug. Fixed!

PR #54, opened 2020-08-14
Add cluster run command

ruflin:
This adds the command to spin up the cluster but have the logs exposed. The reason is that when I spin up the cluster, I always want to see the logs and not having it run in the background to debug potential issues.

I am opening this as draft to have a discussion on this first and it seems on shutdown it somehow gets stuck. Also naming wise users will aks what is the difference between up and run (one background, one not). Lets see what others think.

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -12,13 +12,18 @@ func setupClusterCommand() *cobra.Command {
 		Use:   "up",
 		Short: "Boot up the testing cluster",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := cluster.BootUp()
+			d, err := cmd.Flags().GetBool("d")
+			if err != nil {
+				return err
+			}
+			err = cluster.BootUp(d)
 			if err != nil {
 				return errors.Wrap(err, "booting up the cluster failed")
 			}
 			return nil
 		},
 	}
+	upCommand.Flags().Bool("d", false, "Run cluster as daemon")
 
 	runCommand := &cobra.Command{
 		Use:   "run",
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -20,6 +20,18 @@ func setupClusterCommand() *cobra.Command {
 		},
 	}
 
+	runCommand := &cobra.Command{
+		Use:   "run",
+		Short: "Run the testing cluster",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := cluster.Run()
+			if err != nil {
+				return errors.Wrap(err, "running the cluster failed")
+			}
+			return nil
+		},
+	}
+
 	downCommand := &cobra.Command{
 		Use:   "down",
 		Short: "Take down the testing cluster",
@@ -52,6 +64,7 @@ func setupClusterCommand() *cobra.Command {
 	}
 	cmd.AddCommand(
 		upCommand,
+		runCommand,
 		downCommand,
 		shellInitCommand)
 	return cmd
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -23,7 +23,7 @@ func setupClusterCommand() *cobra.Command {
 			return nil
 		},
 	}
-	upCommand.Flags().Bool("d", false, "Run cluster as daemon")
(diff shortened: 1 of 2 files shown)


ruflin:
SGTM, can adjust the code.

cmd/cluster.go
```diff
@@ -12,13 +12,18 @@ func setupClusterCommand() *cobra.Command {
...
+				return err
+			}
+			err = cluster.BootUp(d)
 			if err != nil {
 				return errors.Wrap(err, "booting up the cluster failed")
 			}
 			return nil
 		},
 	}
+	upCommand.Flags().Bool("d", false, "Run cluster as daemon")
```
ruflin, line 26: @mtojek This gets `--d` working but not `-d`. I assume there is a better way to define this in cobra?

mtojek: Yes, here is a sample with short flag: https://github.com/elastic/elastic-package/blob/master/cmd/testrunner.go#L28

You can add both forms (daemon/d).

ruflin: Hm, now with using `PersistentFlags().BoolP` it stops working completely.

mtojek: Sorry, I meant to use the method `BoolP`, but with `Flags()`. Persistent flags are reserved rather for general purpose flags, independent from commands, e.g. verbose, debug, etc.

cmd/cluster.go
```diff
@@ -12,13 +12,18 @@ func setupClusterCommand() *cobra.Command {
...
+				return err
+			}
+			err = cluster.BootUp(d)
 			if err != nil {
 				return errors.Wrap(err, "booting up the cluster failed")
 			}
 			return nil
 		},
 	}
+	upCommand.Flags().BoolP("daemon","d", false, "Run cluster as daemon")
```
mtojek, line 26: nit: space between "daemon" and "d"

cmd/cluster.go
```diff
@@ -12,13 +12,18 @@ func setupClusterCommand() *cobra.Command {
 		Use:   "up",
 		Short: "Boot up the testing cluster",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := cluster.BootUp()
+			d, err := cmd.Flags().GetBool("d")
+			if err != nil {
+				return err
```
mtojek, line 17: Please wrap the error. See: https://github.com/elastic/elastic-package/blob/master/cmd/testrunner.go#L52

cmd/cluster.go
```diff
@@ -12,13 +12,18 @@ func setupClusterCommand() *cobra.Command {
 		Use:   "up",
 		Short: "Boot up the testing cluster",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := cluster.BootUp()
+			d, err := cmd.Flags().GetBool("d")
```
mtojek, line 15: It's better to refer to parameters using full names, here: `GetBool("daemon")`.

mtojek, line 15: > `d, err`

`deamonMode, err`, WDYT?

internal/cluster/boot.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 // BootUp method boots up the testing cluster.
-func BootUp() error {
+func BootUp(d bool) error {
```
mtojek, line 17: daemonMode

internal/cluster/boot.go
```diff
@@ -84,7 +84,7 @@ func dockerComposeBuild() error {
 	return nil
 }
 
-func dockerComposeUp() error {
+func dockerComposeUpD() error {
```
mtojek, line 87: I suppose you can remove this method?

internal/cluster/boot.go
```diff
@@ -104,6 +104,29 @@ func dockerComposeUp() error {
 	return nil
 }
 
+func dockerComposeUp(d bool) error {
```
mtojek, line 107: deamonMode

mtojek:
Closing in favor of https://github.com/elastic/elastic-package/pull/60

Issue #45, opened 2020-08-12
Add support for advanced logging


mtojek:
Right now the tool uses simple `fmt.Println` calls. We should introduce some more advanced logging, but ones that doesn't affect readability (JSON array is not a human readable format :)

PR #59, opened 2020-08-17
Add support for levelled logging

mtojek:
_Let's first finalize this one: https://github.com/elastic/elastic-package/pull/53 , then jump on this._

Fixes: https://github.com/elastic/elastic-package/issues/45

This PR implements levelled logging and exposes a "verbose" flag.


internal/logger/logger.go
```diff
@@ -0,0 +1,65 @@
...
+// Warnf method logs message with "warn" level and formats it.
+func Warnf(format string, a ...interface{}) {
+	logMessagef("WARN", format, a...)
+}
+
+func logMessage(level string, a ...interface{}) {
+	var all []interface{}
+	all = append(all, fmt.Sprintf("%5s ", level))
+	all = append(all, a...)
+	log.Print(all...)
```
ycombinator, line 57: Should we include a timestamp in the log message? Not sure how valuable they will be as most sub-commands tend to run for relatively short periods of time. Also, there's no concurrency yet so establishing order of operations via log messages is quite straightforward even without timestamps. Just thought I'd ask and get your thoughts on the idea.

mtojek: Hm.. If you don't ming I would rather keep them there. Timestamps may help us finding some bottlenecks in processing large integrations, interacting with Kibana instance or similar. It's better to have them instead of regret :)

ycombinator: Ha, I totally misunderstood that timestamps were already included in the message! I though we would have to add them manually here that's why I even brought up the question. If they are already included then ++ to just keeping them!

PR #64, opened 2020-08-18
System test runner MVP

ycombinator:
This PR fleshes out the implementation of the system tests runner. 

Initially it focusses on packages whose services can be spun up using Docker Compose but it has a provision to be extended later to include other types of services, e.g. cloud integration packages.

## Configuring the package/dataset for a system test
1. Every package must define how it's service must be spun up and torn down, by placing the appropriate files in the `_dev` folder located at the package's root folder.
   * Currently, only Docker Compose-based services are supported. 
   * For this, a `_dev/docker-compose.yml` file must be created. This file may define as many services as needed but the main one is determined by a naming convention: the name of the main service in the `docker-compose.yml` file must be the same name as the package. 

2. Additionally, every package dataset that wants to be system-tested must define a system test configuration file, `config.yml`, located in the `_dev/test/system` folder at the dataset's root folder.

For example, see [this PR](https://github.com/elastic/integrations/pull/263) setting up the above files for the `apache` package and its datasets.

## Running the system test
First, we must spin up the Elastic Stack and export environment variables.
```
$ elastic-package stack up --version 8.0.0-SNAPSHOT -d
$ eval $(elastic-package stack shellinit)
```

Then we can run the system test on a package (or specific dataset(s) within a package with the optional `--dataset` flag).
```
$ cd /path/to/some/package/root/folder
$ elastic-package test system
```

The above command will perform the following steps under the hood:

1. It will find the `_dev` folder at the package root and look inside it to decide how the service for this package must be spun up (and later torn down). 
   * Docker Compose-based services are started up as Docker containers on the same network as the Elastic Stack spun up with `elastic-package stack up`.

2. The service is spun up and certain **context variables** are registered. These variables become available to every dataset's test configuration file. Currently, the following context variables are available:
   * `{{ Service.STDOUT }}`: Available for Docker Compose-based services. Its value will be a file path available in the Agent container. This file will contain the contents of the service's STDOUT stream. It is useful for gathering logs from Docker Compose-based services as they write their logs to STDOUT by default.
   * `{{ Service.Ports.n.InternalPort }}`: Available for Docker-Compose based services. Its value will be the internal port available on the Docker network shared between the service container(s) and Elastic Stack containers. Note that the `n` here is a 0-based index, to allow for a service to listen on multiple internal ports.
   * `{{ Service.Port.InternalPort }}`: Available for Docker-Compose based services. It is simply a convenient alias for `{{ Service.Ports.0.InternalPort }}` as most Docker-Compose based services will only listen on a single internal port.
   * `{{ Service.Hostname }}`: Host name of service, as addressable from Agent container.

3. The dataset's **system test configuration file** is loaded into memory. Any context variables are substituted with their corresponding values. All other dataset and package variables omitted from the system test configuration file are populated from the package and dataset manifest files, respectively.

4. A **policy** is created for the dataset being tested in Ingest Manager. The policy is attached to the test Agent that is already spun up as part of the Elastic Stack and enrolled with Fleet. Within a minute or so the policy is sent to the Agent by Fleet.

5. Agent applies the policy and should start indexing data into the appropriate data stream(s) in Elasticsearch.

6. The system test runner checks if the expected data streams exist and have data (documents) in them.


internal/testrunner/runners/system/system.go
```diff
@@ -3,29 +3,59 @@ package system
...
+	// Step 1. Setup service.
+	// Step 1a. (Deferred) Tear down service.
+	ctxt, err := serviceRunner.SetUp()
+	defer serviceRunner.TearDown(ctxt)
+	if err != nil {
+		return errors.Wrap(err, "could not setup service")
+	}
+
+	// Step 2. Setup test cluster (ES + Kibana + Agent).
+	// Step 2a. Register Agent with Fleet.
```
ycombinator, line 52: @mtojek I would like your advice here: what would be the best way to accomplish these two steps (2. and 2a.) here?

mtojek: Step 2a:

It depends where the agent will be spawned. If it runs in a container then the enrollment should be performed automatically (it's in agent's Dockerfile).
I believe you have to interact with cluster to attach new container with observed product (like mysql or nginx) to the existing network, and detach it after tests.

Step 2:

I think you should assume that the cluster is already up and running, and the Agent in subscribed. You can pick up configuration options through environment variables (shellinit). I assume that the Agent is long-running and we just switch configuration inside.

ycombinator: > I think you should assume that the cluster is already up and running, and the Agent in subscribed. You can pick up configuration options through environment variables (shellinit). I assume that the Agent is long-running and we just switch configuration inside.

This is the part I am most interested in. Is there some way I can _ensure_ that the cluster is up, rather than assuming it is? Are there any functions/methods I could call from here to make that happen?

> I believe you have to interact with cluster to attach new container with observed product (like mysql or nginx) to the existing network, and detach it after tests.

Cool, this is along the lines of what I was hoping for. For this I would need to know the name of the network. Is there some function/method I could call from here to get that (related to the previous paragraph, I suppose)?

mtojek: > This is the part I am most interested in. Is there some way I can ensure that the cluster is up, rather than assuming it is? Are there any functions/methods I could call from here to make that happen idempotently?

I think it' s `cluster.BootUp()`, the only issue we need to figure out is waiting for Kibana. I believe you start with the assumption that the cluster is up and fail in case it's not.

>  For this I would need to know the name of the network. Is there some function/method I could call from here to get that (related to the previous paragraph, I suppose)?

There is no such function at the moment, as it wasn't required, but I suppose we can enforce any name store it in `.env` file in `.elastic-package`.

ycombinator:
@mtojek I just built `elastic-package` from this PR and tried running `elastic-package test system` from the `apache` package folder. I got this error:

```
$ elastic-package test system
system run /Users/shaunak/development/github/integrations/packages/apache/dataset/access/_dev/test/system
Custom build packages directory found: /Users/shaunak/development/github/integrations/build/integrations
Building package-registry
Step 1/3 : FROM docker.elastic.co/package-registry/distribution:snapshot
ERROR: Service 'package-registry' failed to build: Get https://docker.elastic.co/v2/package-registry/distribution/manifests/snapshot: unauthorized: authentication required
Error: error running package system tests: could not setup test cluster: building docker images failed: running command failed: exit status 1
```

Any idea what causes this / how a user of `elastic-package` is expected to authenticate?

ycombinator:
I was able to resolve the issue in my previous comment. It was a one-off issue, so nothing we need to worry about in general with `elastic-package` users.

internal/testrunner/testrunner.go
```diff
@@ -55,12 +76,12 @@ func RegisterRunner(testType TestType, runFunc RunFunc) {
 }
 
 // Run method delegates execution to the registered test runner, based on the test type.
-func Run(testType TestType, testFolderPath string) error {
+func Run(testType TestType, testFolder TestFolder, packageRootPath string) error {
```
mtojek, line 79: I think we can sync up both branches.

I introduced a similar concept of `TestOptions`: https://github.com/elastic/elastic-package/blob/c3fd692e9923f4160b11b3eac569b76014552934/internal/testrunner/testrunner.go to easily add more arguments here.

internal/testrunner/runners/system/system.go
```diff
@@ -2,30 +2,134 @@ package system
...
-	fmt.Println("system run", r.testFolderPath)
+	// Step 1. Setup service.
+	// Step 1a. (Deferred) Tear down service.
+	logger.Info("Setting up service...")
+	serviceRunner, err := servicedeployer.Factory(r.packageRootPath)
+	if err != nil {
+		return errors.Wrap(err, "could not create service runner")
+	}
+
+	ctxt := common.MapStr{}
```
mtojek, line 62: I think it's still early and there is no implementation using `MapStr`. Do you plan to use the MapStr to flatten JSON object representation? If so, I wonder if standars `encoding/json` is sufficient.

mtojek: Also, this is the reason why I don't flatten the expected results in the pipeline tests. Maybe I should introduce it there.

internal/kibana/ingestmanager/client.go
```diff
@@ -0,0 +1,16 @@
+package ingestmanager
+
+type Client struct {
```
mtojek, line 3: I used a dedicated client for Elasticsearch. Maybe there is something similar for Kibana?

internal/testrunner/runners/system/system.go
```diff
@@ -2,30 +2,134 @@ package system
...
+	// Step 4. (TODO in future) Optionally exercise service to generate load.
+
+	// Step 5. Assert that there's expected data in data stream.
+
 	return nil
 }
 
-func init() {
-	testrunner.RegisterRunner(TestType, Run)
+func getStackSettingsFromEnv() stackSettings {
```
mtojek, line 118: I exposed environment variables here: https://github.com/elastic/elastic-package/blob/c3fd692e9923f4160b11b3eac569b76014552934/internal/elasticsearch/client.go Does it make sense to use them here?

internal/testrunner/runners/system/system.go
ycombinator: I'm not 100% sure yet how this usage is going to evolve. My current thinking is to use `ctxt` to pass some arbitrary key-value pair data between the various steps of testing but this is something I'm experimenting with.

For test results, since those are consistently going to be present for each test run, I probably wouldn't use `ctxt` but something else. I haven't gotten that far yet so I'm not sure yet.

internal/kibana/ingestmanager/client.go
ycombinator: I have not found a reliable one so far. I suspect, over time, as Kibana APIs become more "public" we will see one emerge. At that point we can use it here.

internal/testrunner/runners/system/system.go
ycombinator: I'm open to it but I'm not sure how to do that. Can you elaborate what should change here?

mtojek: I suppose it would be merging the other PR and adjust the implementation here to use environment variables defined in a single place (e.g. close to the stack).

ycombinator: Okay, I plan to bring the changes from the other PR into this one today. Let's see how things look at that point.

internal/kibana/ingestmanager/client.go
ycombinator: BTW, looks like there is some community interest in this too: https://github.com/elastic/go-elasticsearch/issues/182

mtojek: Oh, nice :)

internal/install/static_snapshot_yml.go
```diff
@@ -69,4 +69,31 @@ services:
...
+      interval: 1s
+    environment:
+    - "FLEET_ENROLL=1"
+    - "FLEET_ENROLL_INSECURE=1"
+    - "FLEET_SETUP=1"
+    - "KIBANA_HOST=http://kibana:5601"
+    volumes:
+    - type: bind
+      source: ../tmp/
+      target: /tmp/service_logs/
```
mtojek, line 92: Ok, so with this solution, we lose the opportunity to test on real filesystem and have log files distributed across the entire partition. We might need a comment somewhere about this limitation or a task for next iteration (maybe we can solve it somehow).

internal/testrunner/runners/system/system.go
```diff
@@ -5,31 +5,364 @@
...
+		logger.Info("tearing down service...")
+		if err := serviceRunner.TearDown(ctxt); err != nil {
+			logger.Errorf("error tearing down service: %s", err)
+		}
+	}()
+	if err != nil {
+		return errors.Wrap(err, "could not setup service")
+	}
+
+	// Step 2. Configure package (single data stream) via Ingest Manager APIs.
```
mtojek, line 112: Actually you don't use any "install" API like the one that is used in integration tests of package-registry. It looks like it's not required to be involved in the process, so I wonder what does it do actually? For reference: https://github.com/elastic/package-registry/blob/master/main_integration_test.go#L82

Maybe it would be fine to call it for safety reasons, to verify if the package can be installed?

internal/testrunner/runners/system/system.go
```diff
@@ -5,31 +5,364 @@
...
+	im, err := ingestmanager.NewClient(r.stackSettings.kibana.host, r.stackSettings.elasticsearch.username, r.stackSettings.elasticsearch.password)
+	if err != nil {
+		return errors.Wrap(err, "could not create ingest manager client")
+	}
+
+	logger.Info("creating test policy...")
+	p := ingestmanager.Policy{
+		Name:        fmt.Sprintf("ep-test-system-%s-%s", r.testFolder.Package, r.testFolder.Dataset),
```
mtojek, line 120: I didn't see you applied any timestamps here. I wonder if it's not better to include some. Consider a scenario in which developer stops the test runner in the middle and for some reason the cleanup never happened or failed. We can prevent some transient errors in the future.

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Build service container name
+	serviceName, err := getStrFromCtxt(ctxt, "service.name")
+	if err != nil {
+		return ctxt, errors.Wrap(err, "could not get service name")
+	}
+	serviceContainer := fmt.Sprintf("%s_%s_1", r.project, serviceName)
+	ctxt.Put("Service.Hostname", serviceContainer)
+
+	// Redirect service container's STDOUT and STDERR streams to files in local logs folder
```
mtojek, line 70: I wonder if we can simplify this - we still have a volume with log files mounted, so maybe don't depend specifically on stdout and stderr, but just share a volume between agent and the service. I'm fine with discussing this in a different issue.

mtojek:
Ok, I went through the source code, left couple of high level comments. I really like this, it's not full-blown testing framework, but a simple runner. Nicely done! In general I'm for merging a more or less experimental version of the feature and simply not enabling it in the Integrations repository.

My only concerns are around "recording" stdout and stderr of the container instead of sharing a mounted volume between the agent and deployed service. A possible solution to be considered:

1. In the service configuration, we define paths to be exposed as mounts (if there are multiple directories with logs).
2. Service and agent are deployed on demand only for the lifetime of the test (always in pair).
3. They share multiple volumes with logs.

Re 2: I understand it might be hard to determine when to start or stop the service&agent while interacting with multiple test cases, as enrolling and unenrolling every time is time-consuming. On the other hand, we have guarantee that the agent picked up the right configuration.

internal/install/static_snapshot_yml.go
```diff
@@ -69,4 +69,31 @@ services:
...
+      retries: 30
+      interval: 1s
+    environment:
+    - "FLEET_ENROLL=1"
+    - "FLEET_ENROLL_INSECURE=1"
+    - "FLEET_SETUP=1"
+    - "KIBANA_HOST=http://kibana:5601"
+    volumes:
+    - type: bind
+      source: ../tmp/
```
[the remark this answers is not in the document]
ycombinator: Sure, I'll call it `service_logs` so it matches the name of the folder in the agent container (`/tmp/service_logs`).

internal/install/static_snapshot_yml.go
```diff
@@ -69,4 +69,31 @@ services:
...
+    healthcheck:
+      test: ["CMD", "grep", "Agent is starting", "/usr/share/elastic-agent/elastic-agent.log"]
+      retries: 30
+      interval: 1s
+    environment:
+    - "FLEET_ENROLL=1"
+    - "FLEET_ENROLL_INSECURE=1"
+    - "FLEET_SETUP=1"
+    - "KIBANA_HOST=http://kibana:5601"
+    volumes:
```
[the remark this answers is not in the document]
ycombinator: I did not, but will do.

internal/install/static_snapshot_yml.go
```diff
@@ -69,4 +69,31 @@ services:
...
+
+  elastic-agent:
+    image: docker.elastic.co/beats/elastic-agent:${STACK_VERSION}
+    depends_on:
+      elasticsearch:
+        condition: service_healthy
+      kibana:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD", "grep", "Agent is starting", "/usr/share/elastic-agent/elastic-agent.log"]
```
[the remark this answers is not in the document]
ycombinator: I'm not sure I understand this point. Are you suggesting we tell Agent to write its log files to `/tmp` instead of where it writes them by default and change the health check here somehow?

internal/testrunner/runners/system/config.go
```diff
@@ -0,0 +1,57 @@
...
+
+	var c config
+	if err := yaml.Unmarshal(data, &c); err != nil {
+		return nil, errors.Wrapf(err, "unable to parse system test configuration file: %s", configFilePath)
+	}
+
+	return &c, nil
+}
+
+func applyContext(ctxt common.MapStr, data []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
ycombinator: Okay, I will add some comments explaining why.

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	if err := r.stderr.Close(); err != nil {
+		return errors.Wrapf(err, "could not close STDERR file: %s", r.stderrFilePath)
+	}
+	if err := os.Remove(r.stderrFilePath); err != nil {
+		return errors.Wrapf(err, "could not delete STDERR file: %s", r.stderrFilePath)
+	}
+
+	if err := r.stdout.Close(); err != nil {
```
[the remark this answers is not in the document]
ycombinator: Yeah, good idea to move this to a `defer`. Will do!

internal/testrunner/runners/system/servicedeployer/factory.go
```diff
@@ -0,0 +1,27 @@
...
+
+var (
+	// ErrNotFound is returned when the appropriate service runner for a package
+	// cannot be found.
+	ErrNotFound = errors.New("unable to find service runner")
+)
+
+// Factory chooses the appropriate service runner for the given package, depending
+// on service configuration files defined in the package.
+func Factory(packageRootPath string) (Runner, error) {
```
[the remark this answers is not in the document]
ycombinator: I like it! Will change.

internal/testrunner/runners/system/system.go
```diff
@@ -5,31 +5,364 @@
...
+		return errors.Wrap(err, "could not create service runner")
+	}
+
+	tempDir, err := install.TempDir()
+	if err != nil {
+		return errors.Wrap(err, "could not get temporary folder")
+	}
+
+	ctxt := common.MapStr{}
+	ctxt.Put("service.name", r.testFolder.Package)
```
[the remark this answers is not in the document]
ycombinator: Yeah, I think we can do a hybrid here. I will change it to a struct and add any "standard" properties that we expect to exist all the time. But I will also add a `common.MapStr` field to the struct for arbitrary key/values.

internal/testrunner/runners/system/system.go
```diff
@@ -5,31 +5,364 @@
...
+	if agents == nil || len(agents) == 0 {
+		return errors.New("no agents found")
+	}
+	agent := agents[0]
+	origPolicy := ingestmanager.Policy{
+		ID: agent.PolicyID,
+	}
+
+	// Delete old data
+	esClient, err := elasticsearch.Client()
```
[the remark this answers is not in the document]
ycombinator: Good idea to share one client across multiple invocations of the test runner. Will change.

internal/install/static_snapshot_yml.go
ycombinator: Not necessarily. There is no requirement for log files paths in the dataset configuration to be restricted to `/tmp/service_logs`. So, for example, when we write a test for the `system` package, we can refer to standard locations for system log files assuming the Agent is running some flavor of Linux.

internal/testrunner/runners/system/system.go
ycombinator: That's because it is being called by Agent. See https://github.com/elastic/elastic-package/pull/64/files#diff-906969eaf1b600ec57ae5e30f11d1a3dR87, https://github.com/elastic/beats/blob/2a0e0998243db56d0bfd681ff3cb06183166bdd5/dev-tools/packaging/templates/docker/docker-entrypoint.elastic-agent.tmpl#L64, and https://github.com/elastic/beats/blob/master/dev-tools/packaging/templates/docker/docker-entrypoint.elastic-agent.tmpl#L16.

internal/testrunner/runners/system/system.go
ycombinator: Yes, we could do that. Another option would be to not use timestamps so we have the same identifiers for each test run. Then we clean up resources _before_ (and after) the test run starts, in case any resources were left from transient failures from a previous run.

Either way, I agree we need to account for orphan resources due to transient failures from previous runs.

WDYT?

internal/testrunner/runners/system/servicedeployer/compose.go
ycombinator: I think I'm missing something here. Let's use the `apache` container as an example. Let's say we have a shared volume between this container and the `elastic-agent` container. I assume we want Apache log files to be placed by the `apache` container into this volume. But how/where exactly do we make this happen?

ycombinator: I just read your overall PR review comment in which you mention this:

> In the service configuration, we define paths to be exposed as mounts...

I think that answers my question above, "how/where exactly do we make this happen?" We would define them in the `docker-compose.yml` file for the service. Let me try this out and see if we can make this work.

internal/install/static_snapshot_yml.go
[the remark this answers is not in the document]
ycombinator: Hmmm, interesting. It seems to work for `package-registry` and `elasticsearch` but not for `kibana`:

### Starting up only `package-registry` ✔️ 

```
$ elastic-package stack up --version 8.0.0-SNAPSHOT -d -v -s package-registry
2020/09/01 10:34:28 DEBUG Enable verbose logging
Boot up the Elastic stack
2020/09/01 10:34:28 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack build package-registry package-registry_is_ready
2020/09/01 10:34:53 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack down
2020/09/01 10:34:54 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack up -d package-registry package-registry_is_ready
Done
```
```
$ docker ps
CONTAINER ID        IMAGE                                    COMMAND                  CREATED             STATUS                    PORTS                      NAMES
2ff3fbee2ecb        elastic-package-stack_package-registry   "./package-registry …"   18 seconds ago      Up 17 seconds (healthy)   127.0.0.1:8080->8080/tcp   elastic-package-stack_package-registry_1
```

### Starting up only `elasticsearch` ✔️ 

```
$ elastic-package stack up --version 8.0.0-SNAPSHOT -d -v -s elasticsearch
2020/09/01 10:41:34 DEBUG Enable verbose logging
Boot up the Elastic stack
2020/09/01 10:41:34 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack build elasticsearch elasticsearch_is_ready
2020/09/01 10:41:34 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack down
2020/09/01 10:41:35 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack up -d elasticsearch elasticsearch_is_ready
Done
```
```
$ docker ps
CONTAINER ID        IMAGE                                                          COMMAND                  CREATED              STATUS                        PORTS                                NAMES
826522aaf207        docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT   "/tini -- /usr/local…"   About a minute ago   Up About a minute (healthy)   127.0.0.1:9200->9200/tcp, 9300/tcp   elastic-package-stack_elasticsearch_1
```

### Starting up only `kibana` ❌ 

```
$ elastic-package stack up --version 8.0.0-SNAPSHOT -d -v -s kibana
2020/09/01 10:36:42 DEBUG Enable verbose logging
Boot up the Elastic stack
2020/09/01 10:36:42 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack build
2020/09/01 10:37:06 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack down
2020/09/01 10:37:07 DEBUG running command: /usr/local/bin/docker-compose -f /Users/shaunak/.elastic-package/stack/snapshot.yml -p elastic-package-stack up -d
Done
```
```
$ docker ps
CONTAINER ID        IMAGE                                                          COMMAND                  CREATED              STATUS                        PORTS                                NAMES
f00fbb57d861        docker.elastic.co/beats/elastic-agent:8.0.0-SNAPSHOT           "/usr/local/bin/dock…"   40 seconds ago       Up 39 seconds (healthy)                                            elastic-package-stack_elastic-agent_1
a3b35e4a30ad        docker.elastic.co/kibana/kibana:8.0.0-SNAPSHOT                 "/usr/local/bin/dumb…"   About a minute ago   Up About a minute (healthy)   127.0.0.1:5601->5601/tcp             elastic-package-stack_kibana_1
6f35f3833863        docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT   "/tini -- /usr/local…"   2 minutes ago        Up 2 minutes (healthy)        127.0.0.1:9200->9200/tcp, 9300/tcp   elastic-package-stack_elasticsearch_1
0c4cd0260cf1        elastic-package-stack_package-registry                         "./package-registry …"   2 minutes ago        Up 2 minutes (healthy)        127.0.0.1:8080->8080/tcp             elastic-package-stack_package-registry_1
```

Wonder why that is? 🤔 Investigating...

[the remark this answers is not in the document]
ycombinator: Oh, it's because of this check: https://github.com/elastic/elastic-package/blob/ba7825e1b010cab16705f47c5c6d1fad332edd52/internal/stack/boot.go#L186-L188

I will update that to check for `elastic-agent` now as that is the "root" service in the dependency tree now.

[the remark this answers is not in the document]
ycombinator: Updated in 730738123a978601e4e871e2958d16cfd5f8d18a.

internal/kibana/ingestmanager/agents.go
```diff
@@ -0,0 +1,52 @@
+package ingestmanager
```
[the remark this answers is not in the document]
ycombinator: Renamed `agents.go` and `policies.go` in cb87e8f.

internal/testrunner/runners/system/servicedeployer/compose.go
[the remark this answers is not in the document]
ycombinator: Fixed in f24d666.

internal/testrunner/runners/system/config.go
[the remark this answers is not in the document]
ycombinator: I took a crack at this in 38c62ae. Let me know if you want further changes.

internal/testrunner/runners/system/servicedeployer/factory.go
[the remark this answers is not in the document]
ycombinator: Done in fcae0b7. Let me know what you think.

internal/testrunner/runners/system/system.go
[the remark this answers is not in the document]
ycombinator: Actually I went all the way and made the context a struct. See c0c672a.

internal/testrunner/runners/system/system.go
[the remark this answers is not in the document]
ycombinator: Done in 7ee7bd9064a73f3098a8b9e1f54cfdf6b0e351dc. I used it in the Pipeline Test Runner as well so you might want to take a look at that.

internal/install/static_snapshot_yml.go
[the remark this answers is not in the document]
ycombinator: Fixed in fcec6d1.

ycombinator:
@mtojek I'm stopping for the day so posting an update here about addressing your feedback so far.

I've addressed _most_ of your feedback by now. The main item remaining is to explore sharing logs between the service container and agent via a volume instead of redirecting STDOUT and STDERR: https://github.com/elastic/elastic-package/pull/64#discussion_r481066466. I will look into this tomorrow.

Also, while I look into the shared volume implementation, I have questions about a couple of your suggestions: https://github.com/elastic/elastic-package/pull/64#discussion_r481281429 and https://github.com/elastic/elastic-package/pull/64#discussion_r481295862.

internal/install/static_snapshot_yml.go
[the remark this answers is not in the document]
mtojek: Sorry for the confusion, it's a rather a nice-to-have feature to also observe programmatically logs of elastic-agent, filebeat and metricbeat while testing. Many configuration errors are reported deep down by filebeat or metricbeat's internals.
Logs can be collected in the same single directory and exported as volume.

internal/testrunner/runners/system/system.go
mtojek: Sorry, I misguided you. I was thinking about the reference to this API method: https://github.com/elastic/package-storage/blob/production/testing/main_integration_test.go#L108

internal/testrunner/runners/system/system.go
mtojek: > Then we clean up resources before (and after) the test run starts, in case any resources were left from transient failures from a previous run.

As long as we don't have intention to run tests in parallel, this can work pretty well.

internal/common/context.go
```diff
@@ -0,0 +1,45 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package common
+
+// ServiceContext encapsulates context that is both available to a ServiceDeployer and
+// populated by a DeployedService. The fields in ServiceContext may be used in handlebars
+// templates in system test configuration files, for example: {{ Hostname }}.
+type ServiceContext struct {
```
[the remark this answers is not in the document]
mtojek: I wonder if the `ServiceContext` implementation should be stored close to the `ServiceDeployer` implementation, in the `servicedeployer` package. If the package name doesn't fit well, what do you think about `deployment` or something similar.

mtojek (commented):
Yeah, I like the way it looks now, my only concern is related to the place where the context is stored.

Also, it seems that CI failed with:
```
[2020-09-01T23:15:55.736Z] + elastic-package stack up -d
[2020-09-01T23:15:55.736Z] elastic-package has been installed.
[2020-09-01T23:15:55.736Z] Boot up the Elastic stack
[2020-09-01T23:19:17.373Z] Error: booting up the stack failed: running docker-compose failed: running command failed: running Docker Compose up command failed: exit status 1
```

For debugging purposes, you can enable verbose mode in this test (currently it doesn't say too much about the root cause).

Once this PR is merged, we can focus on creating followup issues and prioritize them.

internal/install/static_snapshot_yml.go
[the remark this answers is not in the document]
ycombinator: Thanks for explaining. Let's handle this in a follow up PR.

internal/testrunner/runners/system/system.go
ycombinator: Hmmm, good question — I wonder how I'm able to get things working even without calling this API. 🤔 Let's leave it out of this PR for now and add it if we need it in a follow up PR.

internal/testrunner/runners/system/servicedeployer/deployed_service.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/elastic/elastic-package/internal/common"
+)
+
+// DeployedService defines the interface for interacting with a service that has been deployed.
+type DeployedService interface {
+	// TearDown implements the logic for tearing down a service.
+	TearDown() error
+
+	// GetContext returns the current context from the service.
+	GetContext() common.ServiceContext
```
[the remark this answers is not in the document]
ycombinator: I like the parallel with `SetContext` so I would prefer to leave it as `GetContext`.

internal/common/context.go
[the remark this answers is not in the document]
ycombinator: I agree! I think I originally had it in the `servicedeployer` package but then ran into some import cycle issues, which is when I moved it out into `common`. But I just moved it back into `servicedeployer` now (via 65bf419) and I was able to build `elastic-package`.

internal/testrunner/runners/system/system.go
ycombinator: Actually, what I suggested above is not trivial to implement for policies as their IDs are generated by Kibana. We could retrieve all policies and delete ones that match a certain naming convention, but let's worry about that in a follow up PR.

For now, I took your suggestion and added a timestamp to the policy name (via 9fde2df).

internal/testrunner/runners/system/servicedeployer/deployed_service.go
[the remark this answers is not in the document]
mtojek: Sure, let's leave it :)

For reference - I was referring to https://golang.org/doc/effective_go.html#Getters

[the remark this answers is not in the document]
ycombinator: Aha! Thanks for the reference. I will rename it :)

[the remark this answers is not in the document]
ycombinator: Fixed in 8c1c071.

ycombinator:
CI failure should be fixed by 326120ec4a736ab3c072fffef22037a14ffecd25.

mtojek:
> CI failure should be fixed by [326120e](https://github.com/elastic/elastic-package/commit/326120ec4a736ab3c072fffef22037a14ffecd25).

Out of curiosity, why did it fail?

ycombinator:
> Out of curiosity, why did it fail?

Before 326120e, we were using `7.9.0` as the default version of the stack. In this PR we introduce the Elastic Agent in the Docker Compose file. Elastic Agent 7.9.0's Docker image did not have the ability to set the `--insecure` flag when talking to Kibana. This ability was added in 7.9.1 (https://github.com/elastic/beats/pull/20713), so I needed to bump up the default version to 7.9.1.

mtojek:
Clear now! Thanks for explanation.

ycombinator:
There is one more issue we need to deal with w.r.t. the default stack version and this PR. As of now in this PR the default stack version is set to 7.9.1. However, this PR calls Kibana APIs that are only available starting 7.10.0 — actually these APIs are available in 7.9.0 as well but they use different identifiers in that version, e.g. "config" instead of "policy".

So what should we do about this? We could bump the default stack version to 7.10.0-SNAPSHOT _or_ we could rename the API identifiers to the old ones. Personally I lean towards the former option as we know users are unlikely to use `elastic-package` with 7.9.0 of the stack. What is your preference, @mtojek?

internal/compose/compose.go
```diff
@@ -67,7 +136,7 @@ func (p *Project) Down(opts CommandOptions) error {
 	args = append(args, "down")
 	args = append(args, opts.ExtraArgs...)
 
-	if err := runDockerComposeCmd(args, opts.Env); err != nil {
+	if err := p.runDockerComposeCmd(args, opts.Env, nil); err != nil {
```
mtojek, line 139: I'm admit I'm allergic to passing nil values as it tends to evolve into WinAPI "NULL" syndrome: `call(a, b, 1, 1, NULL, NULL, NULL, NULL, NULL, true, NULL, 1, NULL)` :)

I think you can repack all these args into `dockerComposeOptions`? Less refactoring in the future.

mtojek (approved):
> So what should we do about this? We could bump the default stack version to 7.10.0-SNAPSHOT or we could rename the API identifiers to the old ones. Personally I lean towards the former option as we know users are unlikely to use elastic-package with 7.9.0 of the stack. What is your preference, @mtojek?

I'm for the option that will be less painful for developers (elastic-package users), so let's bump it up to the 7.10.0-SNAPSHOT.

Anyway, I think is in the good shape to be merged. I left a comment about passing options, but you can postpone or ignore it :) LGTM!

mtojek:
@ycombinator let's get it merge and see in the wild how does it operate on real packages.

ycombinator:
> Anyway, I think is in the good shape to be merged.

Thanks. There are quite a few suggestions/ideas in this PR that need follow up PRs:
- Volume mounting: https://github.com/elastic/elastic-package/pull/64#discussion_r481066466
- Observing logs of Agent and underlying Beats while testing: https://github.com/elastic/elastic-package/pull/64#discussion_r480965636
- Calling the package installation API?: https://github.com/elastic/elastic-package/pull/64#discussion_r485172111
- Deleting old test assets before running a test: https://github.com/elastic/elastic-package/pull/64#discussion_r485200235

I will file issues for each of these after I merge this PR and update this comment with references.

mtojek:
Works for me! We can easily share them.

internal/compose/compose.go
ycombinator: Done in e0f21d7.

ycombinator:
Investigating the latest CI failure...

ycombinator:
Updated the health check for the Elastic Agent service in d1886e7 to check for the expected log entry across multiple log files, in case the logs had rotated before the health check started running. This should fix CI.

PR #70, opened 2020-08-20
Implement --version flag on cluster up,update subcommands

### LINKED ISSUE #57: Support multiple stack environments
I'm starting to get used to run `elastic-package cluster up` every time I need an Elastic Stack cluster. In most cases around packages, I'm interested in the 7.9 cluster, but sometimes I want to test against master. It would be nice if a flag could be passed to `cluster up` with the version to be used. Some example:

```
elastic-package up --stack=8.0.0-SNAPSHOT
elastic-package up --stack=7.10.0-SNAPSHOT
elastic-package up --stack=7.9.1
```

Param name is just a suggestion.

ycombinator:
This PR implements an optional `--version` string flag for the `elastic-package cluster up` and `elastic-package cluster update` sub-commands. It can take values such as `7.7.2`, `7.9.0`, `7.10.0-SNAPSHOT`, 8.0.0-SNAPSHOT`, etc. The default value of this flag, i.e. the default version of the Elastic cluster/stack that will be spun up is governed by this constant.

Resolves #57.


internal/cobraext/const.go
```diff
@@ -16,4 +16,7 @@ const (
 
 	VerboseFlagName        = "verbose"
 	VerboseFlagDescription = "verbose mode"
+
+	StackVersionFlagName    = "version"
```
mtojek, line 20: I wonder if we may need a different version somewhere, so it would safer to rename this argument to `--stack-version`.

internal/cluster/version.go
```diff
@@ -0,0 +1,5 @@
+package cluster
+
+const (
+	DefaultVersion = "7.9.0"
```
mtojek, line 4: DefaultStackVersion?

mtojek:
Linter complains about missing comment:

`[2020-08-20T23:42:25.140Z] /var/lib/jenkins/workspace/Beats_elastic-package_PR-70/src/github.com/elastic/elastic-package/internal/cluster/version.go:4:2: exported const DefaultVersion should have comment (or a comment on this block) or be unexported`

cmd/cluster.go
```diff
@@ -18,14 +18,20 @@ func setupClusterCommand() *cobra.Command {
...
 			if err != nil {
 				return errors.Wrap(err, "booting up the cluster failed")
 			}
 			return nil
 		},
 	}
 	upCommand.Flags().BoolP(cobraext.DaemonModeFlagName, "d", false, cobraext.DaemonModeFlagDescription)
+	upCommand.Flags().StringP(cobraext.StackVersionFlagName, "", cluster.DefaultVersion, cobraext.StackVersionDescription)
 
 	downCommand := &cobra.Command{
```
ruflin, line 36: I did not test this but I wonder if down works as expected if non default values for the stack version are used.

PR #71, opened 2020-08-21
Adding fake is_ready service to block on kibana health

### LINKED ISSUE #55: Proposal Indicate to user when `elastic-package cluster up` is ready?
At the moment, when someone runs `elastic-package cluster up` and the command returns, the cluster is not actually ready to use and it can take quite a bit long until Kibana is available. This might be confusing to the user. To debug, `docker ps` has to be used to check the health status.

My proposal would be to only return the up command when the cluster is ready to be worked with. In this test script I implemented a check: https://github.com/elastic/package-storage/blob/production/testing/main_integration_test.go#L54 This could be added to the up command.

ycombinator:
This PR adds a fake `is_ready` service that depends on the `kibana` service being up and healthy. This ensures that `elastic cluster up` does not return until all stack/cluster services are actually up and healthy.

Resolves #55.

### DIFF
--- a/internal/install/static_snapshot_yml.go
+++ b/internal/install/static_snapshot_yml.go
@@ -47,4 +47,10 @@ services:
       interval: 1s
     ports:
       - "127.0.0.1:8080:8080"
+
+  is_ready:
+    image: alpine:3.7
+    depends_on:
+      kibana:
+        condition: service_healthy
 `


internal/install/static_snapshot_yml.go
```diff
@@ -47,4 +47,10 @@ services:
       interval: 1s
     ports:
       - "127.0.0.1:8080:8080"
+
+  is_ready:
+    image: alpine:3.7
```
mtojek, line 52: I wonder if the `scratch` image can be applied here.

PR #73, opened 2020-08-21
Implement pipeline tests

mtojek:
This PR introduces pipeline test runner, which supports two kinds of test inputs: event and raw.

Test cases defined for `nginx` package:
https://github.com/mtojek/integrations/tree/enable-format-command/packages/nginx/dataset/access/_dev/test/pipeline

TODOs:
- [x] Read `event` test input
- [x] Read `raw` test input (with config)
- [x] Install appropriate pipeline
- [x] Simulate pipeline processing
- [x] Compare results withe expected ones
- [x] "Generate" flag

**Samples:**

Successful run

```bash
$ elastic-package test pipeline
Test case: test-access-event.json
Test case: test-access-raw.log
```

Failed run

```bash
elastic-package test pipeline
Test case: test-access-event.json
Expected results are different from current ones:
 {
     "expected": [
         {
             "@timestamp": "2016-12-07T10:04:37.000Z",
             "nginx": {
                 "access": {
                     "remote_ip_list": [
                         "127.0.0.1"
                     ]
                 }
             },
             "related": {
                 "ip": [
                     "127.0.0.1"
                 ]
             },
             "http": {
                 "request": {
                     "method": "get"
                 },
                 "version": "1.1",
                 "response": {
                     "body": {
                         "bytes": 571
                     },
-                    "status_code": 404
+                    "status_code": 123
                 }
             },
             "source": {
                 "address": "127.0.0.1",
                 "ip": "127.0.0.1"
             },
             "event": {
                 "category": [
                     "web"
                 ],
                 "type": [
                     "access"
                 ],
                 "created": "2016-10-25T12:49:34.000Z",
                 "kind": "event",
-                "outcome": "failure"
+                "outcome": "success"
             },
             "user_agent": {
                 "name": "Chrome",
                 "original": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.98 Safari/537.36",
                 "os": {
                     "name": "Mac OS X",
                     "version": "10.12.0",
                     "full": "Mac OS X 10.12.0"
                 },
                 "device": {
                     "name": "Mac"
                 },
                 "version": "54.0.2840.98"
             },
             "url": {
                 "original": "/test1"
             }
         },
         {
             "@timestamp": "2016-12-07T10:05:07.000Z",
             "nginx": {
                 "access": {
                     "remote_ip_list": [
                         "127.0.0.1"
                     ]
                 }
             },
             "related": {
                 "ip": [
                     "127.0.0.1"
                 ]
             },
             "http": {
                 "request": {
                     "method": "get"
                 },
                 "version": "1.1",
                 "response": {
                     "body": {
                         "bytes": 169
                     },
                     "status_code": 404
                 }
             },
             "source": {
                 "address": "127.0.0.1",
                 "ip": "127.0.0.1"
             },
             "event": {
                 "category": [
                     "web"
                 ],
                 "type": [
                     "access"
                 ],
                 "created": "2016-10-25T12:49:34.000Z",
                 "kind": "event",
                 "outcome": "failure"
             },
             "user_agent": {
                 "name": "Firefox",
                 "original": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:49.0) Gecko/20100101 Firefox/49.0",
                 "os": {
                     "name": "Mac OS X",
                     "version": "10.12",
                     "full": "Mac OS X 10.12"
                 },
                 "device": {
                     "name": "Mac"
                 },
                 "version": "49.0."
             },
             "url": {
                 "original": "/taga"
             }
         }
     ]
 }
Test case: test-access-raw.log
Error: error running package pipeline tests: at least one test case failed
```


internal/files/copy.go
```diff
@@ -32,3 +45,12 @@ func CopyAll(sourcePath, destinationPath string) error {
 			filepath.Join(sourcePath, relativePath))
 	})
 }
+
+func shouldDirectoryBeSkips(name string, skippedDirs []string) bool {
```
ycombinator, line 49: Typo: `shouldDirectoryBeSkipped`?

internal/testrunner/runners/pipeline/ingest_pipeline.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/elastic/go-elasticsearch/v7"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+
+	"github.com/elastic/elastic-package/internal/packages"
+)
+
+const defaultPipelineName = "default"
+
+var ingestPipelineTag = regexp.MustCompile("{{ IngestPipeline.+ }}")
```
ycombinator, line 24: Probably want to make this a bit more forgiving?

```suggestion
var ingestPipelineTag = regexp.MustCompile("{{\s*IngestPipeline.+}}")
```

internal/files/copy.go
mtojek: Fixed!

internal/testrunner/runners/pipeline/ingest_pipeline.go
mtojek: Agree, fixed.

internal/elasticsearch/client.go
```diff
@@ -0,0 +1,36 @@
...
+		Username:  username,
+		Password:  password,
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "creating Elasticsearch client failed")
+	}
+	return client, nil
+}
+
+func undefinedEnvError(envName string) error {
```
ycombinator, line 34: Nit: this function seems to something that would be useful generally when dealing with required environment variables, not something specific to the Elasticsearch client. So maybe consider moving it out of this file/package?

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	entryPipeline, pipelineIDs, err := installIngestPipelines(esClient, datasetPath)
+	if err != nil {
+		return errors.Wrap(err, "installing ingest pipelines failed")
+	}
+	defer func() {
+		err := uninstallIngestPipelines(esClient, pipelineIDs)
+		if err != nil {
+			fmt.Printf("uninstalling ingest pipelines failed: %v", err)
```
ycombinator, line 57: Should this use the logger?

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}()
+
+	var failed bool
+	for _, testCaseFile := range testCaseFiles {
+		tc, err := r.loadTestCaseFile(testCaseFile)
+		if err != nil {
+			return errors.Wrap(err, "loading test case failed")
+		}
+		fmt.Printf("Test case: %s\n", tc.name)
```
ycombinator, line 67: Should this use the logger?

internal/testrunner/runners/pipeline/test_result.go
```diff
@@ -0,0 +1,84 @@
...
+
+	expected, err := readExpectedTestResult(testCasePath)
+	if err != nil {
+		return errors.Wrap(err, "reading expected test result failed")
+	}
+
+	report := diff.Diff(string(expected), string(current))
+	if report != "" {
+		fmt.Println("Expected results are different from current ones:")
+		fmt.Println(report)
```
ycombinator, line 54: Should the first or both statements here use the logger?

internal/elasticsearch/client.go
mtojek: It looks like an ordinary helper function, but I wonder what would be the best place. `stack.UndefinedEnvError` ?

internal/install/install.go
```diff
@@ -44,7 +44,7 @@ func EnsureInstalled() error {
 		return errors.Wrap(err, "writing static resources failed")
 	}
 
-	fmt.Println("elastic-package has been installed.")
+	fmt.Fprintln(os.Stderr, "elastic-package has been installed.")
```
ycombinator, line 47: Why the change to `os.Stderr`? Also, should this use the logger?

internal/builder/packages.go
```diff
@@ -80,7 +80,7 @@ func buildPackage(sourcePath string) error {
 	}
 
 	fmt.Printf("Copy package content (source: %s)\n", sourcePath)
```
ycombinator, line 82: Should this use the logger?

internal/install/install.go
mtojek: No, in this case, the output is evalled from Stdout:

`eval "$(elastic-package stack shellinit)"`

 so this `fmt.Println` goes to Stderr. Another option would be replace it with `# comment`

internal/testrunner/runners/pipeline/runner.go
mtojek: It depends which parts do we prefer to keep as default messages/logs. You can take a look at log dumps in the issue description. Do you see something that can be delegated to a logger?

internal/install/install.go
mtojek: One more note on how did I find it. The eval failed only during the first run of a new build, whenever docker images have been updated (reinstallation notice).

ycombinator: Hmmm, I see. Feels a bit hacky, especially since all other output goes to STDOUT but I don't see a better option than what you've done (writing to STDERR) without complicating the code. I'm good with this change.

internal/testrunner/runners/pipeline/runner.go
ycombinator: I see your point. There's logs to help understand what the command is doing and then there's the actual end result output of the command. I would consider this statement here part of the command's output so I'm good with keeping this as-is.

internal/testrunner/runners/pipeline/runner.go
ycombinator: Based on our discussion in https://github.com/elastic/elastic-package/pull/73/files#r475905214, I would consider this to be a logging statement, not part of the output of the command.

internal/testrunner/runners/pipeline/test_result.go
ycombinator: Based on our discussion in https://github.com/elastic/elastic-package/pull/73/files#r475905214, I would consider this to be part of the output of the command, not about what the command is doing during it's execution. So I'm good with what you have here.

internal/builder/packages.go
ycombinator: Based on our discussion in https://github.com/elastic/elastic-package/pull/73/files#r475905214, I would consider this to be a logging statement, not part of the output of the command.

internal/elasticsearch/client.go
ycombinator: Yeah, not sure of the best place, looking at what we have at the moment. Let's just keep it here for now. We can extract it when we need it in another place. Maybe by then there will be a suitable package or we can just stick it in `stack` at the time.

internal/builder/packages.go
mtojek: Fixed

internal/testrunner/runners/pipeline/test_result.go
mtojek: Fixed

internal/testrunner/runners/pipeline/runner.go
mtojek: Fixed

internal/elasticsearch/client.go
mtojek: I adjusted all `fmt.Println` calls, but I think we need to change a bit the logger behavior. Take a look at this:

```
elastic-package test pipeline
Test case: test-access-event.json
2020/08/25 08:46:08  WARN Expected results are different from current ones:
2020/08/25 08:46:08  WARN  {
     "expected": [
         {
             "@timestamp": "2016-12-07T10:04:37.000Z",
```

The verbose logging is disabled, but it prints warnings. This is because the verbose logging is in fact a debug mode check. Should I change this condition to disable/enable all logger calls? In this case it would be:

```
elastic-package test pipeline
Test case: test-access-event.json
Test case: test-access-raw.log
Error: error running package pipeline tests: at least one test case failed
```

(actually you don't know which one failed)

mtojek:
I think the spec requires some fixes too. @ycombinator did I use the wrong directory for `_dev`?

```
elastic-package check
Error: checking package failed: found 1 validation error:
   1. item [_dev] is not allowed in folder [/Users/marcin.tojek/go/src/github.com/elastic/integrations/packages/nginx/dataset/access]
```

internal/testrunner/runners/pipeline/test_case.go
```diff
@@ -0,0 +1,103 @@
...
+func readRawInputEntries(inputData []byte, c testConfig) ([]string, error) {
+	var inputDataEntries []string
+
+	var builder strings.Builder
+	scanner := bufio.NewScanner(bytes.NewReader(inputData))
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		var body string
+		if c.Multiline != nil && c.Multiline.FirstLinePattern != "" {
```
ruflin, line 74: Where does it exactly take this pattern from? How will it work with more complex patterns supported by Agent / Filebeat?

ruflin: I see, the pattern comes from the testcase directory config.

ruflin (commented):
For the pipeline tests, its own line reader is used it with basic support for multiline (if I read this correctly). This is great to reduce the movable parts. One problem with this is, that there might be a different stream config for filebeat and when the package is actually shipped, it does not work but the tests worked. I think long term we unfortunately need to run the agent, load the config into the Agent and ship data to Elasticsearch to confirm config + pipeline + template work as expected. This will also generate a bit a different data output as it contains additional meta information from the agent which might be relevant for some pipelines.

An other part I mentioned above is the template. Not sure which mapping the tests use today, but having the correct mapping in place is important. It is possible that a mapping is not compatible with the ingested data and ingestion stops working. 

All the above should not block this PR from getting merged.

ruflin (approved):

mtojek:
> I think long term we unfortunately need to run the agent, load the config into the Agent and ship data to Elasticsearch to confirm config + pipeline + template work as expected. This will also generate a bit a different data output as it contains additional meta information from the agent which might be relevant for some pipelines.

Yes, there are pros and cons of this method. The cons you've mentioned, although this can be accomplished with system tests. On the other hand, these tests are really fast.

> An other part I mentioned above is the template. Not sure which mapping the tests use today, but having the correct mapping in place is important. It is possible that a mapping is not compatible with the ingested data and ingestion stops working.

I'm not aware of this part unfortunately, but I'm fine with adjusting this. Do I need to set something else?

internal/testrunner/runners/pipeline/ingest_pipeline.go
```diff
@@ -0,0 +1,219 @@
...
+
+	nonce := time.Now().UnixNano()
+
+	mainPipeline := getWithPipelineNameWithNonce(getPipelineNameOrDefault(datasetManifest), nonce)
+	pipelines, err := loadIngestPipelineFiles(datasetPath, nonce)
+	if err != nil {
+		return "", nil, errors.Wrap(err, "loading ingest pipeline files failed")
+	}
+
+	jsonPipelines, err := convertPipelineToJSON(pipelines)
```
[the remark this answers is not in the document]
mtojek: I suspect that the format conversion may introduce some difficulties, but this is the format supported by the library. I'm not sure if I can switch to JSON (if library internals allows for this).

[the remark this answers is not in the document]
ruflin: Probably also something that can be done by the system tests to load it as yaml to keep this part simple.

ruflin:
I'm a bit worried that we might implement multiline support also in the package here. At the same time as you said, these tests are efficient and get the job done. Lets wait until we hit the first more complex multiline logs or cases where additional meta data is used for the pipeline itself.

Good to know we will also partially cover this in the system tests. I like the idea of having some simple and perhaps not perfect but which allows quick iterations and have an end-to-end check in a different place. Will the system tests also generate some example output and validate it? Where will validation happen that for all fields mappings are specified?

For the mapping part, perhaps we should also solve this with the system tests instead. The way it works today is that as soon as the Beat is run, the templates are loaded so all tests run with the templates. In the case of the packages, all the `fields.yml` of a dataset would have to be combined and loaded as a template. Having the template there would cover cases where in the `fields.yml` `foo` is specified as a field but you try to ship `foo.bar` as part of the pipeline. But if you install a package, all this is done for you. If we want to keep this part here simple, lets also skip it but lets make sure we do it as part of the system tests where we also need Kibana etc.

mtojek:
> I'm a bit worried that we might implement multiline support also in the package here. At the same time as you said, these tests are efficient and get the job done. Lets wait until we hit the first more complex multiline logs or cases where additional meta data is used for the pipeline itself.

Agree. Additional fields can be setup via the configuration file:

```
{
    "multiline": {
        "first_line_pattern": "^(?:[0-9]{1,3}\\.){3}[0-9]{1,3}"
    },
    "fields": {
        "@timestamp": "2020-04-28T11:07:58.223Z",
        "ecs": {
            "version": "1.5.0"
        },
        "event.category": [
            "web"
        ]
    }
}
```

> Good to know we will also partially cover this in the system tests. I like the idea of having some simple and perhaps not perfect but which allows quick iterations and have an end-to-end check in a different place. Will the system tests also generate some example output and validate it? Where will validation happen that for all fields mappings are specified?

That's the question to @ycombinator rather, but I think so! It makes testing way easier.

> For the mapping part, perhaps we should also solve this with the system tests instead. The way it works today is that as soon as the Beat is run, the templates are loaded so all tests run with the templates. In the case of the packages, all the `fields.yml` of a dataset would have to be combined and loaded as a template. Having the template there would cover cases where in the `fields.yml` `foo` is specified as a field but you try to ship `foo.bar` as part of the pipeline. But if you install a package, all this is done for you. If we want to keep this part here simple, lets also skip it but lets make sure we do it as part of the system tests where we also need Kibana etc.

Hm.. now I see it. It's a kind of loop hole. On the other hand if the tool needs to install all fields and the pipeline, and so the index in Elasticsearch, it may perform this in a wrong way, inconsistent with what Kibana performs. I think I'm towards leaving this to be covered with system tests.

ycombinator:
> Good to know we will also partially cover this in the system tests. I like the idea of having some simple and perhaps not perfect but which allows quick iterations and have an end-to-end check in a different place. 

Yep, the intent of (ingest) pipeline tests is exactly what you said — to enable quick iteration when a package author is developing an ingest pipeline for their package. I _almost_ see it as a CLI for the Simulate Ingest Pipeline API but with the added ability to record input and expected output to catch regressions if the pipeline changes.

> Will the system tests also generate some example output and validate it? Where will validation happen that for all fields mappings are specified?

TBH, I haven't thought this far ahead yet. Initially the scope of the system tests is simply going to be to check if data is indexed into the data stream. We can certainly build on that further to check for expected structure of the data and that all fields that are in the indexed document are defined in `fields.yml`(s).

internal/testrunner/runners/pipeline/ingest_pipeline.go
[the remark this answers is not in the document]
ycombinator: System tests will not be directly loading any ingest pipelines. The will be relying on Kibana (Ingest Manager) to do this as it would in actual usage.

internal/packages/packages.go
```diff
@@ -19,6 +24,15 @@ type PackageManifest struct {
 	Version string `json:"version"`
 }
 
+// DatasetManifest represents the structure of a dataset's manifest
+type DatasetManifest struct {
```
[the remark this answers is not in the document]
ycombinator: Yes, this might happen once the spec starts to implement file contents' validation, the remaining item in https://github.com/elastic/package-spec/issues/4.

ycombinator:
> I think the spec requires some fixes too. @ycombinator did I use the wrong directory for `_dev`?

@mtojek indeed, we need to implement this proposal: https://github.com/elastic/package-spec/issues/22. Want to take a stab at it — would be good to get someone other than myself more familiar with the spec anyway?

internal/testrunner/runners/pipeline/test_result.go
ycombinator: @mtojek I think you misread my comment 🙂. I was good with what you had here before (`fmt.Println`).

mtojek: I apologize for the confusion. It's fixed back :)

mtojek:
> @mtojek indeed, we need to implement this proposal: elastic/package-spec#22. Want to take a stab at it — would be good to get someone other than myself more involved with package-spec development anyway?

Deal. I can try to move it forward. I think it doesn't block the current PR, pipelines tests won't be triggered as there is no `_dev` folder.

Issue #75, opened 2020-08-24
Add documentation


ycombinator:
Now that the `elastic-package` tool is gaining maturity in some capabilities/commands and is adding new capabilities/commands quite quickly, we should start adding appropriate documentation for the tool. In particular, I think we need documentation that covers these areas:
- An introduction to the tool, covering its use cases, capabilities, and commands.
- Reference documentation for each command. Ideally this will be available as part of the help text for each command, i.e. when `elastic-package help <command>` is run but also be available in the repo itself.
- Use case or HOWTO guides, e.g. how do I test my package?

mtojek:
@ycombinator I think you can mark some action items as done.

ycombinator:
Thanks @mtojek, I marked the first one as done. I think that's the only one for now.

ycombinator:
@mtojek I'd like to email the packages list talking about the testing features in `elastic-package`. Can you write the HOWTO for pipeline testing? Then we can include both HOWTOs (pipeline and system testing) in the email. Thanks!

mtojek:
It looks like there is one item remaining:

>  Reference documentation for each command. Ideally this will be available as part of the help text for each command, i.e. when elastic-package help <command> is run but also be available in the repo itself.

Currently the documentation for commands is limited to a couple of sentences (see: https://github.com/elastic/elastic-package#commands) without going deeper into explanation for switches and arguments. I would be for leaving it as is and
 adding a paragraph to  `elastic-package help <command>` (Markdown docs can get outdated really quickly). For instance, the `export` command may need more words to describe its actions:

```
elastic-package help export
Use export command to export assets relevant for the package from the Elastic stack.

Usage:
  elastic-package export [command]

Available Commands:
  dashboards  Export dashboards from Kibana

Flags:
  -h, --help   help for export

Global Flags:
  -v, --verbose   verbose mode

Use "elastic-package export [command] --help" for more information about a command.
``` 

@ycombinator would you mind sharing your idea for this bullet point?

ycombinator:
@mtojek Yes, we are mostly in agreement. My thinking was that we should flesh out the help text in the command help itself, which is exactly what you are thinking too:

>  I would be for leaving it as is and adding a paragraph to `elastic-package help <command>`

[...]

PR #76, opened 2020-08-27
Fleshing out README

ycombinator:
This PR fleshes out the README by adding a list of supported commands under `elastic-package` and short descriptions for each.

Preview of rendered README: https://github.com/ycombinator/elastic-package/blob/readme/README.md.

Resolves partially #75.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,16 @@ to specify a different version, including SNAPSHOT versions.
 _Context: global_
 
 
+### `elastic-package test`
+
+Use this command to run tests on a package. Currently, the only type of test that is supported are Pipeline Tests.
+
+#### Pipeline Tests
+
+These tests allow you to exercise any Ingest Node Pipelines defined by your packages.
+
+_Context: package_
+
 ### `elastic-package version`
 
 Use this command to print the version of `elastic-package` that you have installed. This is
--- a/README.md
+++ b/README.md
@@ -10,6 +10,13 @@ Currently, `elastic-package` only supports packages of type [Elastic Integration
 Download and build the latest master of `elastic-package` binary:
 
 ```bash
+git clone https://github.com/elastic/elastic-package.git
+make build
+```
+
+Alternatively, you may use `go get` but you will not be able to use the `elastic-package version` command.
+
+```bash
 go get github.com/elastic/elastic-package
 ```
 
--- a/README.md
+++ b/README.md
@@ -83,9 +83,11 @@ _Context: package_
 
 ### `elastic-package promote`
 
-Use this command to promote a package to the Snapshot Package Registry.
+Use this command to promote packages from one stage of the Package Registry to another.
 
-_Context: package_
+:warning: This command is intended primarily for use by administrators. 
+
+_Context: global_
 
 #### GitHub authorization
 
--- a/README.md
+++ b/README.md
@@ -97,7 +97,9 @@ file or passed via the `GITHUB_TOKEN` environment variable.
 Here are the instructions on how to create your own personal access token (PAT):
 https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token
 
-Make sure you have access to public repositories (to open pull requests) and to user data (check authorized account data).
+Make sure you have enabled the following scopes:
+* `public_repo` — to open pull requests on GitHub repositories.
+* `read:user` and `user:email` — to read your user profile information from GitHub in order to populate pull requests appropriately.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -30,8 +30,8 @@ elastic-package help
 
 `elastic-package` current offers the commands listed below. 
 
-Some commands have a global context, meaning that they can be run from anywhere and they will have the 
-same result. Other commands have a package context; these must be run from anywhere under a package's
+Some commands have a _global context_, meaning that they can be run from anywhere and they will have the 
+same result. Other commands have a _package context_; these must be run from anywhere under a package's
 root folder and they will operate on the contents of that package.
 
 For more details on a specific command, run `elastic-package help <command>`.
--- a/README.md
+++ b/README.md
@@ -53,8 +53,9 @@ _Context: global_
 
 ### `elastic-package build`
 
-Use this command to build a package. TODO: what is the purpose of building a package — when would
-a package developer want/need to run this command?
+Use this command to build a package. Built packages are stored in the `build/` folder located at the root folder of the local Git repository checkout that contains your package folder.
+
+Built packages are served up by the Elastic Package Registry running locally (see `elastic-package stack`). They can also be published to the `package-storage` repository.
 
 _Context: package_
 
--- a/README.md
+++ b/README.md
@@ -30,8 +30,8 @@ elastic-package help
 
 `elastic-package` currently offers the commands listed below. 
 
-Some commands have a _global context_, meaning that they can be run from anywhere and they will have the 
-same result. Other commands have a _package context_; these must be run from anywhere under a package's
+Some commands have a _global context_, meaning that they can be executed from anywhere and they will have the 
+same result. Other commands have a _package context_; these must be executed from somewhere under a package's
 root folder and they will operate on the contents of that package.
 
 For more details on a specific command, run `elastic-package help <command>`.
--- a/README.md
+++ b/README.md
@@ -50,12 +50,6 @@ description of what each command does.
 
 _Context: global_
 
-### `elastic-package version`
-
-Use this command to print the version of `elastic-package` that you have installed. This is
-especially useful when reporting bugs.
-
-_Context: global_
 
 ### `elastic-package build`
 
@@ -64,17 +58,6 @@ a package developer want/need to run this command?
 
 _Context: package_
 
-### `elastic-package lint`
-
-Use this command to validate the contents of a package.
-
-_Context: package_
-
-### `elastic-package format`
-
-Use this command to format the contents of a package.
-
-_Context: package_
 
 ### `elastic-package check`
 
@@ -82,12 +65,21 @@ Use this command to run the `format`, `lint`, and `build` commands all at once, 
 
 _Context: package_
 
-### `elastic-package stack`
 
-Use this command to spin up a Docker-based Elastic Stack consisting of Elasticsearch, Kibana, and 
-the Package Registry.
+### `elastic-package format`
 
-_Context: global_
+Use this command to format the contents of a package.
+
+_Context: package_
+
+
+### `elastic-package lint`
+
+Use this command to validate the contents of a package using the 
+[package specification](https://github.com/elastic/package-spec).
+
+_Context: package_
+
 
 ### `elastic-package promote`
 
@@ -107,6 +99,23 @@ https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-a
 Make sure you have enabled the following scopes:
 * `public_repo` — to open pull requests on GitHub repositories.
 * `read:user` and `user:email` — to read your user profile information from GitHub in order to populate pull requests appropriately.
+
+
+### `elastic-package stack`
+
+Use this command to spin up a Docker-based Elastic Stack consisting of Elasticsearch, Kibana, and 
+the Package Registry.
+
+_Context: global_
+
+
+### `elastic-package version`
+
+Use this command to print the version of `elastic-package` that you have installed. This is
+especially useful when reporting bugs.
+
+_Context: global_
+
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,18 +17,80 @@ Change directory to the package under development. Note: an integration is a spe
...
+
+Use this command to print the version of `elastic-package` that you have installed. This is
+especially useful when reporting bugs.
+
+_Context: global_
+
+### `elastic-package build`
+
+Use this command to build a package. TODO: what is the purpose of building a package — when would
+a package developer want/need to run this command?
```
ycombinator, line 56: @mtojek Could you help answer these TODO questions here, please? Thanks!

README.md
```diff
@@ -24,18 +17,80 @@ Change directory to the package under development. Note: an integration is a spe
...
+### `elastic-package stack`
+
+Use this command to spin up a Docker-based Elastic Stack consisting of Elasticsearch, Kibana, and 
+the Package Registry.
+
+_Context: global_
+
+### `elastic-package promote`
+
+Use this command to promote a package to the Snapshot Package Registry.
```
ycombinator, line 87: @mtojek Is this description accurate?

README.md
```diff
@@ -24,25 +17,89 @@ Change directory to the package under development. Note: an integration is a spe
...
 The tool uses the GitHub token to authorize user's call to API. The token can be stored in the `~/.elastic/github.token`
 file or passed via the `GITHUB_TOKEN` environment variable.
 
 Here are the instructions on how to create your own personal access token (PAT):
 https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token
 
-Make sure you have access to public repositories (to open pull requests) and to user data (check authorized account data).
+Make sure you have enabled the following scopes:
+* `public_repo` — to open pull requests on GitHub repositories.
+* `read:user` and `user:email` — to read your user profile information from GitHub in order to populate pull requests appropriately.
```
ycombinator, line 102: @mtojek Could you confirm that this is accurate, please? Thanks!

README.md
mtojek: The `build` command copies package internals to the build directory and encode Kibana dashboards. Now it will also skip `_dev` directory.

README.md
```diff
@@ -24,25 +17,89 @@ Change directory to the package under development. Note: an integration is a spe
...
+### `elastic-package build`
+
+Use this command to build a package. TODO: what is the purpose of building a package — when would
+a package developer want/need to run this command?
+
+_Context: package_
+
+### `elastic-package lint`
+
+Use this command to validate the contents of a package.
```
mtojek, line 62: ... against schema files defined by package-spec repository.

README.md
mtojek: The promotion from Integrations to package-storage happens automagically (Jenkins job).

The `promote` command exposes a package selection form in the TUI, which simplifies promotion between different stages (snapshot, staging, production) including opening relevant pull requests to the package-storage.

README.md
mtojek: I'm not convince if user:email is required, but let's keep it if we need this in the future.

README.md
```diff
@@ -24,25 +17,89 @@ Change directory to the package under development. Note: an integration is a spe
...
+
+### `elastic-package check`
+
+Use this command to run the `format`, `lint`, and `build` commands all at once, in that order.
+
+_Context: package_
+
+### `elastic-package stack`
+
+Use this command to spin up a Docker-based Elastic Stack consisting of Elasticsearch, Kibana, and
```
mtojek, line 80: typo: s/, and/ and/

README.md
```diff
@@ -24,25 +17,89 @@ Change directory to the package under development. Note: an integration is a spe
...
+### `elastic-package check`
+
+Use this command to run the `format`, `lint`, and `build` commands all at once, in that order.
+
+_Context: package_
+
+### `elastic-package stack`
+
+Use this command to spin up a Docker-based Elastic Stack consisting of Elasticsearch, Kibana, and 
+the Package Registry.
```
mtojek, line 81: It supports different versions of the stack?

README.md
```diff
@@ -1,16 +1,9 @@
...
-## Features
-
-TODO
-
-## Supported package types
-
-* [Elastic Integrations](https://github.com/elastic/integrations)
+Currently, `elastic-package` only supports packages of type [Elastic Integrations](https://github.com/elastic/integrations).
 
 ## Getting started
```
mtojek, line 9: If you are cleaning up the README, it might be a good option to refresh the Getting started as well. Currently, with `go get` we will loose the information about the version, so maybe it's better to replace it with `git clone` and `make build`.

We can also mention that the tool is go-gettable (but it will loose information about revision).

README.md
```diff
@@ -24,25 +17,89 @@ Change directory to the package under development. Note: an integration is a spe
...
 
-Run the `help` command and see available actions:
+Run the `help` command and see available commands:
 
 ```bash
 elastic-package help
 ```
 
-## GitHub authorization
+## Commands
```
mtojek, line 29: Do you think we can sort these commands and maybe keep help at the top of them?

mtojek, line 29: Did you miss the test command or is it intended to cover it later on once system tests are done?

mtojek (commented):
Ah, finally a decent README :) Thanks for writing it. I left a couple of comments on various aspects.

README.md
ycombinator: Thanks, I understand what it does internally. I was more wondering why a package developer would want or need to run this command? What use case does it support? Maybe it's supposed to be a pre-requisite for some other commands or something like that?

README.md
ycombinator: Ah yes, thanks for reminding me! I can fix the description. Is this command one we want to document or is it meant more for administrators only? We can also document it but say that it's not meant for general use or something like that.

README.md
ycombinator: No, this was intentional per:

image

README.md
ycombinator: Yes, generally I was trying to avoid mentioning all the flags for each command but I agree, this one is good to mention. Will add!

README.md
mtojek: No more comments, thank you for the explanation :D

README.md
ycombinator: Good point — I will clarify this in the README.

However, in the future (not for this PR), I hope we can have a single command to install `elastic-package` with all it's features. It seems weird from a user's perspective that the `version` command only works if you checkout the source and build from it (but I understand why that is the case currently). In the future, maybe we can think of committing a `version.go` file that's updated every so often via `make build` and then `elastic-package version` reads from this file or something like that.

README.md
ycombinator: I intended to add it when system tests are done but, now that you mention it, we have pipeline tests so maybe we should add it to the README now itself? Or would you prefer to wait until we have defined pipeline test files in a couple of integrations packages first?

ycombinator: Sorted in 51c497473dffecb4d6163351d23d6b2e5bcb5e25.

README.md
ycombinator: Updated Getting Started section in efe6362dd89696e60eb48eb4dd845a67581eea76.

README.md
ycombinator: Referenced the package spec in 51c497473dffecb4d6163351d23d6b2e5bcb5e25.

README.md
ycombinator: Added in 0d6b1e6.

ycombinator:
@mtojek Thanks for the review. I believe I've addressed all your feedback but some of the comments need more discussion. Please re-review when you get a chance. Thanks!

README.md
mtojek: I'm fine with both options, we can wait with announcing them officially.

README.md
mtojek: I'm ok with both options, let's wait with announcing them and silently observe in the wild first.

README.md
mtojek: Hm, I wouldn't keep it secret (do not mention in README), but it's valid point to add a notice about administrator use only.

README.md
```diff
@@ -64,30 +58,28 @@ a package developer want/need to run this command?
 
 _Context: package_
 
-### `elastic-package lint`
 
-Use this command to validate the contents of a package.
+### `elastic-package check`
+
+Use this command to run the `format`, `lint`, and `build` commands all at once, in that order.
```
mtojek, line 64: ... and the future the "test" subcommand.

I wonder if we should include "test" in the "check" chain as it requires the cluster to be up and running.

mtojek (approved):
Only few nit-picks. LGTM

README.md
ycombinator: I'm going to wait to add the `test` command here until https://github.com/elastic/integrations/pull/262 is merged. We can easily update the README at that time.

README.md
ycombinator: Sorry, but I didn't understand. Did you mean to post your answer to a different comment thread?

README.md
ycombinator: Let's keep the README about current state as much as possible. We can always update it easily as/when we make enhancements or changes.

I could go either way on including `test` in `check` but I slightly favor keeping it out. This way `check` can run relatively fast and can be run as an earlier stage in CI.

README.md
ycombinator: Sounds good, thanks. I will add a notice about administrator use only.

README.md
```diff
@@ -24,25 +24,99 @@ Change directory to the package under development. Note: an integration is a spe
...
+[package specification](https://github.com/elastic/package-spec).
+
+_Context: package_
+
+
+### `elastic-package promote`
+
+Use this command to promote a package to the Snapshot Package Registry.
+
+_Context: package_
```
ycombinator, line 88: @mtojek I think this should be `global`, right, since the tool lets the user select multiple packages at once?

mtojek: Correct, I missed this one. It should be `global`.

ycombinator: Fixed in 291401e.

README.md
ycombinator: Added in 291401e.

ycombinator:
Hey @mtojek thanks for the re-review. I've addressed the latest feedback; please take a look. The main item that I need your input on is https://github.com/elastic/elastic-package/pull/76#discussion_r478504388.

README.md
mtojek: > Sorry, but I didn't understand. Did you mean to post your answer to a different comment thread?

Yes, I had poor network connection. I (or JS) must have messed up sth with anchors. 

Anyway, this is good question. I agree with you that there is no direct goal of running `build` command. Currently built artifacts are used to:
1. Build docker image consisting of package-registry with custom packages.
2. Push changes from integrations to the package-storage.
3. Dry-run package-registry (I believe we should move towards package-spec, as that was an ultimate goal). @ruflin do you know if there are any features that block switching to spec validation?

README.md
mtojek: FYI PR is merged.

mtojek (approved):
Re-reviewed, nothing blocking here!

README.md
ruflin: Not now, but in the future I expect this to build the `.zip` file of the package. 

++ on switching to package spec validation and remove the validation code itself from the registry. Each validation part we remove we can check if it is also in the spec.

ruflin: Additional note: We should still have the validation in the package-registry but run it based on package-spec. So far some time we will probably do double validation but this should not be an issue.

README.md
ycombinator: Great! I will add `test` here now.

ycombinator: Added a section in bd93e6a. 

But I think we need to now define what a Pipeline Test is and how a package author can set it up. Let's do this in a follow up PR. I added a checklist item for this in https://github.com/elastic/elastic-package/issues/75. @mtojek Since you worked on the pipeline tests, can you take on this follow up PR?

README.md
ycombinator: Thanks for the clarifications @mtojek. 

> 1. Build docker image consisting of package-registry with custom packages.
> 2. Push changes from integrations to the package-storage.

I have added some text covering these two use cases.

> 3. Dry-run package-registry.

I don't understand what this means, how it's different from 1., and how locally-built packages help here, particularly from a package author's perspective. It sounds like this is an internal use, not something package authors need to care about. So I have not mentioned this in the README text for this command.

ycombinator:
I think, at this point, the contents of this README PR are already going to be helpful to package authors so I'm going to merge it.

There are a couple of rough edges that we might want to polish, but I think those can be done in follow up PRs (see https://github.com/elastic/elastic-package/pull/76#discussion_r480276789 and https://github.com/elastic/elastic-package/pull/76#discussion_r480273409).

README.md
mtojek: HOWTO guides is a good idea. I can write some docs about it.

Issue #72, opened 2020-08-21
stack command: start selected services only


mtojek:
Pipeline tests (see: https://github.com/elastic/elastic-package/issues/15) require only the Elasticsearch instance, so it's not necessary to boot up the entire stack.

Maybe the `stack up` subcommand should expose additional parameters like service names? e.g. `elasticsearch`, `kibana`, `package-registry`

mtojek:
@ycombinator @ruflin do you have any preferences?

ruflin:
+1 on not booting Kibana for these tests. I think it might be confusing the stack provides a flag to not start the stack.

There is also the part that for just the ingest pipeline testing you don't need security and probably other params. I would put it separate for now.

mtojek:
> +1 on not booting Kibana for these tests. I think it might be confusing the stack provides a flag to not start the stack.

It depends on how do you look at it. We can say the `stack up` subcommand allows for selection of stack components.

mtojek:
I will need this one before enabling it in the Integrations repository. Otherwise, people might feel confused why do they need to spin up Kibana for `mage check` if they are not gonna use it.

PR #77, opened 2020-08-27
Select services to boot

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/72

Testing:
```
$ elastic-package stack up --services elasticsearch,package-registry
```

### DIFF
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -77,7 +77,9 @@ func setupStackCommand() *cobra.Command {
 				return cobraext.FlagParsingError(err, cobraext.StackVersionFlagName)
 			}
 
-			err = stack.Update(stackVersion)
+			err = stack.Update(stack.BootOptions{
+				StackVersion: stackVersion,
+			})
 			if err != nil {
 				return errors.Wrap(err, "failed updating the stack images")
 			}
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -117,9 +117,11 @@ func setupStackCommand() *cobra.Command {
 }
 
 func availableServicesAsList() []string {
-	var available []string
+	available := make([]string, len(availableServices))
+	i := 0
 	for aService := range availableServices {
-		available = append(available, aService)
+		available[i] = aService
+		i++
 	}
 	return available
 }
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -10,6 +10,12 @@ import (
 	"github.com/elastic/elastic-package/internal/stack"
 )
 
+var availableServices = map[string]bool{
+	"elasticsearch":    true,
+	"kibana":           true,
+	"package-registry": true,
+}
+
 func setupStackCommand() *cobra.Command {
 	upCommand := &cobra.Command{
 		Use:   "up",
@@ -20,12 +26,26 @@ func setupStackCommand() *cobra.Command {
 				return cobraext.FlagParsingError(err, cobraext.DaemonModeFlagName)
 			}
 
+			services, err := cmd.Flags().GetStringSlice(cobraext.StackServicesFlagName)
+			if err != nil {
+				return cobraext.FlagParsingError(err, cobraext.StackServicesFlagName)
+			}
+
+			err = validateServicesFlag(services)
+			if err != nil {
+				return errors.Wrap(err, "validating services failed")
+			}
+
 			stackVersion, err := cmd.Flags().GetString(cobraext.StackVersionFlagName)
 			if err != nil {
 				return cobraext.FlagParsingError(err, cobraext.StackVersionFlagName)
 			}
 
-			err = stack.BootUp(daemonMode, stackVersion)
+			err = stack.BootUp(stack.BootOptions{
+				DaemonMode:   daemonMode,
+				StackVersion: stackVersion,
+				Services:     services,
+			})
 			if err != nil {
 				return errors.Wrap(err, "booting up the stack failed")
 			}
@@ -33,6 +53,7 @@ func setupStackCommand() *cobra.Command {
 		},
 	}
 	upCommand.Flags().BoolP(cobraext.DaemonModeFlagName, "d", false, cobraext.DaemonModeFlagDescription)
+	upCommand.Flags().StringSliceP(cobraext.StackServicesFlagName, "s", nil, cobraext.StackServicesFlagDescription)
 	upCommand.Flags().StringP(cobraext.StackVersionFlagName, "", stack.DefaultVersion, cobraext.StackVersionDescription)
 
 	downCommand := &cobra.Command{
@@ -89,4 +110,21 @@ func setupStackCommand() *cobra.Command {
 		updateCommand,
 		shellInitCommand)
 	return cmd
+}
+
+func validateServicesFlag(services []string) error {
+	selected := map[string]bool{}
+
+	for _, aService := range services {
+		if _, found := availableServices[aService]; !found {
+			return fmt.Errorf("service \"%s\" is not available", aService)
+		}
+
+		if _, found := selected[aService]; found {
+			return fmt.Errorf("service \"%s\" must be selected at most once", aService)
+		}
+
+		selected[aService] = true
+	}
+	return nil
 }
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -113,7 +113,7 @@ func setupStackCommand() *cobra.Command {
 }
 
 func validateServicesFlag(services []string) error {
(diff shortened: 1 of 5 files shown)


ycombinator:
For specifying multiple datasets as part of the `test` command we used `--dataset` (singular). Should we change that to `--datasets` (plural) to match `--services` here OR should we change `--services` to `--service` here?

cmd/stack.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/elastic/elastic-package/internal/stack"
 )
 
+var availableServices = map[string]bool{
+	"elasticsearch":    true,
+	"kibana":           true,
+	"package-registry": true,
```
ycombinator, line 16: Nit: Neat trick I learnt from Steffen: for such "existence lookup" maps, we can use `map[string]struct{}` as `struct{}` takes up 0 bytes whereas `bool` takes up 1 byte 😅.

cmd/stack.go
```diff
@@ -20,19 +26,34 @@ func setupStackCommand() *cobra.Command {
 				return cobraext.FlagParsingError(err, cobraext.DaemonModeFlagName)
 			}
 
+			services, err := cmd.Flags().GetStringSlice(cobraext.StackServicesFlagName)
```
ycombinator, line 29: Ah, didn't know about `GetStringSlice` - nice! Mind converting the parsing of the `--dataset` flag in the `test` command to use this same mechanism?

internal/stack/boot.go
```diff
@@ -128,12 +137,18 @@ func dockerComposeUp(daemonMode bool, stackVersion string) error {
...
 	}
 
-	if daemonMode {
+	if options.DaemonMode {
 		args = append(args, "-d")
 	}
 
+	if len(options.Services) > 0 {
+		for _, aService := range options.Services {
+			args = append(args, aService, fmt.Sprintf("%s_is_ready", aService))
```
ycombinator, line 146: Nit: you can just append `fmt.Sprintf("%s_is_ready", aService)`. Because of the dependency, `aService` will automatically be spun up.

mtojek: I can't, I checked it :) 

With single `_is_ready` service (e.g. elasticsearch_is_ready), the docker-compose will exit immediately when the "true"-application stops, so you can't achieve an effect of tailing live logs. It will behave exactly like in daemon mode.

ycombinator: Ah, I see. Good to know — thanks!

cmd/stack.go
mtojek: Fixed!

cmd/stack.go
mtojek: You're right, let's switch to this :)

mtojek:
> For specifying multiple datasets as part of the test command we used --dataset (singular). Should we change that to --datasets (plural) to match --services here OR should we change --services to --service here?

I renamed both, dataset and service, to plural forms.

ycombinator:
I was just testing this PR with `--services elasticsearch`. I ran `docker system prune -af` first to make sure I had a clean slate.

```
$ elastic-package stack up --services elasticsearch
elasticsearch uses an image, skipping
elasticsearch_is_ready uses an image, skipping
kibana uses an image, skipping
kibana_is_ready uses an image, skipping
package-registry_is_ready uses an image, skipping
Building package-registry
Step 1/3 : FROM docker.elastic.co/package-registry/distribution:snapshot
snapshot: Pulling from package-registry/distribution
524b0c1e57f8: Pull complete
e96e4b939e26: Pull complete
9815b89d68f4: Pull complete
e5885ed9db3a: Pull complete
d9caa98fa2de: Pull complete
7e79611347e7: Pull complete
516042a87580: Pull complete
61a01535d213: Pull complete
Digest: sha256:6e65e0f1dc7a6305d72814a86385ec71b8e631985c2ca48ecff4e602f441c9c1
Status: Downloaded newer image for docker.elastic.co/package-registry/distribution:snapshot
 ---> 2362c9b35f6e
Step 2/3 : COPY ${STACK_DIR}/package-registry.config.yml /package-registry/config.yml
 ---> 1f27af0c4d89
Step 3/3 : COPY ${STACK_DIR}/development/ /packages/development
 ---> 3637f360bea5

Successfully built 3637f360bea5
Successfully tagged stack_package-registry:latest
Removing network stack_default
WARNING: Network stack_default not found.
Creating network "stack_default" with the default driver
Pulling elasticsearch (docker.elastic.co/elasticsearch/elasticsearch:7.9.0)...
7.9.0: Pulling from elasticsearch/elasticsearch
f1feca467797: Pull complete
f1962334d09b: Pull complete
c3f3fd33311b: Pull complete
c01dc5464de5: Pull complete
c7c56d706a5f: Pull complete
14e364ba972c: Pull complete
8a4689df54ab: Pull complete
daa74c1eb347: Pull complete
bb1b1c773e9d: Pull complete
Digest: sha256:221c7f4a1ed3ed84ea7c1b28fd9fcb4c18e2d9370868da4b87cb78ba82f4d14e
Status: Downloaded newer image for docker.elastic.co/elasticsearch/elasticsearch:7.9.0
Pulling elasticsearch_is_ready (tianon/true:)...
latest: Pulling from tianon/true
c53fb220cbad: Pull complete
Digest: sha256:009cce421096698832595ce039aa13fa44327d96beedb84282a69d3dbcf5a81b
Status: Downloaded newer image for tianon/true:latest
Creating stack_elasticsearch_1 ... done
Creating stack_elasticsearch_is_ready_1 ... done
Attaching to stack_elasticsearch_1, stack_elasticsearch_is_ready_1
...
```

```
$ docker ps -a
CONTAINER ID        IMAGE                                                 COMMAND                  CREATED             STATUS                      PORTS                                NAMES
ca9a4209992c        tianon/true                                           "/true"                  21 seconds ago      Exited (0) 20 seconds ago                                        stack_elasticsearch_is_ready_1
2198e276b702        docker.elastic.co/elasticsearch/elasticsearch:7.9.0   "/tini -- /usr/local…"   45 seconds ago      Up 44 seconds (healthy)     127.0.0.1:9200->9200/tcp, 9300/tcp   stack_elasticsearch_1
```

```
$ docker images
REPOSITORY                                        TAG                 IMAGE ID            CREATED              SIZE
stack_package-registry                            latest              3637f360bea5        About a minute ago   342MB
docker.elastic.co/package-registry/distribution   snapshot            2362c9b35f6e        2 hours ago          342MB
docker.elastic.co/elasticsearch/elasticsearch     7.9.0               0eddd30d8bdf        2 weeks ago          762MB
tianon/true                                       latest              183cb5fd5414        22 months ago        125B
```

Notice that even though I only asked for the `elasticsearch` service to be started, the package registry image was also pulled down and the local package-registry Dockerfile was built. Is this expected? Can it be avoided unless the `package-registry` service is also supposed to be started up?

mtojek:
I admit I had doubts whether should I resolve it or leave it, but as you spotted it and commented, I introduced a simple service dependency adjustment for Kibana.

Fixed.

internal/cobraext/const.go
```diff
@@ -20,6 +20,9 @@ const (
 	VerboseFlagName        = "verbose"
 	VerboseFlagDescription = "verbose mode"
 
+	StackServicesFlagName        = "services"
+	StackServicesFlagDescription = "component services (comma-separated values: elasticsearch, kibana, package-registry)"
```
ycombinator, line 24: How do you feel about introducing a `%s` placeholder instead of hardcoding the service names here and then interpolating it with `availableServices` where this constant is used?

mtojek: Fixed

cmd/stack.go
```diff
@@ -90,3 +115,28 @@ func setupStackCommand() *cobra.Command {
 		shellInitCommand)
 	return cmd
 }
+
+func availableServicesAsList() []string {
+	var available []string
```
ycombinator, line 120: Nit: Not that it matters much here because we are only dealing with a handful of `availableServices` but in general, when the size of the slice can be determined beforehand you can preallocate the memory for it by doing:

```
available := make([]string, len(availableServices)
```

And then you don't even need the `append` because you can do something like:

```
i := 0
for aService := range availableServices {
  available[i] = aService
  i++
}
```

But again, this is just to show a _general_ pattern when converting the keys of a `map[string]interface{}` to a `[]string`. What you have here is totally fine since we are only dealing with a handful of keys.

mtojek: You're right, we can optimize it here :)

Fixed.

Issue #8, opened 2020-06-05
Command: install the integration


mtojek:
The command installs the integration on the agent.

Suggested command: `elastic-package install`

mtojek:
@ycombinator I suppose you will have to install the integration somehow. I would be cool if you manage to cover this subcommand.

ycombinator:
Yes, I will need to handle this capability for system tests anyway so I'm happy to take this issue. Thanks!

PR #80, opened 2020-08-28
Refactoring: extracting out docker compose API

ycombinator:
While working on https://github.com/elastic/elastic-package/pull/64 as well as thinking about https://github.com/elastic/elastic-package/issues/8, I realized it might be better to extract the Docker Compose Golang API out of the `stack` package. This PR does that.

### DIFF
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -18,11 +18,8 @@ type Project struct {
 func NewProject(name string, paths ...string) (*Project, error) {
 	for _, path := range paths {
 		info, err := os.Stat(path)
-		if err != nil && os.IsNotExist(err) {
-			return nil, errors.Wrapf(err, "could not find Docker Compose configuration file: %s", path)
-		}
 		if err != nil {
-			return nil, errors.Wrapf(err, "error finding Docker Compose configuration file: %s", path)
+			return nil, errors.Wrapf(err, "could not find Docker Compose configuration file: %s", path)
 		}
 
 		if info.IsDir() {
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -14,6 +14,12 @@ type Project struct {
 	composeFilePaths []string
 }
 
+type CommandOptions struct {
+	Env       []string
+	ExtraArgs []string
+	Services  []string
+}
+
 // NewProject creates a new Docker Compose project given a sequence of Docker Compose configuration files.
 func NewProject(name string, paths ...string) (*Project, error) {
 	for _, path := range paths {
@@ -36,13 +42,13 @@ func NewProject(name string, paths ...string) (*Project, error) {
 }
 
 // Up brings up a Docker Compose project.
-func (p *Project) Up(extraArgs, env []string, services ...string) error {
+func (p *Project) Up(opts CommandOptions) error {
 	args := p.baseArgs()
 	args = append(args, "up")
-	args = append(args, extraArgs...)
-	args = append(args, services...)
+	args = append(args, opts.ExtraArgs...)
+	args = append(args, opts.Services...)
 
-	if err := runDockerComposeCmd(args, env); err != nil {
+	if err := runDockerComposeCmd(args, opts.Env); err != nil {
 		return errors.Wrap(err, "running Docker Compose up command failed")
 	}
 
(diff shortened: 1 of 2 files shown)


internal/stack/boot.go
```diff
@@ -123,21 +118,19 @@ func dockerComposeUp(daemonMode bool, stackVersion string) error {
 		return errors.Wrap(err, "locating stack directory failed")
 	}
 
-	args := []string{
-		"-f", filepath.Join(stackDir, "snapshot.yml"),
-		"up",
+	c, err := compose.NewProject(dockerComposeProjectName, filepath.Join(stackDir, "snapshot.yml"))
+	if err != nil {
+		return errors.Wrap(err, "could not create docker compose project")
```
mtojek, line 123: (nit, just note about polishing)

It would be cool to unify naming: docker compose vs docker-compose vs Docker Compose (not sure which is, officially, the correct one).

internal/stack/boot.go
```diff
@@ -149,17 +142,16 @@ func dockerComposeDown() error {
...
 	// We set the STACK_VERSION env var here to avoid showing a warning to the user about
 	// it not being set.
-	cmd.Env = append(os.Environ(), fmt.Sprintf("STACK_VERSION=%s", DefaultVersion))
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
-	err = cmd.Run()
-	if err != nil {
+	env := []string{fmt.Sprintf("STACK_VERSION=%s", DefaultVersion)}
+
+	if err := c.Down(nil, env); err != nil {
```
mtojek, line 154: Same here, you can get rid of passing the "nil" value.

mtojek (commented):
I'm generally +1 with the refactoring, although, I suppose, the source code will change a bit after merge from master (I touched `stack` a bit).

I wonder if there is any docker compose wrapping library to nicely interact with API (don't spend time on this, just thinking loud).

internal/stack/boot.go
ycombinator: Looking at https://docs.docker.com/compose/, it appears the name of the project is "Docker Compose". The CLI tool is called `docker-compose`. Since we are referring more to the project here, I will update all references to use "Docker Compose".

ycombinator: Updated in 0882065.

internal/compose/compose.go
```diff
@@ -0,0 +1,113 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/pkg/errors"
+)
+
+// Project represents a Docker Compose project.
+type Project struct {
+	name  string
+	paths []string
```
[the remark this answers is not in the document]
ycombinator: Fixed in b7edb15fe32dc5e04bb624a5f76d2a50d3e952f2.

internal/compose/compose.go
```diff
@@ -0,0 +1,113 @@
...
+// NewProject creates a new Docker Compose project given a sequence of Docker Compose configuration files.
+func NewProject(name string, paths ...string) (*Project, error) {
+	for _, path := range paths {
+		info, err := os.Stat(path)
+		if err != nil && os.IsNotExist(err) {
+			return nil, errors.Wrapf(err, "could not find docker-compose configuration file: %s", path)
+		}
+		if err != nil {
+			return nil, errors.Wrapf(err, "error finding docker-compose configuration file: %s", path)
+		}
```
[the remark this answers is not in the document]
ycombinator: Fixed in db4c7aa.

internal/compose/compose.go
```diff
@@ -0,0 +1,113 @@
...
+	c := Project{
+		name,
+		paths,
+	}
+
+	return &c, nil
+}
+
+// Up brings up a Docker Compose project.
+func (p *Project) Up(extraArgs, env []string, services ...string) error {
```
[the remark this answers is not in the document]
ycombinator: Fixed in 48f0a53.

internal/stack/boot.go
```diff
@@ -79,17 +80,14 @@ func dockerComposeBuild(stackVersion string) error {
...
-	cmd := exec.Command("docker-compose", args...)
-	cmd.Env = append(os.Environ(), fmt.Sprintf("STACK_VERSION=%s", stackVersion))
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
 
-	err = cmd.Run()
-	if err != nil {
+	env := []string{fmt.Sprintf("STACK_VERSION=%s", stackVersion)}
+
+	if err := c.Build(nil, env, "package-registry"); err != nil {
```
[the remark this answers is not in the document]
ycombinator: Fixed in 48f0a53.

internal/stack/boot.go
ycombinator: Fixed in 48f0a53.

ycombinator:
@mtojek Thanks for the review. I've addressed all your feedback. Please re-review when you get a chance!

ycombinator:
> I wonder if there is any docker compose wrapping library to nicely interact with API (don't spend time on this, just thinking loud).

Yes, this was my first thought too and so I looked before putting up this PR. I found https://github.com/docker/libcompose but it is no longer maintained. Besides that I have not found any other one. My thinking was that by doing this refactoring here, we can more easily replace it with a 3rd party library if/when one becomes available.

mtojek (approved):

PR #82, opened 2020-08-31
Fix: daemon bug

mtojek:
This PR fixes bug in the order of command args of `docker-compose up`


mtojek:
I need this one, so I will self-approve it.

PR #91, opened 2020-09-09
Update package-spec to latest version

mtojek:
This PR updates dependency on package-spec to the latest version. I need to include manifest file validation.


mtojek:
self-approved. Need this one for integrations.

ycombinator (commented):
Posthumous LGTM.

PR #92, opened 2020-09-09

mtojek:
Update dependency on package-spec to include additional field type.


ycombinator (approved):
LGTM. WFG.

PR #99, opened 2020-09-16
[System test runner] Assign to the deployed service context

ycombinator:
This PR fixes a subtle bug where we were setting the `STDOUT` and `STDERR` fields on the context passed in to a `ServiceDeployer`'s `SetUp` method. Instead we should be setting these on the context of the `DeployedService` returned from this method.

### DIFF
--- a/internal/testrunner/runners/system/servicedeployer/compose.go
+++ b/internal/testrunner/runners/system/servicedeployer/compose.go
@@ -87,7 +87,7 @@ func (r *DockerComposeServiceDeployer) SetUp(ctxt ServiceContext) (DeployedServi
 		return nil, errors.Wrap(err, "could not create STDOUT file")
 	}
 	service.stdout = outFile
-	ctxt.STDOUT = agentLogsFolder + stdoutFileName
+	service.ctxt.STDOUT = agentLogsFolder + stdoutFileName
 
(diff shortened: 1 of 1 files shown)


internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -87,7 +87,7 @@ func (r *DockerComposeServiceDeployer) SetUp(ctxt ServiceContext) (DeployedServi
 		return nil, errors.Wrap(err, "could not create STDOUT file")
 	}
 	service.stdout = outFile
-	ctxt.STDOUT = agentLogsFolder + stdoutFileName
+	service.ctxt.STDOUT = agentLogsFolder + stdoutFileName
```
mtojek, line 90: In general we shouldn't modify the context as service property, but rather build the new context first and then assign to the service.

Let's keep it here for now as we're in the middle of debugging the issue.

mtojek (approved):
LGTM

mtojek:
Let me merge this one.

internal/testrunner/runners/system/servicedeployer/compose.go
ycombinator: https://github.com/elastic/elastic-package/pull/101

PR #100, opened 2020-09-16
Fix Kibana Fleet URL

ycombinator:
This PR fixes the Kibana Fleet URL in the `kibana.config.yml` file used by the Kibana Docker container that's booted up by `elastic-package stack up`. This URL value is passed to the Elastic Agent container so the hostname in the URL needs to be `kibana` and not `localhost`.


mtojek (approved):
LGTM

mtojek:
Let me merge this change.

Issue #85, opened 2020-09-08
[System test runner] Use volume mounts for logs


ycombinator:
Follow up to https://github.com/elastic/elastic-package/pull/64, based on suggestion in https://github.com/elastic/elastic-package/pull/64#discussion_r481066466:

> I wonder if we can simplify this - we still have a volume with log files mounted, so maybe don't depend specifically on stdout and stderr, but just share a volume between agent and the service. I'm fine with discussing this in a different issue.

mtojek:
@ycombinator it would beneficial for this task to finalize the PR https://github.com/elastic/integrations/pull/263 first.

mtojek:
I did an investigation and tried different options.

Problem: share all logs between Agent and the service under test (SUT).

Requirements:
1. Long running Elastic stack.
2. No parallel tests (at least at the moment).
3. Logs cleanup after SUT shutdown.

Options:

1. Create external volume (`docker volumes create`?).

Before starting the Elastic stack we can create the volume and mount it in Agent and SUT containers during startup.
Cleanup might be hard to perform as we need to sneak into the volume and wipe the content.

2. Boot up SUT as part of the project (`docker-compose up`).

SUT can share the volume created by the previous boot of the Elastic stack.
The problem with cleanup remains - we need to sneak into.

3. Use `tmp/service_logs ` system directory.

The temporary directory is bound to the normal filesystem path. The directory can be shared between Elastic Agent, SUT and test runner (!).
The test runner  can clean the service logs after test execution.

ycombinator:
I'm also in favor of option 3. It is in the same general mold as what we are currently doing, except it uses volumes for sharing the services log files between containers instead of stream redirection.

PR #105, opened 2020-09-17
User volumes for service logs

mtojek:
Closes: https://github.com/elastic/elastic-package/issues/85

I'm opening as draft, because we need to have a service that writes logs to files.

I introduced also service logs direcotyr env: `ELASTIC_PACKAGE_SERVICE_LOGS_DIR`


internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -76,47 +70,10 @@ func (r *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
...
-		return nil, errors.Wrap(err, "could not redirect service container STDOUT and STDERR streams")
-	}
-
-	stackNetwork := fmt.Sprintf("%s_default", stack.DockerComposeProjectName)
-	logger.Debugf("attaching service container %s to stack network %s", serviceContainer, stackNetwork)
-
-	cmd = exec.Command("docker", "network", "connect", stackNetwork, serviceContainer)
-	if err := cmd.Run(); err != nil {
-		return nil, errors.Wrap(err, "could not attach service container to stack network")
-	}
```
ycombinator, line 116: I think we still need to attach the service container to the stack network for metrics data streams to work, no?

mtojek: I forgot about this requirement. I will bring it back.

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -76,47 +70,10 @@ func (r *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
...
-	localLogsFolder := inCtxt.Logs.Folder.Local
-	agentLogsFolder := inCtxt.Logs.Folder.Agent
-
-	service.stdoutFilePath = filepath.Join(localLogsFolder, stdoutFileName)
-	logger.Debugf("creating temp file %s to hold service container %s STDOUT", service.stdoutFilePath, serviceContainer)
-	outFile, err := os.Create(service.stdoutFilePath)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not create STDOUT file")
-	}
-	service.stdout = outFile
```
ycombinator, line 89: This is how a Docker-based service's STDOUT stream was being written to a log file on the local filesystem. How are we capturing a Docker-based service's logs now? 

Related: what would be the syntax in a package's test configuration file to refer to such logs? In the current implementation the syntax is `{{ STDOUT }}` and `{{ STDERR }}`.

mtojek: I considered this as a base assumption for this change - to base only on log file resources, as most apps emit them as two log streams (application, access). I think it simplified the implementation as we don't have to deal with opening and closing file.

In case of stdout, stderr we can "tee" these streams to local files.

I don't insist on such design, just wanted to handle all streams the same way (via filesystem).

ycombinator: Okay, let me try to understand this with the `apache` Docker container. If you look at https://github.com/elastic/integrations/pull/263, the system test configuration file for the `apache/access` log dataset (for the current implementation of the system test runner) looks like this:

https://github.com/ycombinator/integrations/blob/124ce5912eaf010352c7054bbc0c5ec3e84664c4/packages/apache/dataset/access/_dev/test/system/config.yml

How do you see this changing in the new implementation (this PR)?

mtojek: We know in advance that service logs are available in the directory `/tmp/service_logs/` in the agent container. 

I suppose the variable `var.paths` can contain `/tmp/service_logs/access.log*`?

mtojek: Fixed. Additional context properties  are exposed: `LogsFolderLocal` and `LogsFolderAgent`.

internal/testrunner/runners/system/servicedeployer/compose.go
mtojek: Fixed

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -76,47 +77,17 @@ func (r *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
...
-	cmd.Stdout = service.stdout
-	cmd.Stderr = service.stderr
-
-	if err := cmd.Start(); err != nil {
-		return nil, errors.Wrap(err, "could not redirect service container STDOUT and STDERR streams")
-	}
-
+	// Connect service network with stack network (for the purpose of metrics collection)
 	stackNetwork := fmt.Sprintf("%s_default", stack.DockerComposeProjectName)
-	logger.Debugf("attaching service container %s to stack network %s", serviceContainer, stackNetwork)
```
ycombinator, line 111: Maybe bring back the debug logs so we can tell what's going on while troubleshooting?

internal/testrunner/runners/system/servicedeployer/compose.go
ycombinator: Thanks, so my understanding is that the system test configuration file for the `apache/access` log dataset would look like:

```
vars: ~
dataset:
  vars:
    paths:
      - "{{LogsFolderAgent}}/access.log*"
```

And `httpd.conf` would be configured as you showed in your last comment above.

But why would a package author just know that `logs/access_log` from their `httpd.conf` should be referred to as `{{LogsFolderAgent}}/access_log*`? I'm guessing this comes from the volume mount definitions that they would define in the docker-compose file for their service? Perhaps this last piece would be clear to me if you can make a PR to https://github.com/elastic/integrations/pull/263 showing the necessary changes.

internal/testrunner/runners/system/servicedeployer/context.go
```diff
@@ -19,27 +19,12 @@ type ServiceContext struct {
...
-			// Local contains the folder path where log files produced by
-			// the service are stored on the local filesystem, i.e. where
-			// elastic-package is running.
-			Local string
-
-			// Agent contains the folder path where log files produced by
-			// the service are stored on the Agent container's filesystem.
-			Agent string
-		}
-	}
```
ycombinator, line 34: Any particular reason not to keep the same hierarchical structure as before? Just curious if the hierarchy vs. flat is a personal preference or if there's a stronger reason you prefer flat?

internal/testrunner/runners/system/servicedeployer/compose.go
mtojek: > Thanks, so my understanding is that the system test configuration file for the apache/access log dataset would look like:

Correct.

> And httpd.conf would be configured as you showed in your last comment above.

Yes

> But why would a package author just know that `logs/access_log` from their `httpd.conf` should be referred to as `{{LogsFolderAgent}}/access_log*`?

Frankly speaking I considered this as part of standard stack setup - the Elastic Agent has a directory bound to service logs of the service under tests (SUT). As the agent is a long running instance and it's booted up first, it needs to create the volume (no the SUT). The local directory with logs can customized on the SUT side:

```
version: '2.3'

services:
  apache:
    image: docker.elastic.co/integrations-ci/beats-apache:${SERVICE_VERSION:-2.4.20}-1
    ports:
      - 80
    volumes:
      - ${ELASTIC_PACKAGE_SERVICE_LOGS_DIR}:/usr/local/apache2/logs
``` 

To be honest I had in mind to document both `{{LogsFolderAgent}}` and `${ELASTIC_PACKAGE_SERVICE_LOGS_DIR}` in some README or guide.

I'm willing to adjust it if you have an idea on how we can simplify it.

internal/testrunner/runners/system/servicedeployer/context.go
mtojek: I kept it flat, because we've only two configuration options. I adjusted it back so it's again structured.

internal/testrunner/runners/system/servicedeployer/compose.go
mtojek: Fixed, thanks!

internal/testrunner/runners/system/servicedeployer/compose.go
ycombinator: Okay, I understand the complete picture now. Thanks.

It is slightly annoying that these two variables expose the inner workings of how logs get from a service container to the agent container. It would be nice if a user didn't need to know about that detail. What if we make both variables (from a user's perspective) the same and something more generic, e.g. `SERVICE_LOGS_DIR`? Then, as a user, it is more intuitive that I'm mapping my service logs folder (`/usr/local/apache2/logs`) to `${SERVICE_LOGS_DIR}` and then in my system test config, reading from the same folder.

mtojek: Good call, it's definitely a better user experience.

mtojek: Fixed

internal/testrunner/runners/system/test_config.go
```diff
@@ -54,11 +54,15 @@ func newConfig(systemTestFolderPath string, ctxt servicedeployer.ServiceContext)
...
 // most interesting context values will be set by a ServiceDeployer in its SetUp method.
 func applyContext(data []byte, ctxt servicedeployer.ServiceContext) ([]byte, error) {
-	tpl := string(data)
-	result, err := raymond.Render(tpl, ctxt)
+	tmpl, err := raymond.Parse(string(data))
 	if err != nil {
-		return data, errors.Wrap(err, "could not render data with context")
+		return data, errors.Wrap(err, "parsing template body failed")
 	}
+	tmpl.RegisterHelpers(ctxt.Aliases())
```
ycombinator, line 61: ❤️

ycombinator (approved):
LGTM. Very nice — much better than my stream redirection hack 😄 👍 🎉

PR #107, opened 2020-09-18
Sort imports using goimports

mtojek:
This PR introduces `goimports` to keep the Go imports sorted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,7 @@ build:
 
 format:
 	gofmt -s -w .
-
-imports:
+	
 	go get -u golang.org/x/tools/cmd/goimports
 	goimports -local github.com/elastic/elastic-package/ -w .
 
@@ -29,4 +28,4 @@ check-git-clean:
 	git update-index --really-refresh
 	git diff-index --quiet HEAD
 
-check: build format imports lint licenser gomod test check-git-clean
+check: build format lint licenser gomod test check-git-clean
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -5,6 +5,10 @@ build:
 format:
 	gofmt -s -w .
 
+imports:
+	go get -u golang.org/x/tools/cmd/goimports
+	goimports -local github.com/elastic/elastic-package/ -w .
+
```
ycombinator, line 11: Should we just make this part of `format`?

mtojek: Fixed. Thanks!

Makefile
```diff
@@ -4,6 +4,9 @@ build:
 
 format:
 	gofmt -s -w .
+	
+	go get -u golang.org/x/tools/cmd/goimports
+	goimports -local github.com/elastic/elastic-package/ -w .
```
ycombinator, line 9: Sorry, but I just found this on https://godoc.org/golang.org/x/tools/cmd/goimports:

> In addition to fixing imports, goimports also formats your code in the same style as gofmt so it can be used as a replacement for your editor's gofmt-on-save hook.

So maybe we don't even need `gofmt -s -w .` any more?

mtojek: Right, it includes the `gofmt` behavior. Working on this

mtojek: Fixed!

ycombinator (approved):
LGTM. Good to merge as soon as CI is happy.

Issue #108, opened 2020-09-18
Add sample test package


mtojek:
I believe we need some basic fake integration to test our changes and prevent from potential issues.

PR #109, opened 2020-09-18
Add sample test packages

mtojek:
Closes https://github.com/elastic/elastic-package/issues/108

_I'm opening as draft as there is something wrong with system tests runner. Let's wait for the CI._


mtojek:
Stack trace:

```
➜  apache git:(add-sample-test-package) pwd
/Users/marcin.tojek/go/src/github.com/elastic/elastic-package/test/packages/apache
➜  apache git:(add-sample-test-package) elastic-package test system
2020/09/18 12:24:54  INFO setting up service...
Error: error running package system tests: could not create service runner: unable to find service runner
```

ycombinator:
> Error: error running package system tests: could not create service runner: unable to find service runner

Maybe because there's nothing under `_dev/deploy`?

mtojek:
Yes, but there are multiple integrations that don't have any `_dev`.

scripts/test-check-packages.sh
```diff
@@ -0,0 +1,17 @@
...
+
+for d in test/packages/*/; do
+  (
+    cd $d
+    elastic-package check -v
+    elastic-package test -v
+  )
+done
+
+elastic-package stack down -v
```
ycombinator, line 17: I wish bash had `defer`s 🙂

ycombinator (approved):
LGTM. Great idea to add these!

scripts/test-check-packages.sh
mtojek: Hm.. I will try with "trap".

mtojek: I think it works as intended :)

PR #120, opened 2020-09-29
Make --datasets option work

### LINKED ISSUE #118: [System test runner] --datasets flag is not working
```
$ elastic-package test system -v --datasets access
2020/09/28 19:25:21 DEBUG Enable verbose logging
Done
```

ycombinator:
Fixes #118.

Turns out the golang `filepath.Glob` function does not accept any syntax that allows specifying multiple terms in a single glob expression. So instead, we iterate over the datasets, creating a glob expression for each.

### DIFF
--- a/internal/testrunner/testrunner.go
+++ b/internal/testrunner/testrunner.go
@@ -7,6 +7,7 @@ package testrunner
 import (
 	"fmt"
 	"path/filepath"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


mtojek (approved):
LGTM

Did you test it against cases without --datasets option (run for all datasets)?

ycombinator:
Yeah, I tested it with a single dataset, two datasets, and no datasets option. All work as expected.

PR #123, opened 2020-09-30
Pulling in histogram spec change

jonathan-buttner:
Pulling in the package spec change with the histogram field addition.


ycombinator (approved):
LGTM

PR #130, opened 2020-10-07
Support flat and extended manifest files

mtojek:
Issue: https://github.com/elastic/package-spec/issues/55

Changes introduced in the PR:
* use ucfg to support both flat and extended manifest files
* fix visibility of manifest related structures (all visible)
* replace `VarValue` (scalar/list) with `interface{}`
* fix comments in multiple places (mainly: `dataStream`)

### DIFF
--- a/internal/packages/packages.go
+++ b/internal/packages/packages.go
@@ -18,18 +18,18 @@ const (
 	// PackageManifestFile is the name of the package's main manifest file.
 	PackageManifestFile = "manifest.yml"
 
-	// DataStreamManifestFile is the name of the dataStream's manifest file.
+	// DataStreamManifestFile is the name of the data stream's manifest file.
 	DataStreamManifestFile = "manifest.yml"
 )
 
-// VarValue represents a variable value as defined in a package or dataStream
+// VarValue represents a Variable value as defined in a package or data stream
 // manifest file.
 type VarValue struct {
 	scalar string
 	list   []string
 }
 
-// UnmarshalYAML knows how to parse a variable value from a package or dataStream
+// UnmarshalYAML knows how to parse a Variable value from a package or data stream
 // manifest file into a VarValue.
 func (vv *VarValue) UnmarshalYAML(value *yaml.Node) error {
 	switch value.Kind {
@@ -41,7 +41,7 @@ func (vv *VarValue) UnmarshalYAML(value *yaml.Node) error {
 			vv.list[idx] = content.Value
 		}
 	default:
-		return errors.New("unknown variable value")
+		return errors.New("unknown Variable value")
 	}
 	return nil
 }
@@ -57,19 +57,22 @@ func (vv VarValue) MarshalJSON() ([]byte, error) {
 	return nil, nil
 }
 
-type variable struct {
+// Variable is an instance of configuration variable (named, typed).
+type Variable struct {
 	Name    string   `json:"name"`
 	Type    string   `json:"type"`
 	Default VarValue `json:"default"`
 }
 
-type input struct {
+// Input is a single input configuration.
+type Input struct {
 	Type string     `json:"type"`
-	Vars []variable `json:"vars"`
+	Vars []Variable `json:"vars"`
 }
 
-type policyTemplate struct {
-	Inputs []input `json:"inputs"`
+// PolicyTemplate is a configuration of inputs responsible for collecting log or metric data.
+type PolicyTemplate struct {
+	Inputs []Input `json:"inputs"`
 }
 
 // PackageManifest represents the basic structure of a package's manifest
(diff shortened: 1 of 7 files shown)


internal/packages/packages.go
```diff
@@ -5,94 +5,63 @@
...
-	// DataStreamManifestFile is the name of the dataStream's manifest file.
+	// DataStreamManifestFile is the name of the data stream's manifest file.
 	DataStreamManifestFile = "manifest.yml"
 )
 
-// VarValue represents a variable value as defined in a package or dataStream
-// manifest file.
-type VarValue struct {
-	scalar string
-	list   []string
```
ycombinator, line 29: I'm curious why you removed `VarValue` and replaced it with the broader `interface{}` type. AFAICT the only types of values allowed for input configuration variables are scalars (int, string, boolean, etc.) or lists of scalars. We don't support maps/dicts. That was the intent behind creating a stricter `VarValue` type that encapsulated these rules.

internal/packages/packages.go
```diff
@@ -5,94 +5,63 @@
...
+// Variable is an instance of configuration variable (named, typed).
+type Variable struct {
+	Name    string      `config:"name" json:"name" yaml:"name"`
+	Type    string      `config:"type" json:"type" yaml:"type"`
+	Default interface{} `config:"default" json:"default" yaml:"default"`
```
ycombinator, line 28: Now that we're introducing the `config` struct tags for `ucfg`'s benefit, do we still need to keep the `yaml` struct tags?

internal/packages/packages.go
```diff
@@ -5,94 +5,63 @@
...
+// Input is a single input configuration.
+type Input struct {
+	Type string     `config:"type" json:"type" yaml:"type"`
+	Vars []Variable `config:"vars" json:"vars" yaml:"vars"`
 }
 
-type policyTemplate struct {
-	Inputs []input `json:"inputs"`
+// PolicyTemplate is a configuration of inputs responsible for collecting log or metric data.
+type PolicyTemplate struct {
```
ycombinator, line 38: Is the visibility change needed for `ucfg` to work correctly?

internal/packages/packages.go
mtojek: It's because of ucfg which (I suppose) doesn't support UnmarshalYAML and MarshalJSON methods. It seemed to me that the codebase would be smaller, although I'm aware of the risk you brought here.

internal/packages/packages.go
mtojek: It's due to consistency. If somebody wants to use Manifest structure and extract policy templates to a separate variable, it wouldn't be possible (can't use a package visible field). In general exposed fields should use public structures.

internal/packages/packages.go
mtojek: I think you're right, although I have to check whether we don't call `yaml.Marshal` somewhere...

internal/packages/packages.go
ycombinator: Could you keep `VarValue` and make it implement https://godoc.org/github.com/elastic/go-ucfg#ConfigUnpacker somehow?

mtojek: I can try to do this, although I'm not convinced it will work (didn't see this earlier).

I see the benefits of having stricter `VarValue` type, so if you prefer I can switch back from `ucfg` to our implementation and extend it with same trick for expanding config map keys as in package-spec.

ycombinator: Thanks, I appreciate you giving it a try. There are a few examples of implementing `ConfigUnpacker` in the Beats repo. You can find them in your IDE with this regex: `func \(.* Unpack\(.*Config`.

If you succeed, it would allow us to preserve the strictness of what a variable value can be, which is very nice.

If it doesn't work, let's stick with `ucfg` and reduce the strictness of what a variable value can be. I think that's the better tradeoff than implementing some trick to parse flat values with the `yaml` package.

mtojek: I think I got it fixed. Let's see what CI will say.

ycombinator (approved):
LGTM. WFG.

mtojek:
Something didn't work here:

```
[2020-10-12T17:03:59.994Z] 2020/10/12 17:03:59 DEBUG found 0 hits in logs-apache.access-ep data stream
[2020-10-12T17:04:00.940Z] 2020/10/12 17:04:00 DEBUG found 0 hits in logs-apache.access-ep data stream
[2020-10-12T17:04:01.885Z] 2020/10/12 17:04:01 DEBUG found 0 hits in logs-apache.access-ep data stream
[2020-10-12T17:04:02.829Z] 2020/10/12 17:04:02 DEBUG found 0 hits in logs-apache.access-ep data stream
[2020-10-12T17:04:03.777Z] System test for apache/access data stream failed
```

I will investigate it tomorrow.

mtojek:
Ok, I approached the next issue:

```
[2020-10-13T09:42:07.328Z] Error: error running package system tests: locating data stream root failed: verifying manifest file failed (path: /var/lib/jenkins/workspace/Beats_elastic-package_PR-130/src/github.com/elastic/elastic-package/test/packages/apache/data_stream/status/manifest.yml): reading package manifest failed (path: /var/lib/jenkins/workspace/Beats_elastic-package_PR-130/src/github.com/elastic/elastic-package/test/packages/apache/data_stream/status/manifest.yml): unpacking data stream manifest failed (path: /var/lib/jenkins/workspace/Beats_elastic-package_PR-130/src/github.com/elastic/elastic-package/test/packages/apache/data_stream/status/manifest.yml): type mismatch accessing 'streams.0.vars.0.default' (source:'/var/lib/jenkins/workspace/Beats_elastic-package_PR-130/src/github.com/elastic/elastic-package/test/packages/apache/data_stream/status/manifest.yml')
```

mtojek:
Ok, it looks that I fixed it finally. VarVariable can be either `interface{}` or `[]interface{}`. Initially we forgot that's not just string (array of strings).

PR #132, opened 2020-10-08
HOWTO: Writing pipeline tests for a package

mtojek:
Previews:
https://github.com/mtojek/elastic-package/tree/howto_pipeline_tests#pipeline-tests
https://github.com/mtojek/elastic-package/blob/howto_pipeline_tests/docs/howto/pipeline_testing.md

- [x] Copy skeleton for the howto from system tests
- [x] Test case definition: Plain log files
- [x] Test case definition: Input events

PR #137, opened 2020-10-13
README Move context close to headings

mtojek:
Follow-up on https://github.com/elastic/elastic-package/pull/132

> Maybe we should move all the Context: lines below their respective command headings? Separate PR?


ycombinator (approved):
LGTM.

Issue #113, opened 2020-09-21
[System test runner] Implement result assertion


mtojek:
The elastic-package verifies if documents with results are in Elasticsearch, but it doesn't assert any fields of these ones. 

Introduce support for assertions of test results.

mtojek:
I think it would be nice to deliver this feature now :) I will take a look at it.

mtojek:
We're checking type correctness, potential error messages.
The issue for skipped fields has been created: https://github.com/elastic/elastic-package/issues/147

I suppose the next step is smart comparison of expected and collected documents in system tests?

ycombinator:
> I suppose the next step is smart comparison of expected and collected documents in system tests?

Yes, that will complete the scope of this issue.

mtojek:
I think we discussed it some time ago, but it might be a quick win here - 

Add `pattern` property to the string based fields in `fields.yml` files, and perform type assertion based on these fields only. No additional files for storing documents with patterns will be required.

ycombinator:
Yes, I think the quick win makes sense because I think pattern matching is the best we're going to be able to do with system tests anyway (since we're dealing with live data). But, of course, we would support it in pipeline tests as well. And, as you noted, pattern matching will only be applied for fields that define patterns; otherwise they just get the basic existence and type checking you implemented in #143.

And then, in a follow up PR, we can _also_ support exact document matching. The understanding here would be that this only makes sense for pipeline tests.

PR #138, opened 2020-10-13
Check if collected documents contain errors

mtojek:
First step in terms of https://github.com/elastic/elastic-package/issues/113

Check if there are errors in collected documents.

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/pkg/multierror"
 	es "github.com/elastic/go-elasticsearch/v7"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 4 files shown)


mtojek:
... and already detected issues :) 

```
[2020-10-13T13:02:27.232Z] Error: error running package system tests: could not check for expected data in data stream: [0] Provided Grok expressions do not match field value: [AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 172.21.0.2. Set the 'ServerName' directive globally to suppress this message]
```

internal/testrunner/runners/system/runner.go
```diff
@@ -213,16 +214,33 @@ func (r *runner) run() error {
...
+
+		var multiErr multierror.Error
+		for _, hit := range results.Hits.Hits {
+			if hit.Source.Error != nil {
+				multiErr = append(multiErr, errors.New(hit.Source.Error.Message))
+			}
+		}
+
+		if len(multiErr) > 0 {
+			return false, multiErr
```
ycombinator, line 241: Should we wrap `multiErr` with some overall message so it's not just a collection of error messages from ES?

mtojek: I like this idea. Fixed.

PR #127, opened 2020-10-06
Remove setting that's no longer needed

ycombinator:
Now that elastic/kibana#77613 has been resolved, we can get rid of the workaround added in #111.

PR #139, opened 2020-10-13
Fix waiting for agent policy reassignment

nchaulet:
My take on #127 

The `agent-status` is not really reliable to know successfully get a config, this PR change it to use the `policy_revision` that is set when an agent acknowledged a `POLICY_CHANGE`

### DIFF
--- a/internal/kibana/ingestmanager/client_agents.go
+++ b/internal/kibana/ingestmanager/client_agents.go
@@ -69,14 +69,14 @@ func (c *Client) getTotalAgentForPolicy(p Policy) (int, error) {
 	path := fmt.Sprintf("agents?kuery=%s", kuery)
 	statusCode, respBody, err := c.get(path)
 	if err != nil {
-		return -1, errors.Wrapf(err, "could not check agent status; API status code = %d; policy ID = %s; response body = %s", statusCode, p.ID, string(respBody))
+		return 0, errors.Wrapf(err, "could not check agent status; API status code = %d; policy ID = %s; response body = %s", statusCode, p.ID, string(respBody))
 	}
 	var resp struct {
 		Total int `json:"total"`
 	}
 
 	if err := json.Unmarshal(respBody, &resp); err != nil {
-		return -1, errors.Wrap(err, "could not convert agent list (response) to JSON")
+		return 0, errors.Wrap(err, "could not convert agent list (response) to JSON")
 	}
 
 	return resp.Total, nil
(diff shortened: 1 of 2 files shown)


ycombinator:
Thanks for the PR @nchaulet! I'm going to re-test it a few times just to make sure things are not flaky.

internal/kibana/ingestmanager/client_agents.go
```diff
@@ -63,32 +64,51 @@ func (c *Client) AssignPolicyToAgent(a Agent, p Policy) error {
 	return nil
 }
 
+func (c *Client) getTotalAgentForPolicy(p Policy) (int, error) {
+	kuery := url.QueryEscape(fmt.Sprintf("fleet-agents.policy_id:\"%s\"", p.ID))
+	path := fmt.Sprintf("agents?kuery=%s", kuery)
+	statusCode, respBody, err := c.get(path)
+	if err != nil {
+		return -1, errors.Wrapf(err, "could not check agent status; API status code = %d; policy ID = %s; response body = %s", statusCode, p.ID, string(respBody))
```
ycombinator, line 72: Should be okay to return `0` as the first return value here, yes?

ycombinator: Generally speaking the Golang idiom is to return a data type's zero-value as the return value if there is an error also being returned. The expectation is that the caller will check for the error and ignore any other return values.

internal/kibana/ingestmanager/client_agents.go
```diff
@@ -63,32 +64,51 @@ func (c *Client) AssignPolicyToAgent(a Agent, p Policy) error {
...
+			Total int `json:"total"`
 		}
 
 		if err := json.Unmarshal(respBody, &resp); err != nil {
-			return errors.Wrap(err, "could not convert agent status (response) to JSON")
-		}
-
-		if resp.Results.Total == 0 {
-			return fmt.Errorf("no agent is available")
+			return errors.Wrap(err, "could not convert agent lost (response) to JSON")
```
ycombinator, line 108: Did you mean `last` instead of `lost`?

nchaulet: `list` I will fix it

ycombinator:
Going to test for flakiness one final time but this is looking pretty good!

ycombinator (approved):
LGTM. Thanks, again, for the fix, @nchaulet!

PR #140, opened 2020-10-14

mtojek:
Update dependency on package-spec to merge recent changes in package-spec (empty array as default value for variable).


ycombinator (approved):
LGTM.

PR #141, opened 2020-10-14
Add hash to promotion

jonathan-buttner:
This PR adds functionality for including the combined sha1 hash for packages that are being promoted from one environment to the next. This is useful to help verify that the copied directory from one environment to the next has not accidentally changed the contents of any of the package's files. The hash does not include metadata for the file. One shortcoming is that files could be accidentally renamed and the hash for the directory would still come out to be the same.

An example of the addition is here: https://github.com/elastic/package-storage/pull/468

image

image

### DIFF
--- a/internal/promote/pull_requests.go
+++ b/internal/promote/pull_requests.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/go-git/go-git/v5"
 	"github.com/google/go-github/v32/github"
 	"github.com/pkg/errors"
 )
@@ -28,9 +29,12 @@ func OpenPullRequestWithRemovedPackages(client *github.Client, username, head, b
 
 // OpenPullRequestWithPromotedPackages method opens a PR against "base" branch with promoted packages.
 // Head is the branch containing the changes that will be added to the base branch.
-func OpenPullRequestWithPromotedPackages(client *github.Client, username, head, base, sourceStage, destinationStage string, promotedPackages PackageVersions) (string, error) {
+func OpenPullRequestWithPromotedPackages(client *github.Client, username, head, base, sourceStage, destinationStage string, promotedPackages PackageVersions, r *git.Repository) (string, error) {
 	title := fmt.Sprintf("[%s] Promote packages from %s to %s", destinationStage, sourceStage, destinationStage)
-	description := buildPullRequestPromoteDescription(sourceStage, destinationStage, promotedPackages)
+	description, err := buildPullRequestPromoteDescription(sourceStage, destinationStage, promotedPackages, r, head)
+	if err != nil {
+		return "", err
+	}
 	return openPullRequestWithPackages(client, username, head, base, title, description)
 }
 
@@ -72,13 +76,17 @@ func buildPullRequestRemoveDescription(sourceStage, promotionURL string, version
 	return builder.String()
 }
 
-func buildPullRequestPromoteDescription(sourceStage, destinationStage string, versions PackageVersions) string {
+func buildPullRequestPromoteDescription(sourceStage, destinationStage string, versions PackageVersions, r *git.Repository, head string) (string, error) {
 	var builder strings.Builder
 	builder.WriteString(fmt.Sprintf("This PR promotes packages from `%s` to `%s`.\n", sourceStage, destinationStage))
 	builder.WriteString("\n")
 	builder.WriteString("Promoted packages:\n")
-	for _, str := range versions.Strings() {
+	packageStrs, err := versions.StringsWithHash(r, head)
+	if err != nil {
+		return "", err
+	}
+	for _, str := range packageStrs {
 		builder.WriteString(fmt.Sprintf("* `%s`\n", str))
 	}
-	return builder.String()
+	return builder.String(), nil
 }
--- a/internal/promote/pull_requests.go
+++ b/internal/promote/pull_requests.go
@@ -9,8 +9,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/google/go-github/v32/github"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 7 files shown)


ycombinator:
This is a nice improvement! I'm curious how long it takes to compute the hash. Looks like our largest package is `aws`. Is there some way you could run the hashing code on this package and post how long it took?

ycombinator:
You may want to consider replacing the SHA1 hashing algorithm with the much faster [xxHash algorithm](github.com/cespare/xxhash), particularly since we don't care about using this hash for cryptographic purposes.

internal/promote/storage.go
```diff
@@ -51,6 +51,37 @@ func (pr *PackageVersion) String() string {
 	return fmt.Sprintf("%s-%s", pr.Name, pr.Version)
 }
 
+// StringWithHash method returns a string representation of the PackageVersion.
+func (pr *PackageVersion) StringWithHash(r *git.Repository, branch string) (string, error) {
```
mtojek, line 55: Let me refer to responsibility of the `PackageVersion`.

This structure is intended to keep properties of the package revision (like name or version) and would be a good candidate for the signature (SHA1). I believe the design will remain clean if you separate signature computation from the model here. I'm not convinced adding here functions accessing the local repository. 

As Shaunak suggested the signature calculation might be long, so maybe let's put it aside as a dedicated structure responsible for computing hashes for given package revisions. The output with signatures can be passed to the method responsible for opening PRs (https://github.com/elastic/elastic-package/blob/master/cmd/promote.go#L107).

mtojek (commented):
I like the integrity check and it would be great to add it, although I have some suggestions on the solution design.

jonathan-buttner:
> This is a nice improvement! I'm curious how long it takes to compute the hash. Looks like our largest package is `aws`. Is there some way you could run the hashing code in this PR on this package and post how long it took?

Yeah good idea:
Looks like 14.93615ms

```
Promote packages
Current GitHub user: jonathan-buttner
? Which promotion would you like to run snapshot - staging
? Would you like to promote newest versions only and remove older ones? Yes
Cloning repository...
Creating list of packages...
? Which packages would you like to promote aws-0.3.8
Promote packages from snapshot to staging...
Remove packages from snapshot...
hash time: 14.93615ms
Pull request with promoted packages: https://github.com/elastic/package-storage/pull/472
Pull request with removed packages: https://github.com/elastic/package-storage/pull/473
Done
```

```
diff --git a/internal/promote/storage.go b/internal/promote/storage.go
index ebf95fe..a3b76eb 100644
--- a/internal/promote/storage.go
+++ b/internal/promote/storage.go
@@ -9,6 +9,7 @@ import (
        "io/ioutil"
        "path/filepath"
        "sort"
+       "time"

        "github.com/Masterminds/semver"
        "github.com/go-git/go-billy/v5"
@@ -79,7 +80,11 @@ func (pr *PackageVersion) sha1Hash(r *git.Repository, branch string) (string, er
        if err != nil {
                return "", errors.Wrap(err, "failed to retrieve package paths while calculating directory hash")
        }
-       return hashFiles(wt.Filesystem, resources)
+       start := time.Now()
+       hash, err := hashFiles(wt.Filesystem, resources)
+       end := time.Now()
+       fmt.Printf("hash time: %v\n", end.Sub(start))
+       return hash, err
 }

 // PackageVersions is an array of PackageVersion.
```

> You may want to consider replacing the SHA1 hashing algorithm with the much faster [xxHash algorithm](github.com/cespare/xxhash), particularly since we don't care about using this hash for cryptographic purposes.

Yeah good point, I'll pull that algorithm in

internal/promote/storage.go
jonathan-buttner: Sounds good, I'll extract it out and have it called here: https://github.com/elastic/elastic-package/blob/master/internal/promote/pull_requests.go#L75

ycombinator:
Thanks @jonathan-buttner.  A worst case of ~15ms is really not too bad. But if we can improve it without too much effort (replacing the algorithm), then let's try that out. It'd be nice to know what the same metric looks like with xxHash.

jonathan-buttner:
> Thanks @jonathan-buttner. A worst case of ~15ms is really not too bad. But if we can improve it without too much effort (replacing the algorithm), then let's try that out. It'd be nice to know what the same metric looks like with xxHash.

Down to ~6ms 👍 

```
Promote packages
Current GitHub user: jonathan-buttner
? Which promotion would you like to run snapshot - staging
? Would you like to promote newest versions only and remove older ones? Yes
Cloning repository...
Creating list of packages...
? Which packages would you like to promote aws-0.3.8
Promote packages from snapshot to staging...
Remove packages from snapshot...
hash time: 6.698123ms
Pull request with promoted packages: https://github.com/elastic/package-storage/pull/474
Pull request with removed packages: https://github.com/elastic/package-storage/pull/475
Done
```

ycombinator (approved):
LGTM.

internal/promote/pull_requests.go
```diff
@@ -28,9 +30,12 @@ func OpenPullRequestWithRemovedPackages(client *github.Client, username, head, b
 
 // OpenPullRequestWithPromotedPackages method opens a PR against "base" branch with promoted packages.
 // Head is the branch containing the changes that will be added to the base branch.
-func OpenPullRequestWithPromotedPackages(client *github.Client, username, head, base, sourceStage, destinationStage string, promotedPackages PackageVersions) (string, error) {
+func OpenPullRequestWithPromotedPackages(client *github.Client, username, head, base, sourceStage, destinationStage string, promotedPackages PackageVersions, r *git.Repository) (string, error) {
```
mtojek, line 33: This is looking better!

Just one last thought. I think it's possible to leave `OpenPullRequestWithPromotedPackages` as is and let it be responsible only for building PR content (concatenating strings) and calling Github API. What do you think about moving `calculatePackageSignature` to the `signature.go` and creating a new struct, e.g. `SignedPackageVersions`, that later on can be passed here:

`func OpenPullRequestWithPromotedPackages(client *github.Client, username, head, base, sourceStage, destinationStage string, signedPromotedPackages SignedPackageVersions) (string, error) {`

or 

`func OpenPullRequestWithPromotedPackages(client *github.Client, username, head, base, sourceStage, destinationStage string, promotedPackages PackageVersions,  signatures PackageSignatures) (string, error) {`

jonathan-buttner: Yeah that makes sense. I'm going to go with
```
func OpenPullRequestWithPromotedPackages(client *github.Client, username, head, base, sourceStage, destinationStage string, signedPromotedPackages SignedPackageVersions) (string, error) {
```

and have the `SignedPackageVersions` struct be composed of a `PackageVersion`

mtojek (approved):
Good job :) Thank you for adjusting the implementation, ship it!

PR #142, opened 2020-10-15
Better wording

ycombinator:
Changes "current" to "actual" (when comparing test results).


mtojek (approved):
It looks like I introduced the "current" word. Sorry about that, approved of course.

Issue #90, opened 2020-09-09
[System test runner] Assert that fields in event document are in fields.yml


ycombinator:
Follow up to #64.

Currently the system test runner simply checks if documents are present in the appropriate data stream. A next step would be to parse the fields from those documents and check if they match up with the ones defined in the package/dataset's `fields.yml` files.

mtojek:
I'm going with this feature.

mtojek:
This might be a bit complex that I though due to wildcards. See the AWS `package-fields.yml`:

```yaml
- name: aws
  type: group
  fields:
    - name: tags.*
      type: object
      description: |
        Tag key value pairs from aws resources.
    - name: s3.bucket.name
      type: keyword
      description: |
        Name of a S3 bucket.
    - name: dimensions.*
      type: object
      description: |
        Metric dimensions.
    - name: '*.metrics.*.*'
      type: object
      description: |
        Metrics that returned from Cloudwatch API query.
```

It looks like `*.metrics.*.*` is a catch-all object definition.

ycombinator:
Hmmm, might be worth looking into how the `aws` module in Metricbeat is handling this?

ycombinator:
> Hmmm, might be worth looking into how the `aws` module in Metricbeat is handling this?

Looks like it's not :) There is no system test for the `aws` Metricbeat module in Beats: https://github.com/elastic/beats/tree/master/x-pack/metricbeat/module/aws

exekias:
One option to avoid mismatches between Kibana implementation and whatever we do is to is actually use the installed integration. Send events to the index and check the resulting mapping there

ycombinator:
>  Send events to the index and check the resulting mapping there

This might work for most cases. But IIRC there is some setting available in `fields.yml` that causes a field not to be mapped. Let me see if I can find it...

ycombinator:
I think I was thinking of the `default_field: false` setting available on fields in `fields.yml`, but that does not impact whether the field shows up in the ES template mapping or not. So I think we are good here to check against the resulting mapping.

mtojek:
Thank you for your input, @exekias . I'm not sure if I follow the idea. Do you think we can use the Get field mapping API? See: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-field-mapping.html

ycombinator:
I think you can just get the entire mapping for a data stream, no? And then, for each field in the event document, ensure that that field is defined in the mapping.

[...]

PR #143, opened 2020-10-19
Implement basic field assertion (types, presence)

mtojek:
This PR introduces a basic field type assertion during pipeline and system tests. When a pipeline or system test is executed, generated results are examined against correctness with fields schema (fields YAML files). All inconsistencies will be reported.

Issue: https://github.com/elastic/elastic-package/issues/90

Action items (TODOs):

- [x] Add AWS sample data stream with fields
- [x] Load JSON unit test (with ES document) from JSON file
- [x] Validate JSON results in pipeline tests
- [x] Validate JSON results in system tests
- [x] Validate document body
- [x] Validate field type
- [x] Fix all edge cases in pipeline tests
- [x] Fix all edge cases in system tests


ycombinator:
I assume you will choose a different test package than `aws` when it comes to validating JSON body in system tests?

mtojek:
> I assume you will choose a different test package than aws when it comes to validating JSON results in system tests?

Yes, I will try to use the only one present, `apache`, but it would be nice to spend some time and populate the repo (already created such issue in integrations).

mtojek:
Not sure what should I do with entries like:

```
[2020-10-20T16:31:02.897Z] Error: error running package system tests: could not check for expected data in data stream: one or more errors found in documents stored in logs-apache.access-ep data stream: [0] field "agent.hostname" is not defined
```

The field `agent.hostname` is not defined anywhere as it's a common one. Should I accept it even though there is no definition for it in `fields.yml`?

ycombinator:
> The field `agent.hostname` is not defined anywhere as it's a common one. Should I accept it even though there is no definition for it in `fields.yml`?

I assume this field is somehow getting defined in the index template's mapping. Do you how where that's happening?

mtojek:
> I assume this field is somehow getting defined in the index template's mapping. Do you how/where that's happening?

No idea. Maybe @ruflin know some more details? I suspect that there are more hidden definitions.

mtojek:
The system test runner reports tests results under the "Tests" tab:
https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Felastic-package/detail/PR-143/12/tests

Moving on...

mtojek:
I need to fix the `geo_point` field:

```
[0] field "source.geo.location.lon" is not defined
[1] field "source.geo.location.lon" is not defined
[2] field "source.geo.location.lon" is not defined
[3] field "source.geo.location.lon" is not defined
[4] field "source.geo.location.lon" is not defined
[5] field "source.geo.location.lon" is not defined
[6] field "source.geo.location.lat" is not defined
[7] field "source.geo.location.lat" is not defined
```

ycombinator:
@mtojek Could you update the PR description with a short description of what this PR does and why? And/or link it to an issue? This will help if anyone stumbles upon this PR in the future via git blame or something to figure out why we did what we did here 😄 . Thanks!

Makefile
```diff
@@ -22,13 +22,16 @@ licenser:
 gomod:
 	go mod tidy
 
+test-go:
+	go test -v -count 1 ./...
```
ycombinator, line 26: Consider adding an explanation for the `-count 1`, similar to https://github.com/elastic/package-spec/blob/e7102f8cbd392d977bb2a21b12674dbe2ae8b985/code/go/Makefile#L33-L35.

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Validator is responsible for fields validation.
+type Validator struct {
+	schema []fieldDefinition
+}
+
+// CreateValidatorForDataStream method creates a validator for the data stream.
+func CreateValidatorForDataStream(dataStreamRootPath string) (*Validator, error) {
+	fieldsDir := filepath.Join(dataStreamRootPath, "fields")
+	fis, err := ioutil.ReadDir(fieldsDir)
```
ycombinator, line 30: Nit: It's not clear to me what `fis` stand for. But I understand that it will contain a list of fields files. Maybe rename the variable to `fieldsFiles` instead of `fis` to be a bit more descriptive and to parallel `fieldsDir`?

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+			if err != nil {
+				return err
+			}
+			break
+		}
+	}
+	return nil
+}
+
+func (v *Validator) validateElementFormat(key string, val interface{}) error {
```
ycombinator, line 99: Nit: consider renaming this method to `validatScalarElement` or similar so it parallel's `validateMapElement`.

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	err := parseElementValue(key, *definition, val)
+	if err != nil {
+		return errors.Wrap(err, "parsing field value failed")
+	}
+	return nil
+}
+
+func skipValidationForField(key string) bool {
```
ycombinator, line 120: At the very least we should add a comment here explaining why these fields are being skipped during validation. Additionally, should we file an issue to follow up on some way to validate these in the future (related to the discussion started here: https://github.com/elastic/elastic-package/pull/143#issuecomment-712994535)?

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+	return isFieldFamilyMatching("agent", key) ||
+		isFieldFamilyMatching("elastic_agent", key)
+}
+
+func isFieldFamilyMatching(family, key string) bool {
+	return key == family || strings.HasPrefix(key, family+".")
+}
+
+func findElementDefinitionInSlice(root, searchedKey string, fieldDefinitions []fieldDefinition) *fieldDefinition {
```
ycombinator, line 129: Nit: consider renaming `Slice` to something else that is more meaningful to the problem domain like `FieldDefinitions`. `Slice` can be inferred by looking at the data type of the last argument anyway.

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+		key := strings.TrimLeft(root+"."+def.Name, ".")
+		if compareKeys(key, def, searchedKey) {
+			return &def
+		}
+
+		if len(def.Fields) == 0 {
+			continue
+		}
+
+		fd := findElementDefinitionInSlice(key, searchedKey, def.Fields)
```
ycombinator, line 140: To avoid star recursion here, I wonder if we should store the field definitions in a flattened manner. Then we can just iterated over the flattened list and check prefixes or do some sort of glob matching instead of recursion. Could be an optimization for a follow up PR.

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+	return nil
+}
+
+func compareKeys(key string, def fieldDefinition, searchedKey string) bool {
+	k := strings.ReplaceAll(key, ".", "\\.")
+	k = strings.ReplaceAll(k, "*", "[^.]+")
+
+	// Workaround for potential geo_point, as "lot" and "lat" fields are not present in field definitions.
```
ycombinator, line 152: Typo: should be `lon`.

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+	if def.Type == "geo_point" {
+		k += "\\.(lon|lat)"
+	}
+
+	k = fmt.Sprintf("^%s$", k)
+	matched, err := regexp.MatchString(k, searchedKey)
+	if err != nil {
+		panic(err)
```
ycombinator, line 160: Since we're `panic`ing here, maybe add a bit more context with `errors.Wrap`?

internal/fields/validate.go
```diff
@@ -0,0 +1,196 @@
...
+		valid = true // all other types are considered valid not blocking validation
+	}
+
+	if !valid {
+		return fmt.Errorf("field \"%s\" has invalid type, expected: %s, actual Go type: %s", key, definition.Type, reflect.TypeOf(val))
+	}
+	return nil
+}
+
+func ensureSingleElementValue(val interface{}) (interface{}, bool) {
```
ycombinator, line 187: By "single" do you mean "scalar"?

ycombinator, line 187: Took me a few seconds to realize why we need this helper: because any data type in ES can take array values. Maybe add an explanatory comment here for future us?

internal/multierror/error.go
```diff
@@ -0,0 +1,26 @@
...
+	if me == nil {
+		return ""
+	}
+
+	strs := make([]string, len(me))
+	for i, err := range me {
+		strs[i] = fmt.Sprintf("[%d] %v", i, err)
+	}
+	return strings.Join(strs, "\n")
+}
```
ycombinator, line 26: ❤️

internal/testrunner/runners/system/runner.go
```diff
@@ -247,26 +262,32 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
...
-			if hit.Source.Error != nil {
-				multiErr = append(multiErr, errors.New(hit.Source.Error.Message))
+			if message, err := hit.Source.GetValue("error.message"); err != common.ErrKeyNotFound {
+				multiErr = append(multiErr, errors.New(message.(string)))
+				continue
+			}
+
+			err := fieldsValidator.ValidateDocumentMap(hit.Source)
+			if err != nil {
+				multiErr = append(multiErr, err)
```
ycombinator, line 272: Consider adding a `continue` here. I know it's not needed because this is the last check in the loop but  it future-proofs the code in case we add more code to the loop after this (and forget to add the `continue` here at the time).

internal/testrunner/runners/system/runner.go
```diff
@@ -247,26 +262,32 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
...
+				Reason:  fmt.Sprintf("one or more errors found in documents stored in %s data stream", dataStream),
+				Details: multiErr.Error(),
+			}
 		}
 		return true, nil
 	}, 2*time.Minute)
-
 	if err != nil {
-		return resultsWith(result, errors.Wrap(err, "could not check for expected data in data stream"))
+		return resultsWith(result, err)
```
ycombinator, line 285: Any reason to review the `errors.Wrap` here?

ycombinator (commented):
Really nice work on this! Most of my comments are minor.

Makefile
mtojek: Fixed

internal/fields/validate.go
mtojek: This is actually `fileInfos`, but I agree it's not really descriptive. Renamed!

internal/fields/validate.go
mtojek: Renamed. Thanks!

internal/fields/validate.go
mtojek: Fixed

internal/testrunner/runners/system/runner.go
mtojek: This is great argument, really appreciate it. Introduced of course.

internal/testrunner/runners/system/runner.go
mtojek: It is like that because I changed the implementation of `resultsWith` function:

```go
	resultsWith := func(tr testrunner.TestResult, err error) ([]testrunner.TestResult, error) {
		tr.TimeElapsed = time.Now().Sub(startTime)
		if err == nil {
			return []testrunner.TestResult{tr}, nil
		}

		if tcf, ok := err.(testerrors.ErrTestCaseFailed); ok {
			tr.FailureMsg = tcf.Reason
			tr.FailureDetails = tcf.Details
			return []testrunner.TestResult{tr}, nil
		}

		tr.ErrorMsg = err.Error()
		return []testrunner.TestResult{tr}, err
	}
```

You can say, that I'm doing type assertion: `err.(testerrors.ErrTestCaseFailed)`.

internal/fields/validate.go
mtojek: Fixed.

internal/fields/validate.go
mtojek: Created an issue: https://github.com/elastic/elastic-package/issues/147

internal/fields/validate.go
mtojek: You're right here. Renamed to `findElementDefinition`.

internal/fields/validate.go
mtojek: I added a comment.

internal/fields/validate.go
mtojek: I don't remember the exact thought process now, but I had a similar discussion with myself.

 Let's consider a field in results: `apache.access.foo.bar`. If we flatten field definitions (from YML files), we'll end up with 4 records to cover the single result fields -  we need field definitions for `apache`, `apache.access`, `apache.access.foo`, `apache.access.foo.bar`.

I considered this as not a huge gain comparing to the recursive solution (more towards readability?), but I'm happy to adjust it in follow-ups.

ycombinator: ++ to leaving it this way for now. Only reason to try the flattening ideal would be for performance / lower memory usage but until either of those things become an issue, let's not optimize prematurely.

internal/testrunner/runners/system/runner.go
ycombinator: Ah, I see now. Thanks for pointing me to that.

ycombinator:
> I added a couple of fields and it seems that there're new weird ones coming :) (log.input?). Maybe we shouldn't fail on undefined fields and just check the type correctness? I see now a case in which a field change pushed to filebeat/metricbeat (new field added) fails all system tests for integrations.

I imagine such new common fields should not be added frequently so I would actually prefer to be strict, at least initially and see how much of a PITA it becomes for us. If we suddenly see *all* tests failing for integrations, it's a pretty good indication that it's not something wrong with one or two integrations but something more widespread and common. Looking at the failure details it should become pretty clear what that common new field is and then we can add it to our skip list.

Also, depending on how we solve #147, we might not have to worry about this issue of new common fields being added every so often.

mtojek:
Ok, I will adjust the implementation to report all problems and then add missing fields. I'm only worried about future situations in which we have to adjust all integrations (time consuming), but maybe it won't be so hard.

ycombinator:
> Ok, I will adjust the implementation to report all problems and then add missing fields. I'm only worried about future situations in which we have to adjust all integrations (time consuming), but maybe it won't be so hard.

Yes, this is a fair point. I think we'll have to see how it goes. If it starts to become a hassle, I'll be +1 to not failing on undefined fields. But I'm also hoping we can find an elegant solution to #147, something that involves being able to detect these common fields from some other source than the package field definition files.

mtojek:
I modified the code to get all validation errors at once, but it went up to 300 errors :) See: https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Felastic-package/detail/PR-143/33/tests

I will adjust the code to report only unique records.

mtojek:
I managed to clean up all reported fields: https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Felastic-package/detail/PR-143/35/tests

Field families that are not related to the specific integration, but appear in documents: `cloud.*`, `host.*`, `event.*`.

mtojek:
@ycombinator I fixed all issues and addressed your comments. Please take a look when you're free. Keep in mind the number of additional fields I had to put here.

internal/fields/validate.go
```diff
@@ -0,0 +1,217 @@
...
+// in every (most?) documents collected by Elastic Agent, but aren't defined in any integration in `fields.yml` files.
+// FIXME https://github.com/elastic/elastic-package/issues/147
+func skipValidationForField(key string) bool {
+	return isFieldFamilyMatching("agent", key) ||
+		isFieldFamilyMatching("elastic_agent", key) ||
+		isFieldFamilyMatching("cloud", key) || // too many common fields
+		isFieldFamilyMatching("event", key) || // too many common fields
+		isFieldFamilyMatching("host", key) || // too many common fields
+		isFieldFamilyMatching("metricset", key) || // field is deprecated
+		isFieldFamilyMatching("event.module", key) // field is deprecated
```
ycombinator, line 141: This line is now redundant with `isFieldFamilyMatching("event", key)` but I think we should keep it anyway because even if we were to somehow get rid of `isFieldFamilyMatching("event", key)` in the future, we would want this line to be here, correct?

ycombinator (approved):
LGTM. Nice job on this feature!

internal/fields/validate.go
mtojek: Right, maybe it's better to keep it here even though it's redundant.

PR #144, opened 2020-10-26

mtojek:
This PR updates dependency on package-spec to validate dashboards filenames.


ycombinator (approved):
LGTM.

PR #149, opened 2020-10-28
Fix: fail testruns on error or failure

mtojek:
This PR fixes bug in the `test` command. The application should exit with an error code in case of an error or a failure. 

Spotted in https://github.com/elastic/integrations/pull/345


ycombinator (approved):
LGTM

Issue #13, opened 2020-06-05
Command: clean up used resources


mtojek:
The command cleans all built artifacts.

Suggested command: `elastic-package clean`

ycombinator:
This should also clean up the test reports folder (see the discussion here: https://github.com/elastic/elastic-package/pull/158#discussion_r517151935).

PR #158, opened 2020-11-02
Store test reports in build/test-results/ folder

ycombinator:
Addresses https://github.com/elastic/integrations/pull/315#discussion_r511854030.

This PR stores test reports generated by `elastic-package test --report-output file` into `build/test-results/`. Before this PR they were being stored in `~/.elastic-package/tmp/test/reports/`.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -47,10 +47,12 @@ pipeline {
       }
       post {
         always {
-          archiveArtifacts(allowEmptyArchive: true, artifacts: 'src/github.com/elastic/elastic-package/build/test-results/*.xml')
-          junit(allowEmptyResults: false,
+          dir("${BASE_DIR}") {
+            archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/test-results/*.xml')
+            junit(allowEmptyResults: false,
                 keepLongStdio: true,
-                testResults: "src/github.com/elastic/elastic-package/build/test-results/*.xml")
+                testResults: "build/test-results/*.xml")
+          }
         }
       }
     }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -47,10 +47,10 @@ pipeline {
       }
       post {
         always {
-          archiveArtifacts(allowEmptyArchive: true, artifacts: '.elastic-package/build/test-results/*.xml')
+          archiveArtifacts(allowEmptyArchive: true, artifacts: 'src/github.com/elastic/elastic-package/build/test-results/*.xml')
           junit(allowEmptyResults: false,
                 keepLongStdio: true,
-                testResults: ".elastic-package/build/test-results/*.xml")
+                testResults: "src/github.com/elastic/elastic-package/build/test-results/*.xml")
         }
       }
     }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -47,10 +47,10 @@ pipeline {
       }
       post {
         always {
-          archiveArtifacts(allowEmptyArchive: true, artifacts: '.elastic-package/tmp/test/reports/*.xml')
+          archiveArtifacts(allowEmptyArchive: true, artifacts: '.elastic-package/build/test-results/*.xml')
           junit(allowEmptyResults: false,
                 keepLongStdio: true,
-                testResults: ".elastic-package/tmp/test/reports/*.xml")
+                testResults: ".elastic-package/build/test-results/*.xml")
         }
       }
     }
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -65,7 +65,11 @@ func FindBuildPackagesDirectory() (string, bool, error) {
 	if found {
 		path := filepath.Join(buildDir, "integrations") // TODO add support for other package types
 		fileInfo, err := os.Stat(path)
-		if err == nil && fileInfo.IsDir() {
+		if err != nil {
+			return "", false, err
+		}
+
+		if fileInfo.IsDir() {
 			return path, true, nil
 		}
 	}
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -32,8 +32,8 @@ func BuildPackage() error {
 	return nil
 }
 
-// FindBuildPackagesDirectory method locates the target build directory for packages.
-func FindBuildPackagesDirectory() (string, bool, error) {
+// FindBuildDirectory locates the target build directory.
+func FindBuildDirectory() (string, bool, error) {
 	workDir, err := os.Getwd()
 	if err != nil {
 		return "", false, errors.Wrap(err, "locating working directory failed")
@@ -41,7 +41,7 @@ func FindBuildPackagesDirectory() (string, bool, error) {
 
 	dir := workDir
 	for dir != "." {
-		path := filepath.Join(dir, "build", "integrations") // TODO add support for other package types
+		path := filepath.Join(dir, "build")
 		fileInfo, err := os.Stat(path)
 		if err == nil && fileInfo.IsDir() {
 			return path, true, nil
@@ -52,6 +52,24 @@ func FindBuildPackagesDirectory() (string, bool, error) {
 		}
 		dir = filepath.Dir(dir)
 	}
+	return "", false, nil
+}
+
+// FindBuildPackagesDirectory method locates the target build directory for packages.
+func FindBuildPackagesDirectory() (string, bool, error) {
+	buildDir, found, err := FindBuildDirectory()
+	if err != nil {
+		return "", false, err
+	}
+
+	if found {
+		path := filepath.Join(buildDir, "integrations") // TODO add support for other package types
+		fileInfo, err := os.Stat(path)
+		if err == nil && fileInfo.IsDir() {
+			return path, true, nil
+		}
+	}
+
 	return "", false, nil
 }
 
--- a/internal/install/install.go
+++ b/internal/install/install.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/builder"
 )
 
 const (
@@ -18,11 +20,11 @@ const (
 	stackDir          = "stack"
 	packagesDir       = "development"
 	temporaryDir      = "tmp"
+	testReportsDir    = "test-results"
 )
 
 var (
 	serviceLogsDir = filepath.Join(temporaryDir, "service_logs")
-	testReportsDir = filepath.Join("build", "test-results")
 )
 
 const versionFilename = "version"
@@ -96,11 +98,11 @@ func ServiceLogsDir() (string, error) {
 
 // TestReportsDir returns the location of the directory to store test reports.
 func TestReportsDir() (string, error) {
-	configurationDir, err := configurationDir()
+	buildDir, _, err := builder.FindBuildDirectory()
 	if err != nil {
-		return "", errors.Wrap(err, "locating configuration directory failed")
+		return "", errors.Wrap(err, "locating build directory failed")
 	}
-	return filepath.Join(configurationDir, testReportsDir), nil
+	return filepath.Join(buildDir, testReportsDir), nil
 }
 
 func configurationDir() (string, error) {
--- a/internal/install/install.go
+++ b/internal/install/install.go
@@ -60,10 +60,6 @@ func EnsureInstalled() error {
 		return errors.Wrap(err, "creating service logs directory failed")
 	}
 
-	if err := createTestReportsDir(elasticPackagePath); err != nil {
-		return errors.Wrap(err, "creating test reports directory failed")
-	}
-
 	fmt.Fprintln(os.Stderr, "elastic-package has been installed.")
 	return nil
 }
@@ -169,15 +165,6 @@ func writeStaticResource(err error, path, content string) error {
 
 func createServiceLogsDir(elasticPackagePath string) error {
 	dirPath := filepath.Join(elasticPackagePath, serviceLogsDir)
-	err := os.MkdirAll(dirPath, 0755)
-	if err != nil {
-		return errors.Wrapf(err, "mkdir failed (path: %s)", dirPath)
-	}
-	return nil
-}
-
-func createTestReportsDir(elasticPackagePath string) error {
-	dirPath := filepath.Join(elasticPackagePath, testReportsDir)
 	err := os.MkdirAll(dirPath, 0755)
 	if err != nil {
 		return errors.Wrapf(err, "mkdir failed (path: %s)", dirPath)
--- a/internal/install/install.go
+++ b/internal/install/install.go
@@ -22,7 +22,7 @@ const (
 
 var (
 	serviceLogsDir = filepath.Join(temporaryDir, "service_logs")
-	testReportsDir = filepath.Join(temporaryDir, "test", "reports")
+	testReportsDir = filepath.Join("build", "test-results")
 )
 
 const versionFilename = "version"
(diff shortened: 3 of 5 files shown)


mtojek:
AFAIK as I remember I was talking about the `built` directory for the project (Integrations, elastic-package) not to pollute the `.elastic-package` at all. WDYT?

mtojek (commented):
Left one comment

ycombinator:
> AFAIK as I remember I was talking about the `built` directory for the project (Integrations, elastic-package) not to pollute the `~/.elastic-package` at all. WDYT?

Oh I see. In that case I totally misunderstood the comment!

Yeah, I see the logic in keeping test results out of `~/elastic-package`. That folder is more meant for `elastic-package` internals, so polluting it with package-specific artifacts does seem out of place. 

I can see two ways to implement this:
1. We can redesign the `--report-output` flag to take either `stdout` (default) or something that looks like an absolute filesystem path. If the value is neither `stdout` nor an absolute path, we error out. If it's an absolute path, `elastic-package` will treat that path like a folder, creating it as necessary, and then write the test result files to it, _or_
2. We keep the design of `--report-output` as-is and add another flag, say `--report-folder`, that is only valid when`--report-output` is set to `file`. Then we treat the value of `--report-folder` as a path to a folder, creating it as necessary, and then write the test result files to it.

I'm personally in favor of option 1 because the semantics are pretty easy to follow and it results in one less flag (compared to option 2). WDYT?

mtojek:
> I'm personally in favor of option 1 because the semantics are pretty easy to follow and it results in one less flag (compared to option 2). WDYT?

I agree with you, unless we follow a different schema:
1. `--report-output file` uses the build directory by default (there is a method in internals that can discover it). The user can overwrite it with `--report-path` as you suggested, but I wonder if this flag is necessary at this stage (I suspect nobody will into them except Jenkins).

ycombinator:
> unless we follow a different schema: ...

I don't a concrete use case for `file` as a valid value and writing to a default folder. For the problem we're trying to solve - moving the package-specific build/test artifacts outside of `~/.elastic-package` _something_ is going to have to tell `elastic-package` where the appropriate location for those artifacts should be. And, as you noted, currently the only user of this is Jenkins. Which means there's no real value to having a default location at the moment. Finally, if it turns out in the future that there is a use case for a default location, we an re-introduce the `file` value for this flag.

I will make the necessary changes to this PR.

mtojek:
The reason why I insisted on keeping results in the `github.com/elastic/integrations/build/test-results` directory is that it won't leak test results anywhere and pollute the `~/`, especially that these files need to stay after the command finishes (contrary to `~/.elastic-package/tmp` which should be cleaned after the execution*). I don't have any strong arguments in terms of parallel execution.

side-note:

* - maybe we should rethink the `~/.elastic-package/tmp` if it's not better to keep such files in `build` directory, which is known pattern

ycombinator:
> *- maybe we should rethink the `~/.elastic-package/tmp` if it's not better to keep such files in build directory, which is known pattern

This is a good discussion! We're having to come up with a general principle to decide what sorts of stuff belongs under `~/.elastic-package` vs. not. 

I consider `~/.elastic-package` as an internal implementation detail of the `elastic-package` tool. As such:
- it should **never** contain files that might need to be available to something other than itself. This is the reason for moving the test results out of `~/.elastic-package`.
- it should **definitely** contain any files that shouldn't be available to anything other than itself, opening up the potential for misuse or abuse. To me, this is the reason for keeping `/tmp/service_logs` within `~/.elastic-package` as these logs are a temporary location to "transfer" the service container logs to the agent container. However, I do agree that if we find ourselves adding more things to `~/.elastic-package/tmp/` we need to ask ourselves if those files need to be accessible *exclusively* by `elastic-package` or not.

WDYT?

.ci/Jenkinsfile
```diff
@@ -47,10 +47,10 @@ pipeline {
       }
       post {
         always {
-          archiveArtifacts(allowEmptyArchive: true, artifacts: '.elastic-package/tmp/test/reports/*.xml')
+          archiveArtifacts(allowEmptyArchive: true, artifacts: '/tmp/build/test-results/*.xml')
```
ycombinator, line 50: @mtojek I'm open to using a different absolute path here, if you have a better suggestion.

mtojek:
I think I can add more arguments to the list:
* used artifacts (like service logs, test results, etc.) should be scoped to the smallest possible range (e.g. package build scope?)
* we shouldn't mix files used by different application instances if we decide to introduce parallelism.

.ci/Jenkinsfile
mtojek: I suggest to ask @v1v for suggestion here, but it's fine for me. Maybe you can add a reference to the `elastic-package` in the middle (e.g. `/tmp/elastic-package/build/test-results`).

To fully understand the problem - what is the reason you wouldn't like to use the project's internal build directory now? Do you prefer to wait for the result of the discussion around (`~/.elastic-package`)?

ycombinator: I'm happy to use the project's internal build directory. I just don't know how to get the absolute path for it in CI.

v1v: In my opinion, I'd say everything about the generated files should be within the workspace context, nothing about the `HOME` but the workspace, then we don't need to worry about cleaning up those leftover files.

My ideal build process shoudl ook like:

```bash
$ git clone https://github.com/elastic/elastic-package
$ cd elastic-package
$ make clean check
$ find build/test-results -name *.xml -ls
```

`make clean` does remove the `build` folder if it exists, `make check` will do as usual but generated test report files will be created in the ./build/test-results folder

`.gitignore` will help to skip those folders if that's the concern.

ycombinator: So what is the way to generate the absolute path to the `build` folder in the workspace context? Specifically, I need to do it over here: https://github.com/elastic/elastic-package/pull/158/files#diff-cdc2dffbf82ed630cddb43c7641dc0534207dd69911283e6fa573efa299f73e6R21.

scripts/test-check-packages.sh
```diff
@@ -14,10 +14,14 @@ elastic-package stack up -d -v
...
+echo "***** DEBUGGING *****"
+echo $WORKSPACE
+echo "***** DEBUGGING *****"
+
 for d in test/packages/*/; do
   (
     cd $d
     elastic-package check -v
-    elastic-package test -v --report-format xUnit --report-output file
+    elastic-package test -v --report-format xUnit --report-output /tmp/build/test-results/
```
mtojek, line 25: Keep in mind that these tests should be also possible to run using `make check` locally.

ycombinator: Good point. I guess we _do_ have a case for a default location (related to the discussion in https://github.com/elastic/elastic-package/pull/158#issuecomment-720338982). I will revert the recent changes in this PR so the `--report-output` flag once again takes either `stdout` or `file`. And if it's `file`, it will simply write to the build folder determined by calling `FindBuildPackagesDirectory()`.

v1v:
```
REPORT_FOLDER=$(pwd)/build/test-results/
mkdir -p ${REPORT_FOLDER}

for d in test/packages/*/; do

   ...
  elastic-package test -v --report-format xUnit --report-output "${REPORT_FOLDER}"
```

.ci/Jenkinsfile
v1v: See https://github.com/elastic/elastic-package/pull/158/files#discussion_r515950425

scripts/test-check-packages.sh
ycombinator: Thanks @v1v. 

Chatting with @mtojek about this off-PR I realized that the build directory location is tightly coupled to the `elastic-package` tool. That is, it's baked in to the tool's code itself and not configurable from the outside. Putting aside whether this is good or bad, for now it makes sense to stay consistent with this model. 

As a result I'm going to update this PR again so that `--report-output` cannot take a path to the build directory. Instead it will take `file` as a valid value (as it used to before this PR), and if that value is detected, it will write the test results to the build directory that's baked into `elastic-package`.

.ci/Jenkinsfile
```diff
@@ -47,10 +47,11 @@ pipeline {
       }
       post {
         always {
-          archiveArtifacts(allowEmptyArchive: true, artifacts: '.elastic-package/tmp/test/reports/*.xml')
+          sh(label: 'List test results', script: 'pwd && ls -l && ls -l $(pwd)/src/github.com/elastic/elastic-package/build/test-results/')
+          archiveArtifacts(allowEmptyArchive: true, artifacts: '$(pwd)/src/github.com/elastic/elastic-package/build/test-results/*.xml')
```
mtojek, line 51: Did you try with `dir("${BASE_DIR}") { archiveArtifacts( artificats: 'relative' ) }`?

ycombinator: I will try this, thanks!

ycombinator: Looks like my last experiment worked. Turns out I just had a typo in my relative path (`build/build` instead of just `build`). So I'm going with it.

internal/install/install.go
```diff
@@ -92,11 +94,11 @@ func ServiceLogsDir() (string, error) {
 
 // TestReportsDir returns the location of the directory to store test reports.
 func TestReportsDir() (string, error) {
```
mtojek, line 96: I wonder if `TestReportsDir()` should be present in `install` which is related rather to `~/.elastic-package`. The `test-reports` directory won't be in that place anymore.

internal/builder/packages.go
```diff
@@ -55,6 +55,24 @@ func FindBuildPackagesDirectory() (string, bool, error) {
...
+func FindBuildPackagesDirectory() (string, bool, error) {
+	buildDir, found, err := FindBuildDirectory()
+	if err != nil {
+		return "", false, err
+	}
+
+	if found {
+		path := filepath.Join(buildDir, "integrations") // TODO add support for other package types
+		fileInfo, err := os.Stat(path)
+		if err == nil && fileInfo.IsDir() {
```
mtojek, line 68: nit: maybe refactor this block to two separate if-conditions.

.ci/Jenkinsfile
```diff
@@ -47,10 +47,10 @@ pipeline {
       }
       post {
         always {
-          archiveArtifacts(allowEmptyArchive: true, artifacts: '.elastic-package/tmp/test/reports/*.xml')
```
mdelapenya, line 50: I'd use wrap the block into a dir step, so that we capture the base dir (defined in https://github.com/elastic/elastic-package/blob/master/.ci/Jenkinsfile#L8):

```groovy
always {
          dir("${BASE_DIR}"){
                    archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/test-results/*.xml')
                    junit(allowEmptyResults: false,
                          keepLongStdio: true,
                          testResults: "build/test-results/*.xml")
          }
}
```

internal/install/install.go
ycombinator: Fair point, I'll move it out of that package.

.ci/Jenkinsfile
ycombinator: Thanks! Done in b1feb7c.

internal/builder/packages.go
ycombinator: Addressed in 5fe1126.

internal/install/install.go
ycombinator: Done in 2a4b779.

.ci/Jenkinsfile
ycombinator: @mdelapenya I made this change and it seems to have broken the build (it was green before the build). Did I do something wrong: b1feb7c?

ycombinator:
@mtojek @mdelapenya @v1v This PR is ready for re-review. Thanks!

internal/testrunner/reporters/outputs/file.go
```diff
@@ -56,3 +56,12 @@ func reportToFile(pkg, report string, format testrunner.TestReportFormat) error
 
 	return nil
 }
+
+// testReportsDir returns the location of the directory to store test reports.
+func testReportsDir() (string, error) {
```
mtojek, line 61: Might be a side issue - 

Is this directory cleaned up before a next run or does all test results accumulate in a single place?

mtojek (approved):
LGTM!

Are you going to introduce similar changes in https://github.com/elastic/integrations ?

v1v (approved):

ycombinator:
> Are you going to introduce similar changes in https://github.com/elastic/integrations ?

Yep, that's the plan (as usual).

internal/testrunner/reporters/outputs/file.go
ycombinator: Good question. I'm pretty sure the test results will just accumulate. I'm not sure if that's desired behavior or not. I can see both sides of the argument:
- it's good to let them accumulate because then we have results from previous runs around (and maybe `elastic-package clean` should clean them out), _or_
- we should clean up each time so only the latest results are available at any time.

Obviously, in a CI context this doesn't matter as we will have a fresh environment each time.

Unless you have a preference one way or the other, for now I'm okay with letting them accumulate. If we get feedback about this causing issues, we can address it then. WDYT?

mdelapenya: My thoughts on letting tests accumulate. I haven't seeing this behavior in my past experience in the Java world, where opinionated build systems such as Maven/Gradle define the life cycle. If the report-to-file option is enabled, defaulting to false using standard output, they simply refresh test reports on each test run, redirecting the standard output to the file (I always take a look at how Maven does the thing: http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#redirectTestOutputToFile).

If we do this in another way, I think it could lead to unexpected reports from previous executions in the local machine, although, as you said, in CI this shouldn't be a problem. But I'd prefer having a standard best practice, where the "oracles" (or the behaviours contributors expect -internal and external) are clear.

So I'd go with both approaches: `elastic-package clean` should clean the test reports, and the test command should start a new file.

ycombinator: Thanks for these insights, @mdelapenya — much appreciated!

Accordingly, I will update this PR to make the test command clean the test folder so it writes completely new files. We have a separate issue to implement `elastic-package clean`; I have made a note there: https://github.com/elastic/elastic-package/issues/13#issuecomment-722174887.

mdelapenya: Happy to help 😃

ycombinator:
@mdelapenya @mtojek I've updated this PR per the feedback in https://github.com/elastic/elastic-package/pull/158#discussion_r517820614. Please re-review when you get a chance. Thanks!

internal/testrunner/reporters/outputs/file.go
```diff
@@ -65,3 +56,28 @@ func testReportsDir() (string, error) {
 	}
 	return filepath.Join(buildDir, "test-results"), nil
 }
+
+// CleanTestReportsDir removes creates the test reports folder if it already exists, then
```
mdelapenya, line 60: Super nit:
```suggestion
// CleanTestReportsDir removes the test reports folder if it already exists, then
```

internal/testrunner/reporters/outputs/file.go
```diff
@@ -65,3 +56,28 @@ func testReportsDir() (string, error) {
...
+// creates it, effectively resulting in a clean (aka empty) test reports folder.
+func CleanTestReportsDir() error {
```
mdelapenya, line 62: I'm OK with the logic, although I'd like to know how the reports are generated. If the generation process always writes the reports to the proper place, creating or overwriting the content, then I wonder if the clean one should throw an error (ignoring them if they occur).

Again, my concern is already covered by this PR, so please fee free to merge and discuss in a follow-up not to block it.

Thanks!

ycombinator: Yeah, lets move forward with this PR and then address any more concerns as they come up when we start using this more.

mdelapenya (approved):

mtojek (approved):

Issue #160, opened 2020-11-02
Bug: redeploy package-registry on the live Elastic up


mtojek:
Steps to reproduce:

```
$ elastic-package stack up --services=elasticsearch -d
$ elastic-package stack up --services=package-registry -d
```

The second command restarts the entire stack, but shouldn't.

ycombinator:
I just ran `elastic-package stack up --services=package-registry -v` and the only container I have running now is the `package-registry`. Am I misunderstanding this bug?

mtojek:
Please to boot up Elasticsearch and then add the package-registry to the stack.

ycombinator:
Yes, sorry, I completely missed the first step 🤦 .

mtojek:
I see this as a blocker for redeploying changes while developing a new package (currently the restart of package-registry is required).

ycombinator:
I think you can generalize this bug. Booting up any component(s) of the stack destroys all previously-running components. I assume the desired behavior should be additive?

mtojek:
Yes or no.

If you redeploy Kibana without redeploying Elasticsearch, it simply won't work (due to encryption?).

PR #161, opened 2020-11-02
Fix: do not stop other containers than selected

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/160

### DIFF
--- a/internal/stack/boot.go
+++ b/internal/stack/boot.go
@@ -48,13 +48,6 @@ func BootUp(options Options) error {
 		return errors.Wrap(err, "building docker images failed")
 	}
 
-	if len(options.Services) == 0 { // stop Docker containers
-		err = dockerComposeDown()
-		if err != nil {
-			return errors.Wrap(err, "stopping docker containers failed")
-		}
-	}
-
 	err = dockerComposeUp(options)
 	if err != nil {
 		return errors.Wrap(err, "running docker-compose failed")
--- a/internal/stack/boot.go
+++ b/internal/stack/boot.go
@@ -48,9 +48,11 @@ func BootUp(options Options) error {
 		return errors.Wrap(err, "building docker images failed")
 	}
 
-	err = dockerComposeDown()
-	if err != nil {
-		return errors.Wrap(err, "stopping docker containers failed")
+	if len(options.Services) == 0 { // stop Docker containers
+		err = dockerComposeDown()
+		if err != nil {
+			return errors.Wrap(err, "stopping docker containers failed")
+		}
 	}
 
 	err = dockerComposeUp(options)


internal/stack/boot.go
```diff
@@ -48,9 +48,11 @@ func BootUp(options Options) error {
...
 	}
 
-	err = dockerComposeDown()
-	if err != nil {
-		return errors.Wrap(err, "stopping docker containers failed")
+	if len(options.Services) == 0 { // stop Docker containers
+		err = dockerComposeDown()
+		if err != nil {
+			return errors.Wrap(err, "stopping docker containers failed")
+		}
```
ycombinator, line 55: I understand the logic here but I have a broader question: what is the purpose of stopping all the elastic-package-controlled containers (when no specific services are provided)?

mtojek: I used it to replace a single ES instance with entire Kibana stack, but I assume it's a single use case.

mtojek: Another use case:

If you run `elastic-package stack up -d` twice, it will restart (stop and start), the entire stack. Do you suggest we should get rid of `dockerComposeDown()` at all?

ycombinator: I'm more wondering if we should make the logic of `elastic-package stack up` a bit smarter. Meaning, maybe we should look at what's running (including the versions) and compare it what the user has just requested to boot up (again, including the versions). And then do a diff between the two to figure out a) which services should be stopped, b) which ones can stay running as-is, and c) which ones should be booted up.

That should take care of all scenarios, I believe?

mtojek: Yes, it looks reasonable. I believe it can be simplified:

a) determine which services should be stopped (= ALL - (ones requested with `--services`))
b) use docker-compose to stop these services (if they're stopped, nothing happen)
c) use docker-compose to start services (if they're up, nothing will happen)

With this approach we won't introduce too much logic into elastic-package and we'll depend strictly on docker-compose behavior.

ycombinator: Sounds good to me!

mtojek: I implemented this logic locally, but unfortunately it doesn't meet requirements:

1. Having a fully deployed cluster (Elasticsearch, Kibana, Package Registry) restart only Package Registry.

I'm not sure how can I define it using CLI, as we define services to stay active: `elastic-package stack up -d -v -s elasticsearch,kibana,package-registry`.

mtojek: Summary:

I tried to implement the logic described above, but failed due unfulfilled requirement - restart of the package-registry only.

Justification for the current changes:

With the current changes in the PR, it's possible:

1. Stack the Elastic stack: `elastic-package stack up -d -v`.
2. Modify integration source (change package description?).
3. Build the integrations: `elastic-package build` (from inside the apache).
4. Restart package-registry: `elastic-package stack up -d -v -s package-registry` (from inside the Integrations repository or apache). 

Re 4: If you run `elastic-package stack up -d -v` it will not only restart the package-registry, but also other depending services (Kibana and Elastic-Agent), which is not an expected behavior.

mtojek:
I think I will merge the PR for the purpose of the session today (use case I described before). We can come back and iterate on this.

PR #162, opened 2020-11-04
Add pattern matching to field values

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/113

This PR adds pattern matching to field values. A pattern can specified in the field definition in the `fields.yml` file.

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -190,7 +190,19 @@ func parseElementValue(key string, definition fieldDefinition, val interface{}) 
 	var valid bool
 	switch definition.Type {
 	case "date", "ip", "constant_keyword", "keyword", "text":
-		_, valid = val.(string)
+		var valStr string
+		valStr, valid = val.(string)
+		if !valid || definition.Pattern == "" {
+			break
+		}
+
+		valid, err := regexp.MatchString(definition.Pattern, valStr)
+		if err != nil {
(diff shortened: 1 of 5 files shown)


mtojek:
Draft because of waiting for this one first: https://github.com/elastic/package-spec/pull/80

internal/fields/validate.go
```diff
@@ -190,7 +190,19 @@ func parseElementValue(key string, definition fieldDefinition, val interface{})
...
+		if !valid || definition.Pattern == "" {
+			break
+		}
+
+		valid, err := regexp.MatchString(definition.Pattern, valStr)
+		if err != nil {
+			return errors.Wrap(err, "invalid pattern")
+		}
+		if !valid {
+			return fmt.Errorf("field \"%s\" has invalid format, pattern: %s, actual: %s", key, definition.Pattern, valStr)
```
ycombinator, line 204: I think we need to start making our error messages, especially validation ones, more user-friendly (related: https://github.com/elastic/package-spec/issues/75). So I would suggest changing this to something like:

```suggestion
			return fmt.Errorf("field \"%s\"''s value, %s, does not match the expected pattern: %s", key, valStr, definition.Pattern)
```

mtojek: Good idea, I updated the PR.

ycombinator (approved):
LGTM.

Issue #148, opened 2020-10-27
Command: export Kibana assets


ycombinator:
New package developers will create assets such as visualizations, dashboards, etc. in Kibana for their packages. Then they will need to export these assets as JSON files from Kibana. 

We should provide a sub-command that does the following:
- Talks to Kibana and exports the relevant assets for the package.
- Creates the appropriate asset files in appropriate locations in the package.
- Prefixes each asset ID (including its references in other asset files) with the package name, all lowercased.

mtojek:
Related: https://github.com/elastic/elastic-package/issues/12

mtojek:
I'm going with this issue.

PR #165, opened 2020-11-09
Export dashboards from Kibana

mtojek:
Resolves: https://github.com/elastic/elastic-package/issues/148

This PR adds support for exporting dashboards and referenced objects from the Kibana instance.

Zrzut ekranu 2020-11-12 o 13 58 21

Sample usage:

Apache -
```bash
elastic-package export dashboards --id=apache-Logs-Apache-Dashboard-ecs
```

System-
```bash
elastic-package export dashboards --id=system-79ffd6e0-faa0-11e6-947f-177f697178b8
```

still TODOs:
- [x] Adjust titles (strip ECS suffix)
- [x] Adjust IDs to current naming convention
- [x] Adjust HTML links
- [ ] ~Test: export existing/installed dashboard~ I added a unit test to check the object transformation. Calling `elastic-package export` on real data adds `migrationVersion` fields to .json files, which depens on the current Kibana instance. If the Kibana version changes, it will fail these tests.


ycombinator:
Before we get to the code review, let's look at the developer experience. If I'm following correctly, a package developer will use this sub-command something like this:

1. Package developer creates multiple visualizations, dashboards, etc. in Kibana, as needed for their integration.
2. Package developer goes to Kibana > Stack Management > Kibana Saved Objects in the Kibana UI, then notes down the IDs of the dashboards they want to export.
3. Package developer runs `elastic-package export dashboards --id=<id_1>,...` or they run the command separately for each ID.

I wonder if we can offer a better developer experience here. How about we provide a TUI similar to the one you provided in `elastic-package promote`, where `elastic-package` calls a Kibana API for the _list_ of all available dashboards and presents a multiple-choice menu to select from. Each item would show the dashboard title (similar to what the user would see in step 2 in the Kibana UI above). Then the developer can select more comfortably the relevant dashboards, and then the `elastic-package` tool can go to work to export these from Kibana.

WDYT?

mtojek:
I thought about a different path:

1. Developer works on a dashboard.
2. Developer copies ID from the URL and uses it as command argument.

I think we can support both modes, either via command arg or TUI. I will adjust the code.

mtojek:
Done.

I've added a nice TUI prompt for selecting dashboards.

cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "finding dashboards failed")
+	}
+
+	if len(savedDashboards) == 0 {
+		return []string{}, nil
+	}
+
+	dashboardsPrompt := &survey.MultiSelect{
+		Message:  "Which dashboards would you like to export",
```
ycombinator, line 82:
```suggestion
		Message:  "Which dashboards would you like to export?",
```

internal/packages/packages.go
```diff
@@ -97,6 +97,19 @@ type DataStreamManifest struct {
 	} `config:"streams" json:"streams" yaml:"streams"`
 }
 
+// MustFindPackageRoot finds and returns the path to the root folder of a package.
+// It fails with an error if the package root can't be found.
+func MustFindPackageRoot() (string, error) {
```
ycombinator, line 102: 👍

cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if len(dashboardIDs) == 0 {
+		dashboardIDs, err = promptDashboardIDs(kibanaClient)
+		if err != nil {
+			return errors.Wrap(err, "prompt for dashboard selection failed")
+		}
+
+		if len(dashboardIDs) == 0 {
+			fmt.Println("No dashboard has been found.")
```
ycombinator, line 57: Should this be `found` or `selected`?

Also, let's make this error message actionable, i.e. tell the user what they need to do next to make the error go away.

cmd/export.go
```diff
@@ -0,0 +1,102 @@
...
+	err = export.Dashboards(kibanaClient, dashboardIDs)
+	if err != nil {
+		return errors.Wrap(err, "dashboards export failed")
+	}
+
+	cmd.Println("Done")
+	return nil
+}
+
+func promptDashboardIDs(kibanaClient *kibana.Client) ([]string, error) {
```
ycombinator, line 71: Should this be part of the `internal` package instead of the `cmd` package? Just wondering what our guiding principle is for when to put code in one vs. the other?

cmd/export.go
ycombinator, line 57:
```suggestion
			fmt.Println("No dashboards were found in Kibana.")
```

internal/export/dashboards.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func applyTransformations(ctx *transformationContext, objects []common.MapStr) ([]common.MapStr, error) {
+	return transformObjects(ctx, objects,
+		filterUnsupportedTypes,
+		decodeObject,
+		stripObjectProperties,
+		standardizeObjectProperties,
+		standardizeObjectID)
+}
```
ycombinator, line 63: I like what you have done here with the transformation functions. Most of them are specific to Kibana objects, so I'm wondering if we should define them under `internal/export/kibana` instead of just `internal/export`? I'm okay with doing this later when we have exports of non-Kibana assets.

ycombinator, line 63: Nit: this is probably just personal preference so feel free to ignore this comment. 

Looking at the signature and intended usage for `transformObjects` it seems like a good candidate for a fluent pattern. So something like:

```
return objectTransformer().
  withContext(ctx).
  withTransforms(filterUnsupportedTypes, decodeObject, ...).
  transform(objects)
```

internal/kibana/dashboards.go
```diff
@@ -0,0 +1,75 @@
...
+// Export method exports selected dashboards using the Kibana Export API.
+func (c *Client) Export(dashboardIDs []string) ([]common.MapStr, error) {
+	logger.Debug("Export dashboards using the Kibana Export API")
+
+	var query strings.Builder
+	query.WriteByte('?')
+	for _, dashboardID := range dashboardIDs {
+		query.WriteString("dashboard=")
+		query.WriteString(dashboardID)
+		query.WriteByte('&')
```
ycombinator, line 33: Not a huge deal but wouldn't you end up with a trailing `&`?

internal/kibana/saved_objects.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	sort.Slice(foundObjects, func(i, j int) bool {
+		return sort.StringsAreSorted([]string{strings.ToLower(foundObjects[i].Title), strings.ToLower(foundObjects[j].Title)})
+	})
+	return foundObjects, nil
+}
+
+func (c *Client) findDashboardsNextPage(page int) (*savedObjectsResponse, error) {
+	request, err := http.NewRequest(http.MethodGet, fmt.Sprintf("%s%s%d", c.host, "/api/saved_objects/_find?type=dashboard&fields=title&per_page=5&page=", page), nil)
```
ycombinator, line 94: If you try to export dashboards from the Kibana UI it gives the option of also exporting any related objects (visualizations, etc.). Is that option already enabled with this API call or do we need an additional parameter for it?

ycombinator, line 94: Any particular reason to restrict ourselves to 5 dashboards per page?

Also, regardless of what number we choose, let's make it a `const`.

ycombinator (commented):
Left a few comments/questions but really nice work overall!

cmd/export.go
mtojek: Actually I vote for `cmd` as it interacts with `stdin`. The same design principle is applied to `promote`.

internal/kibana/dashboards.go
mtojek: Yes, but it doesn't mean in terms of processing.

internal/kibana/saved_objects.go
mtojek: I didn't know that, but it means that it's enabled by default :) The API call exports also related objects.

cmd/export.go
mtojek: Fixed

cmd/export.go
mtojek: Fixed

internal/kibana/saved_objects.go
mtojek: Fixed

internal/export/dashboards.go
mtojek: Hm.. maybe postpone it for later once we have a conflict between data sources. Right now there is Kibana only, probably when we have Elasticsearch we'll have to adjust it and maybe change the package API.

mtojek: Good idea. I modified the source code to the suggested form.

ycombinator (approved):
LGTM. This is going to be a super useful feature!

PR #167, opened 2020-11-10
Refactoring: move test errors out of test runners folder

ycombinator:
We should keep the `runners` folder for implementations of various test runners. So this PR moves the error definition code out of this folder.


mtojek (approved):
LGTM

Issue #163, opened 2020-11-05
Refactor: pass IngestManagerClient via dependency injection


mtojek:
The `IngestManagerClient` should be passed similarly to `ESClient` via options.

jsoriano:
Elasticsearch clients are already injected from the commands. We would be missing the injection of some kibana clients:
```
internal/servicedeployer/kubernetes.go: kibanaClient, err := stack.NewKibanaClient()
internal/stack/clients.go:func NewKibanaClient(customOptions ...kibana.ClientOption) (*kibana.Client, error) {
internal/testrunner/runners/asset/runner.go:    kibanaClient, err := stack.NewKibanaClient()
internal/testrunner/runners/system/runner.go:   kib, err := stack.NewKibanaClient()
```

PR #168, opened 2020-11-10
Asset loading test runner

### LINKED ISSUE #115: Asset loading tests
This test should test if all assets in a package are loaded up as expected. Assets include: Elasticsearch index templates, Elasticsearch ILM policies, Elasticsearch ingest pipelines, Kibana visualizations, Kibana saved searches, and Kibana dashboards.

Scope: per package dataset.

Type of test: integration.

Components involved: Kibana; Elasticsearch; Package Registry?

Process: load dataset assets; call Elasticsearch and Kibana APIs to retrieve loaded assets.

Assertions: assets retrieved from APIs match those expected to be loaded.

Package author must define: nothing.

Tooling needed: test runner in `elastic-package` tool.

ycombinator:
This PR implements a new type of test runner: `asset`. This test runner tests whether a package's assets (ES index templates, Kibana dashboards, etc.) are loaded up as expected.

## Sample usage

```
$ elastic-package test asset
Run asset tests for the package
--- Test results for package: apache - START ---
╭─────────┬─────────────┬───────────┬────────────────────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME                                                      │ RESULT │ TIME ELAPSED │
├─────────┼─────────────┼───────────┼────────────────────────────────────────────────────────────────┼────────┼──────────────┤
│ apache  │             │ asset     │ dashboard apache-Logs-Apache-Dashboard is loaded               │ PASS   │ 6.101128287s │
│ apache  │             │ asset     │ dashboard apache-Metrics-Apache-HTTPD-server-status is loaded  │ PASS   │        927ns │
│ apache  │             │ asset     │ visualization apache-HTTPD-CPU is loaded                       │ PASS   │        620ns │
│ apache  │             │ asset     │ visualization apache-HTTPD-Hostname-list is loaded             │ PASS   │        512ns │
│ apache  │             │ asset     │ visualization apache-HTTPD-Load1-slash-5-slash-15 is loaded    │ PASS   │        526ns │
│ apache  │             │ asset     │ visualization apache-HTTPD-Scoreboard is loaded                │ PASS   │        430ns │
│ apache  │             │ asset     │ visualization apache-HTTPD-Total-accesses-and-kbytes is loaded │ PASS   │        424ns │
│ apache  │             │ asset     │ visualization apache-HTTPD-Uptime is loaded                    │ PASS   │        416ns │
│ apache  │             │ asset     │ visualization apache-HTTPD-Workers is loaded                   │ PASS   │        403ns │
│ apache  │             │ asset     │ visualization apache-access-unique-IPs-map is loaded           │ PASS   │        404ns │
│ apache  │             │ asset     │ visualization apache-browsers is loaded                        │ PASS   │        413ns │
│ apache  │             │ asset     │ visualization apache-error-logs-over-time is loaded            │ PASS   │        410ns │
│ apache  │             │ asset     │ visualization apache-operating-systems is loaded               │ PASS   │        407ns │
│ apache  │             │ asset     │ visualization apache-response-codes-of-top-URLs is loaded      │ PASS   │        439ns │
│ apache  │             │ asset     │ visualization apache-response-codes-over-time is loaded        │ PASS   │        453ns │
│ apache  │             │ asset     │ search apache-HTTPD is loaded                                  │ PASS   │        598ns │
│ apache  │             │ asset     │ search apache-access-logs is loaded                            │ PASS   │        420ns │
│ apache  │             │ asset     │ search apache-errors-log is loaded                             │ PASS   │        514ns │
│ apache  │ access      │ asset     │ index_template logs-apache.access is loaded                    │ PASS   │        490ns │
│ apache  │ access      │ asset     │ ingest_pipeline logs-apache.access-0.2.7 is loaded             │ PASS   │        462ns │
│ apache  │ error       │ asset     │ index_template logs-apache.error is loaded                     │ PASS   │        405ns │
│ apache  │ error       │ asset     │ ingest_pipeline logs-apache.error-0.2.7 is loaded              │ PASS   │        434ns │
│ apache  │ status      │ asset     │ index_template metrics-apache.status is loaded                 │ PASS   │        440ns │
╰─────────┴─────────────┴───────────┴────────────────────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: apache - END   ---
Done
```

### TODO
* [x] Install package
* [x] Remove package
* [x] Add assertions on asset existence
* [x] Move `--data-stream` global test flag to pipeline and system tests
* [x] Add HOWTO doc
* [x] Add to README doc
* [x] Update CI configuration

Resolves #115.


ycombinator:
@mtojek This PR is ready for review, when you get a chance.

Before this PR we had 2 test runners, both of which operated at the data stream level and needed test configuration. This PR introduces a test runner that operates at the package level and also doesn't need test configuration. As a result, the code for the test runner framework is now a bit of a mess IMO. I would like to clean this up but in a follow up PR. WDYT?

internal/kibana/ingestmanager/client_packages.go
```diff
@@ -0,0 +1,46 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/packages"
+)
+
+// InstallPackage installs the given package in Fleet.
+func (c *Client) InstallPackage(pkg packages.PackageManifest) ([]packages.Asset, error) {
```
mtojek, line 17: We may need to unify our approaches to Kibana "subclients" as it looks like we duplicate some code. With this approach: https://github.com/elastic/elastic-package/blob/master/internal/kibana/client.go I encapsulated `dashboards` API in a separate `.go` file which shared a Kibana client. Do you have any preferences here?

The reason why I insist on the other approach is that `dashboards` API is not included within the `fleet` API (btw. we should rename it to fleet) and at some point we need to unify approaches.

Let's discuss the final design, so that this client is added in the right place in the source tree and we can adjust the rest of code.

internal/kibana/ingestmanager/client_packages.go
```diff
@@ -0,0 +1,46 @@
...
+	return managePackage(pkg, "install", c.post)
+}
+
+// RemovePackage removes the given package from Fleet.
+func (c *Client) RemovePackage(pkg packages.PackageManifest) ([]packages.Asset, error) {
+	return managePackage(pkg, "remove", c.delete)
```
mtojek, line 23: I have a gut feeling that clients in `ingestmanager` are too much complex. Not sure if we need as many layers to process a single request. Maybe not depending on common abstraction would be simpler?

Please take a look at https://github.com/elastic/elastic-package/blob/master/internal/kibana/saved_objects.go . In this case I need to implement paging and the exact code responsible for calls to Kibana is wrapped up in ca. 20 LOCs: https://github.com/elastic/elastic-package/blob/master/internal/kibana/saved_objects.go#L95-L118

I'm happy to discuss it!

internal/packages/assets.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	assets = append(assets, a...)
+
+	a, err = loadFileBasedAssets(kibanaAssetsFolderPath, AssetTypeKibanaSavedSearch)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not load kibana saved search assets")
+	}
+	assets = append(assets, a...)
+
+	a, err = loadFileBasedAssets(kibanaAssetsFolderPath, AssetTypeKibanaMap)
```
mtojek, line 74: Thinking loud: shouldn't elastic-package discover available types from the package directory? In case we add a new type the command will still operate (open-closed principle).

internal/packages/assets.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "reading package manifest file failed")
+	}
+
+	dataStreamManifestPaths, err := filepath.Glob(filepath.Join(pkgRootPath, "data_stream", "*", DataStreamManifestFile))
+	if err != nil {
+		return nil, errors.Wrap(err, "could not read data stream manifest file paths")
+	}
+
+	assets := make([]Asset, 0)
```
mtojek, line 95: `var assets []Asset`

internal/packages/assets.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error finding kibana %s assets folder", assetType)
+	}
+
+	files, err := ioutil.ReadDir(assetsFolderPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not read %s files", assetType)
+	}
+
+	assets := make([]Asset, 0)
```
mtojek, line 143: `var assets []Asset`

internal/testrunner/common.go
```diff
@@ -0,0 +1,40 @@
...
+		Password string
+	}
+	Kibana struct {
+		Host string
+	}
+}
+
+// GetStackSettingsFromEnv loads Elastic stack connnection settings from
+// environment variables and returns them.
+func GetStackSettingsFromEnv() StackSettings {
```
mtojek, line 23: I think this is a good moment to put it in the `stack` package and expose it as `stack.CurrentSettings()`.

internal/testrunner/runners/asset/runner.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// String returns the name of the test runner.
+func (r runner) String() string {
+	return "asset loading"
+}
+
+// CanRunPerDataStream returns whether this test runner can run on individual
+// data streams within the package.
+func (r runner) CanRunPerDataStream() bool {
```
mtojek, line 52: Let's keep it consistent with README: https://github.com/elastic/elastic-package#elastic-package-test . A context can be global, package or data stream.

internal/testrunner/runners/asset/runner.go
```diff
@@ -0,0 +1,167 @@
...
+		return []testrunner.TestResult{tr}, err
+	}
+
+	pkgManifest, err := packages.ReadPackageManifest(filepath.Join(r.packageRootPath, packages.PackageManifestFile))
+	if err != nil {
+		return resultsWith(result, errors.Wrap(err, "reading package manifest failed"))
+	}
+
+	// Install package
+	im, err := ingestmanager.NewClient(r.stackSettings.Kibana.Host, r.stackSettings.Elasticsearch.Username, r.stackSettings.Elasticsearch.Password)
```
mtojek, line 101: https://github.com/elastic/elastic-package/issues/163

test/packages/apache/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 1.0.0
 name: apache
 title: Apache
-version: 0.1.4
+version: 0.2.7
```
mtojek, line 4: Is this change on purpose?

scripts/test-check-packages.sh
```diff
@@ -10,6 +10,13 @@ cleanup() {
 
 trap cleanup EXIT
 
+for d in test/packages/*/; do
+  (
+    cd $d
+    elastic-package build -v
```
mtojek, line 16: Could you please double-check if the `check` command doesn't include `build` step?

internal/testrunner/testrunner.go
```diff
@@ -141,26 +148,34 @@ func FindTestFolders(packageRootPath string, testType TestType, dataStreams []st
...
-	var testTypes []TestType
-	for t := range runners {
-		testTypes = append(testTypes, t)
+// TestRunners method returns registered test runners.
+func TestRunners() map[TestType]TestRunner {
+	return runners
+}
+
+// NoConfigTestRunners returns test runners that don't need configuration.
+func NoConfigTestRunners() []TestRunner {
```
mtojek, line 170: I admit I got lost here, why some test runners are defined differently (as no-config runners). I assume we can iterate on this in follow-ups.

internal/packages/assets.go
ycombinator: Changed in f027e01.

internal/packages/assets.go
ycombinator: Changed in f027e01.

internal/testrunner/common.go
ycombinator: Done in 23fb729.

internal/kibana/ingestmanager/client_packages.go
ycombinator: Agreed. In the beginning all we had was the ingestmanager (now fleet) client. But now that we have a more general kibana client, we should unify them. I took at stab at this in #210.

mtojek (commented):
@ycombinator please let me know whether I can take another look at this PR.

ycombinator:
> please let me know whether I can take another look at this PR.

Yeah, I'm still iterating on it. I'll re-request your review when it's ready for you to take a look. Thanks!

internal/packages/assets.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return assets, nil
+}
+
+func loadKibanaAssets(pkgRootPath string) ([]Asset, error) {
+	kibanaAssetsFolderPath := filepath.Join(pkgRootPath, "kibana")
+
+	assets, err := loadFileBasedAssets(kibanaAssetsFolderPath, AssetTypeKibanaDashboard)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not load kibana dashboard assets")
```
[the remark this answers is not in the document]
ycombinator: Done in 6bda6ef.

internal/testrunner/runners/asset/runner.go
ycombinator: I'm not sure I follow what to do here. Would you mind elaborating?

In my mind, `elastic-package test` is a command with `package` context. It just allows for extra granularity, if the user desires, via the `--data-stream` flag.

internal/testrunner/testrunner.go
ycombinator: Yes, this was something I was experimenting with early on in this PR but is no longer necessary. I've removed it in 0e527da. Thanks for flagging it!

test/packages/apache/manifest.yml
ycombinator: Yes, I had to make it because the package registry image loaded up by the tests installs Apache version 0.2.6. So, in order for the asset test runner to be able to install the local apache package (i.e. the one being tested), I had to ensure that the local apache package's version was greater than the one in the registry. Otherwise the Install Package API would not allow the local package to be loaded up.

I admit that this is not a very maintainable change. Every time the package registry image is refreshed with a newer version of `apache`, we will need to bump up the version here as well. Maybe we should just set it here to something like `999.999.999`?

ycombinator:
I see that CI is failing. I will look into this tomorrow.

test/packages/apache/manifest.yml
mtojek: Oh, I see :) In this case you can set it to sth weird like 999. (at least the major). Nice catch!

internal/testrunner/runners/asset/runner.go
mtojek: I'm sorry but I can't remind myself what I was thinking about... Let's just ignore this comment.

mtojek (approved):
The implementation looks good to me.

Let's move forward with this PR. Did you take a chance to run it against real integrations?

nit: just rebase the PR against master branch

ycombinator:
Seeing this error now:

```
$ elastic-package test -v --report-format xUnit --report-output file --defer-cleanup 1s
2021/01/14 15:49:34 DEBUG Enable verbose logging
Run test suite for the package
Run asset tests for the package
2021/01/14 15:49:34 DEBUG installing package...
2021/01/14 15:49:34 DEBUG POST http://127.0.0.1:5601/api/fleet/epm/packages/apache-999.999.999
2021/01/14 15:49:34 DEBUG
2021/01/14 15:49:43 DEBUG waiting for 1s before tearing down...
Error: error parsing --data-streams flag: flag accessed but not defined: data-streams
```

Will look into it tomorrow.

ycombinator:
Hi @mtojek, I made a few small changes since your last review to make CI pass. Could you please take another look when you get a chance? Thanks!

internal/packages/assets.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Supported asset types.
+const (
+	AssetTypeElasticsearchIndexTemplate  AssetType = "index_template"
+	AssetTypeElasticsearchIngestPipeline AssetType = "ingest_pipeline"
+
+	AssetTypeKibanaSavedSearch   AssetType = "search"
+	AssetTypeKibanaVisualization AssetType = "visualization"
+	AssetTypeKibanaDashboard     AssetType = "dashboard"
+	AssetTypeKibanaMap           AssetType = "map"
```
mtojek, line 30: + Lens

internal/testrunner/runners/system/runner.go
```diff
@@ -51,17 +51,6 @@ type runner struct {
 	wipeDataStreamHandler   func() error
 }
 
-type stackSettings struct {
```
mtojek, line 54: Is the `stackSettings` (old and new) structure used anywhere? It looks like a leftover after refactoring. All stack settings like host, login, password don't need to be exposed beyond ES/Kibana clients.

internal/packages/packages.go
```diff
@@ -200,6 +205,15 @@ func ReadDataStreamManifest(path string) (*DataStreamManifest, error) {
 	return &m, nil
 }
 
+// GetPipelineNameOrDefault returns the name of the data stream's pipeline, if one is explicitly defined in the
+// data stream manifest. If not, the default pipeline name is returned.
+func (dsm *DataStreamManifest) GetPipelineNameOrDefault() string {
```
mtojek, line 210: 👍 for refactoring

internal/packages/packages.go
```diff
@@ -223,5 +237,5 @@ func isDataStreamManifest(path string) (bool, error) {
 	if err != nil {
 		return false, errors.Wrapf(err, "reading package manifest failed (path: %s)", path)
 	}
-	return m.Title != "" && (m.Type == "logs" || m.Type == "metrics"), nil
+	return m.Title != "" && (m.Type == dataStreamTypeLogs || m.Type == dataStreamTypeMetrics), nil
```
mtojek, line 240: nit: Actually this function isn't perfect. If there is another dataset type, it will simply fail (OCP broken). I wonder if there is a different way to recognize the DataStream. If you find any, feel free to adjust the implementation.

mtojek (approved):
Thank you for the heads up! I left some minor comments, but wouldn't like to block your PR from merging. These ones can be addressed in follow up PRs (up to you).

I'm thrilled to see this running in the Integrations. Nice job!

internal/testrunner/runners/system/runner.go
ycombinator: Good catch. I've removed all stack settings code in cef5f22e598073727598a02d7324ffe5484f4f1c. So now the ES/Kibana clients directly call `os.Getenv` in their constructors to get the stack settings.

internal/packages/assets.go
ycombinator: Addressed in 133e50ee0364bb76f289086a0ca4aa16b92c6967.

internal/packages/packages.go
ycombinator: Yeah, it's not perfect but I think this check here might have some good side-effects. In the asset test runner we check if we are dealing with a `logs` data stream and, if so, check that ingest pipeline assets have also been loaded. Similarly, whenever a new data stream type is added in the future we might need to have special handling for it in `elastic-package`. So being strict here about which data stream types `elastic-package` knows how to handle may be a good thing. 

BTW, I was checking the data stream manifest spec to see if we have defined the data stream `type` field as an `enum` and we have. However, while doing this I noticed that there is a data stream type we have not considered in `elastic-package`: [`traces`](https://github.com/elastic/package-spec/blob/4e314c0a650da34a8f2d6f104dd659664b5377f8/versions/1/data_stream/manifest.spec.yml#L117-L120). It was added for APM in https://github.com/elastic/package-spec/pull/78. I'm guessing the team is not using `elastic-package` yet so they are not running into any issues with this type not being supported. However, I've created https://github.com/elastic/elastic-package/issues/223 to track adding support for this type in `elastic-package`.

ycombinator:
@mtojek I've addressed all your feedback. Please re-review when you get a chance. If the PR looks good to you, please merge it on my behalf as I'll be away from work for the next week. Otherwise I will address any further feedback when I'm back. Thanks!

mtojek (approved):

Issue #169, opened 2020-11-10
[Pipeline tests] Add the ability to ignore fields with dynamic values


marc-gr:
With pipelines that populate some fields dynamically (e.g. : `event.ingested` is populated with `_ingest.timestamp`) pipeline tests will fail, since there is no way with the current config options to ignore this kind of fields.

Would be nice to add some config option that would help in this situations, either by ignoring the field completely, or just by checking its existence and not caring about the value, for example.

An example log line and pipeline to be able to test this is

auditd.log
```log
type=MAC_IPSEC_EVENT msg=audit(1485893834.891:18877201): op=SPD-delete auid=4294967295 ses=4294967295 res=1 src=192.168.2.0 src_prefixlen=24 dst=192.168.0.0 dst_prefixlen=16
```

pipeline.yml
```yaml
---
description: Pipeline for parsing Linux auditd logs
processors:
- set:
    field: event.ingested
    value: '{{_ingest.timestamp}}'
- grok:
    field: message
    pattern_definitions:
      AUDIT_TYPE: "type=%{NOTSPACE:auditd.log.record_type}"
      AUDIT_NODE: "node=%{IPORHOST:auditd.log.node} "
      AUDIT_PREFIX: "^(?:%{AUDIT_NODE})?%{AUDIT_TYPE} msg=audit\\(%{NUMBER:auditd.log.epoch}:%{NUMBER:auditd.log.sequence}\\):(%{DATA})?"
      AUDIT_KEY_VALUES: "%{WORD}=%{GREEDYDATA}"
    patterns:
    - "%{AUDIT_PREFIX} %{AUDIT_KEY_VALUES:auditd.log.kv} old auid=%{NUMBER:auditd.log.old_auid}
      new auid=%{NUMBER:auditd.log.new_auid} old ses=%{NUMBER:auditd.log.old_ses}
      new ses=%{NUMBER:auditd.log.new_ses}"
    - "%{AUDIT_PREFIX} %{AUDIT_KEY_VALUES:auditd.log.kv} msg=['\"]%{AUDIT_KEY_VALUES:auditd.log.sub_kv}['\"]"
    - "%{AUDIT_PREFIX} %{AUDIT_KEY_VALUES:auditd.log.kv}"
    - "%{AUDIT_PREFIX}"
    - "%{AUDIT_TYPE} %{AUDIT_KEY_VALUES:auditd.log.kv}"
- kv:
    field: auditd.log.kv
    field_split: "\\s+"
[...]

PR #171, opened 2020-11-10
Add support for dynamic fields

mtojek:
Resolves: https://github.com/elastic/elastic-package/issues/169

This PR modifies pipeline test runner to verify dynamic fields.

I'm opening this as draft as I need changes in https://github.com/elastic/package-spec/pull/82 .

### DIFF
--- a/docs/howto/pipeline_testing.md
+++ b/docs/howto/pipeline_testing.md
@@ -75,15 +75,17 @@ The pipeline test runner transforms log lines into file input events and sends t
             "web"
         ]
     },
-    "ignored_fields": [
-        "event.ingested"
-    ]
+    "dynamic_fields": {
+        "url.original": "^/.*$"
+    }
 }
 ```
 
 The `multiline` section configures the log file reader to correctly detect multiline log entries using the `first_line_pattern`. Use this property if your logs may be split into multiple lines, e.g. Java stack traces.
 
 The `fields` section allows for customizing extra fields to be added to every read log entry (e.g. `@timestamp`, `ecs`). Use this property to extend your logs with data that can't be extracted from log content, but it's fine to have same field values for every record (e.g. timezone, hostname).
+
+The `dynamic_fields` section allows for marking fields as dynamic (every time they have different non-static values), so that pattern matching instead of strict value check is applied. 
 
 #### Input events
 
--- a/docs/howto/pipeline_testing.md
+++ b/docs/howto/pipeline_testing.md
@@ -74,7 +74,10 @@ The pipeline test runner transforms log lines into file input events and sends t
         "event.category": [
             "web"
         ]
-    }
+    },
+    "ignored_fields": [
+        "event.ingested"
+    ]
 }
 ```
 
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -226,7 +226,7 @@ func verifyDynamicFields(result *testResult, config *testConfig) error {
 			}
 
 			if !matched {
-				multiErr = append(multiErr, fmt.Errorf("dynamic field \"%s\" doesn't match the pattern \"%s\": %s",
+				multiErr = append(multiErr, fmt.Errorf("dynamic field \"%s\" doesn't match the pattern (%s): %s",
 					key, pattern, valStr))
 			}
 		}
@@ -235,7 +235,7 @@ func verifyDynamicFields(result *testResult, config *testConfig) error {
 	if len(multiErr) > 0 {
 		return testrunner.ErrTestCaseFailed{
 			Reason:  "one or more problems with dynamic fields found in documents",
-			Details: multiErr.Error(),
+			Details: multiErr.Unique().Error(),
 		}
 	}
 	return nil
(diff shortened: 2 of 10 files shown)


mtojek:
A couple of updates here:
1. Updated depedency on the package-spec (support for dynamic fields).
2. Pipeline tests: expected results in .json files don't contain dynamic fields (as they take variable values).
3. Pipeline tests: once static parts of results are compared (diff), there is another check for dynamic fields with patterns.
4. Pipeline tests: pattern matching for dynamic fields works only for string based values (in other cases it's skipped).

Sample failing output:

```bash
➜  nginx git:(169-ignore-fields) ✗ elastic-package test pipeline -v
2020/11/12 20:44:08 DEBUG Enable verbose logging
--- Test results for package: nginx - START ---
╭─────────┬────────────────────┬───────────┬────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────┬──────────────╮
│ PACKAGE │ DATA STREAM        │ TEST TYPE │ TEST NAME              │ RESULT                                                                              │ TIME ELAPSED │
├─────────┼────────────────────┼───────────┼────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┼──────────────┤
│ nginx   │ access             │ pipeline  │ test-access-event.json │ PASS                                                                                │ 132.061159ms │
│ nginx   │ access             │ pipeline  │ test-access-raw.log    │ FAIL: test case failed: one or more problems with dynamic fields found in documents │ 127.889974ms │
│ nginx   │ access             │ pipeline  │ test-access-single.log │ PASS                                                                                │ 123.057553ms │
│ nginx   │ error              │ pipeline  │ test-error-raw.log     │ PASS                                                                                │ 120.399112ms │
│ nginx   │ ingress_controller │ pipeline  │ test-ingest-raw.log    │ PASS                                                                                │  127.86579ms │
╰─────────┴────────────────────┴───────────┴────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────┴──────────────╯

FAILURE DETAILS:

nginx/access test-access-raw.log:
[0] dynamic field "url.original" doesn't match the pattern (^fake.*$): /
[1] dynamic field "url.original" doesn't match the pattern (^fake.*$): /adsasd
[2] dynamic field "url.original" doesn't match the pattern (^fake.*$): /favicon.ico
[3] dynamic field "url.original" doesn't match the pattern (^fake.*$): /taga
[4] dynamic field "url.original" doesn't match the pattern (^fake.*$): /test
[5] dynamic field "url.original" doesn't match the pattern (^fake.*$): /test1
--- Test results for package: nginx - END   ---
Done
Error: one or more test cases failed
```

docs/howto/pipeline_testing.md
```diff
@@ -74,7 +74,10 @@ The pipeline test runner transforms log lines into file input events and sends t
         "event.category": [
             "web"
         ]
-    }
+    },
+    "ignored_fields": [
+        "event.ingested"
+    ]
```
ycombinator, line 80: I think this section needs to be updated?

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -158,31 +161,86 @@ func (r *runner) loadTestCaseFile(testCaseFile string) (*testCase, error) {
...
+	if err != nil {
+		return err
+	}
+
 	err = verifyFieldsInTestResult(result, fieldsValidator)
 	if err != nil {
 		return err
 	}
 	return nil
 }
```
ycombinator, line 197: In this implementation it looks like the written test result will not contain dynamic fields. I think it would be nice if the written test results were exactly the same as we get from the Simulate API, that way the user is not surprised by certain fields are missing from it.

Would it be possible to keep the dynamic fields in the test results, write them as-is? Maybe `compareResults` could exclude dynamic fields in the comparison or something like that?

ycombinator:
> Issue: #169

Consider using `Resolves: #169` or similar so we don't end up with dangling open issues even after they have actually been resolved.

internal/testrunner/runners/pipeline/runner.go
mtojek: It should be possible, but we need to keep in mind that such fields will change their values every time the results are regenerated.

ycombinator: Personally I think that's okay - we see similar behavior with `data.json` files in Beats when they are regenerated and, if anything, I think it will get some assurance that the tests did indeed run and refreshed the files with data from the latest run.

mtojek: Fixed

docs/howto/pipeline_testing.md
mtojek: Yes, thanks! Fixed

docs/howto/pipeline_testing.md
```diff
@@ -82,6 +85,8 @@ The `multiline` section configures the log file reader to correctly detect multi
 
 The `fields` section allows for customizing extra fields to be added to every read log entry (e.g. `@timestamp`, `ecs`). Use this property to extend your logs with data that can't be extracted from log content, but it's fine to have same field values for every record (e.g. timezone, hostname).
 
+The `dynamic_fields` section allows for marking fields as dynamic (every time they have different non-static values), so that pattern matching instead of strict value check is applied.
```
ycombinator, line 88: ❤️

ycombinator (approved):
LGTM.

Issue #151, opened 2020-10-28
Use "build" command to build README files


mtojek:
To strip `GenerateDocs` step from Integrations, we should support it here.

PR #174, opened 2020-11-18
Build command: render README files

mtojek:
Use build command to render README files from README templates.

Issue: https://github.com/elastic/elastic-package/issues/151

TODOs:
- [x] Check if README is up-to-date
- [x] Render sample event
- [x] Render exported fields


internal/packages/packages.go
```diff
@@ -163,6 +163,11 @@ func FindDataStreamRootForPath(workDir string) (string, bool, error) {
 	return "", false, nil
 }
 
+// ReadPackageManifestForPackage reads and parses the package manifest file for the given package.
+func ReadPackageManifestForPackage(packageRoot string) (*PackageManifest, error) {
```
ycombinator, line 167: Nit: this function name reads a bit awkwardly. How about `ReadPackageManifestFromPackageRoot`?

cmd/lint.go
```diff
@@ -33,6 +37,14 @@ func lintCommandAction(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "locating package root failed")
 	}
 
+	ok, err := docs.IsReadmeUpToDate()
+	if err != nil {
+		return errors.Wrapf(err, "can't check if %s file is up-to-date", docs.ReadmeFile)
+	}
+	if !ok {
+		return fmt.Errorf("%s file is outdated. Rebuild the package with 'elastic-package build'", docs.ReadmeFile)
```
ycombinator, line 45: ❤️ Love the actionable error message here!

cmd/build.go
```diff
@@ -24,7 +25,12 @@ func setupBuildCommand() *cobra.Command {
...
-	err := builder.BuildPackage()
+	err := docs.UpdateReadme()
+	if err != nil {
+		return errors.Wrapf(err, "updating %s file failed", docs.ReadmeFile)
+	}
+
+	err = builder.BuildPackage()
 	if err != nil {
 		return errors.Wrap(err, "building package failed")
 	}
```
ycombinator, line 36: This is a more general comment, applicable to all `elastic-package` commands I suppose. 

When a command completes successfully, how do you feel about printing to STDOUT any side effects the command has had so the user knows where to look for things changed/created by the command? For example, lets take this command here: `elastic-package build`. Now it has two side effects: update the README and build the package. WDYT about printing for the user that the command completed and did these two things, maybe even showing the relevant paths?

internal/packages/packages.go
mtojek: Renamed.

cmd/build.go
mtojek: Ok, it sounds as a reasonable idea. I added relevant `Println`s.

For reference:

```
$ elastic-package build
Build the package
README.md file rendered: /Users/marcin.tojek/go/src/github.com/elastic/elastic-package/test/packages/apache/docs/README.md
Package built: /Users/user/go/src/github.com/elastic/elastic-package/build/integrations/apache/0.1.4
Done
$ cd ../nginx
$ cd elastic-package build
Build the package
Package built: /Users/user/go/src/github.com/elastic/elastic-package/build/integrations/nginx/0.2.4
Done
```

ycombinator (approved):
This is very nice! LGTM.

PR #180, opened 2020-11-23
Refactoring test runner

ycombinator:
This PR defines a `TestRunner` interface. All test runners must implement this interface. This will give consumers of test runners access to their lifecycle methods (`Run` and `TearDown`) as well as to informational methods (`Type` and `String`).

Related: https://github.com/elastic/elastic-package/pull/175#issuecomment-730549015
Related: #168

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -36,12 +36,10 @@ const (
 )
 
 type runner struct {
-	testFolder      testrunner.TestFolder
-	packageRootPath string
-	stackSettings   stackSettings
-	esClient        *es.Client
+	options       testrunner.TestOptions
+	stackSettings stackSettings
 
-	// Execution order of following handlers is defined in runner.tearDown() method.
+	// Execution order of following handlers is defined in runner.TearDown() method.
 	deleteTestPolicyHandler func()
 	resetAgentPolicyHandler func()
 	shutdownServiceHandler  func()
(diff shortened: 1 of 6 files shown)


internal/testrunner/testrunner.go
```diff
@@ -34,10 +30,23 @@ type TestOptions struct {
...
+
+	// String returns the human-friendly name of the test runner.
+	String() string
+
+	// Run executes the test runner.
+	Run(TestOptions) ([]TestResult, error)
+
+	// TearDown cleans up any test runner resources. It must be called
+	// after the test runner has finished executing.
+	TearDown()
```
mtojek, line 46: is it fair for `TearDown` method to return an error?

mtojek (approved):
The CI failed, but I believe it's not related to these changes. Nice job!

internal/testrunner/testrunner.go
ycombinator: Yeah, I had it returning an error at one point but looking at the current implementations of `TearDown`, none of them actually return an error so I took it out.

But, since you're also thinking about it, I'm once again leaning towards allowing `TearDown` to return an error. I will add it back, thanks!

ycombinator: Added in 7de930cc20dedf76ae7752210988e9413ff5b71a.

internal/testrunner/runners/system/runner.go
```diff
@@ -59,23 +57,47 @@ type stackSettings struct {
...
-	}
-	defer r.tearDown(options)
+func (r runner) Run(options testrunner.TestOptions) ([]testrunner.TestResult, error) {
+	r.options = options
+	r.stackSettings = getStackSettingsFromEnv()
+
 	return r.run()
 }
 
+func (r runner) TearDown() {
```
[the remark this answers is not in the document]
ycombinator: Changed to pointer receivers in 7de930cc20dedf76ae7752210988e9413ff5b71a.

mtojek (approved):

PR #187, opened 2020-11-25
Add a way to select the stream under test in system test runner

### LINKED ISSUE #186: [System test runner] Allow input selection from config.yml
For data streams that have multiple input options ([example](https://github.com/elastic/integrations/blob/9e4efb410dc4495a4805b69afc1bcbb79d9d8829/packages/cisco/data_stream/ios/manifest.yml#L40-L41)) in their `manifest.yml`, I would like a way to select the stream under test via the test's `config.yml`.

Perhaps allowing the input type to be specified from the config.yml would address this? Currently the system test runner always enables the first stream in the list.

```
input: logfile
vars: ~
data_stream:
  vars:
    paths:
      - "{{SERVICE_LOGS_DIR}}/*.log"
```

andrewkroh:
This provides a means to select a stream to test by input type.

Closes #186

### DIFF
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -75,10 +75,11 @@ Next, we must define configuration for each data stream that we want to system t
             config.yml
 ```
 
-The `config.yml` file allows you define values for package and data stream-level variables. For example, the `apache/access` data stream's `config.yml` is shown below.
+The `config.yml` file allows you to define values for package and data stream-level variables. For example, the `apache/access` data stream's `config.yml` is shown below.
 
 ```
 vars: ~
+input: logfile
 data_stream:
   vars:
     paths:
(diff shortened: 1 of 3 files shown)


ycombinator:
Thanks for this change! Would you mind updating the HOWTO for system testing as well: https://github.com/elastic/elastic-package/blob/master/docs/howto/system_testing.md?

andrewkroh:
I added a section to describe the `input` option to the howto.

docs/howto/system_testing.md
```diff
@@ -75,10 +75,11 @@ Next, we must define configuration for each data stream that we want to system t
             config.yml
 ```
 
-The `config.yml` file allows you define values for package and data stream-level variables. For example, the `apache/access` data stream's `config.yml` is shown below.
+The `config.yml` file allows you to define values for package and data stream-level variables. For example, the `apache/access` data stream's `config.yml` is shown below.
```
ycombinator, line 78: Thanks.

ycombinator (approved):
LGTM. Thanks, again, for implementing this!

PR #188, opened 2020-11-25
Include stderr with returned docker error

andrewkroh:
Include the stderr output when `docker network connect` fails. This helps with debugging failures.

As an example, here's the output.

    Starting elastic-package-service_foo_1 ... done
    2020/11/24 21:52:31 DEBUG attaching service container elastic-package-service_apache_1 to stack network elastic-package-stack_default
    Error: error running package system tests: could not complete test run: could not setup service: could not attach service container to the stack network (stderr="Error response from daemon: No such container: elastic-package-service_apache_1\n"): exit status 1

### DIFF
--- a/internal/testrunner/runners/system/servicedeployer/compose.go
+++ b/internal/testrunner/runners/system/servicedeployer/compose.go
@@ -5,6 +5,7 @@
 package servicedeployer
 
 import (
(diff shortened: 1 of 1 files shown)


mtojek (approved):
Nice!

Issue #190, opened 2020-11-25
Support stack with service mode


mtojek:
Apparently we've forgotten about this mode as we're focused on the system test runner.

Use case:
A developer wants to run the Elastic stack and the service (like Apache) to design Kibana dashboards.

Temporary workaround:
Provide instructions to interact with the `docker` command to let the service connect to the stack.

Ideal solution:
`elastic-package stack` provides a subcommand to connect to the Elastic stack.

/cc @ycombinator @ChrsMark

ycombinator:
++ we already have the information for spinning up the integration service (e.g. Apache), in `_dev/deploy` so this shouldn't be too difficult to build out by reusing parts of the implementation from the system test runner!

mtojek:
It's not bad, it's just two commands :)

_(I assume that the Elastic stack is up and running, especially the Elastic Agent)._

Boot up the docker-compose with Apache:

```bash
SERVICE_LOGS_DIR=/Users/JohnDoe/.elastic-package/tmp/service_logs docker-compose -p elastic-package-service up -d
```

Connect Apache container to the `elastic-package-stack` network:

```bash
docker network connect elastic-package-stack_default elastic-package-service_apache_1
```

Verify in the Elastic Agent container (run from the Agent container):

```bash
curl http://elastic-package-service_apache_1:80/server-status -I
HTTP/1.1 200 OK
Date: Wed, 25 Nov 2020 11:49:58 GMT
Server: Apache/2.4.20 (Unix)
Content-Type: text/html; charset=ISO-8859-1
```

```bash
ls /tmp/service_logs/
access.log  error.log  httpd.pid
```

mtojek:
I'm going with this one.

mtojek:
Suggested command:

```
elastic-package service up
elastic-package service down
```

PR #191, opened 2020-11-25
HOWTO: connect service with Elastic stack

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/190

Preview: https://github.com/mtojek/elastic-package/blob/docs-run-service/docs/howto/connect_service_with_elastic_stack.md

### DIFF
--- a/docs/howto/connect_service_with_elastic_stack.md
+++ b/docs/howto/connect_service_with_elastic_stack.md
@@ -10,6 +10,41 @@ The procedure assumes an existence of two Docker networks - stack network and se
 the stack network to the service container. The Elastic Agent will have an access to the service metrics endpoint and its
 logs (via mounted volume).
 
+## Prerequisites
+
+### Service Docker Compose definition
+
+The service can be defined using Docker files (YML format), which describe Docker images required for the service to be up.
+The running service produces metrics and logs, which later on can be collected by the Elastic Agent. Such instance is
+a good candidate for system tests or manual testing.
+
+#### Support for logs collection
+
+As the Elastic Agent is running in a separate container, there is shared volume for logs required. The volume is shared
+between the Elastic Agent and the service container.
+
+In the Elastic Agent container logs are exposed under `/tmp/service_logs`:
+
+```yaml
+volumes:
+- type: bind
+  source: ../tmp/service_logs/
+  target: /tmp/service_logs/
+```
+
+We need to define a similar volume in the service `docker-compose.yml` file:
+
+```yaml
+volumes:
+- ${SERVICE_LOGS_DIR}:/usr/local/apache2/logs
+```
+
+`SERVICE_LOGS_DIR` is a special environment variable, which points to the local Docker host directory, shared with
+the Elastic Agent container.
+
+Review the [docker-compose.yml](https://github.com/elastic/elastic-package/blob/master/test/packages/apache/_dev/deploy/docker/docker-compose.yml#L11)
+file for Apache to see a real example.
+
 ## User guide (based on the Apache integration)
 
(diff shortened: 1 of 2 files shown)


docs/howto/connect_service_with_elastic_stack.md
```diff
@@ -0,0 +1,73 @@
...
+stack instead of linking it temporarily for the execution of system tests. It's especially useful if you're adjusting Kibana
+dashboards or want to examine metrics data collected for the service in the Elasticsearch index.
+
+The procedure assumes an existence of two Docker networks - stack network and service network. The goal is to expose
+the stack network to the service container. The Elastic Agent will have an access to the service metrics endpoint and its
+logs (via mounted volume).
+
+## User guide (based on the Apache integration)
+
+1. If you haven't done it already, please boot up the Elastic stack:
```
ChrsMark, line 15: Shouldn't the note about logs volume mounting be mentioned early on?
That you need to add the volume mount part in agent service so as to have logs to be parsed by agent?
```yaml
volumes:
  - ${SERVICE_LOGS_DIR}:/usr/local/apache2/logs
```
```

File: docs/howto/connect_service_with_elastic_stack.md [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+
+```bash
+elastic-package stack up -d -v
```
ChrsMark, line 18: Also let's make it more clear that the compose file is located at `~/.elastic-package/stack/snapshot.yml ` for those that are not familiar with the internals.

ChrsMark (approved):
lgtm! Went over it live while testing manually my stuff and it can cover the manual testing process. Left some comments about some fuzzy points that i was able to by pass but would be better to clarify them more. 

Let me know what you think!

docs/howto/connect_service_with_elastic_stack.md
```diff
@@ -0,0 +1,73 @@
...
+Server: Apache/2.4.20 (Unix)
+Content-Type: text/html; charset=ISO-8859-1
+```
+
+Check presence of application log files:
+
+```bash
+ls /tmp/service_logs
+access.log  error.log  httpd.pid
+```
```
ChrsMark, line 73: Also add a last but not least comment that the integration should point to the targeted service with the service name like: `elastic-package-service_apache_1:80` to make it easier for testers to set the integration from within the Fleet UI.

Screenshot 2020-11-26 at 17 26 46

docs/howto/connect_service_with_elastic_stack.md
mtojek: I put an additional section describing the service Docker Compose files. Regarding the Elastic Agent definition, the volume property is already there:

https://github.com/elastic/elastic-package/blob/master/internal/install/static_snapshot_yml.go#L96

docs/howto/connect_service_with_elastic_stack.md
mtojek: I added the "Advanced" section to cover it, but in general this method assumes that you don't interact with `snapshot.yml`, but can use `docker` command to achieve the goal.

docs/howto/connect_service_with_elastic_stack.md
mtojek: Good catch! I mentioned the point about the service address.

mtojek:
@ChrsMark would you mind re-reviewing this PR?

ChrsMark (approved):
LGTM! thanks!

docs/howto/connect_service_with_elastic_stack.md
ChrsMark: nice!

PR #193, opened 2020-11-30
Testing: Add config option to accept numeric keywords

adriansr:
Keyword fields are not necessarily ingested as JSON strings. It's common to ingest numbers (integer or floating point) as keyword type.

~~This updates the fields validator to accept numeric values. They are converted to string so that any defined patterns can be checked.~~

This adds a new configuration option, `numeric_keyword_fields`, to pipeline test cases and system tests so that selected fields can have numeric type and be ingested into a text-like field.

Example configuration:

```
{
    "dynamic_fields": {
        "event.ingested": ".*"
    },
    "numeric_keyword_fields": [
        "network.iana_number",
        "event.code",
        "syslog.facility"
    ]
}
```

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -211,17 +211,7 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 
 	var valid bool
 	switch definition.Type {
-	case "constant_keyword", "keyword":
-		// Accept numeric fields as valid keywords.
-		if _, isStr := val.(string); !isStr {
-			f64, ok := val.(float64)
-			if !ok {
-				break
-			}
-			val = fmt.Sprintf("%g", f64)
-		}
-		fallthrough
-	case "date", "ip", "text":
+	case "date", "ip", "text", "constant_keyword", "keyword":
 		var valStr string
 		valStr, valid = val.(string)
 		if !valid || definition.Pattern == "" {
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -233,7 +233,7 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 			return errors.Wrap(err, "invalid pattern")
 		}
 		if !valid {
-			return fmt.Errorf("field \"%s\"'s value, %s, does not match the expected pattern: %s", key, valStr, definition.Pattern)
+			return fmt.Errorf("field %q's value, %s, does not match the expected pattern: %s", key, valStr, definition.Pattern)
 		}
 	case "float", "long", "double":
 		_, valid = val.(float64)
@@ -242,7 +242,7 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 	}
 
 	if !valid {
-		return fmt.Errorf("field \"%s\"'s Go type, %T, does not match the expected field type: %s (field value: %v)", key, val, definition.Type, val)
+		return fmt.Errorf("field %q's Go type, %T, does not match the expected field type: %s (field value: %v)", key, val, definition.Type, val)
 	}
 	return nil
 }
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -211,7 +211,7 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 
 	var valid bool
 	switch definition.Type {
-	case "date", "ip", "text", "constant_keyword", "keyword":
+	case "date", "ip", "constant_keyword", "keyword", "text":
 		var valStr string
 		valStr, valid = val.(string)
 		if !valid || definition.Pattern == "" {
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"strings"
 
@@ -212,7 +211,17 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 
 	var valid bool
 	switch definition.Type {
-	case "date", "ip", "constant_keyword", "keyword", "text":
+        case "constant_keyword", "keyword":
+                // Accept numeric fields as valid keywords.
+                if _, isStr := val.(string); !isStr {
+                    f64, ok := val.(float64)
+                    if !ok {
+                        break
+                    }
+                    val = fmt.Sprintf("%f", f64)
+                }
+                fallthrough
+	case "date", "ip", "text":
 		var valStr string
 		valStr, valid = val.(string)
 		if !valid || definition.Pattern == "" {
@@ -233,7 +242,7 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 	}
 
 	if !valid {
-		return fmt.Errorf("field \"%s\"''s Go type, %s, does not match the expected field type: %s", key, reflect.TypeOf(val), definition.Type)
+		return fmt.Errorf("field \"%s\"'s Go type, %T, does not match the expected field type: %s (field value: %v)", key, val, definition.Type, val)
 	}
 	return nil
 }
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -22,17 +22,23 @@ import (
 
 // Validator is responsible for fields validation.
 type Validator struct {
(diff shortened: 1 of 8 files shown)


internal/fields/validate.go
```diff
@@ -212,7 +211,17 @@ func parseElementValue(key string, definition FieldDefinition, val interface{})
 
 	var valid bool
 	switch definition.Type {
-	case "date", "ip", "constant_keyword", "keyword", "text":
+        case "constant_keyword", "keyword":
```
mtojek, line 214: I suspect we may rerun `make format` for this file.

internal/fields/validate.go
```diff
@@ -212,7 +211,17 @@ func parseElementValue(key string, definition FieldDefinition, val interface{})
...
+                // Accept numeric fields as valid keywords.
+                if _, isStr := val.(string); !isStr {
```
mtojek, line 216: It looks like the PR is an answer to the https://github.com/elastic/integrations/issues/422 . Actually I'm a fan of the stricter validation, but I'm fine with adjusting the tool if necessary.

Any thoughts about this PR? @ycombinator @andrewkroh

internal/fields/validate.go
adriansr: Done

internal/fields/validate.go
adriansr: This is the opposite case. 

https://github.com/elastic/integrations/issues/422 wants to ingest strings into numeric fields, which is dangerous because you'll get an exception if ES can't convert the string to the number type (or if coercion is disabled). It's better to validate and convert those fields in a pipeline.

What this PR does is allow numeric values to be ingested as keywords, which is a common pattern for any field that is numeric but has a symbolic meaning and doesn't require `range` queries. In this case the only risk is that if no mapping is present in the index, ES will create the field as `long` instead of `keyword`. But I don't think that's possible with ingest manager.

ycombinator (approved):
Ah, nice enhancement for even stricter validation! Thanks and LGTM.

internal/fields/validate.go
```diff
@@ -233,7 +242,7 @@ func parseElementValue(key string, definition FieldDefinition, val interface{})
 	}
 
 	if !valid {
-		return fmt.Errorf("field \"%s\"''s Go type, %s, does not match the expected field type: %s", key, reflect.TypeOf(val), definition.Type)
+		return fmt.Errorf("field \"%s\"'s Go type, %T, does not match the expected field type: %s (field value: %v)", key, val, definition.Type, val)
```
ycombinator, line 245: Nice use of `%T` to get rid of the explicit `reflect`. TIL!

internal/fields/validate.go
mtojek: Thank you for the explanation, @adriansr. It would be great to add a similar justifying comment in the code. As it's a non-intuitive conversion between fields, let's a unit test around this function (`parseElementValue`).

I wonder if we should describe it in the docs/howto guides, but I leave it up to you to decide if it's useful or not.

internal/fields/validate.go
andrewkroh, line 245:
```suggestion
		return fmt.Errorf("field %q's Go type, %T, does not match the expected field type: %s (field value: %v)", key, val, definition.Type, val)
```

andrewkroh (approved):
I prefer seeing JSON strings in the `_source` for fields mapped as keyword types, but since it's completely safe to map numbers to keywords (the inverse not being true) I'm ok loosening the restrictions to make developing packages a little simpler.

internal/testrunner/runners/pipeline/test_config.go
```diff
@@ -20,6 +20,10 @@ type testConfig struct {
 	Multiline     *multiline             `json:"multiline"`
 	Fields        map[string]interface{} `json:"fields"`
 	DynamicFields map[string]string      `json:"dynamic_fields"`
+
+	// NumericKeywordFields holds a list of fields that have keyword
+	// type but can be ingested as numeric type.
+	NumericKeywordFields []string `json:"numeric_keyword_fields"`
```
mtojek, line 26: This way you're violating the package spec by introducing an additional field. Please go first with the PR to the package-spec. Unfortunately you can't simply define a field with an extension to spec.

EDIT:

Just another thought: I might not be on the same page, but could you explain why do you need this special property? I thought that this is a general rule that can be applied to all numeric fields. /cc @ycombinator

adriansr:
In the end I think it makes more sense for this to be a configuration option to be set on a per-test-case basis, so I've modified the pipeline test runner to accept this new option.

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -256,6 +261,54 @@ func verifyDynamicFields(result *testResult, config *testConfig) error {
 	return nil
 }
 
+func convertNumericKeywordFields(result *testResult, config *testConfig) error {
```
mtojek, line 264: I wouldn't say that this should part of the pipeline test runner. I suppose a similar validation should be part of the system runner as well.

Please consider to put it in a common accessible place.

mtojek (changes requested):

adriansr:
Let me explain the rationale behind these changes.

Across the Beats code base, there are some instances of fields that are defined as `keyword` type and they would be ingested as a numeric field the document's JSON. According to [ES docs:](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html) _Not all numeric data should be mapped as a numeric field data type. Elasticsearch optimizes numeric fields, such as integer or long, for range queries. However, keyword fields are better for term and other term-level queries._

Examples of such fields are:
- network.iana_number
- vlan.id
- dns.id
- observer.*.interface.id
- x509.version_number
- event.code: In the case of some external event sources
- {file,user}.{uid,gid}: for some Beats datasets, under a UNIX-like OS
- syslog.{facility,severity}: For Filebeat's syslog input (these need conversion to ECS, which uses `long`).
- elasticsearch.slowlog.{total_hits,total_shards}: These two really look like they should be a `long`.

There's probably many others, but it's difficult to spot them. I worry especially about cases where we are receiving an arbitrary set of keys (ex. via JSON) from an external source and storing those verbatim in the event with keyword type.

I see three possibilities for handling this:
- Do not accept numeric keywords. This will force us to modify the source of events (module/Beat) so that those numbers are converted to string before being published. This is not too complicated when the processing is performed in an ingest pipeline, but it will be more time consuming for fields generated outside a pipeline.
- Always accept numeric keywords. The original idea of this PR, after your feedback I realised it could be counter-intuitive.
- Make it an opt-in feature. The new version of this PR. By default validation will be strict but we it can be disabled on a per-field per-test-case basis.

mtojek:
Thank you for the detailed explanation and possible options. I would for the first or third option:

> Do not accept numeric keywords.
> Make it an opt-in feature.

Actually I like the third one when the validation is configurable (strictness level). Keep in mind that this option requires updating the package-spec: https://github.com/elastic/package-spec/

adriansr:
@mtojek can you have another look?

internal/fields/validate.go
```diff
@@ -23,17 +22,23 @@ import (
...
 // Validator is responsible for fields validation.
 type Validator struct {
-	schema []FieldDefinition
+	schema               []FieldDefinition
+	numericKeywordFields map[string]struct{}
 }
 
 // CreateValidatorForDataStream function creates a validator for the data stream.
-func CreateValidatorForDataStream(dataStreamRootPath string) (*Validator, error) {
+func CreateValidatorForDataStream(dataStreamRootPath string, numericKeywordFields []string) (*Validator, error) {
```
mtojek, line 30: What do you think about introducing `ValidatorOptions` that will contain `DataStreamRootPath string` and `NumericKeywordFields []string`?

I would be then:

```go
validator, _ := CreateValidatorForDataStream(ValidatorOptions{
  DataStreamRootPath: dsrp,
  NumericKeywordFields: nkf,
})

In case of further extensions we won't need to modify existing code. Another option is to refactor it to fluent interface (e.g. `WithDataStreamRootPath()`, `WithNumericKeywordFields()`).
```

mtojek (commented):
just a short one about the clean code

mtojek (approved):
LGTM
nit: it would be nice to describe this feature in the howto docs (not blocking this PR)

PR #195, opened 2020-12-02
Build service containers for system tests

andrewkroh:
This changes the `docker-compose up` command to include `--build` so that images
will be rebuilt if there were changes. While developing tests you are often making
changes to the service container and it's nice if the changes are automatically
detected, rebuilt, and deployed on system test.


mtojek (approved):
LGTM, but leaving the final word for @ycombinator .

ycombinator (approved):
LGTM.

PR #196, opened 2020-12-02
Updating dependencies

ycombinator:
Updating dependency on elastic/package-spec due to https://github.com/elastic/package-spec/pull/87. Other dependencies are updated too.


mtojek (approved):
WFG

Issue #197, opened 2020-12-02
Bug: defer-cleanup is ignored after test runner failure


mtojek:
I executed:

`elastic-package test system -v --defer-cleanup=120s`

Output:

```
2020/12/02 16:29:31 DEBUG tearing down service...
2020/12/02 16:29:31 DEBUG tearing down service using docker compose runner
2020/12/02 16:29:31 DEBUG running command: /usr/local/bin/docker-compose -f /Users/marcin.tojek/go/src/github.com/elastic/integrations/packages/nginx/_dev/deploy/docker/docker-compose.yml -p elastic-package-service down
Stopping elastic-package-service_nginx_1 ... done
Removing elastic-package-service_nginx_1 ... done
Removing network elastic-package-service_default
2020/12/02 16:29:33 DEBUG deleting data in data stream...
--- Test results for package: nginx - START ---
╭─────────┬─────────────┬───────────┬───────────┬────────────────────────────────────────────────────────────────────────────────────────┬───────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT                                                                                 │  TIME ELAPSED │
├─────────┼─────────────┼───────────┼───────────┼────────────────────────────────────────────────────────────────────────────────────────┼───────────────┤
│ nginx   │ access      │ system    │           │ FAIL: one or more errors found in documents stored in logs-nginx.access-ep data stream │ 21.662550831s │
│ nginx   │ error       │ system    │           │ FAIL: one or more errors found in documents stored in logs-nginx.error-ep data stream  │   23.3848039s │
│ nginx   │ stubstatus  │ system    │           │ PASS                                                                                   │ 19.313296295s │
╰─────────┴─────────────┴───────────┴───────────┴────────────────────────────────────────────────────────────────────────────────────────┴───────────────╯

FAILURE DETAILS:

nginx/access :
[0] field "input.type" is undefined
[1] field "log.file.path" is undefined
[2] field "log.offset" is undefined
nginx/error :
[0] field "ecs.version" is undefined
[1] field "input.type" is undefined
[2] field "log.file.path" is undefined
[3] field "log.offset" is undefined
--- Test results for package: nginx - END   ---
Done
Error: one or more test cases failed
```

The application didn't wait for the specified period of time. It looks like a regression to me. /cc @ycombinator

PR #198, opened 2020-12-03
Fix: expose defer cleanup option

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/197


ycombinator (approved):
LGTM. Thanks for fixing.

Looks like the flag was being defined in the command and used in the test runners but the plumbing in the middle was missing. 🤦

PR #203, opened 2020-12-15
Add sort to system test query

andrewkroh:
While executing system tests multiple times I was getting different
results on each run. The root cause was that the test samples 10
of the events that are written to ES. And on each run it was getting
different events back, validating them, and returning a unique set
of errors.

I changed the query to sort on the `@timestamp` and to also sample
more events (up to 500). This way we get more repeatable tests.


ycombinator (approved):
LGTM. Thanks!

Issue #208, opened 2020-12-18
[System test runner] Support for multiple tests per data stream


andrewkroh:
Data streams can have multiple input options such as logfile, syslog over UDP, and syslog over TCP. In order to have test coverage over all of the supported input configurations we need a way to execute more than one system test per data stream.

Current a single test is declared in `data_stream/<name>/_dev/test/system/config.yml`. If we can expand this to allow more than one test definition then we'll be able to expand test coverage.

- Relates: https://github.com/elastic/integrations/issues/377

ruflin:
@andrewkroh Does each of these have a separate stream template?

andrewkroh:
Yes, as an example this package has a template for the UDP input and a logfile input. https://github.com/elastic/integrations/tree/0d55adaa4ea33e94b05f5c5106cc22c845b720ae/packages/checkpoint/data_stream/firewall/agent/stream

ruflin:
👍 Each template should allow its own test. I'm wondering if we should go so far to allow each template multiple tests?

andrewkroh:
Implemented in https://github.com/elastic/elastic-package/pull/209.

PR #209, opened 2020-12-21
Support multiple system tests per data_stream

adriansr:
Updates the system tests runner so that multiple `test-*-config.yml` files can be included in `<data_stream>/_dev/tests/system`. This allows testing multiple inputs and/or different configuration options per data stream.

Example test for two inputs:

test-tcp-input.yml:
```yaml
  input: tcp
  service: bluecoat-director-tcp
  vars: ~
  data_stream:
    vars:
      tcp_host: 0.0.0.0
      tcp_port: 9514
```

test-udp-input.yml:
```yaml
  input: udp
  service: bluecoat-director-udp
  vars: ~
  data_stream:
    vars:
      udp_host: 0.0.0.0
      udp_port: 9514
```

The `service` tags allows to uses different containers for the tests. In this case, to generate traffic using different protocols.

Sample `docker-compose.yml` in `<package>/_dev/deploy/docker`:
```yaml
version: '2.3'
services:
  bluecoat-director-tcp:
    tty: true
    build: .
    volumes:
      - ./logs:/logs/:ro
    command: /stream -proto tcp -dest elastic-agent:9514 /logs/generated.log
  bluecoat-director-udp:
    tty: true
    build: .
    volumes:
      - ./logs:/logs/:ro
    command: /stream -proto udp -delay 100ms -dest elastic-agent:9514 /logs/generated.log 
```

Related #208 
Related https://github.com/elastic/package-spec/pull/101


ycombinator:
This is a great addition, @adriansr. I was wondering how long before we needed it 🙂 . I was on PTO all last week so I'm only seeing this PR now. I should be able to review it today or tomorrow latest.

internal/testrunner/runners/system/runner.go
```diff
@@ -83,93 +87,201 @@ func (r *runner) TearDown() error {
...
+		}
+		return true, nil
+	}
+}
+
+func (r *runner) runTest(config testConfig, cfgIdx int, ctxt servicedeployer.ServiceContext) ([]testrunner.TestResult, error) {
+	// Determine test name
+	name := config.Name
+	if name == "" {
+		name = config.Input
```
ycombinator, line 248: I understand that the use case for this change is to test with different inputs but I think your solution is more generic than that. So I would not fallback to `config.Input` if `config.Name` is not defined. I think we can require developers to specify `config.Name` always, so if it's empty I think we should error out.

internal/testrunner/runners/system/test_config.go
```diff
@@ -49,16 +51,24 @@ func newConfig(systemTestFolderPath string, ctxt servicedeployer.ServiceContext)
...
+
+	// Handle the case were test config is not a list of cases but a single case
+	if len(c) == 0 {
+		var single testConfig
+		if err := cfg.Unpack(&single); err != nil {
+			return nil, errors.Wrapf(err, "unable to unpack system test configuration file: %s", configFilePath)
+		}
+		c = append(c, single)
+	}
+	return c, nil
```
ycombinator, line 71: This is nice but I'm wondering if we should simplify and require that the test config should always be a list of cases. That way it's also a hint to the package developers that multiple test cases are allowed. Of course, this would mean updating existing system test configurations.

ycombinator (changes requested):
Just left a couple of minor comments.

internal/testrunner/runners/system/test_config.go
mtojek: I think this PR has been started too early. It's nice to have a PoC showing the concept, but ideally we kick off with a proposal in package-spec, as it should be adjusted first.

What do you think about keeping the same order as in pipeline tests, which support multiple test cases per data_stream? See: https://github.com/elastic/integrations/tree/master/packages/apache/data_stream/access/_dev/test/pipeline
I'm referring to the name pattern: `test-name.config.yml`.

Regarding the `config.yml`, I think we have to adjust all integrations, but as there is dependency versioning in place, I don't think we need to rollout changes synchronically.

IMHO the order of steps is the following:
1. PR to the package-spec introducing new `config.yml`.
2. This PR with updated package-spec dependency.
3. PR to the Integrations repository polishing all `config.yml`s

mtojek (commented):
left a general comment

mtojek (changes requested):
@adriansr I'm afraid that changes you pushed to the package-spec blocked the other team (https://github.com/elastic/elastic-package/pull/216) and will be a problem in Integrations as the elastic-package with new spec won't accept any existing `config.yml` files.

I'm afraid we need to solve it ASAP.

adriansr:
@mtojek this PR is ready for review and incorporates the changes from #216 (because it also needs to update the package-spec dependency).

mtojek:
I see it's ready for review, but once you push this, you will also block updating the elastic-package in Integrations without adjusting multiple files. I'm not convinced about the timeline and wouldn't like to block people. Do you plan to adjust all integrations immediately?

I'm referring to this line:

```
var systemTestConfigFilePattern = regexp.MustCompile(`^test-([a-z0-9_.-]+)-config.yml$`)
```

internal/testrunner/runners/system/test_config.go
```diff
@@ -19,10 +20,11 @@ import (
 	"github.com/elastic/elastic-package/internal/testrunner/runners/system/servicedeployer"
 )
 
-const configFileName = "config.yml"
+var systemTestConfigFilePattern = regexp.MustCompile(`^test-([a-z0-9_.-]+)-config.yml$`)
 
 type testConfig struct {
 	Input      string                       `config:"input"`
+	Service    string                       `config:"service"`
```
ycombinator, line 27: @mtojek This is more of a note for us for the future: as we add more service deployers we will need to make sure this field continues to work as expected with the new service deployers as well (not just the Docker compose one that we have today).

test/packages/multiinput/data_stream/test/agent/stream/tcp.yml.hbs
```diff
@@ -0,0 +1,22 @@
+tcp:
+host: "{{tcp_host}}:{{tcp_port}}"
```
ycombinator, line 2: Is the indentation here correct?

test/packages/multiinput/data_stream/test/agent/stream/udp.yml.hbs
```diff
@@ -0,0 +1,22 @@
+udp:
+host: "{{udp_host}}:{{udp_port}}"
```
ycombinator, line 2: Is the indentation here correct?

ycombinator (approved):
Left a couple of minor comments/questions on indentation. Besides those, this PR LGTM.

Would be nice to have a sister PR ready to go in https://github.com/elastic/integrations that updates all existing system test configs per the new naming convention and syntax introduced in this PR. That way you can just update that PR with the elastic-package dependency once this PR here is merged and we can quickly get all integrations updated as well.

test/packages/multiinput/data_stream/test/agent/stream/tcp.yml.hbs
adriansr: Turns out is not an indentation problem. That `tcp` key is just meaningless. Seems like a few packages include this mistake and we've just been repeating it in some places.

mtojek (approved):

test/packages/multiinput/data_stream/test/agent/stream/tcp.yml.hbs
ycombinator: Okay, thanks. Can you make an issue (or PR, whatever is easier), probably in the integrations repo to clean up the meaningless keys?

PR #210, opened 2020-12-30
Merging ingestmanager client into generic kibana client

ycombinator:
When the system tests feature was added to `elastic-package`, we created a client to communicate with the Kibana Ingest Manager API, `ingestmanager.Client`. Some time later, we created a more general Kibana client to communicate with the Kibana Saved Objects and Dashboard Export APIs, `kibana.Client`.  

It makes sense to combine these two clients into one, as noted in https://github.com/elastic/elastic-package/pull/168#discussion_r525030994. This PR attempts to do so.


internal/testrunner/runners/system/runner.go
```diff
@@ -180,25 +180,25 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 	}
 
 	// Step 2. Configure package (single data stream) via Ingest Manager APIs.
-	im, err := ingestmanager.NewClient(r.stackSettings.kibana.host, r.stackSettings.elasticsearch.username, r.stackSettings.elasticsearch.password)
```
mtojek, line 183: I admit that I didn't analyze the source deeper, but are these Kibana/ES properties still required in stack settings.

internal/testrunner/runners/system/runner.go
```diff
@@ -373,26 +373,26 @@ func createPackageDatastream(
...
+			DataStream: kibana.DataStream{
 				Type:    ds.Type,
 				Dataset: fmt.Sprintf("%s.%s", pkg.Name, ds.Name),
 			},
 		},
 	}
 
 	// Add dataStream-level vars
-	dsVars := ingestmanager.Vars{}
+	dsVars := kibana.Vars{}
```
mtojek, line 395: Thinking loud (nothing to be done immediately): The `Vars` structure may look to be a generic one. It may happen in the future that we need to handle another `Vars` structure, so some of these structs will require a relevant prefix added to their names.

internal/kibana/policies.go
```diff
@@ -52,7 +52,7 @@ func (c *Client) CreatePolicy(p Policy) (*Policy, error) {
 func (c *Client) DeletePolicy(p Policy) error {
 	reqBody := `{ "agentPolicyId": "` + p.ID + `" }`
 
-	statusCode, respBody, err := c.post("agent_policies/delete", []byte(reqBody))
+	statusCode, respBody, err := c.post("/api/fleet/agent_policies/delete", []byte(reqBody))
```
mtojek, line 55: What do you think about extracting all URL prefixes to the common place? I know it's an improvement comparing to what we have now.

mtojek (commented):
Thank you for working on this. Apart from system tests, the part that can be tested only manually is `export dashboards` (as it uses the kibana client).

I'm worried about the failed Jenkins job:

```
[2020-12-30T00:43:26.104Z] Error: error running package system tests: could not complete test run: could not create test policy: could not create policy; API status code = 404; response body = {"statusCode":404,"error":"Not Found","message":"Not Found"}
```

I wonder if there was something missed during refactoring.

internal/kibana/policies.go
ycombinator: Sure, but just want to clarify what you mean by URL prefixes: just the `/api/fleet` part (and similar prefixes from other API URLs) or something more?

mtojek: Yes, I meant exactly that: '/api/fleet' or '/api/saved_objects' etc. Just to prevent missing mappings in the future (because they were hidden in multiple places in the source).

internal/testrunner/runners/system/runner.go
ycombinator: Good catch, these are not required any more. Removed in 32fe1b2.

internal/kibana/policies.go
ycombinator: Done in 17ab10c70570def207fd49eb186875a0d941b258.

internal/kibana/policies.go
```diff
@@ -52,7 +52,7 @@ func (c *Client) CreatePolicy(p Policy) (*Policy, error) {
 func (c *Client) DeletePolicy(p Policy) error {
 	reqBody := `{ "agentPolicyId": "` + p.ID + `" }`
 
-	statusCode, respBody, err := c.post("agent_policies/delete", []byte(reqBody))
+	statusCode, respBody, err := c.post(fmt.Sprintf("%s/agent_policies/delete", FleetAPI), []byte(reqBody))
```
mtojek, line 55: nit: Hm.. I'm wondering if it's not better to extract also the resource part (agent_policies, package_policies) as they appear in multiple places. Once it's changed, I think you can switch from `fmt.Sprintf` to a simple concatenation (URL prefix + operation).

internal/kibana/prefixes.go
```diff
@@ -0,0 +1,16 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package kibana
+
+const (
+	// CoreAPI is the prefix for all Kibana Core API resources.
+	CoreAPI = "/api/kibana"
```
mtojek, line 9: WDYT about renaming it to `api.go` or `url_prefixes.go`? Prefixes sounds a bit vague.

mtojek (commented):
Changes look good to me. I'm worried about the CI job output that revealed a couple of unknown errors: https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-210/6/pipeline

Would you mind looking at them?

mtojek:
It seems to be an unnecessary `t.Log()` call. I must have missed it in previous reviews:

```
		t.Run(test.key, func(t *testing.T) {
			err := parseElementValue(test.key, test.definition, test.value)
			if err != nil {
				t.Log(err)
			}
			if test.fail {
				require.Error(t, err)
			} else {
				require.NoError(t, err)
			}
		})
```

You can leave as is or cleanup the test output - up to you :) It means that the problem is:

```
[2021-01-05T00:31:57.673Z] internal/kibana/dashboards.go:35:10: Sprintf format %s has arg query of wrong type strings.Builder
```

internal/kibana/policies.go
ycombinator: I think it's more readable this way, actually. And while the prefixes are more likely to change (due to project name changes, etc.), I think the resource paths are less likely to change. So I would prefer to keep it as-is.

mtojek (approved):

internal/kibana/prefixes.go
ycombinator: Done in 93223e5.

ycombinator:
> It seems to be an unnecessary `t.Log()` call. 

Removed in 7ce568b4e78ae5f8d9a6661d867f9c5809755ee6.

PR #212, opened 2021-01-05
Fix: don't skip _dev/build directories

mtojek:
This PR fixes the issue with skip `_dev/build` directories. I spotted this one while investigating other issues and noticed warnings that README files are static and not generated from templates.


test/packages/apache/docs/README.md
```diff
@@ -218,7 +218,6 @@ An example event for `status` looks as following:
 | apache.status.cpu.load | CPU Load. | scaled_float |
 | apache.status.cpu.system | System cpu. | scaled_float |
 | apache.status.cpu.user | CPU user load. | scaled_float |
-| apache.status.hostname | Apache hostname. | keyword |
```
ycombinator, line 221: Is this change intentional? At first glance it seems unrelated to this PR.

mtojek: It's related, because it's the first time the README has been rendered correctly :)

ycombinator (approved):

Issue #192, opened 2020-11-26
Add more guidance for undefined environment variable ELASTIC_PACKAGE_ELASTICSEARCH_HOST


mtojek:
User story:

A developer may forget to boot up the local cluster or evaluate "stack shellinit". Currently the tool fails with the following message:

```
Error: fetching Elasticsearch client instance failed: undefined environment variable: ELASTIC_PACKAGE_ELASTICSEARCH_HOST 
```

We can consider different options:
1. Leave as is.
2. Add more guidance: suggest to use "shellinit".
3. Use environment defaults.
4. Try to connect to the localhost and in case of failure, output proper guidance.

/cc @exekias @ycombinator @ChrsMark

exekias:
Thanks for opening!

I think at the very least we could output some instructions to run `shellinit` after `stack up` is done. I wonder what's the most common case. If most users are just running this in docker on the localhost, perhaps option 4 makes sense? That would improve the default experience, while still work for more advanced use cases.

ycombinator:
I think we should do two things:

- point 2 above: that is, when any `elastic-package` command that depends on environment variables is executed, we should check as early as practical in that command's execution if the necessary environment variables are set; if not, we show the appropriate error message and suggest that the user runs `elastic-package shellinit` if they are started the stack with `elastic-package stack up`, otherwise they should simply set the required variables to appropriate values.

- at the end of `elastic-package stack up`, suggest that the user runs `elastic-package shellinit` to use the stack with other `elastic-package` commands.

mtojek:
I'm going with this feature.

PR #214, opened 2021-01-11
Add guidance for missing environment variables

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/192

Changes:
- clean error messages
- show shellinit guidance for `stack up`
- show shellinit guidance for undefined environment variable errors


internal/stack/errors.go
```diff
@@ -8,5 +8,6 @@ import "fmt"
 
 // UndefinedEnvError formats an error reported for undefined variable.
 func UndefinedEnvError(envName string) error {
-	return fmt.Errorf("undefined environment variable: %s", envName)
+	return fmt.Errorf("undefined environment variable: %s. If you have started the Elastic stack using the elastic-package tool, "+
```
ycombinator, line 11:
```suggestion
	return fmt.Errorf("undefined environment variable: %s. If you started the Elastic stack using the elastic-package tool, "+
```

internal/stack/errors.go
```diff
@@ -8,5 +8,6 @@ import "fmt"
...
+		"please load stack environment variables using 'eval \"$(elastic-package stack shellinit)\"' or set their values manually", envName)
```
ycombinator, line 12: Nit:

```suggestion
		`please load stack environment variables using 'eval "$(elastic-package stack shellinit)"' or set their values manually`, envName)
```

internal/stack/errors.go
ycombinator: I'm good with this solution for now but I'm wondering how difficult it would be to detect programmatically if the stack was started up using `elastic-package`. Again, okay to implement something "smarter" like that in a follow up PR if it's not trivial. The changes you have here already are an improvement so I'm ++ to going forward with them for now.

cmd/stack.go
```diff
@@ -48,6 +48,8 @@ func setupStackCommand() *cobra.Command {
 				return cobraext.FlagParsingError(err, cobraext.StackVersionFlagName)
 			}
 
+			cmd.Println("Remember to load stack environment variables using 'eval \"$(elastic-package stack shellinit)\"'.")
```
ycombinator, line 51: Nit:

```suggestion
			cmd.Println(`Remember to load stack environment variables using 'eval "$(elastic-package stack shellinit)"'.`)
```

ycombinator (approved):
Left a couple of nitpicks and a suggestion for a follow up PR.

internal/stack/errors.go
mtojek: I know what you're suggesting, but I think I would stick to the if-sentence, to present what is expected in both cases.

Issue #189, opened 2020-11-25
Clarify which packages are being loaded into the package registry


ycombinator:
Currently, when `elastic-package stack up` is run, the package registry is rebuilt and packages from the local filesystem (if present) are copied into it. Depending on _where_ this command is run from, packages are picked up from different local filesystem folders.

### Case I: Command is run from outside the `integrations` repo folder

```
$ pwd
/Users/shaunak
$ elastic-package stack up -d -v
...
Building package-registry
Step 1/3 : FROM docker.elastic.co/package-registry/distribution:snapshot
 ---> 7d18928145b3
Step 2/3 : COPY ${STACK_DIR}/package-registry.config.yml /package-registry/config.yml
 ---> Using cache
 ---> 377e72d31f43
Step 3/3 : COPY ${STACK_DIR}/development/ /packages/development
 ---> Using cache
 ---> 4701b437cecb

Successfully built 4701b437cecb
Successfully tagged elastic-package-stack_package-registry:latest
...
```

Notice that packages are picked up from `${STACK_DIR}/development`.

### Case II: Command is run from inside the `integrations` repo folder (or one of it's sub-folders)

```
$ pwd
/Users/shaunak/development/github/integrations
$ elastic-package stack up -d -v
2020/11/25 14:35:49 DEBUG Enable verbose logging
Boot up the Elastic stack
Custom build packages directory found: /Users/shaunak/development/github/integrations/build/integrations
...
Building package-registry
Step 1/3 : FROM docker.elastic.co/package-registry/distribution:snapshot
 ---> 7d18928145b3
Step 2/3 : COPY ${STACK_DIR}/package-registry.config.yml /package-registry/config.yml
 ---> Using cache
 ---> 377e72d31f43
Step 3/3 : COPY ${STACK_DIR}/development/ /packages/development
 ---> Using cache
 ---> df38cea82eb7

Successfully built df38cea82eb7
Successfully tagged elastic-package-stack_package-registry:latest
...
```

Notice that packages are picked up from the `integrations/build/integrations` folder and also from the `${STACK_DIR}/development` folder.

---

I think it would be clearer if we could print out in both cases exactly which local folders packages are being picked up from. Something very obvious like:

> Packages from the following folders will be loaded into the package registry:
> - /path/to/folder/1
> - /path/to/folder/2

fearful-symmetry:
JUST ran into this and wasted like, over an hour being confused.

mtojek:
[...]

PR #215, opened 2021-01-11
Print loaded directories with packages

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/189

This PR adds log messages showing which directories will be included in the package-registry custom Docker image.

_I'm not sure if this is the expectation, but hopefully it will help understand the logic behind `stack up`._


internal/stack/boot.go
```diff
@@ -43,6 +43,13 @@ func BootUp(options Options) error {
 		}
 	}
 
+	fmt.Println("Packages from the following directories will be loaded into the package-registry:")
+	fmt.Println("- built-in packages (package-storage:snapshot Docker image)")
+
+	if found {
+		fmt.Printf("- %s (copy of: %s)\n", stackPackagesDir, buildPackagesPath)
```
ycombinator, line 50: I think we can just talk about `buildPackagesPath` here. Not sure if we should talk about `stackPackagePath` as it exposes `elastic-package` internals?

ycombinator (commented):
Left a minor comment/question.

internal/stack/boot.go
mtojek: Maybe you're right, I will leave only the `buildPackagesPath`

mtojek: Fixed.

ycombinator (approved):
LGTM.

PR #216, opened 2021-01-11
Update package_spec

AaronLamb1:
Bringing in the most recent package spec change for `unsigned_long` https://github.com/elastic/package-spec/pull/103


jonathan-buttner:
The reason that the config.yml files had to change name is because of this PR: https://github.com/elastic/package-spec/pull/101/files#diff-b53b009efa91873f46aeec782b4cc40f8d0c9675d2ffcc1f184adae44cd72192R6

That change to the spec requires that the test files config.yml files be in the form of `test-blah-config.yml`. Unfortunately this means that to bring in the `unsigned_long` we have to bring in that change too. I believe the functionality to support multiple `config.yml` files is being worked on here: https://github.com/elastic/elastic-package/pull/209 Until that PR is merged we won't be able to update the `elastic-package` tool with the most recent package-spec.

image

How do we feel about pushing through this `test-test-config.yml` change in the interim to unblock package-spec changes?

@mtojek @ycombinator

ycombinator:
Each config is supposed to be for a different system test within a data stream. So maybe just call them `test-default-config.yml` instead of `test-test-config.yml`?

mtojek (changes requested):
@AaronLamb1 Please clean up the https://github.com/elastic/package-spec/pull/103 first. I noticed that you pushed lots of dependencies which IMHO wasn't necessary. Please run `go mod tidy` against them.

mtojek:
> Each config is supposed to be for a different system test within a data stream. So maybe just call them test-default-config.yml instead of test-test-config.yml?

I think that @adriansr should introduce these changes first, down to Integrations. They're blocking right now.

jonathan-buttner:
> @AaronLamb1 Please clean up the [elastic/package-spec#103](https://github.com/elastic/package-spec/pull/103) first. I noticed that you pushed lots of dependencies which IMHO wasn't necessary. Please run `go mod tidy` against them.

@mtojek here's the PR to clean up go.mod: https://github.com/elastic/package-spec/pull/104

mtojek:
I think that @adriansr pushed his changes, including elastic-package. @jonathan-buttner @AaronLamb1 Could you please double-check if this PR is still valid?

jonathan-buttner:
closing because Adrian's PR pulled in the new spec 👍

PR #219, opened 2021-01-13
Collect Elastic stack logs

mtojek:
This PR introduces `elastic-package stack dump` command which can save container logs to local files.

Issue: https://github.com/elastic/integrations/issues/380

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ elastic-package
 
 # Build directory
 /build
+
+# stack dump
+elastic-stack-dump/
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -81,7 +81,7 @@ func setupStackCommand() *cobra.Command {
 	upCommand.Flags().BoolP(cobraext.DaemonModeFlagName, "d", false, cobraext.DaemonModeFlagDescription)
 	upCommand.Flags().StringSliceP(cobraext.StackServicesFlagName, "s", nil,
 		fmt.Sprintf(cobraext.StackServicesFlagDescription, strings.Join(availableServicesAsList(), ", ")))
-	upCommand.Flags().StringP(cobraext.StackVersionFlagName, "", stack.DefaultVersion, cobraext.StackVersionDescription)
+	upCommand.Flags().StringP(cobraext.StackVersionFlagName, "", stack.DefaultVersion, cobraext.StackVersionFlagDescription)
 
 	downCommand := &cobra.Command{
 		Use:   "down",
@@ -121,7 +121,7 @@ func setupStackCommand() *cobra.Command {
 			return nil
 		},
 	}
-	updateCommand.Flags().StringP(cobraext.StackVersionFlagName, "", stack.DefaultVersion, cobraext.StackVersionDescription)
+	updateCommand.Flags().StringP(cobraext.StackVersionFlagName, "", stack.DefaultVersion, cobraext.StackVersionFlagDescription)
 
 	shellInitCommand := &cobra.Command{
 		Use:   "shellinit",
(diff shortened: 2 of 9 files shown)


ycombinator:
Why not call it `elastic-package stack logs`?

mtojek:
> Why not call it `elastic-package stack logs`? Seems a bit more intuitive (to me anyway).

I was thinking about leaving an open gate for further improvement: https://github.com/elastic/integrations/issues/380#issuecomment-721014545

ycombinator:
> I was thinking about leaving an open gate for further improvement: elastic/integrations#380 (comment) . 

Ah, I see. That makes sense then. ++ to keeping it as `dump`. I will review the rest of this PR shortly.

.gitignore
```diff
@@ -22,3 +22,6 @@ elastic-package
 
 # Build directory
 /build
+
+# stack dump
+elastic-stack-dump/
```
ycombinator, line 27: Do you need this? We are already excluding `/build` and from what I can tell `elastic-stack-dump` lives under `build/`.

internal/cobraext/const.go
```diff
@@ -39,6 +39,9 @@ const (
 	StackServicesFlagName        = "services"
 	StackServicesFlagDescription = "component services (comma-separated values: %s)"
 
-	StackVersionFlagName    = "version"
-	StackVersionDescription = "stack version"
+	StackVersionFlagName        = "version"
+	StackVersionFlagDescription = "stack version"
```
ycombinator, line 43: Good catch, thanks for fixing!

internal/cobraext/const.go
```diff
@@ -39,6 +39,9 @@ const (
...
+
+	StackDumpOutputFlagName        = "output"
+	StackDumpOutputFlagDescription = "output location for "
```
ycombinator, line 46: This looks like an incomplete sentence?

cmd/stack.go
```diff
@@ -136,6 +136,28 @@ func setupStackCommand() *cobra.Command {
...
+			})
+			if err != nil {
+				return errors.Wrap(err, "dump failed")
+			}
+
+			cmd.Println("Done")
+			return nil
+		},
+	}
+	dumpCommand.Flags().StringP(cobraext.StackDumpOutputFlagName, "", "elastic-stack-dump", cobraext.StackDumpOutputFlagDescription)
```
ycombinator, line 159: How about changing the default value to `build/elastic-stack-dump`? It's consistent with `build/test-results`, which have a similar feel IMO.

cmd/stack.go
```diff
@@ -136,6 +136,28 @@ func setupStackCommand() *cobra.Command {
...
+			}
+
+			err = stack.Dump(stack.DumpOptions{
+				Output: output,
+			})
+			if err != nil {
+				return errors.Wrap(err, "dump failed")
+			}
+
+			cmd.Println("Done")
```
ycombinator, line 155: Consider printing out where the user can find the stack dump directory before "Done".

cmd/stack.go
mtojek: I did it on purpose, because the `elastic-package stack` is executed in the global scope, so it may happen there won't be the `build` directory. Do you agree with this behavior?

internal/cobraext/const.go
mtojek: Thanks, fixed

cmd/stack.go
mtojek: Fixed

cmd/stack.go
ycombinator: Hmm, I see your point. 

Perhaps I should explain my concern: I was rather hoping we would have a single top-level directory where we can collect all runtime artifacts rather than multiple top-level directories, increasing the level of "pollution" in a project using `elastic-package`. 

> ... so it may happen there won't be the `build` directory. 

In this case couldn't we create the `build` directory if it doesn't exist?

That said, I don't feel too strongly about this either way. We can always revisit the behavior and change if it becomes a problem later.

ycombinator (approved):

PR #220, opened 2021-01-13
System tests: start all services in compose project

adriansr:
This reverts part a change introduced by #209 as some tests may require more than one container running at the same time.

Relates https://github.com/elastic/integrations/issues/520


mtojek (approved):

mtojek:
Let me merge this PR.

Issue #218, opened 2021-01-12
Allow tests to be disabled


ycombinator:
In https://github.com/elastic/integrations/pull/514, @adriansr had to disable some packages' system tests in order to make progress on his feature. To do this, he had to remove the system test configuration files entirely because that is the only way to disable tests at the moment.

Instead, it would be nice if the configuration files accepted a `skip` or similar field which, if present, would skip the test. Further, the value of this field could be an arbitrary string representing the reason for skipping/disabling the test. Finally, this capability could be applied not just to system tests but also other types of tests.

* [x] `package-spec` changes to allow `skip` field: https://github.com/elastic/package-spec/pull/105
* [x] `elastic-package` changes to respect `skip` field in test runners: https://github.com/elastic/elastic-package/pull/221

ycombinator:
Resolved via elastic/package-spec#105 and #221.

PR #221, opened 2021-01-15
Add ability to skip tests

ycombinator:
This PR adds the ability for package developers to skip tests. 

### TODO
* [x] Implement skipping for asset loading test runner. Depends on #168.
* [x] Extract common `skip` struct from various test runners.
* [x] Add some skipped tests in test packages.

Related: #218.

### DIFF
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -107,6 +107,12 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 		}
 		tr.Name = tc.name
 
+		if tc.config.Skip != nil {
+			tr.Skipped = true
+			results = append(results, tr)
+			continue
+		}
+
 		result, err := simulatePipelineProcessing(r.options.ESClient, entryPipeline, tc)
 		if err != nil {
 			err := errors.Wrap(err, "simulating pipeline processing failed")
@@ -168,6 +174,13 @@ func (r *runner) loadTestCaseFile(testCaseFile string) (*testCase, error) {
 	config, err := readConfigForTestCase(testCasePath)
 	if err != nil {
 		return nil, errors.Wrapf(err, "reading config for test case failed (testCasePath: %s)", testCasePath)
+	}
+
+	if config.Skip != nil {
+		return &testCase{
+			name:   testCaseFile,
+			config: &config,
+		}, nil
 	}
 
 	ext := filepath.Ext(testCaseFile)
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -108,6 +108,10 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 		tr.Name = tc.name
 
 		if tc.config.Skip != nil {
+			logger.Warnf("skipping %s test for %s/%s: %s (details: %s)",
+				TestType, r.options.TestFolder.Package, r.options.TestFolder.DataStream,
+				tc.config.Skip.Reason, tc.config.Skip.Link)
+
 			tr.Skipped = true
 			results = append(results, tr)
 			continue
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -112,7 +112,7 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 				TestType, r.options.TestFolder.Package, r.options.TestFolder.DataStream,
 				tc.config.Skip.Reason, tc.config.Skip.Link.String())
 
-			tr.Skipped = true
+			tr.Skipped = tc.config.Skip
 			results = append(results, tr)
 			continue
 		}
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -110,7 +110,7 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 		if tc.config.Skip != nil {
 			logger.Warnf("skipping %s test for %s/%s: %s (details: %s)",
 				TestType, r.options.TestFolder.Package, r.options.TestFolder.DataStream,
-				tc.config.Skip.Reason, tc.config.Skip.Link)
+				tc.config.Skip.Reason, tc.config.Skip.Link.String())
 
 			tr.Skipped = true
 			results = append(results, tr)
--- a/internal/testrunner/runners/pipeline/test_config.go
+++ b/internal/testrunner/runners/pipeline/test_config.go
@@ -19,8 +19,7 @@ import (
 const configTestSuffix = "-config.json"
 
 type testConfig struct {
-	// Skip allows this test to be skipped.
-	Skip *testrunner.SkipConfig `config:"skip"`
+	testrunner.SkippableConfig `config:",inline"`
 
 	Multiline     *multiline             `json:"multiline"`
 	Fields        map[string]interface{} `json:"fields"`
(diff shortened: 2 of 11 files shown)


ycombinator:
I need to figure out why CI is not showing the `apache/error` system test as skipped.

mtojek:
> The xUnit formatter also appears to be correct, but I will probably need to play around with the attributes/elements and their values to get skipped tests working properly in CI.

If you prefer to play with it in follow up (bugfix), I'm good with it!

internal/testrunner/runners/pipeline/test_config.go
```diff
@@ -12,11 +12,16 @@ import (
...
 	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/testrunner"
 )
 
 const configTestSuffix = "-config.json"
 
 type testConfig struct {
+	// Skip allows this test to be skipped.
+	Skip *testrunner.SkipConfig `config:"skip"`
```
mtojek, line 23: I see that you placed `Skip` section in every config file, so I was wondering if it's a good idea to create a `commonTestConfig`?

internal/testrunner/testrunner.go
```diff
@@ -80,6 +80,45 @@ type TestResult struct {
...
 	ErrorMsg string
+
+	// If the test was skipped, the reason it was skipped and a link for more
+	// details.
+	Skipped *SkipConfig
+}
+
+// ResultComposer wraps a TestResult and provides convenience methods for
+// manipulating this TestResult.
+type ResultComposer struct {
```
mtojek, line 91: I like the idea of ResultComposer which is reusable.

internal/testrunner/runners/system/runner.go
```diff
@@ -105,53 +105,28 @@ func (r *runner) TearDown() error {
...
+func (r *runner) newResult(name string) testrunner.ResultComposer {
+	return testrunner.ResultComposer{
 		TestResult: testrunner.TestResult{
 			TestType:   TestType,
 			Name:       name,
 			Package:    r.options.TestFolder.Package,
 			DataStream: r.options.TestFolder.DataStream,
 		},
-		startTime: time.Now(),
+		StartTime: time.Now(),
```
mtojek, line 116: Maybe a constructor which will set the startTime? I saw you set this value in multiple places.

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -107,6 +107,16 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 		}
 		tr.Name = tc.name
 
+		if tc.config.Skip != nil {
```
mtojek, line 110: I think I have a general concern, but I'm sure if it can be solved. I'm worried that one day we'll add a new test runner, in which we'll forget to add the "Skip" config section. Maybe we should limit this to minimum. Do we need a skip check for assets tests (same for install)?

mtojek (approved):
Minors only and a general comment about the Skip section, but nothing blocking I believe.

internal/testrunner/runners/pipeline/runner.go
ycombinator: This is a good point. Maybe this goes well with your other suggestion of a `commonTestConfig`. Let me try something and update this PR.

internal/testrunner/runners/pipeline/test_config.go
ycombinator: In 4e25215, I introduced a `SkippableConfig` struct that could be embedded in any test config structs that allow skipping. Let me know what you think of this approach.

internal/testrunner/runners/system/runner.go
ycombinator: Done in d93be29.

internal/testrunner/runners/pipeline/runner.go
ycombinator: In 4e25215, I introduced a `SkippableConfig` struct that could be embedded in any test config structs that allow skipping. This doesn't necessarily address the concern you are bringing up (a future test runner forgetting to include the skip config section) but I think it can help make it a bit easier to add a skip config section to a future test runner. I think to truly address your concern a larger refactor might be needed which would allow a higher-level generic test runner of some sort to handle skipping before delegating the actual running of a test to specific types of test runners.

ycombinator:
@mtojek This PR is now failing CI with an unrelated error: https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-221/20/pipeline#step-122-log-693.

The same / similar error is occurring on `master` as well: https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/master/22/pipeline/#step-108-log-544

I would suggest that you review this PR for the latest changes but let's not merge it even when it passes review. I think we need to take care of the error on `master` separately first, then we can rebase this PR on `master` and see if it passes CI as well.

ycombinator:
I believe the CI error I mentioned in my previous comment will be resolved once we have a new Kibana 7.11 Docker image. Related: https://github.com/elastic/kibana/pull/89644.

ycombinator:
jenkins, run the tests please

internal/testrunner/runners/pipeline/test_config.go
mtojek: Composition, it should do the trick as well :)

internal/testrunner/runners/pipeline/runner.go
mtojek: > I think to truly address your concern a larger refactor might be needed which would allow a higher-level generic test runner of some sort to handle skipping before delegating the actual running of a test to specific types of test runners.

... and we may think about such refactoring exercise soon, as we have a better clarity of what do we expect from test runners and what kind of tests we need. Definitely we can postpone it at the moment.

test/packages/apache/data_stream/error/_dev/test/system/test-default-config.yml
```diff
@@ -1,3 +1,6 @@
+skip:
+  reason: testing skip
+  link: https://github.com/elastic/integrations/issues/123456789
 vars: ~
 data_stream:
   vars:
```
mtojek, line 6: I assume that this test will fail if it runs, that's why it's skipped = PASS

mtojek (approved):

test/packages/apache/data_stream/error/_dev/test/system/test-default-config.yml
ycombinator: This used to be a test that used to pass before I added the skip. Are you suggesting that I should make it fail, so we have a way to know something is not working correctly if the skip logic in the test runner breaks?

internal/testrunner/runners/pipeline/runner.go
ycombinator: Agreed. Added https://github.com/elastic/elastic-package/issues/238 to track the refactoring.

test/packages/apache/data_stream/error/_dev/test/system/test-default-config.yml
mtojek: Yes, that's point :) I think the quick way is to introduce some faulty variables? Don't have to implement anything special. But I wouldn't say it's blocking comment. For me this PR is ready to be merged.

ycombinator: Thanks for confirming, I will make this change now. Should be fairly quick to test.

ycombinator: I'm going to do a series of 3 commits to test this change:
1. A commit that comments out the `skip` section, effectively re-enabling the `apache.error` system test. With this commit the test should pass.
2. A commit that breaks the test. With this commit the test should fail.
3. A commit that adds back the `skip` section. With this commit the test should be skipped.

mtojek (approved):
LGTM!

PR #224, opened 2021-01-18
Fix asset test runner

mtojek:
This PR adjusts the asset test runner with a couple of fixes.

Changes:
* ingest pipeline may be undefined to logs-type packages
* use asset ID from inside of a Kibana saved object file
* format loaded assets in failure message


mtojek:
self, approved. I need to fix https://github.com/elastic/integrations/pull/536 .

Issue #89, opened 2020-09-09
[System test runner] Add more service deployers


ycombinator:
Follow up to #64. 

Currently the system test runner only supports the Docker Compose service deployer. That is, it can only test packages whose services can be spun up using Docker Compose. We should add more service deployers to enable system testing of packages such as `system` (probably a no-op or minimal service deployer), `aws` (probably some way to pass connection parameters and credentials via environment variables and/or something that understands terraform files), `kubernetes`.

mtojek:
For reference:

Testing on Kubernetes written by @ChrsMark : https://github.com/elastic/integrations/blob/master/testing/environments/kubernetes/README.md

mtojek:
We need to cover following providers:

- [ ] AWS
- [ ] Azure
- [ ] CloudFoundry
- [ ] GCP
- [ ] Kubernetes/OpenShift 

(correct me if I missed any of them)

mtojek:
@kaiyan-sheng @narph @ChrsMark 

Would you mind describing here use cases for AWS, Azure and Kubernetes? I'm looking forward to seeing how these cloud/infra providers can be used for testing integrations.

ChrsMark:
Thanks for the ping @mtojek, I will try to provide a scenario, with line comments/thoughts that would cover our k8s needs.

# Vanilla Kubernetes
1. Run `elastic-package k8s up` to bring up a k8s cluster, I don't think we should care where it is, on GKE or locally on minikube or kind. Maybe it will be better to have it running on GKE for now to avoid an extra step of minikube/kind installation (?). In this step all the required prerequisites should happen, like installing `kube_state_metrics` from which `state_*` metricsets will collect metrics from.
2. Run `elastic-package test k8s` (the syntax is abstract here for the sake of the example) so as to deploy agent on the running k8s cluster and enrol it to the Elastic Stack. Elastic stack should be running maybe on the same k8s cluster so as to have easier networking configuration to my mind (similar to the approach mentioned at [testing on k8s](https://github.com/elastic/integrations/blob/master/testing/environments/kubernetes/README.md)). After the test is completed the cluster is still up and the agents are still shipping metrics. To clean this up we need to run the next command to bring the whole cluster down.
3. Run `elastic-package k8s down` which will destroy the cluster including the Elastic stack and Agents.

Note: I think this scenario will can be expanded to test other packages like `istio` and `ingress-controller` by adding them as extra flags in step 1.

# OCP
[...]

PR #227, opened 2021-01-19
Implement terraform based system tests

mtojek:
~I'm opening this as draft to present the progress on the TF-based system tests.~

~Please do not consider this as final form, as I will modify it soon.~

Issue: https://github.com/elastic/elastic-package/issues/89

How to test this:

```
export AWS_ACCESS_KEY_ID="secret"
export AWS_SECRET_ACCESS_KEY="secret"

cd test/packages/aws
elastic-package test system --data-streams ec2_metrics -v
```

Tf definitions in: `data_stream/ec2_metrics/_dev/deploy/tf/main.tf` (or other .tf files)

EDIT:

I switched from SNS to EC2 metrics during tests as metrics tend to appear quicker.


mtojek:
Status update:

I managed to write the TF service deployer, but it needs to nice Terraform definition, which will actually perform something useful. A test scenario consist of two parts:
1. Test configuration (mainly vars): https://github.com/mtojek/elastic-package/blob/89-terraform-definitions/test/packages/aws/data_stream/sns/_dev/test/system/test-default-config.yml
2. Terraform definition which will bring up required AWS resources: https://github.com/mtojek/elastic-package/blob/89-terraform-definitions/test/packages/aws/data_stream/sns/_dev/deploy/tf/main.tf

@kaiyan-sheng I was wondering if you wouldn't like to play with it, share some feedback and write some Terraform files which can usually do sth useful (most wanted). I wrote a sample which creates SNS topic, but due to lack of messages, I don't see any CloudWatch metrics :(

FYI I haven't enabled the AWS account in the CI due to security reasons and cost-savings. Let's first check it on our workstations.

kaiyan-sheng:
@mtojek Will do(next week)! Thank you! I think if we can make the sns working, we can merge this and then add more tests to cover more datasets in other PRs.

mtojek:
@ycombinator With precious help of @kaiyan-sheng I managed to successfully follow the entire flow. This PR is ready for review now, but it depends on https://github.com/elastic/elastic-package/pull/228 .

test/packages/aws/docs/README.md
```diff
@@ -1311,10 +1311,14 @@ An example event for `ec2` looks as following:
 | aws.ec2.cpu.total.pct | The percentage of allocated EC2 compute units that are currently in use on the instance. | scaled_float |
 | aws.ec2.diskio.read.bytes | Bytes read from all instance store volumes available to the instance. | long |
 | aws.ec2.diskio.read.bytes_per_sec | Bytes read per second from all instance store volumes available to the instance. | long |
+| aws.ec2.diskio.read.count | The number of disk IO reads | long |
```
kaiyan-sheng, line 1314: Thanks for adding the fields!!!

kaiyan-sheng: Oh I just realized this is added in `elastic-package`. How can we sync `elastic/integrations/packages/aws` with `elastic-package/test.packages/aws`?

test/packages/aws/data_stream/ec2_metrics/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,9 @@
+vars:
+  access_key_id: '{{AWS_ACCESS_KEY_ID}}'
+  secret_access_key: '{{AWS_SECRET_ACCESS_KEY}}'
```
kaiyan-sheng, line 3: Can we also add session_token in vars?

internal/install/install.go
```diff
@@ -89,6 +98,21 @@ func ServiceLogsDir() (string, error) {
 	return filepath.Join(configurationDir, serviceLogsDir), nil
 }
 
+// ServiceDeployerComposeFile function returns the path to the selected service deployer's Docker Compose definition.
+func ServiceDeployerComposeFile(deployerName string) (string, error) {
```
ycombinator, line 102: Nit: I think you can just call this `TerraformDeployerComposeFile` and remove the `deployerName` argument. If/when we have more deployers that use a docker compose file for setup, we can generalize this then.

internal/install/install.go
```diff
@@ -139,6 +163,22 @@ func writeStackResources(elasticPackagePath string) error {
 	return nil
 }
 
+func writeDeployerResources(elasticPackagePath string) error {
```
ycombinator, line 166: Nit: since this is specific to terraform (for now at least), consider renaming it to `writeTerraformDeployerResources`. We can generalize this later if/when we have another deployer that needs resources to be written.

internal/install/static_terraform_deployer_yml.go
```diff
@@ -0,0 +1,20 @@
...
+
+package install
+
+const terraformDeployerYml = `version: '2.3'
+services:
+  terraform:
+    build: .
+    tty: true
+    environment:
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID} # TODO Extract to the env-file
```
ycombinator, line 13: Do you plan to address this TODO in this PR or a follow up PR? If it's for a follow up PR, can you make an issue for it so we don't lose track of it and post the link to the issue in the comment?

internal/testrunner/runners/system/runner.go
```diff
@@ -174,6 +176,10 @@ func (r *runner) run() (results []testrunner.TestResult, err error) {
 	return results, nil
 }
 
+func createTestRunID() string {
+	return fmt.Sprintf("%d", rand.Intn(99999-10000)+10000)
```
ycombinator, line 180: Maybe seed the PRNG each time with something like `rand.Seed(time.Now().UnixNano())`?

Also, nit: for an extra bit of readability, consider defining min and max constants and then using those in the `rand.Intn` call.

internal/testrunner/runners/system/servicedeployer/factory.go
```diff
@@ -5,27 +5,66 @@
...
+	PackageRootPath    string
+	DataStreamRootPath string
+}
+
+// Factory chooses the appropriate service runner for the given data stream, depending
+// on service configuration files defined in the package or data stream.
+func Factory(options FactoryOptions) (ServiceDeployer, error) {
+	devDeployPath, err := findDevDeployPath(options)
+	if err != nil {
+		logger.Errorf("can't find _dev/deploy directory")
```
ycombinator, line 35: You're defining `devDeployDir` as a constant earlier. Might as well use it here. 🙂

internal/testrunner/runners/system/servicedeployer/factory.go
```diff
@@ -5,27 +5,66 @@
...
 	if _, err := os.Stat(dockerComposeYMLPath); err == nil {
 		return NewDockerComposeServiceDeployer(dockerComposeYMLPath)
 	}
 
+	// Is the service defined using Terraform definition files?
+	terraformDirPath := filepath.Join(devDeployPath, "tf")
+	if _, err := os.Stat(terraformDirPath); err == nil {
+		return NewTerraformServiceDeployer(terraformDirPath)
+	}
 	return nil, ErrNotFound
```
ycombinator, line 50: Nit: please add newline to separate the final return statement from the two deployer cases.

internal/testrunner/runners/system/servicedeployer/factory.go
```diff
@@ -5,27 +5,66 @@
...
 
 	// Is the service defined using a docker compose configuration file?
-	dockerComposeYMLPath := filepath.Join(packageDevPath, "deploy", "docker", "docker-compose.yml")
+	dockerComposeYMLPath := filepath.Join(devDeployPath, "docker", "docker-compose.yml")
 	if _, err := os.Stat(dockerComposeYMLPath); err == nil {
 		return NewDockerComposeServiceDeployer(dockerComposeYMLPath)
 	}
 
+	// Is the service defined using Terraform definition files?
+	terraformDirPath := filepath.Join(devDeployPath, "tf")
```
ycombinator, line 46: With the addition of the terraform deployer and the logic in this `Factory` function, we now have a precedence ordering. If a data set somehow defines folders for both types of deployers we will choose the docker compose one. Maybe this is something we should document in the system testing HOWTO document?

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -0,0 +1,89 @@
...
+	tfEnvironment := tsd.buildTerraformExecutorEnvironment(inCtxt)
+	opts := compose.CommandOptions{
+		Env:       tfEnvironment,
+		ExtraArgs: []string{"--build", "-d"},
+	}
+	if err := p.Up(opts); err != nil {
+		return nil, errors.Wrap(err, "could not boot up service using docker compose")
+	}
+
+	// Build service container name
```
ycombinator, line 68: If I'm following correctly, we're creating a container for the terraform executor here, not the service being tested, right? If so, could you please fix the comment and variable names?

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -0,0 +1,89 @@
...
+	serviceContainer := fmt.Sprintf("%s_terraform_1", service.project)
+	outCtxt.Hostname = serviceContainer
+
+	// Connect service network with stack network (for the purpose of metrics collection)
```
ycombinator, line 72: Apologies if you explained this to me already a couple days ago, but what is the reason for connecting the terraform executor to the stack network? Will any of the stack containers (e.g. agent) need to talk to the terraform executor container?

internal/testrunner/runners/system/servicedeployer/terraform_env.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"os"
+)
+
+const (
+	awsAccessKeyID     = "AWS_ACCESS_KEY_ID"
+	awsSecretAccessKey = "AWS_SECRET_ACCESS_KEY"
+	awsProfile         = "AWS_PROFILE"
+	awsRegion          = "AWS_REGION"
```
ycombinator, line 16: It would be really nice if we could keep any mention of integration-specific variables/constants out of the elastic-package source code. I know there is [another place](https://github.com/elastic/elastic-package/pull/227/files#r565746042) where this is happening as well. Similar to the other place, could you please add a TODO comment here with a link to an issue for cleaning this up?

scripts/test-check-packages.sh
```diff
@@ -5,6 +5,9 @@ set -euxo pipefail
 cleanup() {
   r=$?
 
+  # Dump stack logs
+  elastic-package stack dump -v --output build/elastic-stack-dump/check
+
```
ycombinator, line 10: Is this something you added just for debugging this PR or do you want to keep it?

ycombinator (changes requested):
This is looking good. I left mostly minor comments. 

The big thing for me that needs to be addressed as part of this PR is updating the system test runner HOWTO documentation. Currently it talks only about the Docker Compose deployer. Please update it to talk about the Terraform deployer as well so users know what choices they have.

test/packages/aws/docs/README.md
mtojek: The goal of these packages in the elastic-package repo is to have a playground to test new feature before they reach out to Integrations. Syncing is actually the other way round - once we apply similar changes in the Integrations, I will copy over it from Integrations to the elastic-package :)

test/packages/aws/data_stream/ec2_metrics/_dev/test/system/test-default-config.yml
mtojek: Done, added support for session token.

internal/install/install.go
mtojek: Fixed

scripts/test-check-packages.sh
mtojek: I added this on purpose because during in the tests the stack is booted twice, but we dumped logs once. I needed this "dump" call to verify if AWS secrets are not leaking.

internal/testrunner/runners/system/servicedeployer/terraform.go
mtojek: It was for next use cases, where we may want to proxy traffic from an external endpoint, but as long as I removed ports I think we can remove it too.

I think we'll have to check the progress of research on running the Elastic Agent in the k8s cluster and see how we can combine it with elastic-package.

internal/install/install.go
mtojek: Fixed

internal/install/static_terraform_deployer_yml.go
mtojek: https://github.com/elastic/elastic-package/issues/235

internal/testrunner/runners/system/servicedeployer/terraform_env.go
mtojek: https://github.com/elastic/elastic-package/issues/235

internal/testrunner/runners/system/servicedeployer/terraform.go
mtojek: My bad, fixed. Thanks!

internal/testrunner/runners/system/runner.go
mtojek: Fixed

internal/testrunner/runners/system/servicedeployer/factory.go
mtojek: Fixed

internal/testrunner/runners/system/servicedeployer/factory.go
mtojek: Added, but I'm wondering why is this preference here :) ?

internal/testrunner/runners/system/servicedeployer/factory.go
mtojek: I described the change in docs (in this PR) and also described a bit the terraform service deployer (but without diving into details).

mtojek:
@ycombinator Thanks a lot for the review! I addressed your comments and introduced small changes in README. Please take a look and re-review whenever you've time.

internal/testrunner/runners/system/servicedeployer/factory.go
ycombinator: Total personal preference, I suppose :) To me the two cases seem to be their own separate logical blocks from everything around them. But this is one of those things you can totally ignore in the future. In general, feel free to ignore anything I prefix with `Nit: `.

test/packages/aws/docs/README.md
kaiyan-sheng: 👍 Sounds good!

docs/howto/system_testing.md
```diff
@@ -39,19 +38,33 @@ Packages have a specific folder structure (only relevant parts shown).
...
+        deploy/
+          <service deployer>/
+            <service deployer files>
+```
+
+`<service deployer>` - a name of the supported service deployer: `docker` or `tf` (terraform).
+
+### Docker Compose service deployer
 
 The `docker-compose.yml` file defines the integration service(s) for the package. If your package has a logs data stream, the log files from your package's integration service must be written to a volume. For example, the `apache` package has the following definition in it's integration service's `docker-compose.yml` file.
```
ycombinator, line 69: This sentence made sense earlier because the sample file structure above showed a `docker-compose.yml` file. Now the sample file structure above just says `<service deployer files>` so this sentence doesn't make as much sense.

I would add another introductory sentence to bridge this gap, something like:

> When using the Docker Compose service deployer, the `<service deployer files>` must include a `docker-compose.yml` file.

docs/howto/system_testing.md
```diff
@@ -39,19 +38,33 @@ Packages have a specific folder structure (only relevant parts shown).
...
+<package root>/
+  data_stream/
+    <data stream>/
+      _dev/
+        deploy/
+          <service deployer>/
+            <service deployer files>
+```
+
+`<service deployer>` - a name of the supported service deployer: `docker` or `tf` (terraform).
```
ycombinator, line 65: Might be worth clarifying that the `docker` name is for the Docker Compose service deployer and the `tf` name is for the Terraform service deployer. It's pretty obvious but it doesn't hurt to be explicit and tie these names to the sections that follow.

docs/howto/system_testing.md
```diff
@@ -66,7 +79,42 @@ services:
 
 Here, `SERVICE_LOGS_DIR` is a special keyword. It is something that we will need later.
 
-### Data stream-level configuration
+### Terraform service deployer
+
+The `*.tf` files define the infrastructure using the Terraform syntax. The terraform based service can be handy to boot up
```
ycombinator, line 84: Similar to my comment on the Docker Composer service deployer, I would add an introductory sentence to bridge the gap between what's shown in the sample file structure and this sentence here.

docs/howto/system_testing.md
```diff
@@ -66,7 +79,42 @@ services:
...
+resources on AWS and use them for testing (e.g. spawn EC2 instance and collect its metrics).
```
ycombinator, line 85: Should we specifically mention AWS here? Or can we be more generic and say resources supported by Terraform or something like that?

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -0,0 +1,76 @@
...
+		Env:       tfEnvironment,
+		ExtraArgs: []string{"--build", "-d"},
+	}
+	if err := p.Up(opts); err != nil {
+		return nil, errors.Wrap(err, "could not boot up service using docker compose")
+	}
+
+	// Build Terraform executor container name
+	serviceContainer := fmt.Sprintf("%s_terraform_1", service.project)
+	outCtxt.Hostname = serviceContainer
```
ycombinator, line 67: Do we need to set `outCtxt.Hostname` here? AFAICT it's not useful because the Agent configuration will never really need to talk to the Terraform executor container. Instead it will need to use the `outCtxt.CustomProperties` to talk to the service that is deployed by the Terraform executor.

docs/howto/system_testing.md
```diff
@@ -39,19 +38,33 @@ Packages have a specific folder structure (only relevant parts shown).
...
 <package root>/
   _dev/
     deploy/
-      docker/
-        docker-compose.yml
+      <service deployer>/
+        <service deployer files>
+```
+
+or the data-stream's level:
```
ycombinator, line 53:
```suggestion
or the data stream's level:
```

docs/howto/system_testing.md
```diff
@@ -39,19 +38,33 @@ Packages have a specific folder structure (only relevant parts shown).
   manifest.yml
 ```
 
-To define a system test we must define configuration at two levels: the package level and each data stream's level.
+To define a system test we must define configuration on at least one level: a package or a data stream's one.
 
-### Package-level configuration
-
-First, we must define the configuration for deploying a package's integration service. As mentioned in the [_Limitations_](#Limitations) section above, only packages whose integration services can be deployed via Docker Compose are supported at the moment.
+First, we must define the configuration for deploying a package's integration service. We can define it on either the package-level:
```
ycombinator, line 43:
```suggestion
First, we must define the configuration for deploying a package's integration service. We can define it on either the package level:
```

internal/testrunner/runners/system/servicedeployer/factory.go
ycombinator: Hmm, maybe I missed it but I'm not seeing anything in the docs in this PR that talk about the precedence ordering between the service deployers. I'm looking for something in the docs that would explain that only one service deployer should be defined for a package and if multiple ones are defined the Docker Compose service deployer will take precedence (or we can say the results are unpredictable).

ycombinator: How about also adding a check in this code to see if more than one service deployer is defined and then return an error in that case?

docs/howto/system_testing.md
mtojek: Added here and for the TF service deployer

docs/howto/system_testing.md
mtojek: Good point, docker and docker compose are different products.

docs/howto/system_testing.md
mtojek: Fixed

docs/howto/system_testing.md
mtojek: Fixed

internal/testrunner/runners/system/servicedeployer/terraform.go
mtojek: We can try this, I hope everything will still work (it's not expected value somewhere). Fixed.

ycombinator (approved):
LGTM.

internal/testrunner/runners/system/servicedeployer/factory.go
mtojek: I admit I forgot about his idea. Let me quickly implement this one.

ycombinator: Hahaha, seems like I forgot about it too 😄. Thanks for finding it again — just re-request my review on the PR as usual when you have this implemented. Thanks!

mtojek: Ok, now it's fixed :D

ycombinator (approved):
LGTM.

PR #229, opened 2021-01-20
Add option to notify test service after agent policy is applied

andrewkroh:
This adds a config option, `service_notify_signal`, to the system test runner to
allow the test runner to notify the service via a signal (e.g. SIGHUP) after the
policy has been applied by the Agent.

This is useful when the test service is connecting to the listening agent or streaming data
to the agent.

It helps address two issues that I encountered.
  1. While repeatedly testing the agent stays up with the old policy applied, but
      before the new policy is applied and the old ES data cleared the test service would
      connect and deliver data to the "old" agent/policy.
  2. For UDP streams there is no reliable way to know that the agent is ready and listening.

An alternative would be to start the test service container after the policy is applied, but
this could affect tests where the agent connects to the service (e.g. a prometheus exporter)
since the test might expect the service to be ready at startup.

### Test Logs

```
2021/01/21 09:03:13 DEBUG Wait until the policy (ID: 5df0cc00-5bf1-11eb-b44a-31c22f285a18) is assigned to all agents...
2021/01/21 09:03:15 DEBUG running command: /usr/local/bin/docker-compose -f /Users/akroh/go/src/github.com/elastic/elastic-package/test/packages/multiinput/_dev/deploy/docker/docker-compose.yml -p elastic-package-service kill -s SIGHUP test-udp
Killing elastic-package-service_test-udp_1 ... done
2021/01/21 09:03:16 DEBUG checking for expected data in data stream...
2021/01/21 09:03:16 DEBUG found 30 hits in logs-multiinput.test-ep data stream
2021/01/21 09:03:16 DEBUG reassigning original policy back to agent...
```

### DIFF
--- a/internal/testrunner/runners/system/servicedeployer/compose.go
+++ b/internal/testrunner/runners/system/servicedeployer/compose.go
@@ -104,6 +104,21 @@ func (r *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
 	return &service, nil
 }
 
+// Signal sends a signal to the service.
+func (s *dockerComposeDeployedService) Signal(signal string) error {
+	p, err := compose.NewProject(s.project, s.ymlPath)
+	if err != nil {
+		return errors.Wrap(err, "could not create docker compose project for service")
+	}
+
+	opts := compose.CommandOptions{ExtraArgs: []string{"-s", signal}}
+	if s.ctxt.Name != "" {
+		opts.Services = append(opts.Services, s.ctxt.Name)
+	}
+
+	return errors.Wrapf(p.Kill(opts), "could not send %q signal", signal)
+}
+
 // TearDown tears down the service.
 func (s *dockerComposeDeployedService) TearDown() error {
 	logger.Debugf("tearing down service using docker compose runner")
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -183,6 +183,20 @@ func (p *Project) Build(opts CommandOptions) error {
 	return nil
 }
 
+// Kill sends a signal to a service container.
+func (p *Project) Kill(opts CommandOptions) error {
+	args := p.baseArgs()
+	args = append(args, "kill")
+	args = append(args, opts.ExtraArgs...)
+	args = append(args, opts.Services...)
+
+	if err := p.runDockerComposeCmd(dockerComposeOptions{args: args, env: opts.Env}); err != nil {
+		return errors.Wrap(err, "running Docker Compose kill command failed")
+	}
+
+	return nil
+}
+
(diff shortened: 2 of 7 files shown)


andrewkroh:
Failing build is fixed by https://github.com/elastic/elastic-package/pull/230.

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -104,6 +104,21 @@ func (r *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
...
 }
 
+// Signal sends a signal to the service.
+func (s *dockerComposeDeployedService) Signal(signal string) error {
+	p, err := compose.NewProject(s.project, s.ymlPath)
+	if err != nil {
+		return errors.Wrap(err, "could not create docker compose project for service")
+	}
+
+	opts := compose.CommandOptions{ExtraArgs: []string{"-s", signal}}
```
mtojek, line 114: Would you mind adding a sample package to the `test/packages` (a copy from integrations it's also fine), so we can check if this improvement works?

mtojek (commented):
>An alternative would be to start the test service container after the policy is applied, but
this could affect tests where the agent connects to the service (e.g. a prometheus exporter)
since the test might expect the service to be ready at startup.

It looks like a classic "chicken and egg" problem. Most likely we'd have to ignore potential connectivity errors at the beginning. Maybe it's a better idea, but let's defer it at the moment.

Usually in such cases we depended on infinite stream of events (e.g. healthcheck calling `curl :80/`). I assume that you wouldn't like to follow this scheme, right?

internal/testrunner/runners/system/servicedeployer/compose.go
andrewkroh: I updated the multiinput test to exercise the feature. I added test log output to the PR description for reference.

andrewkroh:
> Usually in such cases we depended on infinite stream of events (e.g. healthcheck calling curl :80/). I assume that you wouldn't like to follow this scheme, right?

That sounds interesting, could you elaborate a little more what you were thinking. Like making the test service wait on some condition to become true? The only option I can think of is to access the API in kibana to monitor where the agent policy is applied.

mtojek:
The Apache integration is tested with the regular traffic generated by Docker healthcheck: https://github.com/elastic/integrations/blob/master/packages/apache/_dev/deploy/docker/Dockerfile#L5 . Maybe you can also adapt a similar logic (the healthcheck is a trigger).

mtojek (approved):
I think this feature/configuration option is useful and we may also use it in other places. If there isn't any simpler solution, let's not block it.

PR #230, opened 2021-01-20
Add tool dependencies to go.mod

andrewkroh:
Changes in tools and their dependencies can cause previously passing
builds to start failing because of 'make check'. This adds the project's tool
dependencies to go.mod so this no longer happens. An alternative would
be invoke use `GO111MODULE=off go get` when fetching tool dependencies
in the Makefile since this would not modify go.mod.

The root of the problem is the Makefile calls `go get -u` which adds the tool
dependency to go.mod AND updates any transitive dependencies of the tool.
Then `go mod tidy` is executed which removes the entries from go.mod but
leaves behind the go.sum updates to the transitive dependencies that the
project uses. Since go.sum is modified `make check` fails.

Reference: https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,15 +8,15 @@ clean:
 	rm -rf build
 
 format:
-	go get -u golang.org/x/tools/cmd/goimports
+	go get golang.org/x/tools/cmd/goimports
 	goimports -local github.com/elastic/elastic-package/ -w .
 
 lint:
-	go get -u golang.org/x/lint/golint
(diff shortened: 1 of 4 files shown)


andrewkroh:
This is intended to fix the failing build of master. Like https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/master/16/.

mtojek (approved):

PR #236, opened 2021-01-28
Remove whitespace from services list in help text

justinkambic:
# Summary

I had difficulty getting the service selection feature of this utility to work. My goal is to update the **help** text to help other new users avoid any difficulty.

## Problem

I was trying to use the `--help` flag for `elastic-package stack up` to stand up only package-registry and Elasticsearch. The **help** utility prints a message like:

```
  -s, --services strings   component services (comma-separated values: package-registry, elasticsearch, kibana)
```

But when I entered a command like:

```
elastic-package stack up -v --version 8.0.0-SNAPSHOT --services package-registry, elasticsearch
```

or

```
elastic-package stack up -v --version 8.0.0-SNAPSHOT --services "package-registry, elasticsearch"
```

I received error messages like:

```
2021/01/28 10:09:51 DEBUG Enable verbose logging
Boot up the Elastic stack
Error: validating services failed: service "" is not available
```

and

```
2021/01/28 10:10:11 DEBUG Enable verbose logging
Boot up the Elastic stack
Error: validating services failed: service " elasticsearch" is not available
```

## Proposed solution

I think we can improve the **help** message by having the example reflect the command that worked for me. I entered a command like:

```
elastic-package stack up -v --version 8.0.0-SNAPSHOT --services "package-registry,elasticsearch"
```

My change aims to make the output of this **help** message look like:

```
  -s, --services strings   component services (comma-separated values: package-registry, elasticsearch, kibana)
```

### Additional thought

It might also be good to put the entire example list in quotes, as this seemed to be necessary for me when running the utility. 

Example:

```
  -s, --services strings   component services (comma-separated values: "package-registry,elasticsearch,kibana")
```

### DIFF
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -52,11 +52,11 @@ func setupStackCommand() *cobra.Command {
 			}
 
 			services, err := cmd.Flags().GetStringSlice(cobraext.StackServicesFlagName)
-			common.TrimStringSlice(services)
-
 			if err != nil {
 				return cobraext.FlagParsingError(err, cobraext.StackServicesFlagName)
 			}
+
+			common.TrimStringSlice(services)
 
 			err = validateServicesFlag(services)
 			if err != nil {
@@ -85,7 +85,7 @@ func setupStackCommand() *cobra.Command {
 	}
 	upCommand.Flags().BoolP(cobraext.DaemonModeFlagName, "d", false, cobraext.DaemonModeFlagDescription)
 	upCommand.Flags().StringSliceP(cobraext.StackServicesFlagName, "s", nil,
-		fmt.Sprintf(cobraext.StackServicesFlagDescription, `"`+strings.Join(availableServicesAsList(), ",")+`"`))
+		fmt.Sprintf(cobraext.StackServicesFlagDescription, strings.Join(availableServicesAsList(), ",")))
 	upCommand.Flags().StringP(cobraext.StackVersionFlagName, "", stack.DefaultVersion, cobraext.StackVersionFlagDescription)
 
 	downCommand := &cobra.Command{
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -49,8 +49,10 @@ func setupStackCommand() *cobra.Command {
 			if err != nil {
 				return cobraext.FlagParsingError(err, cobraext.DaemonModeFlagName)
 			}
-
 			services, err := cmd.Flags().GetStringSlice(cobraext.StackServicesFlagName)
+
+			stripServicesWhitespace(services)
+
 			if err != nil {
 				return cobraext.FlagParsingError(err, cobraext.StackServicesFlagName)
 			}
@@ -82,7 +84,7 @@ func setupStackCommand() *cobra.Command {
 	}
 	upCommand.Flags().BoolP(cobraext.DaemonModeFlagName, "d", false, cobraext.DaemonModeFlagDescription)
 	upCommand.Flags().StringSliceP(cobraext.StackServicesFlagName, "s", nil,
-		fmt.Sprintf(cobraext.StackServicesFlagDescription, strings.Join(availableServicesAsList(), ",")))
+		fmt.Sprintf(cobraext.StackServicesFlagDescription, fmt.Sprintf("\"%s\"", strings.Join(availableServicesAsList(), ","))))
 	upCommand.Flags().StringP(cobraext.StackVersionFlagName, "", stack.DefaultVersion, cobraext.StackVersionFlagDescription)
 
 	downCommand := &cobra.Command{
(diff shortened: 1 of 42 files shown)


ycombinator:
> I think we can improve the help message by having the example reflect the command that worked for me.
> It might also be good to put the entire example list in quotes, as this seemed to be necessary for me when running the utility.

I like both your suggestions, @justinkambic. 

Additionally, I think we should strip whitespaces at the point in the code where we consume the values from the flag. 

I think this PR implements the first suggestion. Do you want to take a stab at implementing the other two? Or if you prefer I can push up a commit with those changes.

justinkambic:
@ycombinator I have contributed a few lines of my finest golang. LMK if the style doesn't work or feel free to recommend whatever changes.

cmd/stack.go
```diff
@@ -202,3 +204,9 @@ func validateServicesFlag(services []string) error {
 	}
 	return nil
 }
+
+func stripServicesWhitespace(services []string) {
```
ycombinator, line 208: Nothing about this helper function is specific to services so you could probably name this more generically, like `trimStringSliceSpace` or something. And the change the argument name to `strs` or something.

ycombinator: And then once you've done that, we should probably call this function from anywhere we also call `cmd.Flags().GetStringSlice(...)` in the codebase. So I'd suggest moving this function definition to `internal/common/helpers.go`. You'll need to rename it to start with an uppercase letter so it's exported. Then you can call it from here and other places via `common.TrimStringSliceSpace(...)`. Finally, you'll need to add an import at the top of this file's `import` section for `"github.com/elastic/elastic-package/internal/common"`.

cmd/stack.go
```diff
@@ -82,7 +84,7 @@ func setupStackCommand() *cobra.Command {
 	}
 	upCommand.Flags().BoolP(cobraext.DaemonModeFlagName, "d", false, cobraext.DaemonModeFlagDescription)
 	upCommand.Flags().StringSliceP(cobraext.StackServicesFlagName, "s", nil,
-		fmt.Sprintf(cobraext.StackServicesFlagDescription, strings.Join(availableServicesAsList(), ", ")))
+		fmt.Sprintf(cobraext.StackServicesFlagDescription, fmt.Sprintf("\"%s\"", strings.Join(availableServicesAsList(), ","))))
```
ycombinator, line 87: Nit: to prevent needing escaping the inner double quote.
```suggestion
		fmt.Sprintf(cobraext.StackServicesFlagDescription, fmt.Sprintf(`"%s"`, strings.Join(availableServicesAsList(), ","))))
```

ycombinator: Or you could just avoid the `fmt.Sprintf` altogether.
```suggestion
		fmt.Sprintf(cobraext.StackServicesFlagDescription, `"` + strings.Join(availableServicesAsList(), ",") + `"`))
```

cmd/stack.go
justinkambic: ok, will revise, thanks for feedback!

justinkambic: @ycombinator is there _already_ a file named `helpers.go` that I should be using or am I making it?

ycombinator: Sorry, you're making it.

justinkambic: Ok have a look at https://github.com/elastic/elastic-package/pull/236/commits/70a2fa4b8c2fa0f1d5f08f432ac26173fa4b40a6.

ycombinator (approved):
LGTM. I'm going to flag @mtojek as a second reviewer, just in case I missed something.

P.S. I pushed up a commit (7e1ad20) with a unit test for the `TrimStringSlice` helper function.

cmd/stack.go
```diff
@@ -82,7 +85,7 @@ func setupStackCommand() *cobra.Command {
 	}
 	upCommand.Flags().BoolP(cobraext.DaemonModeFlagName, "d", false, cobraext.DaemonModeFlagDescription)
 	upCommand.Flags().StringSliceP(cobraext.StackServicesFlagName, "s", nil,
-		fmt.Sprintf(cobraext.StackServicesFlagDescription, strings.Join(availableServicesAsList(), ", ")))
+		fmt.Sprintf(cobraext.StackServicesFlagDescription, `"`+strings.Join(availableServicesAsList(), ",")+`"`))
```
mtojek, line 88: Could you change the format defined in `StackServicesFlagDescription` instead of concatenating `"`?

cmd/stack.go
```diff
@@ -51,6 +52,8 @@ func setupStackCommand() *cobra.Command {
 			}
 
 			services, err := cmd.Flags().GetStringSlice(cobraext.StackServicesFlagName)
+			common.TrimStringSlice(services)
```
mtojek, line 55: First check the error, then trim whitespaces

mtojek (commented):
I didn't expect it may be interpreted this way, but looking at help hint you're right :) I left minor comments.

justinkambic:
Feedback implemented. Thank you @ycombinator for adding a test. LMK if escaped quotes are still undesirable, seemed ok in that case.

ycombinator (approved):
LGTM.

justinkambic:
thank you for fixing CI @ycombinator

Issue #235, opened 2021-01-28
Terraform deployer: implement env-file


mtojek:
Issue: https://github.com/elastic/elastic-package/pull/227#discussion_r565746042 https://github.com/elastic/elastic-package/pull/227#discussion_r565754848

> It would be really nice if we could keep any mention of integration-specific variables/constants out of the elastic-package source code. I know there is another place where this is happening as well. Similar to the other place, could you please add a TODO comment here with a link to an issue for cleaning this up?

The goal is to extract all cloud provider specific environment variables from the Docker image and just pass an env file with related ones.

mtojek:
@ycombinator Actually we need to design a clean way of passing these environment variables. It would mean that we'll have to load them in Docker runtime (not in service compose definition) and most likely we've to add a `env-file` with variable names in the "terraform" directory.

mtojek:
I'm going with this.

mtojek:
I research different methods in terms of user experience and security.

Method 1.

I started with a basic `env` file, which contains a list of global variables, that should be passed to the Terraform Docker application. `env` file's content:

```
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_SESSION_TOKEN
```

The file is processed by the system test runner and used to create a local `_dev/deploy/tf/env-file`:

```
AWS_ACCESS_KEY_ID="1111"
AWS_SECRET_ACCESS_KEY="2222"
AWS_SESSION_TOKEN="3333"
```

Unfortunately this file must be stored in the file system (`~/.elastic-package/tmp/env-file`) and is replaced every time per data stream's system test. It's a security threat as we're storing credentials in some local file and if the elastic-package doesn't exit correctly (process interrupted or crashed), the file is a left-over.

Method 2.

In the `_dev/deploy/tf` directory we add a new `.yml` file, e.g. `aws.yml`:

```yaml
version: '2.3'
services:
  terraform:
    environment:
      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
      - AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
      - AWS_PROFILE=${AWS_PROFILE}
      - AWS_REGION=${AWS_REGION:-us-east-1}
```

[...]

PR #241, opened 2021-02-02
Terraform service deployer: support user-defined environment variables

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/235

This PR modifies the existing TF service deployer and removes all references to global variables required by cloud providers (e.g. `AWS_...`). It means that the user is no longer required to modify and release the elastic-package in case of enabling new cloud provider (different environment variables).

Environment variables are not stored in the filesystem (no temporary files).

~I will update the `go.mod` once https://github.com/elastic/package-spec/pull/123 is merged.~


ycombinator (approved):
Fantastic job extracting all the env vars from the tooling to the data stream. So clean now! 

LGTM.

Issue #239, opened 2021-02-01
Support Kubernetes service deployer


mtojek:
Follow-up on: https://github.com/elastic/elastic-package/issues/89

This issue might be blocked by lack of recommendation for the managed fleet and Kubernetes cluster. Not sure if it's a temporary situation, but maybe we need to support unmanaged mode as well.

I suppose that technical details depend on the decision above (managed vs unmanaged). 

cc @ChrsMark @ycombinator

ChrsMark:
@mtojek how about supporting both cases (managed==fleet + unmanaged==standalone)? To my mind, standalone version will be a way to go for many people out there doing k8s ops based on infra as code principles, and hence I expect us having both ways supported.

mtojek:
We may consider to support both modes, maybe even break it into two iterations.

Speaking about unmanaged - let's build the design together:

1. Installed `kubectl` and cluster-context configured. 
The version/format of kubectl/kubeconfig seems to be important, not sure if we can can replace it with a docker image for `kubectl`. With `kubectl` you can select the target cluster, either local one deployed with `kind` or remote one.
This way we can easily use GCP clusters (`gcloud container clusters get-credentials cluster-name` will re–configure the `kubectl`).

2. `_dev/deploy/k8s/`: place to store `deployment.yaml` files, describing the standalone Elastic-Agent deployment (with configuration).

3. System test runner uses kubectl to deploy the Elastic Agent (standalone) in the Kubernetes cluster.

Question:

1. Where should be placed the Elastic stack? New one on Kubernetes? Reuse of the Docker-based stack? I'm not sure if we can connect the Elastic Agent in the Kubernetes cluster to the Kibana deployed in the Docker container (IP/routing).

mtojek:
Speaking about managed:

I would assume that the Elastic Agent is deployed as part of the Elastic stack and we pass all required configuration options/secrets. It's not distributed as DaemonSet, but it can reach out to the cluster via HTTP API.

This method might be easier to implement in the elastic-package, but it's not the original way of running the Agent.

ChrsMark:
> We may consider to support both modes, maybe even break it into two iterations.
> 
> Speaking about unmanaged - let's build the design together:
> 
> 1. Installed `kubectl` and cluster-context configured.
>    The version/format of kubectl/kubeconfig seems to be important, not sure if we can can replace it with a docker image for `kubectl`. With `kubectl` you can select the target cluster, either local one deployed with `kind` or remote one.
[...]

PR #243, opened 2021-02-05
Try: check seccomp settings

mtojek:
Please do not merge the draft PR. I'm checking CI settings.


mtojek:
It looks that we're able to use seccomp options and enabled priviledged mode.

PR #245, opened 2021-02-05
Add Kubernetes as test package (pod data stream)

mtojek:
This PR partially copies the Kubernetes package to the `test/packages` directory.

Comment: This is a raw copy of the package from Integrations. I just modified the version to `999.999.999`, so it has the highest priority.


ChrsMark:
@mtojek if I understand correctly you need this as a base to work on the new k8s service deployer right?

mtojek:
> @mtojek if I understand correctly you need this as a base to work on the new k8s service deployer right?

Yes, that's the plan. We used introduced such packages as playground for testing purposes.

ChrsMark (approved):

PR #246, opened 2021-02-05
Implement Kubernetes service deployer

mtojek:

How to use it:

```bash
elastic-package stack up -d
eval "$(elastic-package stack shellinit)"

kind create cluster

cd test/packages/kubernetes
elastic-package test system --data-streams pod -v
```

TODOs:

- [x] Check kubectl context == `kind-kind`
- [x] Attach `kind` container to the Elastic stack network
- [x] Install Elastic Agent in the cluster
- [x] Install custom definitions if present
- [x] Test execution: List cluster agent only
- [x] Uninstall custom definitions
- [x] CI: install `kind` and `kubectl`
- [x] Modify the pod system test to use the Kubernetes service deployer
- [x] Update package-spec to support new files
- [x] Merge and use https://github.com/elastic/package-spec/pull/130
~- [ ] Update system docs~ I'll extract this to a separate PR, so we can push the implementation and write docs based on the pushed state.


mtojek:
jenkins, run the tests please

mtojek:
Ok, I spent some time investigating the problem with failing tests and found out the root cause. I'm not sure whether I should adjust the logic in the test runner (let's wait for a discussion in the other issue).

For reference: https://github.com/elastic/beats/issues/23928

Nevertheless, I believe this PR is ready for review.

internal/install/static_kubernetes_elastic_agent_yml.go
```diff
@@ -0,0 +1,125 @@
...
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package install
+
+const kubernetesDeployerElasticAgentYml = `---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kind-fleet-agent-clusterscope
```
ChrsMark, line 11: you are using `pod` dataset which actually queries from `kubelet`'s `stats/summary` endpoint. This one should be considered as node-scope since it will be performed on all nodes through the Daemonset. 
cluater-scope goes for `state_*` metricsets as well as `apiserver`.

mtojek:
I managed to reproduce the current issue with system tests. It's also related to the Agent's issue: https://github.com/elastic/beats/issues/23928#issuecomment-776028086

ChrsMark:
@mtojek regarding the issue with policy token, I guess that you are planning to use only one policy for the tests, right? If you want to actually enrol both `cluster-scope` and `node-scope` agents you will need 2 different policies as you can see at https://github.com/elastic/beats/blob/master/deploy/kubernetes/elastic-agent-kubernetes.yaml. I wonder if this should be considered here when you just bring stack and k8s up, since it would be useful for the development/pseudoproduction phase in order to simulate a scenario of deploying all possible datastreams on your Agents running on k8s cluster. Would that make sense?

mtojek:
Thank you for taking a look, Christos!

The problem with https://github.com/elastic/beats/blob/master/deploy/kubernetes/elastic-agent-kubernetes.yaml is that it deploys two agents for node and cluster. I'm not sure if we need two separate instances to test the integration (actually it's about endpoints/API not the actual deployment). Is this the way to go? Always in-pair?
I don't know why does the YAML file defines two instances of agent instead of two separate YAML files (one per agent). In this PR the pod endpoint I used, was available to the agent in cluster scope.

I'd like to stick to the principle that we should test integrations not the deployment model. It's more like an end-to-end testing, which we're not doing here.

I'm happy to adjust the PR to better fit our needs, but I'd like to understand if this model won't meet our requirements - won' be possible to test some Kubernetes resources.

ChrsMark:
Well the model of deploying 2 instances was inherited from Metricbeat (https://github.com/elastic/beats/blob/v7.9.0/deploy/kubernetes/metricbeat-kubernetes.yaml).

However, now I'm thinking of it again and it should be ok to test all datastreams with only one single Agent, both manually but also with system tests so you can just by-pass my previous comment and go ahead with your desing. Thanks!

internal/install/static_kubernetes_elastic_agent_yml.go
```diff
@@ -0,0 +1,125 @@
...
+  template:
+    metadata:
+      labels:
+        app: kind-fleet-agent-clusterscope
+        group: fleet
+    spec:
+      serviceAccountName: kind-fleet-agent
+      containers:
+        - name: kind-fleet-agent-clusterscope
+          image: docker.elastic.co/beats/elastic-agent:{{ STACK_VERSION }}
```
ycombinator, line 29: IIUC, when the k8s deployer is used, we'll spin up a container in the k8s cluster for the Elastic Agent. Additionally, we will still have the Elastic Agent that is spun up by `elastic-package stack up`, right?

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -0,0 +1,269 @@
...
+	cmd := exec.Command("kubectl", "config", "current-context")
+	errOutput := new(bytes.Buffer)
+	cmd.Stderr = errOutput
+	output, err := cmd.Output()
+	if err != nil {
+		return errors.Wrapf(err, "kubectl command failed (stderr=%q)", errOutput.String())
+	}
+	currentContext := string(bytes.TrimSpace(output))
+
+	if currentContext != "kind-kind" {
```
ycombinator, line 168:
```suggestion
	if currentContext != kindContext {
```

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	for _, c := range networkDescriptions[0].Containers {
+		if c.Name == kindControlPlaneContainerName {
+			logger.Debugf("container %s is already attached to the %s network", kindControlPlaneContainerName, stackNetwork)
+			return nil
+		}
+	}
+
+	logger.Debugf("attach service container %s (ID: %s) to stack network %s", kindControlPlaneContainerName, controlPlaneContainerID, stackNetwork)
```
ycombinator, line 224:
```suggestion
	logger.Debugf("attach kind control plane container %s (ID: %s) to stack network %s", kindControlPlaneContainerName, controlPlaneContainerID, stackNetwork)
```

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "can't unmarshal network inspect for %s (stderr=%q)", stackNetwork, errOutput.String())
+	}
+
+	if len(networkDescriptions) != 1 {
+		return fmt.Errorf("expect single network inspect record, got %d entries", len(networkDescriptions))
+	}
+
+	for _, c := range networkDescriptions[0].Containers {
+		if c.Name == kindControlPlaneContainerName {
```
ycombinator, line 218: Under what circumstances would the kind control plane container be **already** attached to the stack network?

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -0,0 +1,269 @@
...
+		args = append(args, definitionPath)
+	}
+
+	cmd := exec.Command("kubectl", args...)
+	errOutput := new(bytes.Buffer)
+	cmd.Stderr = errOutput
+	if err := cmd.Run(); err != nil {
+		return errors.Wrapf(err, "kubectl apply failed (stderr=%q)", errOutput.String())
+	}
+	return nil
```
ycombinator, line 268: We are doing something similar in `verifyKindContext`. Does it make sense to extract this part into its own function  for running generic `kubectl` commands?

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -0,0 +1,269 @@
...
+func (s kubernetesDeployedService) Context() ServiceContext {
+	return s.ctxt
+}
+
+func (s kubernetesDeployedService) SetContext(sc ServiceContext) error {
+	s.ctxt = sc
+	return nil
+}
+
+var _ DeployedService = new(kubernetesDeployedService)
```
ycombinator, line 73: Maybe I'm missing something obvious or some idiom, but why do we need this line?

internal/install/static_kubernetes_elastic_agent_yml.go
mtojek: Right, I'm starting an additional instance of the Elastic Agent (lazy initialization) and leaving it up there for next tests.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
mtojek: Thanks, fixed

internal/testrunner/runners/system/servicedeployer/kubernetes.go
mtojek: I adjusted the comment to the similar form.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
mtojek: As the Elastic Agent is loaded lazily, since execution of the second system test, the container is already connected.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
mtojek: https://golang.org/doc/faq#guarantee_satisfies_interface

internal/testrunner/runners/system/servicedeployer/kubernetes.go
mtojek: Maybe these two functions are not good candidates to get unified, but this might a good point to extract it to the `kind` package.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
ycombinator: Cool, I was not aware of this. Thanks!

internal/testrunner/runners/system/servicedeployer/kubernetes.go
ycombinator: I'm not opposed to this but I'm wondering if it would be better (clean slate approach) to instead utilize the `TearDown` method to detach the network and remove any containers started up in `SetUp` method?

internal/testrunner/runners/system/servicedeployer/kubernetes.go
ycombinator: ++ to `kind` package. That would be even better!

internal/testrunner/runners/system/servicedeployer/kubernetes.go
mtojek: My intention was to cut the total execution time, not to waste time on booting up the agent, but I can adjust to it.

ycombinator: Ah okay, let's keep it this way. If we see any issues we can adjust to do more teardown and re-setup.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
mtojek: Fixed. Introduced `kind`, `kubectl` and added some changes to `docker` package.

mtojek:
jenkins run the tests please

ycombinator (approved):
LGTM. 

This was not an easy service deployer to build, IMO. Very good job at iterating on it and then ultimately arriving at this PR.

mtojek:
Thank you for reviewing. Let's hold on with merging this PR immediately as we need the changes in Agent to go through first: https://github.com/elastic/beats/pull/24001 . Not sure in which branch it will land finally, in the worst case I will bump the stack version to 7.12.0-SNAPSHOT.

mtojek:
I assume that the next revision of Elastic Agent will contain the bugfix (default policy selection) and I'll be able to merge this PR.

mtojek:
The CI job has finally passed. I'll merge this one and work on enabling it in integrations (install kind).

PR #247, opened 2021-02-08
Allow installing older versions of the package

ycombinator:
### What does this PR do

This PR allows the system test runner to install the version of the package under test, even if that version may be older than the latest version of the same package available in the package registry. 

### Why is this PR important

Package developers may want to create new patch versions from older minor versions of a package, even though a newer minor version of the same package might already have been released. For example, `system-0.9.2` and `system-0.10.5` may already have been released. A package developer might want to create+test+release a `system-0.9.3` package next that builds upon the `system-0.9.2` package but does not or cannot include all the changes from the `system-0.10.5` package.

### DIFF
--- a/internal/kibana/packages.go
+++ b/internal/kibana/packages.go
@@ -16,7 +16,9 @@ import (
 // InstallPackage installs the given package in Fleet.
 func (c *Client) InstallPackage(pkg packages.PackageManifest) ([]packages.Asset, error) {
 	path := fmt.Sprintf("%s/epm/packages/%s-%s", FleetAPI, pkg.Name, pkg.Version)
-	statusCode, respBody, err := c.post(path, nil)
+	reqBody := []byte(`{"force":true}`) // allows installing older versions of the package being tested
+
+	statusCode, respBody, err := c.post(path, reqBody)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ycombinator:
To test the changes in this PR:

- First, I will push a commit that deliberately bumps **down** a test package's version to something that's known to be lower than versions already released in the package registry. This commit should fail CI with an error saying that the older version of the test package cannot be installed.
- Then I will push a commit with the code change that will allow installing older versions of test packages. This commit should make CI pass again.

ycombinator:
As expected (see [previous comment](https://github.com/elastic/elastic-package/pull/247#issuecomment-775476867)), the first commit has [failed CI](https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-247/5/pipeline#step-150-log-580) with the expected error:

```
[2021-02-08T21:36:56.117Z] Error: error running package asset tests: could not complete test run: could not install package: could not install package; API status code = 400; response body = {"statusCode":400,"error":"Bad Request","message":"apache-0.0.1 is out-of-date and cannot be installed or updated"}
```

mtojek (approved):
Ship it!

Issue #87, opened 2020-09-08
[System test runner] Install package?


ycombinator:
Follow up to https://github.com/elastic/elastic-package/pull/64, based on discussion in https://github.com/elastic/elastic-package/pull/64#discussion_r485172111:

> Actually you don't use any "install" API like the one that is used in integration tests of package-registry. It looks like it's not required to be involved in the process, so I wonder what does it do actually? ... I was thinking about the reference to this API method: https://github.com/elastic/package-storage/blob/production/testing/main_integration_test.go#L108 

> Hmmm, good question — I wonder how I'm able to get things working even without calling this API. 🤔 Let's leave it out of this PR for now and add it if we need it in a follow up PR.

PR #250, opened 2021-02-15
CLI to install a package

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/87
Fixes: https://github.com/elastic/elastic-package/issues/8

This PR exposes CLI action to install the current package (work dir), available in the Package Registry.

It introduces also the "check package conditions" feature to verify if the package meets Elastic stack conditions. It will be used by https://github.com/elastic/integrations/issues/537#issuecomment-779253207 to check in advance whether the stack can expose the package.

How to test the PR:

1. Go to the `test/packages/kubernetes`.
2. `elastic-package build`
3. `elastic-package stack up -d` (to include the package)
4. `elastic-package install` (to install the package)

Check constraints (Kibana is not required to be up):

```
➜  apache git:(87-install-package) elastic-package install -c kibana.version=7.8.0 -v
2021/02/16 13:00:00 DEBUG Enable verbose logging
Check conditions for package
2021/02/16 13:00:00 DEBUG Verify Kibana version (constraint: ^7.9.0, requirement: 7.8.0)
Error: checking conditions failed: Kibana constraint unsatisfied: [0] 7.8.0 does not have same major version as 7.9.0
➜  apache git:(87-install-package) elastic-package install -c kibana.version=7.11.0 -v
2021/02/16 13:00:04 DEBUG Enable verbose logging
Check conditions for package
2021/02/16 13:00:04 DEBUG Verify Kibana version (constraint: ^7.9.0, requirement: 7.11.0)
2021/02/16 13:00:04 DEBUG Constraint kibana.version = ^7.9.0 is satisfied
Done
```


cmd/install.go
```diff
@@ -0,0 +1,95 @@
...
+  package`
+
+func setupInstallCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install",
+		Short: "Install the package",
+		Long:  installLongDescription,
+		RunE:  installCommandAction,
+	}
+	cmd.Flags().BoolP(cobraext.UninstallPackageFlagName, "u", false, cobraext.UninstallPackageFlagDescription)
```
ycombinator, line 30: It feels a little odd (to me) that we use `elastic-package install -u` to uninstall a package. I think it would be clearer to create an `elastic-package uninstall` command instead. Even in terms of implementation this would be cleaner IMO — for example, it doesn't really make sense to check any compatibility conditions when uninstalling a package.

mtojek: I agree with uninstall. 

Regarding the compatibility conditions... I need this feature somewhere in elastic-package, so that I can implement the support matrix between packages and stacks (https://github.com/elastic/integrations/issues/537#issuecomment-779253207). I planned to use the tool to check if the CI should spawn the stack, because the package should be compatible.

If you have other ideas for the support check or a better place to put it, feel free to share them.

ycombinator: No no, I think the compatibility check the way you have it implemented for `elastic-package install` is just fine. I was saying it is not needed for `elastic-package uninstall`, that's all.

ycombinator: This might helps clarify: I would remove the `-u` option from this PR and make it just about installing a package. Then put up a separate PR that implements `elastic-package uninstall` (and in this PR I think we don't need the compatibility checks).

mtojek: I removed uninstallation from this PR.

internal/packages/conditions.go
```diff
@@ -0,0 +1,71 @@
...
+
+type packageRequirements struct {
+	kibana struct {
+		version *semver.Version
+	}
+}
+
+// CheckConditions method compares the given values with conditions in manifest.
+// The method is useful to check in advance if the package is compatible with particular stack version.
+func CheckConditions(manifest PackageManifest, keyValuePairs []string) error {
```
ycombinator, line 28: I really like how generic this function is. I have been thinking of adding more conditions as part of https://github.com/elastic/package-registry/issues/670 and I think something like what you have here will come in very handy!

internal/packages/conditions.go
```diff
@@ -0,0 +1,71 @@
...
+	requirements, err := parsePackageRequirements(keyValuePairs)
+	if err != nil {
+		return errors.Wrap(err, "can't parse given keyValue pairs as package conditions")
+	}
+
+	// So far, Kibana is the only supported constraint
```
ycombinator, line 34: Nit: consistency.
```suggestion
	// So far, Kibana is the only supported condition
```

ycombinator (approved):
LGTM.

Issue #226, opened 2021-01-18
test asset: validate fields in sample event


mtojek:
The elastic-package should validate if sample events (JSON files) contain fields defined in the `fields` folder. I spotted in the Integrations repository, that some files don't only contain events, but also wrapping fields with underscores:

```
{
    "_index": "metrics-aws.sns-default-000001",
    "_id": "Ds9vXHIBpGMSUzkZmyod",
    "_version": 1,
    "_score": null,
    "_source": {
        "@timestamp": "2020-05-28T17:58:27.154Z",
        "service": {
            "type": "aws"
        },
        "stream": {
            "dataset": "aws.sns",
            "namespace": "default",
            "type": "metrics"
        },
        "ecs": {
            "version": "1.5.0"
        },
```

For reference:
https://github.com/elastic/integrations/blob/master/packages/aws/data_stream/sns/sample_event.json#L6

ycombinator:
Just to clarify: this issue is about making sure that we don't end up with underscore-delimited fields like `_source`, etc. in `sample-event.json` files, correct?

mtojek:
Yes, most likely it will result also in adjustments in current integrations.

mtojek:
Based on the comment https://github.com/elastic/elastic-package/pull/251#issuecomment-781361651 , I will create a new test runner, let's say - `static`, which can be used to check static resources, which are not installed in Kibana (they're just static).

mtojek:
There is another constraint (just discovered). All "per data stream" test runners require to have a corresponding `_/dev/test/<test_type>` available. I wouldn't like to hardcode a hack for Static Test Runner not to require such directory, so I'll try to introduce a similar mechanism like `CanRunPerDataStream`, e.g `TestFolderRequired`.

ycombinator:
You mean like this: https://github.com/elastic/elastic-package/pull/168#discussion_r525055005? 😉

mtojek:
A kind of :) I started with a bit different implementation, we can discuss it on code review.

PR #251, opened 2021-02-16
Test sample event

mtojek:
This PR enables field validation for sample events.

Related: https://github.com/elastic/elastic-package/issues/226


mtojek:
@ycombinator The status of this PR is WIP, because I'd like to discuss one issue. I thought it might be a quick win, but I hit the wall while introducing the improvement to the asset test runner. This test runner operates in package scope and I would need one running in a data stream scope. I'm wondering whether I should introduce another runner for static files (e.g. sample events), let's say - `test static`?

Another option would be placing it somewhere here https://github.com/elastic/elastic-package/blob/master/internal/docs/sample_event.go and fail during the build phase.

ycombinator:
When I think of package assets, I think of things that can be installed into the Elastic Stack, e.g. ingest pipelines, kibana visualizations, etc., things that Kibana would install when the package is installed. 

By that measure, the sample event JSON isn't a package asset. It's more similar to the package README or even the CHANGELOG, in that it's something that's _informational_ for the package consumer.

So I would suggest implementing this as another test runner instead of making it part of the asset test runner. This new test runner would be able to operate at the data stream level (so it would return `true` for [`CanRunPerDataStream()`](https://github.com/elastic/elastic-package/blob/master/internal/testrunner/testrunner.go#L49)).

WDYT?

mtojek:
Resolving in favor of new test runner.

PR #256, opened 2021-02-18
docs: Kubernetes service deployer

mtojek:
This PR introduces docs for the Kubernetes service deployer, which has already been enabled in the Integrations repository.

### DIFF
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -62,7 +62,10 @@ or the data stream's level:
             <service deployer files>
 ```
 
-`<service deployer>` - a name of the supported service deployer: `docker` (Docker Compose service deployer) or `tf` (Terraform service deployer).
+`<service deployer>` - a name of the supported service deployer:
+* `docker` - Docker Compose
+* `k8s` - Kubernetes
+* `tf` - Terraform
 
 ### Docker Compose service deployer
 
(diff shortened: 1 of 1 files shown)


ChrsMark (commented):
lgtm.
May to add a complete `bash` snippet with all the commands in the row so as to execute the tests, ie:
```
kind create cluster
elastic-package stack up -d -v
elastic-package k8s system tests
....
```

mtojek:
> May to add a complete bash snippet with all the commands in the row so as to execute the tests

Done, thanks

ChrsMark (approved):
Awesome! Glad to see that happening!

docs/howto/system_testing.md
```diff
@@ -118,6 +121,30 @@ data "aws_ami" "latest-amzn" {
 
 Notice the use of the `TEST_RUN_ID` variable. It contains a unique ID, which can help differentiate resources created in potential concurrent test runs.
 
+### Kubernetes service deployer
+
+The Kubernetes service deployer requires the `_dev/deploy/k8s` directory to be present. It can include additional `*.yaml` files to deploy
```
ycombinator, line 126: I just noticed here that we are expecting `*.yaml` files under `_dev/deploy/k8s`. Sorry I missed it in the [spec PR](https://github.com/elastic/package-spec/pull/130/files#diff-16e6de8c0f8acc5369cc32adf906667fdf37e0e570af8ae5a43a142f31edf668R5). Any reason we don't expect `*.yml` files like we do everywhere else?

ycombinator: BTW, this is not a blocker for this documentation PR as this documentation reflects the reality. 

But if there is no good reason to go with `*.yaml` here instead of `*.yml` like we do everywhere else, I would prefer that we switch to `*.yml` files under `_dev/deploy/k8s` too (starting with the spec, etc.)

docs/howto/system_testing.md
```diff
@@ -118,6 +121,30 @@ data "aws_ami" "latest-amzn" {
...
+custom applications in the Kubernetes cluster (e.g. Nginx deployment). If no resource definitions (`*.yaml` files ) are needed,
+the `_dev/deploy/k8s` directory must contain an `.empty` file (to preserve the `k8s` directory under version control).
+
+The Kubernetes service deployer needs the [kind](https://kind.sigs.k8s.io/) to be installed and the cluster to be up and running:
```
ycombinator, line 130:
```suggestion
The Kubernetes service deployer needs [kind](https://kind.sigs.k8s.io/) to be installed and the cluster to be up and running:
```

docs/howto/system_testing.md
mtojek: Yes, it's on purpose as in the Kubernetes universe they're handling `*.yaml` files (not `*.yml`).

ycombinator: Gotcha, thanks!

ycombinator (approved):
LGTM.

PR #258, opened 2021-02-19
Support data stream type == traces

### LINKED ISSUE #223: Support `traces` data stream type
In https://github.com/elastic/package-spec/pull/78, a new data stream type, `traces`, was added for supporting APM packages. This may have implications on `elastic-package`. We should figure out what those might be and support this new type.

cc: @elastic/apm-server.

ycombinator:
Looking in all three active branches of the `package-storage` repo, I have confirmed that the `apm` package is the only one that is using `type: traces` in [one of its data streams](https://github.com/elastic/package-storage/blob/c5925eb82898dfc3e879a521871c7383513804c7/packages/apm/0.1.0-dev.5/data_stream/traces/manifest.yml#L2). I looked at this data stream and it defines [ingest pipelines](https://github.com/elastic/package-storage/tree/snapshot/packages/apm/0.1.0-dev.5/data_stream/traces/elasticsearch/ingest_pipeline).

This PR updates `elastic-package` to support `type: traces` for the above use case. Specifically:
- it updates the check for valid data streams to include data streams of `type: traces`, and
- it allows running pipeline tests on data streams of `type: traces`

Resolves #223.

### DIFF
--- a/internal/packages/assets.go
+++ b/internal/packages/assets.go
@@ -120,7 +120,7 @@ func loadElasticsearchAssets(pkgRootPath string) ([]Asset, error) {
 		}
 		assets = append(assets, asset)
 
-		if dsManifest.Type == dataStreamTypeLogs {
+		if dsManifest.Type == dataStreamTypeLogs || dsManifest.Type == dataStreamTypeTraces {
 			elasticsearchDirPath := filepath.Join(filepath.Dir(dsManifestPath), "elasticsearch", "ingest_pipeline")
 			pipelineFiles, _ := ioutil.ReadDir(elasticsearchDirPath)
 			if pipelineFiles == nil || len(pipelineFiles) == 0 {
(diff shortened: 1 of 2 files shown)


jalvz:
thanks @ycombinator . what were the consequences of this missing? i've been testing this for a while and couldn't notice any issue with traces.

jalvz (approved):

ycombinator:
>  what were the consequences of this missing? i've been testing this for a while and couldn't notice any issue with traces.

At this point the consequences are only theoretical because the `apm` package does not contain any [pipeline tests](https://github.com/elastic/elastic-package/blob/master/docs/howto/pipeline_testing.md). If it did and `elastic-package test pipeline --data-stream traces` was run, it wouldn't actually run those tests because prior to this PR we only ran pipeline tests for data streams of type `logs`.

So in a practical sense, this PR is more forward looking. If/when the `apm` package defines pipeline tests for it's data streams' ingest pipelines, particularly ones for the `traces` data stream, the changes in this PR will allow those tests to run.

jalvz:
got it, thanks @ycombinator !

mtojek (approved):
One more comment to what @ycombinator said - to fully use all these features (including automated testing), it would be great to move the "apm" package to the Integrations repository.

PR #260, opened 2021-02-23
Updating package-spec dependency

ycombinator:
This PR updates the dependency on the `package-spec` to bring in this change: https://github.com/elastic/package-spec/pull/139.

Related: https://github.com/elastic/package-spec/issues/106


ycombinator:
CI is failing with this error:

```

[2021-02-23T00:59:28.131Z] ERROR: for elastic-agent_is_ready  Container "c1071590ccd8" is unhealthy.

[2021-02-23T00:59:28.131Z] Encountered errors while bringing up the project.

[2021-02-23T00:59:28.131Z] Error: booting up the stack failed: running docker-compose failed: running command failed: running Docker Compose up command failed: exit status 1

[2021-02-23T00:59:28.131Z] + cleanup

[2021-02-23T00:59:28.131Z] + r=1
```

I believe this is due to recent changes in `elastic-agent` behavior which are going to be fixed in a new Docker image shortly. So we will simply retry CI again in a day or so.

mtojek (approved):
The change looks good to me, but please wait for the green status.

mtojek:
Blocked by: https://github.com/elastic/beats/pull/24163

mtojek:
@ycombinator I think it failed due to some issues around `event.ingested`.

ycombinator:
@mtojek I ended up making some more code changes to this PR even after applying the changes from https://github.com/elastic/elastic-package/pull/263.  So I have re-requested your review.  Thanks!

mtojek (approved):
Just a loose thought:

Personally I think that this is an overkill here (switch from JSON to YAML), because eventually we'll end up with correcting these files mostly on our own :)

IMHO you could go without deprecation period, but of course it's clearer.

Apart from this, LGTM!

ycombinator:
> Personally I think that this is an overkill here (switch from JSON to YAML), because eventually we'll end up with correcting these files mostly on our own :)

> IMHO you could go without deprecation period, but of course it's clearer.

Yeah, I agree about this when it comes to packages in `integrations`. The deprecation period is more for packages that are being developed outside of the `integrations` repo that we don't really know about. I doubt any of them are using pipeline tests but, just in case they are, better to be safe!

PR #261, opened 2021-02-23
Bump up stack version to 7.13.0-SNAPSHOT

mtojek:
This PR bumps up the stack to 7.13.0-SNAPSHOT.

PR #263, opened 2021-02-24
Temporarily hardcode agent image reference

mtojek:
This PR will temporarily hardcode the Docker image reference for Elastic Agent to the stable image. Unfortunately this fix will cause that 7.12.0-SNAPSHOT will be the only working one, but most likely will unblock development.

Related: https://github.com/elastic/beats/issues/24198

### DIFF
--- a/internal/install/static_snapshot_yml.go
+++ b/internal/install/static_snapshot_yml.go
@@ -75,6 +75,7 @@ services:
         condition: service_healthy
 
   elastic-agent:
+    # FIXME Must remove this workaround once https://github.com/elastic/beats/issues/24198 is fixed
     image: docker.elastic.co/beats/elastic-agent@sha256:75c9fbf835a7d24166bceb82f28f4c6c5d00048ba9f0a25b945818e4cf2bba69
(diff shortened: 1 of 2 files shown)


internal/install/static_snapshot_yml.go
```diff
@@ -75,7 +75,7 @@ services:
         condition: service_healthy
 
   elastic-agent:
-    image: docker.elastic.co/beats/elastic-agent:${STACK_VERSION}
+    image: docker.elastic.co/beats/elastic-agent@sha256:75c9fbf835a7d24166bceb82f28f4c6c5d00048ba9f0a25b945818e4cf2bba69
```
ycombinator, line 78: Could you add a screaming comment here saying this is a temporary hack and should be removed once https://github.com/elastic/elastic-package/pull/261 goes green? Just in case we somehow forget about it. 😅

mtojek: Added :D

ycombinator (approved):
LGTM.

PR #262, opened 2021-02-23
Implement static test runner

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/226

This PR implements static test runner with its first responsibility - linting sample events.

PR #264, opened 2021-02-25
Support YAML configs for pipeline tests: Forgot a field

ycombinator:
Follow up to #260.


mtojek (approved):

mtojek:
Let me merge it, so I can include it in https://github.com/elastic/elastic-package/pull/262

PR #265, opened 2021-03-01
Fix: correctly filter AWS instance type

mtojek:
This PR modifies the test package "AWS" to prevent the following problem:

```
Error: Error launching source instance: UnsupportedOperation: Microsoft SQL Server Enterprise Edition is not supported for the instance type 't1.micro'.
	status code: 400, request id: 3cfb03a1-c3f5-44fe-9fbe-fa66c5ca1902

  on main.tf line 7, in resource "aws_instance" "i":
   7: resource "aws_instance" "i" {
```


mtojek:
self-approved, we need this to fix master branch.

PR #267, opened 2021-03-02
Update package-spec to support input groups

mtojek:
This PR updates dependency on package-spec, so it will support input groups.

Issue: https://github.com/elastic/package-spec/issues/132#issuecomment-788007017


ycombinator (approved):
LGTM.

PR #268, opened 2021-03-02
Fix compatibility check snapshot

mtojek:
This PR modifies the version compatibility check to properly handle SNAPSHOTs (just ignore them).

We need this change to enable testing against the Elastic stack v7.12.0-SNAPSHOT (which hasn't been released). So far we can depend only on release versions:

https://github.com/elastic/integrations/blob/master/.ci/Jenkinsfile#L17-L18

### DIFF
--- a/internal/packages/conditions_test.go
+++ b/internal/packages/conditions_test.go
@@ -1,0 +1,59 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package packages
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCheckConditions_InvalidRelease(t *testing.T) {
+	manifest := PackageManifest{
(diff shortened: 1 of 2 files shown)


internal/packages/conditions_test.go
```diff
@@ -0,0 +1,59 @@
...
+	manifest := PackageManifest{
+		Conditions: Conditions{
+			Kibana: KibanaConditions{Version: "^7.11.0"},
+		},
+		PolicyTemplates: nil,
+	}
+
+	err := CheckConditions(manifest, []string{"kibana.version=7.11.1-SNAPSHOT"})
+	assert.NoError(t, err)
+}
```
ycombinator, line 59: Ni: you could probably replace the individual test functions with a single [table-driven test function](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) but this is okay too!

ycombinator (approved):
LGTM.

internal/packages/conditions_test.go
mtojek: Yeah, maybe next time, 4 occurrences are still easy to touch ;)

PR #269, opened 2021-03-02
Accept strings that parse as numbers as numeric types

jsoriano:
`elastic-package test` fails if a document contains a string value with a number in a field with numeric type, such as:
```
[5] parsing field value failed: field "process.pid"'s Go type, string, does not match the expected field type: long (field value: 87)
```
But these values are accepted by Elasticsearch.

Modify the check to accept also strings that can be parsed as numbers.

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -10,14 +10,14 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
-
-	"github.com/elastic/elastic-package/internal/multierror"
 
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
 
 	"github.com/elastic/elastic-package/internal/common"
+	"github.com/elastic/elastic-package/internal/multierror"
 )
 
 // Validator is responsible for fields validation.
@@ -267,7 +267,13 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 			return fmt.Errorf("field %q's value, %s, does not match the expected pattern: %s", key, valStr, definition.Pattern)
 		}
 	case "float", "long", "double":
-		_, valid = val.(float64)
+		switch val := val.(type) {
+		case float64, int:
+			valid = true
+		case string:
+			_, err := strconv.ParseFloat(val, 64)
+			valid = err == nil
+		}
 	default:
 		valid = true // all other types are considered valid not blocking validation
 	}
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -271,6 +271,7 @@ func parseElementValue(key string, definition FieldDefinition, val interface{}) 
 		case float64, int:
 			valid = true
 		case string:
+			// Elasticsearch can ingest numbers represented as strings in the JSON document.
 			_, err := strconv.ParseFloat(val, 64)
 			valid = err == nil
 		}
--- a/internal/fields/validate_test.go
+++ b/internal/fields/validate_test.go
@@ -80,7 +80,6 @@ func Test_parseElementValue(t *testing.T) {
 			definition: FieldDefinition{
 				Type: "long",
 			},
-			fail: true,
 		},
 
 		// keyword and constant_keyword (string)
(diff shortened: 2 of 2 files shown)


mtojek:
We already have this feature, but the strict mode is enabled by default. If you prefer to enable conversion, please use test config (`numeric_keyword_fields`): https://github.com/elastic/integrations/blob/2b70a5437d506109aa9bc93516aebcb839999e01/packages/zoom/data_stream/webhook/_dev/test/pipeline/test-user.json-config.yml

jsoriano:
@mtojek oh ok, thanks!

ycombinator:
@mtojek Maybe we should add documentation for the test config options somewhere under https://github.com/elastic/elastic-package/tree/master/docs/howto?

mtojek:
It's a good idea to do. Frankly speaking I though we have already documented it somewhere, but it looks like it's missing.

ycombinator:
I'll make a PR for the docs.

jsoriano:
Umm, after doing some tests with `numeric_keyword_fields`, I have found that this is intended for the opposite of what I need :thinking: 

`numeric_keyword_fields` helps when the field is defined as `keyword`, but it is present as number in the document.

I need something for fields that are defined as numbers, but are present as strings in the documents, for example:
```
            "process": {
                "pid": "34"
            },
```
For `process.pid` defined as `long`, as is in ECS.

Should we add another option for this? Or should I reopen this PR? Or should I be doing something different? :slightly_smiling_face: 

For context, I am needing this to migrate the test files for PostgreSQL (https://github.com/elastic/integrations/pull/747).

ycombinator:
I would be in favor of reopening this PR. From what I can tell the code in this PR is only converting string values to numbers if the field is defined as a number. So there is enough strictness already IMO. In fact, now I'm wondering if we could apply the same logic for numeric values that are intended to be keywords and not need the `numeric_keyword_fields` setting too. 🤔 

The other option would be to introduce another setting, similar to `numeric_keyword_fields`, but I think it's not necessary given that the field definition already tells us that we want a field to be treated like a number.

@mtojek WDYT?

mtojek:
Ok, I see the conversion direction now, missed that earlier (sorry about that).

Actually I would be in favor of preserving strict checks, but I understand this is a behavior that is enabled by default (dynamic field mapping)? We can merge this PR, but then it won't be consistent with `numeric_keyword_fields` concept, so maybe @ycombinator's advise is correct and we should get rid of it.

Let's review this PR to unblock @jsoriano  and think what would be the solution for our cases.

/cc @andrewkroh as the Security Team uses `numeric_keyword_fields` in few places.

internal/fields/validate.go
```diff
@@ -267,7 +267,13 @@ func parseElementValue(key string, definition FieldDefinition, val interface{})
 			return fmt.Errorf("field %q's value, %s, does not match the expected pattern: %s", key, valStr, definition.Pattern)
 		}
 	case "float", "long", "double":
-		_, valid = val.(float64)
+		switch val := val.(type) {
+		case float64, int:
+			valid = true
+		case string:
```
ycombinator, line 273: Maybe add a comment here saying why this is okay (because ES is okay with it)?

ycombinator (approved):
Left a minor suggestion for adding a comment. Otherwise LGTM.

mtojek (approved):

ycombinator:
>> It's a good idea to do. Frankly speaking I though we have already documented it somewhere, but it looks like it's missing.

> I'll make a PR for the docs.

https://github.com/elastic/elastic-package/pull/270

andrewkroh:
I think the processing pipeline should be using a `convert` processor (or similar feature) to ensure that the JSON types match the Elasticsearch types. The idea being that if we declare a value as a number in our fields.yml then we ought be delivering the value as a number. Relying on type coercion at indexing could result in ingest failures and then event loss, but if you handle this earlier you at least have an option on how to handle the data type exception (like you could use `on_failure` to drop the offending value or allow `convert` to fail, but always remove the source value).

If we do choose to go away from the strict type checking for strings that become numbers then it makes no sense to keep the `numeric_keyword_fields` option. This option allows for a validation exception when the _source contains a number but the mapping is as a keyword. This type of conversion carries no risk of an ES mapping exception. So if we kept this type of check by default but allowed the more risking string to number coercion then we'd have an "imbalance"

mtojek:
> I think the processing pipeline should be using a convert processor (or similar feature) to ensure that the JSON types match the Elasticsearch types

I would prefer that option as well, but maybe I'm not familiar with historical reasons/customs/etc and it's natural to depend on dynamic mappings in Elasticsearch. @jsoriano did you consider just using `convert`? 

As I said before I prefer strictness in this area, but I'm ok to merge this PR and then deprecate the `numeric_keyword_fields` option. Most likely it's something that needs be planned.

jsoriano:
Sounds good to me. I didn't try to modify the pipeline, I just imported it from beats, but I agree that it would be better to make it more strict, I'll give a try to add explicit conversions.

jsoriano:
https://github.com/elastic/integrations/pull/747 updated to don't need this. Thanks for the suggestion!

jsoriano:
@mtojek @andrewkroh I am finding this issue on a metricset, I have tried to add an ingest pipeline and a processor in the stream, but none of them seemed to work. Should any of these work for a metrics data stream? or the only way is to fix this in Metricbeat?
```
postgresql/statement default:
[0] parsing field value failed: field "postgresql.statement.query.id"'s Go type, string, does not match the expected field type: long (field value: 1691311383)
```

andrewkroh:
I'm not aware of any reasons that a data stream for metrics would not allow an ingest pipeline to be used. So that's interesting and curious.

But I think Metricbeat should be sending a number for this value if that is its intent.

jsoriano:
Open PR for Metricbeat: https://github.com/elastic/beats/pull/24359

PR #272, opened 2021-03-03
Elastic Agent: use fleet server

mtojek:
This PR enables the Fleet server in the Elastic stack.

Unsolved issues:
- [ ] correctly handle another default policy (verify system tests)
- [ ] correct healthcheck that will mark that the Agent has a policy assigned
- [ ] assign the default policy (should be done by fleet server?)

Issue: https://github.com/elastic/beats/issues/24310

PR #273, opened 2021-03-03
Elastic Agent: set FLEET_URL and FLEET_INSECURE

mtojek:
This PR sets the FLEET_URL to Kibana host. 

Either this or https://github.com/elastic/elastic-package/pull/272 will eventually work.

Issue: https://github.com/elastic/beats/issues/24310

### DIFF
--- a/internal/install/static_snapshot_yml.go
+++ b/internal/install/static_snapshot_yml.go
@@ -75,8 +75,7 @@ services:
         condition: service_healthy
 
   elastic-agent:
-    # Temporary workaround for: https://github.com/elastic/beats/issues/24310
-    image: docker.elastic.co/beats/elastic-agent@sha256:6182d3ebb975965c4501b551dfed2ddc6b7f47c05187884c62fe6192f7df4625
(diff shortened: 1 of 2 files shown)


internal/install/static_snapshot_yml.go
```diff
@@ -91,6 +90,7 @@ services:
     - "FLEET_ENROLL=1"
     - "FLEET_ENROLL_INSECURE=1"
     - "FLEET_SETUP=1"
+    - "FLEET_URL=http://kibana:5601"
```
ruflin, line 93: @blakerouse Is this required for the case where fleet-server runs as part of the Elastic Agent?

blakerouse: With `FLEET_SERVER_ENABLE` set to true, `FLEET_URL` is ignored. `FLEET_URL` is required when `FLEET_SERVER_ENABLE` is not set.

ruflin (approved):
Tested with `elastic-package stack up --version=7.13.0-SNAPSHOT -v` and works as expected.

Issue #271, opened 2021-03-03
In case of emergency - select Docker image references in the Elastic stack


mtojek:
There was [another issue](https://github.com/elastic/beats/issues/24310) introduced to Elastic Agent, which caused problems with booting up the Elastic stack (unhealthy container with elastic-agent).

The goal of this issue is to figure out a workaround to quickly specify Docker image references.

mtojek:
Let's discuss multiple options:

1. Global environment variables

If these values are set, they will override default values for Docker Compose template.

```
ELASTIC_PACKAGE_STACK_ELASTICSEARCH_IMAGE=docker.elastic.co/elasticsearch/elasticsearch:7.13.0-SNAPSHOT
ELASTIC_PACKAGE_STACK_KIBANA_IMAGE=docker.elastic.co/kibana/kibana:7.13.0-SNAPSHOT
ELASTIC_PACKAGE_STACK_ELASTIC_AGENT_IMAGE=docker.elastic.co/beats/elastic-agent@sha256:75c9fbf835a7d24166bceb82f28f4c6c5d00048ba9f0a25b945818e4cf2bba69
```

(I suggested `image` instead of version because of format different between snapshot/release and sha256)

2. Local file with key-values.

A local config file passed as CLI option that contains additional settings. Something similar to https://github.com/elastic/elastic-package/issues/179 . I'm not sure if I'm convinced to this form - can't figure out the best place in the filesystem to keep it.

3. Additional CLI options

so it would be:

```
elastic-package stack up --elasticsearch-image-ref "docker.elastic.co/elasticsearch/elasticsearch:7.13.0-SNAPSHOT" --elastic-agent-image-ref "docker.elastic.co/beats/elastic-agent@sha256:75c9fbf835a7d24166bceb82f28f4c6c5d00048ba9f0a25b945818e4cf2bba69"
```

---

With options 1 and 3 we may control it without push&deployment of new revisions of elastic-package. Just have to open a PR to integrations to temporarily modify the Jenkinsfile. 

---

4. Tiny `versions` file kept in elastic-package's repository

The file will be available online (master branch) and the elastic-package will have to fetch it if not present locally or `elastic-package update` has been triggered. Later on this file will be used to fetch latest/safest references to images (either `7.13.0-SNAPSHOT` or `sha256@....`.

cc @ycombinator

ycombinator:
[...]

PR #274, opened 2021-03-04
Temporary switch to stable image of Elastic Agent

mtojek:
This PR is a quick mitigation to the problem of fault Elastic Agent image reported in https://github.com/elastic/beats/issues/24310.

Unfortunately it block older stacks from execution until https://github.com/elastic/elastic-package/issues/271 is fixed.


mtojek:
self-approved to unblock pipelines.

PR #275, opened 2021-03-04
Use configuration file to select Elastic stack version

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/271

This PR modifies the behavior of Elastic stack orchestrator (Docker Compose) and Kubernetes agent's deployer to consider selected versions.

TODO:

- [x] Select right image references for `docker-compose up`
- [x] Select right image references for `docker-compose pull`
- [x] Select right image references for Kubernetes agent deployer

### DIFF
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -58,9 +58,9 @@ func (ac *ApplicationConfiguration) DefaultStackImageRefs() ImageRefs {
 // StackImageRefs function selects the appropriate set of Docker image references for the given stack version.
 func (ac *ApplicationConfiguration) StackImageRefs(version string) ImageRefs {
 	refs := ac.c.Stack.ImageRefOverridesForVersion(version)
-	refs.ElasticAgent = stringOrDefault(refs.ElasticAgent, fmt.Sprintf("%s:%s", elasticAgentImageName, DefaultStackVersion))
-	refs.Elasticsearch = stringOrDefault(refs.Elasticsearch, fmt.Sprintf("%s:%s", elasticsearchImageName, DefaultStackVersion))
-	refs.Kibana = stringOrDefault(refs.Kibana, fmt.Sprintf("%s:%s", kibanaImageName, DefaultStackVersion))
+	refs.ElasticAgent = stringOrDefault(refs.ElasticAgent, fmt.Sprintf("%s:%s", elasticAgentImageName, version))
+	refs.Elasticsearch = stringOrDefault(refs.Elasticsearch, fmt.Sprintf("%s:%s", elasticsearchImageName, version))
+	refs.Kibana = stringOrDefault(refs.Kibana, fmt.Sprintf("%s:%s", kibanaImageName, version))
 	return refs
 }
 
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -4,9 +4,10 @@
 
 package install
 
+// ApplicationConfiguration represents the configuration of the elastic-package.
 type ApplicationConfiguration struct{}
 
-// StackImageRefs function selects the appropriate set of Docker image references for the default stack version.
+// DefaultStackImageRefs function selects the appropriate set of Docker image references for the default stack version.
 func (ac *ApplicationConfiguration) DefaultStackImageRefs() (*ImageRefs, error) {
 	return ac.StackImageRefs(DefaultStackVersion)
 }
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -4,25 +4,40 @@
 
 package install
 
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/elastic/go-ucfg"
+	"github.com/elastic/go-ucfg/yaml"
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/common"
+)
+
 // ApplicationConfiguration represents the configuration of the elastic-package.
-type ApplicationConfiguration struct{}
+type ApplicationConfiguration struct {
+	c common.MapStr
+}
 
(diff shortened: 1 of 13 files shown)


mtojek:
One note: I didn't write the override for 8.0.0-SNAPSHOT because simply I don't know which sha256 will work here :(

mtojek:
> One note: I didn't write the override for 8.0.0-SNAPSHOT because simply I don't know which sha256 will work here :(

Thanks @ruflin for pointing to the working snapshot for elastic-agent. I suppose that with this change we have a good workaround for emergency situations.

internal/install/application_configuration.go
```diff
@@ -0,0 +1,95 @@
...
+type ApplicationConfiguration struct {
+	c configFile
+}
+
+type configFile struct {
+	Stack stack `yaml:"stack"`
+}
+
+type stack struct {
+	ImageRefOverrides map[string]ImageRefs `yaml:"imageRefOverrides"`
```
ycombinator, line 26: Should we switch the yaml setting to snake case? Looking at other settings in package yaml files used for configuration (e.g. pipeline test case configuration), we use snake case.

internal/install/application_configuration_yml.go
```diff
@@ -0,0 +1,22 @@
...
+	elasticAgentImageName  = "docker.elastic.co/beats/elastic-agent"
+	elasticsearchImageName = "docker.elastic.co/elasticsearch/elasticsearch"
+	kibanaImageName        = "docker.elastic.co/kibana/kibana"
+)
+
+const applicationConfigurationYmlFile = "config.yml"
+
+const applicationConfigurationYml = `stack:
+  imageRefOverrides:
+    7.13.0-SNAPSHOT:
```
ycombinator, line 17: I understand what you are trying to do over here but I'm wondering if we need this level in the setting hierarchy.

I was thinking the config that is shipped with `elastic-package` would look something like this:

```yaml
stack:
  images:
    "elastic-agent": "docker.elastic.co/beats/elastic-agent:7.13.0-SNAPSHOT"
    elasticsearch: "docker.elastic.co/elasticsearch/elasticsearch:7.13.0-SNAPSHOT"
    kibana: "docker.elastic.co/kibana/kibana:7.13.0-SNAPSHOT"
```

These would replace the `DefaultStackVersion` hardcoded version we have in the code today. Instead, the code would read the config and get the image names (including versions) from there.

Individual images could then be overridden by repo-level or local-level configurations that have the same structure (as mentioned in [this comment](https://github.com/elastic/elastic-package/issues/271#issuecomment-789980898)).

mtojek: The reason why I implement it like this is because we need a place to fixtures for various stack version. Use cases:
* elastic-agent in 7.13.0 doesn't work with latest snapshot - we need to use a stable image
* kibana in 8.0.0-SNAPSHOT is broken, we know about this, we need to set a specific version

I'm afraid we need a support matrix somewhere (per version stack), not sure where should we keep it. I'm happy to discuss it further!

ycombinator: Okay, I see your point now, especially considering install tests again older versions.

I'm good with this structure but I'd keep the values in this file (the one supplied by `elastic-package`) clean — no specific SHAs, etc. IMO, overrides should be specified either at the repo level or the local development level, in similarly structured files as this one. `elastic-package` can discover these files by walking up the folder tree until it finds `~/.elastic-package/config.yml` and `~/.elastic-package/config.dev.yml`.

Nit: maybe we should remove the `DefaultVersion` const from the code and move it into this file, under the setting `stack.default_version` or similar? That way all information related to stack versions is in the same file and we only need to update this one file after every stack release.

mtojek: Let me challenge it, I believe we're close to the final form.

It's a common use case to start the start wherever you want with: `elastic-package stack up`, which should pick up latest snapshots. If there is a problem with latest SNAPSHOTS, without fixtures in `~/.elastic-package/config.yml` it will fail.

I think this is the only concern I have. It would be great if we can make `elastic-package stack up` working just anywhere.

internal/install/application_configuration.go
mtojek: Fixed!

internal/install/application_configuration_yml.go
ycombinator: Discussed this with @mtojek off-PR. We're on the same page now. Current model is good as-is. Reasons:
- It is the same amount of work for package authors to update `elastic-package` to get the override image IDs and then back to latest ones as it would be if we created a `.elastic-package/config.yml` in the `integrations` repo.
- Creating a `.elastic-package/config.yml` in the `integrations` repo would benefit developers who develop packages in that repo, but not other package developers using `elastic-package`.
- By providing the overrides via the `elastic-package` tool we also get an opportunity to get package developers to update their version of `elastic-package` from time to time.

ycombinator (approved):
LGTM. WFG.

internal/install/application_configuration_yml.go
```diff
@@ -0,0 +1,17 @@
...
+const (
+	elasticAgentImageName  = "docker.elastic.co/beats/elastic-agent"
+	elasticsearchImageName = "docker.elastic.co/elasticsearch/elasticsearch"
+	kibanaImageName        = "docker.elastic.co/kibana/kibana"
+)
+
+const applicationConfigurationYmlFile = "config.yml"
+
+const applicationConfigurationYml = `stack:
+  image_ref_overrides:
```
ycombinator, line 16: Maybe put some commented out structure here so it's easy to know how to change it when we need to make overrides? Also maybe include a comment about informing package developers whenever overrides are defined or removed here?

ycombinator (changes requested):
Oops, gave my LGTM a bit pre-maturely. I just made another comment.

ycombinator (approved):
LGTM.

PR #276, opened 2021-03-05
Disable the new default fleet-server

ruflin:
Fleet-server setup was enabled by default. This disables it for now until the new setup is fully working.


ruflin:
@mtojek How do we deal with different versions having a different setup? I don't expect this to happen too often but it will happen for 7.12 to 7.13 and I could foresee that in minors we introduce "optimisation" flags which we would also want to use in newer versions?

mtojek:
Ideally these flags should be ignored by older stacks, if this is insufficient we'll have to think about something more sophisticated.

mtojek:
BTW If you want to experiment with this PR, please revert changes around `sha256` for elastic-agent (to use latest snapshots).

ruflin:
Closing as this also fails because https://github.com/elastic/beats/pull/24319 is not part of the snapshot builds yet. Closing.

Issue #253, opened 2021-02-17
test runner: ctrl+c should tear down the service


mtojek:
It's a common use case to break test execution with ctrl+c. I would be great to perform a cleanup after that.

PR #277, opened 2021-03-05
Catch signal while running test runner

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/253

This PR modifies the behavior of test runner to catch the signal, so we can cleanup resources while `ctrl+c` is pressed.

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -367,16 +367,17 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 	}
 
 	// Assign policy to agent
-	logger.Debug("assigning package data stream to agent...")
-	if err := kib.AssignPolicyToAgent(agent, *policy); err != nil {
-		return result.WithError(errors.Wrap(err, "could not assign policy to agent"))
-	}
 	r.resetAgentPolicyHandler = func() error {
 		logger.Debug("reassigning original policy back to agent...")
 		if err := kib.AssignPolicyToAgent(agent, origPolicy); err != nil {
 			return errors.Wrap(err, "error reassigning original policy to agent")
 		}
 		return nil
+	}
+
+	logger.Debug("assigning package data stream to agent...")
+	if err := kib.AssignPolicyToAgent(agent, *policy); err != nil {
+		return result.WithError(errors.Wrap(err, "could not assign policy to agent"))
 	}
 
 	// Signal to the service that the agent is ready (policy is assigned).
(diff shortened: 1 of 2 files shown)


mtojek:
Hm.. I found another issue:

```
2021/03/05 15:05:18 ERROR can't tear down the test runner: error cleaning up test policy: could not delete policy; API status code = 500; response body = {"statusCode":500,"error":"Internal Server Error","message":"Cannot delete agent policy that is assigned to agent(s)"}
```

@ruflin  Should it be HTTP 500 or HTTP 4x?

internal/testrunner/runners/system/runner.go
```diff
@@ -379,6 +375,11 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 		return nil
 	}
 
+	logger.Debug("assigning package data stream to agent...")
+	if err := kib.AssignPolicyToAgent(agent, *policy); err != nil {
+		return result.WithError(errors.Wrap(err, "could not assign policy to agent"))
+	}
+
```
ycombinator, line 382: Just curious, why did this block need to be moved?

internal/testrunner/testrunner.go
```diff
@@ -253,6 +255,20 @@ func Run(testType TestType, options TestOptions) ([]TestResult, error) {
...
+		<-ch
+		logger.Info("Signal caught!")
+
+		err := tearDown()
+		if err != nil {
+			logger.Errorf("can't tear down the test runner: %v", err)
+		}
+		os.Exit(1)
+	}()
+
```
ycombinator, line 271: This definitely works so I don't want to block this PR.

But in case some day `elastic-package` gets more long-running sub-commands, we might want to consider moving this logic to a higher level (not sure where exactly yet) so that a SIGTERM or SIGINT could invoke tear down, if implemented, on that sub command.

BTW, we already have another long-running sub command: `elastic-package stack up` (without the `-d` flag). But it looks like SIGINT is already being handled by the underlying lib.

mtojek: I can improve on this in follow-up issues, but this particular one I wanted to focus on improving developer experience of a person writing system tests (and rerunning tests multiple times).

ycombinator: ++ to follow up PRs, as needed.

internal/testrunner/runners/system/runner.go
mtojek: Assigning a policy is a long operation (due to condition "wait for the policy to get assigned"). If you stop here, but in the mean time an agent gets assigned, you won't be able to delete policy. Just in case it's safer to reassign the original policy.

ycombinator (approved):
LGTM

PR #280, opened 2021-03-09
Use different API to check assigned policy revisions

mtojek:
This PR modifies test runner to use different API methods to check agents which picked up latest policy revisions (kuery has been removed).

This is a workaround for https://github.com/elastic/kibana/issues/94053 .

### DIFF
--- a/internal/kibana/agents.go
+++ b/internal/kibana/agents.go
@@ -62,50 +62,46 @@ func (c *Client) AssignPolicyToAgent(a Agent, p Policy) error {
 		return fmt.Errorf("could not assign policy to agent; API status code = %d; response body = %s", statusCode, string(respBody))
 	}
 
-	err = c.waitUntilPolicyAssigned(p)
+	err = c.waitUntilPolicyAssigned(a, p)
 	if err != nil {
 		return errors.Wrap(err, "error occurred while waiting for the policy to be assigned to all agents")
 	}
 	return nil
 }
 
-func (c *Client) waitUntilPolicyAssigned(p Policy) error {
+func (c *Client) waitUntilPolicyAssigned(a Agent, p Policy) error {
 	for {
-		agents, err := c.ListAgents()
+		agent, err := c.getAgent(a.ID)
 		if err != nil {
-			return errors.Wrap(err, "can't list available agents")
+			return errors.Wrap(err, "can't get the agent")
 		}
 
-		agentsWithPolicy := filterAgentsByPolicy(agents, p)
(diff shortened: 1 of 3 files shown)


internal/testrunner/runners/system/runner.go
```diff
@@ -554,6 +559,10 @@ func filterAgents(allAgents []kibana.Agent, ctx servicedeployer.ServiceContext)
 
 	var filtered []kibana.Agent
 	for _, agent := range allAgents {
+		if agent.PolicyRevision == 0 {
+			continue // For some reason Kibana doesn't always return a valid policy revision (eventually it will be present and valid)
```
ycombinator, line 563: Not blocking this PR but it would be good to understand why this is happening and try to remove this condition in a follow up PR. @nchaulet maybe you can shed some light here?

internal/kibana/agents.go
```diff
@@ -62,62 +62,46 @@ func (c *Client) AssignPolicyToAgent(a Agent, p Policy) error {
...
+		time.Sleep(2 * time.Second)
 	}
-	return resp.Total, nil
+	return nil
 }
 
-func (c *Client) waitUntilPolicyAssigned(p Policy) error {
-	totalAgents, err := c.getTotalAgentForPolicy(p)
+func (c *Client) getAgent(agentID string) (*Agent, error) {
+	statusCode, respBody, err := c.get(fmt.Sprintf("%s/agents/%s", FleetAPI, agentID))
```
ycombinator, line 91: Does this API exist in older versions of Kibana? IOW, just wondering if this change is backwards compatible.

mtojek: I think it was already there, we just used a different API.

internal/testrunner/runners/system/runner.go
mtojek: This is actually causing some flakiness for Kubernetes integration. I would be for leaving it here (to unblock our delivery pipeline), but it definitely needs an explanation.

internal/kibana/agents.go
ycombinator: I quickly tested this PR with 7.11.0 and it works. So we're good on backwards compatibility.

internal/testrunner/runners/system/runner.go
mtojek: Is it by design or it just works this way? I'm concerned if it makes sense to have an agent with policy assigned, but no revision. Shouldn't it always be 1 on startup?

PR #285, opened 2021-03-09
Remove reference to integrations folder

ycombinator:
This PR removes a reference to the `integrations` folder (resulting from checking out the `elastic/integrations` repository locally) from the README. This is because `elastic-package` should be usable regardless of whether packages are coming from the `elastic/integrations` repository or not.


mtojek (approved):

jalvz (commented):
lgtm, thanks!

Issue #284, opened 2021-03-09
Pipeline tests should support drop processors


P1llus:
When using drop processors in ingest pipelines the pipeline tests is unable to verify the output of `*-expected.json`

When drop processors hit, the returning json object is `null`, resulting -expected.json to look like this:

```json
{
	"expected": [{
			"network": {
				"community_id": "1:HQ1oJYZ+9SJOoeju7badiLfvwls=",
				"type": "ipv4",
				"bytes": 568,
				"iana_number": "6",
				"packets": 8
			}
		},
		null,
		{
			"network": {
				"community_id": "1:HQ1oJYZ+9SJOoeju7badiLfvwls=",
				"type": "ipv4",
				"bytes": 568,
				"iana_number": "6",
				"packets": 8
			}
		}
	]
}
```

Because of the `null` value, when running `elastic-package test pipeline --generate`, the following error will occur:
`verifying test result failed: comparing test results failed: can't adjust test results: can't unmarshal event: unexpected end of JSON input`

P1llus:
Similar with `elastic-package check`, maybe @mtojek would be able to provide some light if we want/can resolve it?

```
Format the package
Done
Lint the package
Error: checking package failed: linting package failed: found 3 validation errors:
   1. file "/home/marius/go/src/github.com/elastic/integrations/packages/aws/data_stream/vpcflow/_dev/test/pipeline/test-tcp-flag-sequence.log-expected.json" is invalid: field expected.0: Invalid type. Expected: object, given: null
   2. file "/home/marius/go/src/github.com/elastic/integrations/packages/aws/data_stream/vpcflow/_dev/test/pipeline/test-tcp-flag-sequence.log-expected.json" is invalid: field expected.2: Invalid type. Expected: object, given: null
   3. file "/home/marius/go/src/github.com/elastic/integrations/packages/aws/data_stream/vpcflow/_dev/test/pipeline/test-tcp-flag-sequence.log-expected.json" is invalid: field expected.4: Invalid type. Expected: object, given: null
```

mtojek:
Well, it's a officially supported processor, so we can't say no :) I suggest you to temporarily skip this problematic use case, but we'll have to fix it.

BTW Nice finding! :)

cc @ycombinator

mtojek:
I'm going with this one

P1llus:
[...]

PR #287, opened 2021-03-10
Pipeline tests: support dropped events

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/284

This PR modifies the pipeline test runner to support potentially "null" results of processing simulation.

It needs changes in: https://github.com/elastic/package-spec/pull/147


internal/testrunner/runners/pipeline/runner.go
```diff
@@ -248,6 +250,19 @@ func (r *runner) verifyResults(testCaseFile string, config *testConfig, result *
 	return nil
 }
 
+// stripEmptyTestResults function removes events which are empty strings. These strings can represent
+// documents processed by a pipeline which potentially used a "drop" processor (to drop the event at all).
+func stripEmptyTestResults(result *testResult) *testResult {
+	var tr testResult
+	for _, event := range result.events {
+		if len(event) == 0 {
```
ycombinator, line 258: I think we are expecting `event` to be `nil` for empty test results. If so, should we make this check stricter by explicitly checking for `nil`? Otherwise theoretically `[]` would also match (although I know practically we would not see such an event).

```suggestion
		if event == nil {
```

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -248,6 +250,19 @@ func (r *runner) verifyResults(testCaseFile string, config *testConfig, result *
 	return nil
 }
 
+// stripEmptyTestResults function removes events which are empty strings. These strings can represent
+// documents processed by a pipeline which potentially used a "drop" processor (to drop the event at all).
+func stripEmptyTestResults(result *testResult) *testResult {
```
ycombinator, line 255: Maybe add a unit test for this function?

mtojek: Added

internal/testrunner/runners/pipeline/runner.go
mtojek: Fixed

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -248,6 +250,19 @@ func (r *runner) verifyResults(testCaseFile string, config *testConfig, result *
 	return nil
 }
 
+// stripEmptyTestResults function removes events which are empty strings. These strings can represent
```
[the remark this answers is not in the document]
mtojek: Nils, fixed

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -248,6 +250,19 @@ func (r *runner) verifyResults(testCaseFile string, config *testConfig, result *
 	return nil
 }
 
+// stripEmptyTestResults function removes events which are nils. These strings can represent
```
ycombinator, line 253:
```suggestion
// stripEmptyTestResults function removes events which are nils. These nils can represent
```

ycombinator (approved):
LGTM.

PR #288, opened 2021-03-11
Write sample event from system tests if --generate is specified

### LINKED ISSUE #150: Automatically create sample_event.json
The `sample_event.json` file contains a representative sample of documents stored in data stream.

The goal of this task is to research and design an automatic approach to collect sample events.

ycombinator:
Resolves #150 

With this PR, running `elastic-package test system --generate` (note the `--generate` flag) will generate `sample_event.json` files for each of the data streams of the package. The files will be written in the respective data stream folders.

### Example usage
```sh
$ pwd
/Users/shaunak/development/github/elastic-package/test/packages/apache/data_stream/access
```
```sh
$ ls -l
total 8
drwxr-xr-x  3 shaunak  staff   96 Mar  2 15:47 _dev
drwxr-xr-x  3 shaunak  staff   96 Feb  1 08:06 agent
drwxr-xr-x  3 shaunak  staff   96 Feb  1 08:06 elasticsearch
drwxr-xr-x  5 shaunak  staff  160 Feb  1 08:06 fields
-rw-r--r--  1 shaunak  staff  480 Feb  1 08:06 manifest.yml
```
```sh
$ elastic-package test system --data-streams access  --generate
--- Test results for package: apache - START ---
╭─────────┬─────────────┬───────────┬───────────┬────────┬───────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME │ RESULT │  TIME ELAPSED │
├─────────┼─────────────┼───────────┼───────────┼────────┼───────────────┤
│ apache  │ access      │ system    │ default   │ PASS   │ 15.622799344s │
╰─────────┴─────────────┴───────────┴───────────┴────────┴───────────────╯
--- Test results for package: apache - END   ---
Done
```
```sh
$ ls -l
total 16
drwxr-xr-x  3 shaunak  staff    96 Mar  2 15:47 _dev
drwxr-xr-x  3 shaunak  staff    96 Feb  1 08:06 agent
drwxr-xr-x  3 shaunak  staff    96 Feb  1 08:06 elasticsearch
drwxr-xr-x  5 shaunak  staff   160 Feb  1 08:06 fields
-rw-r--r--  1 shaunak  staff   480 Feb  1 08:06 manifest.yml
-rw-r--r--  1 shaunak  staff  2012 Mar 10 16:55 sample_event.json
```


internal/testrunner/runners/system/runner.go
```diff
@@ -220,6 +221,16 @@ func (r *runner) hasNumDocs(
 				multiErr = append(multiErr, errs...)
 				continue
 			}
+
+			// Generate sample event from first doc
```
mtojek, line 225: I do love simplicity of this solution :)

I'm only concerned about the fact that it's in `hasNumDocs`, which shouldn't modify something externally. Keep in mind that the function is actually called twice (on clear and on check). Will the file be written twice?

I'm wondering if we can replace `hasNumDocs` with `areDocsAvailable`, and define a set of callbacks for:
- minDocsNumFunc
- verifyDocumentFunc(data Document) error
- onFirstDocumentFunc(data Document) error - which can generate sample docs.

I'm happy to discuss it as it's actually a refactoring task.

ycombinator: Ah, I didn't even realize the name of the function in which I added this code. I also didn't realize this function was being called twice. Thanks for pointing out these two things.

I completely agree that we need to refactor this code. Let me take a stab at it and have you review it.

ycombinator: I did some refactoring in 3b0c4c3. Let me know what you think.

ycombinator: I did some refactoring in 3b0c4c3.  Broadly speaking, the changes I made were:
- renamed `hasNumDocs` to `checkNumDocs`
- `checkNumDocs` now returns the number of documents + the documents themselves + any error
- don't perform field validation when `checkNumDocs` is called for clearing out old docs in the data stream
- don't write `sample_event.json` when `checkNumDocs` is called for clearing out old docs in the data stream

Let me know what you think.

ycombinator: Hold on, I thought of a way to simplify things further.

ycombinator: Okay, some more refactoring done in 33163be.

mtojek (approved):
Another useful feature, good job :)

PR #295, opened 2021-03-23
Check if user remote exists when promoting

jsoriano:
`promote` needs both the elastic and the user repositories to exist. If the user repository doesn't exist, it fails when trying to push:
```
Error: pushing changes failed: pushing branch failed: repository not found
```
It is not clear for the user what repository is missing.

With these changes, the existence of the user remote is checked when the remote is configured. If it doesn't exist, it fails immediately, including the missing repository in the error message:
```
Error: cloning source repository failed (branch: snapshot): checking user remote (jsoriano, url: https://github.com/jsoriano/package-storage): repository not found
```

Code is refactored to avoid fetching anything before doing this check, and to fetch only the latest commit.

### DIFF
--- a/internal/promote/storage.go
+++ b/internal/promote/storage.go
@@ -98,37 +98,72 @@ func (prs PackageVersions) Strings() []string {
 
 // CloneRepository method clones the repository and changes branch to stage.
 func CloneRepository(user, stage string) (*git.Repository, error) {
-	r, err := git.Clone(memory.NewStorage(), memfs.New(), &git.CloneOptions{
-		URL:           fmt.Sprintf(repositoryURL, "elastic"),
-		RemoteName:    remoteName,
-		ReferenceName: plumbing.NewBranchReferenceName(stage),
-	})
+	// Initialize repository
+	r, err := git.Init(memory.NewStorage(), memfs.New())
 	if err != nil {
-		return nil, errors.Wrap(err, "cloning package-storage repository failed")
+		return nil, errors.Wrap(err, "initializing repository")
 	}
 
-	err = r.Fetch(&git.FetchOptions{
-		RemoteName: remoteName,
+	// Add remotes
+	userRepositoryURL := fmt.Sprintf(repositoryURL, user)
+	userRemote, err := r.CreateRemote(&config.RemoteConfig{
+		Name: user,
+		URLs: []string{
+			userRepositoryURL,
+		},
+	})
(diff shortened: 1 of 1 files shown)


internal/promote/storage.go
```diff
@@ -98,37 +98,72 @@ func (prs PackageVersions) Strings() []string {
...
+		Name: user,
+		URLs: []string{
+			userRepositoryURL,
+		},
 	})
 	if err != nil {
-		return nil, errors.Wrap(err, "cloning package-storage repository failed")
+		return nil, errors.Wrap(err, "creating user remote failed")
+	}
+	remote, err := r.CreateRemote(&config.RemoteConfig{
```
ycombinator, line 118: Maybe call this `upstreamRemote` so it parallel's `userRemote`? In the same vein maybe we should rename `remoteName` to `upstream`?

internal/promote/storage.go
```diff
@@ -98,37 +98,72 @@ func (prs PackageVersions) Strings() []string {
...
+		Name: remoteName,
+		URLs: []string{
+			fmt.Sprintf(repositoryURL, "elastic"),
```
ycombinator, line 121: Maybe use `remoteName` instead of `"elastic"` here? I know they are not the same thing but they are loosely related.

internal/promote/storage.go
```diff
@@ -98,37 +98,72 @@ func (prs PackageVersions) Strings() []string {
...
+		},
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "creating remote failed")
 	}
 
-	err = r.Fetch(&git.FetchOptions{
-		RemoteName: remoteName,
+	// Check if user remote exists
```
ycombinator, line 128: Along the same lines as this, do you think it would be good to check if the upstream remote exists before the fetch and checkout step below?

internal/promote/storage.go
jsoriano: Renamed.

internal/promote/storage.go
jsoriano: Using `upstream` now.

internal/promote/storage.go
jsoriano: Upstream is going to be indirectly checked when doing the fetch, I think this would be enough.

ycombinator (approved):
LGTM

Issue #244, opened 2021-02-05
Fail on empty data streams cmdline option


mtojek:
The following command should fail as there are no data streams selected using the `--data-streams` switch:

```
➜  kubernetes git:(239-add-kubernetes-as-test-package) ✗ elastic-package test system --data-streams -v
Run system tests for the package
--- Test results for package: kubernetes - START ---
No test results
--- Test results for package: kubernetes - END   ---
Done
```

PR #296, opened 2021-03-24
Validate data streams flag

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/244

### DIFF
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -222,7 +222,7 @@ func validateDataStreamsFlag(packageRootPath string, dataStreams []string) error
 		path := filepath.Join(packageRootPath, "data_stream", dataStream)
 		fileInfo, err := os.Stat(path)
 		if err != nil {
-			return errors.Wrap(err, "stat directory failed")
+			return errors.Wrapf(err, "stat directory failed (path: %s)", path)
 		}
 
 		if !fileInfo.IsDir() {
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -6,6 +6,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cmd/testrunner.go
```diff
@@ -210,3 +216,18 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 		return nil
 	}
 }
+
+func validateDataStreamsFlag(packageRootPath string, dataStreams []string) error {
+	for _, dataStream := range dataStreams {
+		path := filepath.Join(packageRootPath, "data_stream", dataStream)
+		fileInfo, err := os.Stat(path)
+		if err != nil {
+			return errors.Wrap(err, "stat directory failed")
```
ycombinator, line 225: Maybe include the `path` in this error?

cmd/testrunner.go
```diff
@@ -210,3 +216,18 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 		return nil
 	}
 }
+
+func validateDataStreamsFlag(packageRootPath string, dataStreams []string) error {
+	for _, dataStream := range dataStreams {
```
ycombinator, line 221: In #244 the issue was that no data streams were specified even though the `--data-streams` flag was specified. If I'm following this code right, that issue will remain even with this code because this `for` loop would be skipped completely and a `nil` will be returned from this function.

mtojek: Actually the root cause were somewhere else. CLI can complain about missing argument for the `data-streams` flag:

```elastic-package test pipeline  -v --data-streams
Error: flag needs an argument: --data-streams```

but it can't complain in this situation:

```
elastic-package test pipeline  --data-streams -v
```

(-v is considered to be a data stream)

cmd/testrunner.go
mtojek: Fixed

ycombinator (approved):
LGTM.

PR #292, opened 2021-03-16
Implement "publish" command

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/21

This PR  implements "publish" command, which will replace "mage UpdatePackageStorage". It guarantees feature parity with https://github.com/elastic/integrations/tree/master/dev/update-package-storage

This improvement will introduce a better logging of performed actions and will allow for executing publish independently for every package (in parallel).

 

TODOs:

- [x] Read package context (especially manifest)
- [x] Check if the package has been built
- [x] Refactor "promote" to reuse as much as possible
- [x] Checkout package-storage
- [x] Find greatest package revision released (revision + stage)
- [x] Copy greatest package to staging as new revision
- [x] Branch out from snapshot
- [x] Commit the above
- [x] Copy over the built package
- [x] Commit the above
- [x] Push to remote Git branch
- [x] Open PR
- [x] Fix: check if package is already released

Sample PR:
https://github.com/elastic/package-storage/pull/1054

Testing:

Unfortunately this time we need to perform some testing:
1. Publish new package to the package storage (select any from `test/packages`).
2. Publish updated package to the package storage (e.g. `nginx-999.999.999`).
3. Try to publish package while the package has been already published (a candidate in integrations?).
4. Try to publish package while the PR is open (e.g. `nginx-999.999.999`).
5. Promote packages between stages, as refactoring impacts that feature. 

Sample output:

```
elastic-package publish -v
elastic-package has been installed.
2021/03/23 14:09:37 DEBUG Enable verbose logging
Publish the package
Current GitHub user: mtojek
Use build directory: /Users/marcin.tojek/go/src/github.com/elastic/elastic-package/build/integrations/nginx/999.999.999
Clone package-storage repository
Find latest package revision of "nginx" in package-storage
2021/03/23 14:09:55 DEBUG Change stage to production
2021/03/23 14:10:04 DEBUG Find revisions of the "nginx" package
2021/03/23 14:10:04 DEBUG Found package revision: nginx-0.2.3
2021/03/23 14:10:04 DEBUG Found package revision: nginx-0.2.4
2021/03/23 14:10:04 DEBUG Found package revision: nginx-0.3.6
2021/03/23 14:10:04 DEBUG Found package revision: nginx-0.3.7
2021/03/23 14:10:04 DEBUG Found package revision: nginx-0.3.9
2021/03/23 14:10:04 DEBUG Change stage to staging
2021/03/23 14:10:13 DEBUG Find revisions of the "nginx" package
[...]

PR #297, opened 2021-03-24
Refactor promote command in favor of publish

mtojek:
This PR refactors "promote" command in favor of future "publish" command, which will reuse the storage part.

Extracted from: https://github.com/elastic/elastic-package/pull/292 to make the PR reviewable.

### DIFF
--- a/internal/cobraext/const.go
+++ b/internal/cobraext/const.go
@@ -36,9 +36,6 @@ const (
 	ReportOutputFlagName        = "report-output"
 	ReportOutputFlagDescription = "output location for test report"
 
-	SkipPullRequestFlagName        = "skip-pull-request"
-	SkipPullRequestFlagDescription = "skip opening a new pull request"
-
 	StackServicesFlagName        = "services"
 	StackServicesFlagDescription = "component services (comma-separated values: \"%s\")"
 
--- a/internal/cobraext/const.go
+++ b/internal/cobraext/const.go
@@ -36,8 +36,8 @@ const (
 	ReportOutputFlagName        = "report-output"
 	ReportOutputFlagDescription = "output location for test report"
 
-	VerboseFlagName        = "verbose"
-	VerboseFlagDescription = "verbose mode"
+	SkipPullRequestFlagName        = "skip-pull-request"
+	SkipPullRequestFlagDescription = "skip opening a new pull request"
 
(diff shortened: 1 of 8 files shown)


internal/cobraext/const.go
```diff
@@ -36,8 +36,8 @@ const (
 	ReportOutputFlagName        = "report-output"
 	ReportOutputFlagDescription = "output location for test report"
 
-	VerboseFlagName        = "verbose"
-	VerboseFlagDescription = "verbose mode"
+	SkipPullRequestFlagName        = "skip-pull-request"
+	SkipPullRequestFlagDescription = "skip opening a new pull request"
```
ycombinator, line 40: Where are these two used? I think maybe these are not supposed to be in this PR?

mtojek: You're right, didn't clean it properly.

internal/storage/storage.go
```diff
@@ -39,16 +42,44 @@ type PackageVersion struct {
 	Name    string
 	Version string
 
+	root   string
```
ycombinator, line 45: I didn't understand why this is needed for promotion.

mtojek: Same reason, missed during extraction. Fixed.

ycombinator (approved):
Refactoring LGTM. Thanks for extracting it out into its own PR — made it a lot easier to review.

We should get these changes into the `integrations` repo and let a few promotions happen with them before merging the second PR. WDYT?

mtojek:
> We should get these changes into the integrations repo and let a few promotions happen with them before merging the second PR. WDYT?

Actually this part is not actively used by elastic/integrations. It's more about running `elastic-package promote`, but it means that users have to download and use the latest version. I think in this situration we can take this risk and push the `publish` command. I don't think we'll get an immediate feedback.

mtojek:
The CI job go stuck for some reason. I will merge this one.

ycombinator:
> Actually this part is not actively used by elastic/integrations. It's more about running `elastic-package promote`, but it means that users have to download and use the latest version. I think in this situration we can take this risk and push the `publish` command. I don't think we'll get an immediate feedback.

Ah yes, I forgot, `elastic-package promote` is a manual step. I agree, we can take the risk.

Issue #179, opened 2020-11-20
Let me change container ports and other settings


fearful-symmetry:
So, right now I use a remote docker host (_remote_ meaning a server in my closet) for all my for all my development, since docker4mac is still terrible. This works great, but it means that any container that exposes some service needs to expose it _publicly_:

```
     ports:
       - "0.0.0.0:5601:5601"
```

`elastic-package` doesn't seem to like me doing that, as it keeps the config in `~/.elastic-package` and it'll blow away any changes I make at will. I don't really feel like mainting my own `.yml` either. Perhaps we can have `elastic-package` maybe read some port override from an environment variable or custom dotfile?

mtojek:
@ycombinator we need to think about a convenient API to cover this suggestion.

ycombinator:
How about two new commands:

`elastic-package stack config show`
set <setting name> <setting value>`.

ycombinator:
We could use environment variables but I think that could get a bit unwieldy if the number of user-settable settings grows beyond a handful. 

So instead, borrowing from `kubectl`, how about introducing the following new sub-commands.

### API

* `elastic-package stack config show` or `elastic-package stack config view`. This command will print out the list of settings that can be configured and their current values.
* `elastic-package stack config set <setting name> <setting value>`. This command will set a single setting, thus overriding its default value.
* `elastic-package stack config unset <setting name>`. This command will unset a single setting.

### Example Usage

* `elastic-package stack config set kibana.port.public 6701`: sets the public (exposed) port to 6701.

### Implementation

* There is a new internal file somewhere under `~/.elastic-package/stack`, say `~/.elastic-package/stack/settings.yml`. It contains any stack-related settings that can be set (i.e. overridden) by the user. By default the file is empty, so settings that are "hardcoded" are used instead.
* Running `elastic-package stack config set <setting name> <setting value>` sets the given setting in `~/.elastic-package/stack/settings.yml`.
* Running `elastic-package stack config unset <setting name>` removes the given setting from `~/.elastic-package/stack/settings.yml`.
* Running `elastic-package stack config show` shows _all_ possible user-settable settings and their current values. This means starting with the "hardcoded" default values for each setting, then overlaying any values for that setting from  `~/.elastic-package/stack/settings.yml`, then displaying the results.
* When `elastic-package stack up` is run, the contents of `~/.elastic-package/stack/settings.yml` are overlaid on the "hardcoded" default values for settings, and the results are used when booting up the stack.

### Question

[...]

Issue #231, opened 2021-01-22
Replace ioutil usage


ycombinator:
With Go 1.16, the `ioutil` package is being deprecated: https://www.srcbeat.com/2021/01/golang-ioutil-deprecated/. We should start replacing its usage.

mistikel:
I would like to take it

PR #301, opened 2021-03-26
Add profile system to elastic-package stack

fearful-symmetry:
This PR addresses https://github.com/elastic/elastic-package/issues/179 and adds an `elastic stack profiles` command to manage multiple configs used by `docker-compose`

This can be used to manage and select config profiles that correspond do to the config files under `~/.elastic-package/stack/`

```
elastic-package profiles new my_profile #create a new profile from default

elastic-package profiles new --from my_profile other_profile #create a new profile from an existing profile

elastic-package profiles list

elastic-package profiles delete other_profile

export ELASTIC_PACKAGE_PROFILE=my_profile # set a default profile to use on startup

elastic-package stack -p my_profile up # or specify a profile in the command
```

This is a follow-up to my earlier PR that added the `locations` package. This should be the "final" PR to add a profile system. It's still a tad large, but at this point I don't know if it can be paired down much further, as the profile system has its fingers in a lot of the docker-compose code.

There's still a few bits of design I'd like input on. 

Do we want to keep the command structure as-is? This only applies to the logic of `stack` so I made it a sub-command, but I guess we do could do `elastic-package profiles ...`

Also, do we want to lasso the `deployer` configs into this? This is limited to the `stack` config used by `docker-compose`, but there's no reason we can't expand it. Also, never used terraform and I don't have much experience with K8s in general, so I'm not sure if we want to touch those configs. I tried to make this fairly modular, so it shouldn't be too hard to expand the files managed by profiles.

There's also the issue of `application_config`, which I sort of left in. I'm not sure if we just want to flat-out remove the code that creates `~/.elastic-package/config.yml`?


ycombinator:
There's a lot to digest in this feature (which is great!) so I'll start by focussing first on the interface and the overall design before diving into the implementation.

> Do we want to keep the command structure as-is? This only applies to the logic of `stack` so I made it a sub-command, but I guess we do could do `elastic-package profiles` ...

I'd be okay with `elastic-package profiles`. If we ever need profiles for another command of `elastic-package` besides `stack`, that would give us the flexibility. If we go with `elastic-package stack profiles` now, I can see how that might box us into a corner and it might be awkward to get out of it later.

> Also, do we want to lasso the `deployer` configs into this? This is limited to the `stack` config used by `docker-compose`, but there's no reason we can't expand it. Also, never used terraform and I don't have much experience with K8s in general, so I'm not sure if we want to touch those configs. I tried to make this fairly modular, so it shouldn't be too hard to expand the files managed by profiles.

I'm good with keeping this out of scope for now. As you said, as long as the design is fairly modular we can expand it as needed later.

> There's also the issue of `application_config`, which I sort of left in. I'm not sure if we just want to flat-out remove the code that creates `~/.elastic-package/config.yml`?

My understanding (@mtojek can correct me if I'm wrong) is that the application configs are the ones set by `elastic-package` itself at build time. I can imagine a user being able to override these locally via profiles (this PR's feature). WDYT of that?

fearful-symmetry:
> is that the application configs are the ones set by elastic-package itself at build time. 

@ycombinator I'm mostly looking at the code here: https://github.com/elastic/elastic-package/pull/275 which was basically a shim while we dealt with a bunch of elastic-agent bugs.

fearful-symmetry:
Hmm, okay, `os.Read*` is new to 1.16 since ioutil is getting deprecated. CI is using 15.8 and `beats` is 15.9. Not sure how to proceed here...

fearful-symmetry:
Okay, so adding `.go-version` doesn't seem to work. Not sure if we want to revert to the soon-to-be-depricated ioutil functions or what.

ycombinator:
@fearful-symmetry Don't bother removing the deprecated `ioutil` functions in this PR for now. We have an issue to track its removal _en masse_ throughout this codebase whenever we're ready to move to Go 1.16: https://github.com/elastic/elastic-package/issues/231. So we'll just sweep up any usages introduced in this PR at that time.

mtojek (commented):
My biggest concern with this PR is that it's huge! I think we need to split it intro reviewable pieces. Per feature? What do you think about the following order:

1. List profiles and create new profile for elastic-package stack
2. Create new profile from existing one.
3. Delete profiles.
4. Boot up the stack using selected profile.

So we would end up with 4 PRs.

fearful-symmetry:
@mtojek that might work? A huge chunk of this code is basally reorganizing (see `internal/locations`) to get everything to play nice, so we're still gonna end up with a fairly large first PR and a bunch of tiny PRs after that. The two new libraries `internal/profile` and `internal/location` sort of need to happen together, and they're the biggest chunk of the whole thing.

mtojek:
Well, 30 files is definitely too much for a single PR and will probably slow us down with merging this in. If you are able to split this codebase more into features, so that we have 7-8 files per PR, then it will make reading this easier. 

> The two new libraries internal/profile and internal/location sort of need to happen together

I see your point, but we need to cut this into reviewable pieces. I don't mind if two libraries need to go together, but for me the most important part is value added to the codebase (new feature). Thus, I suggest to split it.

fearful-symmetry:
> I see your point, but we need to cut this into reviewable pieces. 

I definitely agree, I'm just worried about merging something half-broken. 

I think the best way to do this would be like this:

1. `internal/locations`
2. `profile/profile`
3. Everything else.

That _should_ work, as a good part of the PR is just the renames associated with creating the `locations` library. `profile` needs `locations` but `locations` should be able to stand on its own. How does that sound @mtojek ?

mtojek:
Ok, let's try such approach. I personally prefer to open pull requests which add a real value for the end user, not just a library or simple refactoring, but you're the driver here :) I'm looking forward to reviewing the first PR.

fearful-symmetry:
> Ok, let's try such approach. I personally prefer to open pull requests which add a real value for the end user, not just a library or simple refactoring, but you're the driver here 

Agreed, but I think it's the only way to break the whole feature addition down into manageable chunks, as 6-7 of the files are just changes associated with `locations`. @mtojek

ycombinator:
I'd like to see this split into multiple PRs as well, even if it means the first PR or two are mostly refactoring / prep for later PRs that actually implement the profiles feature.

Alternatively, we also have the option of creating a feature branch and targeting the PRs to it initially if we don't want to land half-baked code into `master` directly.

ycombinator:
@fearful-symmetry @mtojek Before we start breaking this up into smaller PRs and reviewing them, it would be good to settle on the overall design and interface for this feature.  @fearful-symmetry raised some questions in https://github.com/elastic/elastic-package/pull/301#issue-601838072 and I added some comments in https://github.com/elastic/elastic-package/pull/301#issuecomment-808448281. 

Do either/both of you have anything to add? It would be good to reach some consensus on these points before we start reviewing smaller PRs. Otherwise we might merge a couple of initial PRs and then have to change course later, which wouldn't be a great use of time IMO.

mtojek:
Let me respond to raised question too:

> Do we want to keep the command structure as-is? This only applies to the logic of stack so I made it a sub-command, but I guess we do could do elastic-package profiles ...

This is a great question. We may leverage from keeping profiles independent from the stack only in the future. I would be for keeping it as a separate feature.

> Also, do we want to lasso the deployer configs into this? This is limited to the stack config used by docker-compose, but there's no reason we can't expand it. Also, never used terraform and I don't have much experience with K8s in general, so I'm not sure if we want to touch those configs. I tried to make this fairly modular, so it shouldn't be too hard to expand the files managed by profiles.

I agree with Shaunak, let's not introduce additional complexity here. We can iterate on this later on, even leave it for further improvement if it becomes a problem (missing feature).

> There's also the issue of application_config, which I sort of left in. I'm not sure if we just want to flat-out remove the code that creates ~/.elastic-package/config.yml?

This piece of code is a safety switch to be used in case of problems with stable SNAPSHOTs. If we know that the SNAPSHOT is faulty, we can mark a particular revision as the recommended one. Unfortunately it's required due to many drawbacks of current unified building process (one piece can block the entire snapshot).

fearful-symmetry:
@mtojek / @ycombinator I'm reviving this PR, since it's been merged against master, and we just have the latest round of changes here. Apologies for the delay, been a bit backlogged with 7.13 work.

ycombinator:
Hey @fearful-symmetry, thanks for reviving this PR!

I started reviewing it and noticed that we still need to address the CLI design discussion from https://github.com/elastic/elastic-package/pull/301#issuecomment-808448281. Specifically, the current implementation has the CLI syntax as `elastic-package stack profiles ...`, where `...` could be `new`, `delete`, or `list`. We were discussing whether this should instead be `elastic-package profiles ...` to give us the opportunity in the future to use profiles for features of `elastic-package` beyond just `elastic-package stack`.

I'm curious what you and @mtojek think of this change.

fearful-symmetry:
@ycombinator yah, that's something I was contemplating too. My thinking was that since `profiles` is currently limited to the behavior in `stack`, we should keep it there, and then just move/export it if we expand it to other things.

ycombinator:
> My thinking was that since `profiles` is currently limited to the behavior in `stack`, we should keep it there, and then just move/export it if we expand it to other things.

If we change it later, that will change the user interface. Users who are used to running `elastic-package stack profiles` will need to know to use `elastic-package profiles` instead. Personally I'd prefer to start with `elastic-package profiles ...` so we have room to add non-stack profile settings later without affecting the user interface.  

@mtojek do you have a preference in either direction?

fearful-symmetry:
> Personally I'd prefer to start with elastic-package profiles ... so we have room to add non-stack profile settings later without affecting the user interface.

Yah, I think you might be right there.

ycombinator:
@mtojek do you have any opinions on the discussion in the last few comments, starting from https://github.com/elastic/elastic-package/pull/301#issuecomment-825496716? Just want to make sure we are not blocking @fearful-symmetry here.

fearful-symmetry:
Alright @mtojek / @ycombinator I've moved the profile command. You're right--it probably makes more sense here.

ycombinator:
Hey @fearful-symmetry, looks like there's a merge conflict. Once you resolve it I'll proceed with the code review. Thanks!

ycombinator:
> I've moved the profile command.

Could you update the PR description to reflect the changes, please? Thanks!

fearful-symmetry:
@ycombinator alright, everything should be updated now!

README.md
```diff
@@ -122,6 +122,12 @@ Use this command to validate the contents of a package using the package specifi
 
 The command ensures that the package is aligned with the package spec and the README file is up-to-date with its template (if present).
 
+### `elastic-package profiles`
+
+_Context: global_
+
+
+
```
ycombinator, line 130: This section is generated from code now. Please see the code for some of the other commands in the `cmd` package and then implement something similar for the `profiles` command.

internal/profile/profile_json.go
```diff
@@ -0,0 +1,56 @@
...
+	User        string
+	Version     string `json:"elastic_package_version"`
+	From        string
+}
+
+// PackageProfileMetaFile is the filename of the profile metadata file
+const PackageProfileMetaFile ConfigFile = "profile.json"
+
+// CreateProfileMetadata creates the body of the profile.json file
+func CreateProfileMetadata(profileName string, profilePath string) (*SimpleFile, error) {
```
mtojek, line 31: WDYT about renaming this file to `metadata.go`?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+		return errors.Wrapf(err, "stat file failed (path: %s)", profile.ProfilePath)
+	}
+
+	//write the resources
+	return profile.writeProfileResources()
+
+}
+
+// updateExistingDefaultprofile migrates the old default profile to profile_old
+func updateExistingDefaultprofile(path string) error {
```
mtojek, line 62: nit: DefaultProfile (typo)

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+
+// ErrNotAProfile is returned in cases where we don't have a valid profile directory
+var ErrNotAProfile = errors.New("Is is not a profile")
+
+// ConfigFile is a type for for the config file names in a managed profile config
+type ConfigFile string
+
+// ManagedProfileFiles is the list of all files managed in a profile
+// If you create a new file that's managed by a profile, it needs to go in this list
+var ManagedProfileFiles = map[ConfigFile]NewConfig{
```
mtojek, line 31: Why is this var exposed? Do you plan to use it externally?

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+	KibanaConfigFile:              NewKibanaConfig,
+	PackageRegistryDockerfileFile: NewPackageRegistryDockerfile,
+	PackageRegistryConfigFile:     NewPackageRegistryConfig,
+	SnapshotFile:                  NewSnapshotFile,
+	PackageProfileMetaFile:        CreateProfileMetadata,
+	KibanaHealthCheckFile:         NewKibanaHealthCheck,
+}
+
+// ConfigProfile manages a stack config
+type ConfigProfile struct {
```
mtojek, line 41: Can we rename it to just `Profile`?

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+
+}
+
+// Fetch returns an absolute path to the given file
+func (profile ConfigProfile) Fetch(file ConfigFile) string {
+	return profile.configFiles[file].FilePath
+}
+
+// UpdateFileBodies updates the string contents of the config files
+func (profile *ConfigProfile) updateFileBodies(newBody map[ConfigFile]*SimpleFile) {
```
mtojek, line 123: nit: `overwrite` instead of `updateFileBodies`?

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+		profileName: profileName,
+		ProfilePath: profilePath,
+		configFiles: configMap,
+	}
+	return newProfile, nil
+}
+
+// LoadProfile loads an existing profile
+func LoadProfile(elasticPackagePath string, profileName string) (*ConfigProfile, error) {
+
```
mtojek, line 72: nit: empty line

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+		err := cfgFiles.WriteConfig()
+		if err != nil {
+			return errors.Wrap(err, "error writing config file")
+		}
+	}
+	return nil
+}
+
+// metadata returns the metadata struct for the profile
+func (profile ConfigProfile) metadata() (Metadata, error) {
```
mtojek, line 221: WDYT about returning nil instead of plain Metadata in case of an error?

internal/profile/static_package_registry_dockerfile.go
```diff
@@ -0,0 +1,39 @@
...
+// PackageRegistryDockerfileFile is the dockerfile for the Elastic package registry
+const PackageRegistryDockerfileFile ConfigFile = "Dockerfile.package-registry"
+
+const packageRegistryDockerfile = `FROM ` + PackageRegistryBaseImage + `
+
+COPY ${PROFILE_NAME}/package-registry.config.yml /package-registry/config.yml
+COPY development/ /packages/development
+`
+
+type packageRegistryCfg struct {
```
mtojek, line 24: Is it used?

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+			updatedProfile := profile.configFiles[key]
+			updatedProfile.FileBody = toReplace.FileBody
+			profile.configFiles[key] = updatedProfile
+		}
+	}
+
+}
+
+// ProfileAlreadyExists checks to see if a profile with this name already exists
+func (profile ConfigProfile) profileAlreadyExists() (bool, error) {
```
mtojek, line 141: nit: AlreadyExists (profile is redundant)

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error reading in profile %s", profileName)
+	}
+
+	return profile, nil
+
+}
+
+// Fetch returns an absolute path to the given file
+func (profile ConfigProfile) Fetch(file ConfigFile) string {
```
mtojek, line 118: I would rename it to `ConfigPath` or sth similar. Fetch is too ambiguous.

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+		configFiles: configMap,
+	}
+
+	exists, err := profile.profileAlreadyExists()
+	if err != nil {
+		return nil, errors.Wrapf(err, "error checking if profile %s exists", profileName)
+	}
+
+	if !exists {
+		return nil, fmt.Errorf("Profile %s does not exist", profile.ProfilePath)
```
mtojek, line 105: nit: profile (lowercase)

internal/stack/shellinit.go
```diff
@@ -37,18 +36,9 @@ type kibanaConfiguration struct {
...
-	}
-
+func ShellInit(elasticStackProfile *profile.ConfigProfile) (string, error) {
 	// Read Elasticsearch username and password from Kibana configuration file.
-	kibanaConfigurationPath := filepath.Join(locationManager.StackDir(), "kibana.config.yml")
-	body, err := ioutil.ReadFile(kibanaConfigurationPath)
-	if err != nil {
-		return "", errors.Wrap(err, "reading Kibana configuration file failed")
-	}
+	body, err := ioutil.ReadFile(elasticStackProfile.Fetch(profile.KibanaConfigFile))
```
mtojek, line 41: err check?

cmd/profiles.go
```diff
@@ -0,0 +1,126 @@
...
+	env := os.Getenv(cobraext.ProfileNameEnvVar)
+	if env == "" {
+		return profile.DefaultProfile
+	}
+	return env
+
+}
+
+func availableProfilesAsAList() ([]string, error) {
+
```
mtojek, line 110: nit: empty line

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	//print later, after we've run into any errors, to avoid some ugly-half printing if something fails halfway through
+	header := []string{"Name", "Date Created", "User", "elastic-package version", "Path"}
+	for _, headerVal := range header {
+		fmt.Printf("%-30s ", headerVal)
+	}
+	fmt.Printf("\n")
+	for _, iter := range profiles {
+		fmt.Printf("%-30s %-30s %-30s %-30s %-30s\n", iter.Name, iter.DateCreated.Format(time.RFC3339), iter.User, iter.Version, iter.From)
```
mtojek, line 178: Could you please move all the printing to the `cmd` level and use internal just to read/return available profiles?

internal/profile/files.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// NewConfig is a generic function type to return a new Managed config
+type NewConfig = func(profileName string, profilePath string) (*SimpleFile, error)
+
+// SimpleFile defines a file that's managed by the profile system
+// and doesn't require any  rendering
+type SimpleFile struct {
```
mtojek, line 18: Does it have to be exposed externally?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		// If there's changes and we've selected CreateNew, move the old path
+		// TODO: do we want this to pe appended with some kind of version string instead?
+		if localChanges && !overwriteExisting {
+			if localChanges && profileName == DefaultProfile {
+				fmt.Printf("WARNING: default profile has been changed by user or updated by elastic-package. The current profile will be moved to default_old.\n")
+			}
+			updateExistingDefaultprofile(elasticPackagePath)
+			os.Mkdir(profile.ProfilePath, 0755)
```
mtojek, line 47: err check?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+	} else {
+		os.Mkdir(profile.ProfilePath, 0755)
```
mtojek, line 50: err check?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+func DeleteProfileFromDefaultLocation(profileName string) error {
+	loc, err := locations.NewLocationManager()
+	if err != nil {
+		return errors.Wrap(err, "error finding stack dir location")
+	}
+	return DeleteProfile(loc.StackDir(), profileName)
+}
+
+// DeleteProfile deletes a given config profile.
+func DeleteProfile(elasticPackagePath string, profileName string) error {
```
mtojek, line 142: Why are `DeleteProfileFromDefaultLocation` and `DeleteProfile` both exposed?

mtojek (commented):
I did the initial round of reviews. Let's see what can be improved first.

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+	newProfile := &ConfigProfile{
+		profileName: profileName,
+		ProfilePath: profilePath,
+		configFiles: configMap,
+	}
+	return newProfile, nil
+}
+
+// LoadProfile loads an existing profile
+func LoadProfile(elasticPackagePath string, profileName string) (*ConfigProfile, error) {
```
ycombinator, line 71: I think this function could be unexported?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+	loc, err := locations.NewLocationManager()
+	if err != nil {
+		return errors.Wrap(err, "error finding stack dir location")
+	}
+
+	return CreateProfileFrom(loc.StackDir(), profileName, from)
+}
+
+// CreateProfileFrom creates a new profile by copying over an existing profile
+func CreateProfileFrom(elasticPackagePath string, newProfileName string, fromProfileName string) error {
```
ycombinator, line 95: I think this function could be unexported?

internal/profile/install_profile.go
ycombinator: I think `DeleteProfile` could be unexported?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+func PrintProfilesFromDefaultLocation() error {
+	loc, err := locations.NewLocationManager()
+	if err != nil {
+		return errors.Wrap(err, "error finding stack dir location")
+	}
+	return PrintProfiles(loc.StackDir())
+}
+
+// PrintProfiles lists known profiles
+func PrintProfiles(elasticPackagePath string) error {
```
ycombinator, line 164: I think this function could be unexported?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+		localChanges, err := profile.localFilesChanged()
+		if err != nil {
+			return errors.Wrapf(err, "error checking for changes in %s", profile.ProfilePath)
+		}
+
+		// If there's changes and we've selected CreateNew, move the old path
+		// TODO: do we want this to pe appended with some kind of version string instead?
+		if localChanges && !overwriteExisting {
+			if localChanges && profileName == DefaultProfile {
+				fmt.Printf("WARNING: default profile has been changed by user or updated by elastic-package. The current profile will be moved to default_old.\n")
```
ycombinator, line 44: Since we're using `WARNING` here this looks like a good candidate for a logging statement. See https://github.com/elastic/elastic-package/blob/cb436d9f36d9c0b675184217df0c643f46064125/internal/logger/logger.go#L52-L55

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/configuration/locations"
+)
+
+// CreateProfile installs a new profile at the given package path location.
+// overwriteExisting determines the behavior if a profile with the given name already exists.
+// On true, it'll overwrite the profile, on false, it'll copy the existing profile over to profileName_old
+func CreateProfile(elasticPackagePath string, profileName string, overwriteExisting bool) error {
```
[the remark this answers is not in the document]
ycombinator: Maybe I missed something but I don't see this function being called from anywhere except test code.

internal/install/install.go
```diff
@@ -99,16 +110,8 @@ func writeStackResources(elasticPackagePath *locations.LocationManager) error {
 		return errors.Wrapf(err, "creating directory failed (path: %s)", elasticPackagePath.PackagesDir())
 	}
 
-	err = writeStaticResource(err, filepath.Join(elasticPackagePath.StackDir(), "kibana.config.yml"), kibanaConfigYml)
-	err = writeStaticResource(err, filepath.Join(elasticPackagePath.StackDir(), "healthcheck.sh"), kibanaHealthcheckSh)
-	err = writeStaticResource(err, filepath.Join(elasticPackagePath.StackDir(), "snapshot.yml"), snapshotYml)
-	err = writeStaticResource(err, filepath.Join(elasticPackagePath.StackDir(), "package-registry.config.yml"), packageRegistryConfigYml)
-	err = writeStaticResource(err, filepath.Join(elasticPackagePath.StackDir(), "Dockerfile.package-registry"), packageRegistryDockerfile)
+	return profile.CreateProfile(elasticPackagePath.StackDir(), profile.DefaultProfile, false)
```
ycombinator, line 113: Since we're no longer keeping profiles specific to `elastic-package stack` commands, does it make sense now to store profile files under something other than the stack dir? Maybe a new dir that's a sibling of the stack dir, e.g. `profiles` or something?

internal/profile/profile.go
```diff
@@ -0,0 +1,265 @@
...
+var ManagedProfileFiles = map[ConfigFile]NewConfig{
+	KibanaConfigFile:              NewKibanaConfig,
+	PackageRegistryDockerfileFile: NewPackageRegistryDockerfile,
+	PackageRegistryConfigFile:     NewPackageRegistryConfig,
+	SnapshotFile:                  NewSnapshotFile,
+	PackageProfileMetaFile:        CreateProfileMetadata,
+	KibanaHealthCheckFile:         NewKibanaHealthCheck,
+}
+
+// ConfigProfile manages a stack config
```
ycombinator, line 40: Shouldn't say stack config anymore as we've decided that profiles could be used for configuration beyond just stack-related configuration.

internal/profile/profile_json.go
```diff
@@ -0,0 +1,56 @@
...
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/version"
+)
+
+// Metadata stores the data associated with a given profile
+type Metadata struct {
```
ycombinator, line 19: I could be wrong but I think this type can be unexported? AFAICT it's not being used from anywhere outside the `profile` package.

ycombinator: Scratch that, looks like it's needed in `cmd/profiles.go`.

internal/profile/install_profile.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	var profiles []Metadata
+	// TODO: this should read a profile.json file or something like that
+	for _, item := range dirList {
+		if !item.IsDir() {
+			continue
+		}
+		profile, err := LoadProfile(elasticPackagePath, item.Name())
+		if err == ErrNotAProfile {
```
ycombinator, line 198: Consider using [`errors.Is`](https://blog.golang.org/go1.13-errors#TOC_3.2.) here. It'll be more future-proof, in case `err` here ever decides to wrap `ErrNotAProfile`.

internal/cobraext/const.go
```diff
@@ -33,6 +33,17 @@ const (
 	GenerateTestResultFlagName        = "generate"
 	GenerateTestResultFlagDescription = "generate test result file"
 
+	ProfileFlagName        = "profile"
+	ProfileFlagDescription = "select a profile to use for the stack configuration. Can also be set with " + ProfileNameEnvVar
+
+	ProfileNameEnvVar = "ELASTIC_PACKAGE_PROFILE"
+
+	ProfileNewFlagName        = "new"
+	ProfileNewFlagDescription = "create a new profile"
```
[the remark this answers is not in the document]
fearful-symmetry: Those two look like they never got used. Good catch.

internal/compose/compose.go
```diff
@@ -124,6 +124,7 @@ type CommandOptions struct {
 
 // NewProject creates a new Docker Compose project given a sequence of Docker Compose configuration files.
 func NewProject(name string, paths ...string) (*Project, error) {
+	// TODO: a lot of the checks in NewProject don't need to happen any more, we might want to rethink how we do this.
```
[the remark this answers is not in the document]
fearful-symmetry: +1

internal/profile/install_profile.go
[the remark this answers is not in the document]
fearful-symmetry: @ycombinator this is used in `install`:

`return profile.CreateProfile(elasticPackagePath.StackDir(), profile.DefaultProfile, false)`

internal/profile/profile_json.go
fearful-symmetry: I'm trying to keep some kind of naming scheme with the files that define individual files, i.e, `static_kibana_config_yml.go` is `kibana.config.yml`.

internal/profile/profile.go
fearful-symmetry: We'd need to return `*Metadata`, so I'm not sure it makes that much of a difference, considering we don't really need to return a pointer for any particular reason, unless I'm missing something.

internal/profile/static_package_registry_dockerfile.go
fearful-symmetry: Nope!

internal/install/install.go
fearful-symmetry: We could. My thinking was that the less we moved stuff around, the less weird paths we'll need to send to docker, since we already have stuff like `../../tmp/service_logs/`, which I'm not a fan of as-is.

internal/profile/profile.go
fearful-symmetry: Yah, good point. There's probably a few of those hanging around.

internal/profile/install_profile.go
fearful-symmetry: Ah, didn't even know that existed!

fearful-symmetry:
Thanks for all the feedback, I ended up shuffling a lot of the `Create` commands around so we don't have a dozen different exported methods everywhere.

internal/profile/install_profile.go
```diff
@@ -0,0 +1,194 @@
...
+		localChanges, err := profile.localFilesChanged()
+		if err != nil {
+			return errors.Wrapf(err, "error checking for changes in %s", profile.ProfilePath)
+		}
+
+		// If there's changes and we've selected CreateNew, move the old path
+		// TODO: do we want this to pe appended with some kind of version string instead?
+		if localChanges && !options.OverwriteExisting {
+			if localChanges && options.Name == DefaultProfile {
+				logger.Warn("WARNING: default profile has been changed by user or updated by elastic-package. The current profile will be moved to default_old.")
```
ycombinator, line 61: IMO, we can drop the `WARNING` prefix as the `logger.Warn` method will already print `WARN`.
```suggestion
				logger.Warn("default profile has been changed by user or updated by elastic-package. The current profile will be moved to default_old.")
```

internal/install/version_file.go
```diff
@@ -9,22 +9,27 @@ import (
...
+	versionFile, err := ioutil.ReadFile(versionPath)
 	if os.IsExist(err) {
 		return false, nil // old version, no version file
 	}
 	if err != nil {
 		return false, errors.Wrap(err, "reading version file failed")
 	}
 	v := string(versionFile)
+	if version.CommitHash == "undefined" && strings.Contains(v, "undefined") {
+		fmt.Printf("WARNING: CommitHash is undefined, in both %s and the compiled binary, config may be out of date, and elastic-package was not properly built.\n", versionPath)
```
ycombinator, line 31: Let's use `logger.Warnf` here, instead of `fmt.Printf`.

Also, I wouldn't say the last bit about `elastic-package` not being built properly. In our README we allow users to install `elastic-package` using `go get`, and with this method the commit hash is undefined but the package is still built properly otherwise.

cmd/stack.go
```diff
@@ -66,12 +67,30 @@ func setupStackCommand() *cobraext.Command {
 				return cobraext.FlagParsingError(err, cobraext.StackVersionFlagName)
 			}
 
+			profileName, err := cmd.Flags().GetString(cobraext.ProfileFlagName)
+			if err != nil {
+				return cobraext.FlagParsingError(err, cobraext.ProfileFlagName)
+			}
+
+			usrProfile, err := profile.LoadProfileFromDefaultLocation(profileName)
+			if err == profile.ErrNotAProfile {
```
ycombinator, line 76: Please use `errors.Is`.

cmd/stack.go
```diff
@@ -92,7 +111,27 @@ func setupStackCommand() *cobraext.Command {
...
 			cmd.Println("Take down the Elastic stack")
 
-			err := stack.TearDown()
+			profileName, err := cmd.Flags().GetString(cobraext.ProfileFlagName)
+			if err != nil {
+				return cobraext.FlagParsingError(err, cobraext.ProfileFlagName)
+			}
+
+			usrProfile, err := profile.LoadProfileFromDefaultLocation(profileName)
+			if err == profile.ErrNotAProfile {
```
ycombinator, line 120: Please use `errors.Is`.

internal/stack/compose.go
```diff
@@ -59,19 +46,16 @@ func dockerComposePull(options Options) error {
 		Services: withIsReadyServices(withDependentServices(options.Services)),
 	}
 
+	fmt.Printf("Docker-compose opts: %#v\n", opts)
```
ycombinator, line 49: Was this for debugging or did you intend to leave this in? If it's the latter, maybe we should use `logger.Debugf` instead?

internal/profile/files.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// NewConfig is a generic function type to return a new Managed config
+type NewConfig = func(profileName string, profilePath string) (*simpleFile, error)
+
+// simpleFile defines a file that's managed by the profile system
+// and doesn't require any  rendering
+type simpleFile struct {
```
mtojek, line 18: if the `simpleFile` is unexported, I assume that all its methods don't have to be exported ?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	os.Mkdir(newProfile.ProfilePath, 0755)
+
+	newProfile.overwrite(fromProfile.configFiles)
+	return newProfile.writeProfileResources()
+
+}
+
+// LoadProfileFromDefaultLocation loads an existing profile from the default elastic-package config dir
+func LoadProfileFromDefaultLocation(profileName string) (*Profile, error) {
```
mtojek, line 137: Can we drop the suffix `...FromDefaultLocation`? There is no other option available, so it's a bit confusing.

internal/profile/install_profile.go
```diff
@@ -0,0 +1,194 @@
...
+	loc, err := locations.NewLocationManager()
+	if err != nil {
+		return nil, errors.Wrap(err, "error finding stack dir location")
+	}
+
+	return loadProfile(loc.StackDir(), profileName)
+}
+
+// DeleteProfileFromDefaultLocation deletes a profile from the default elastic-package config dir
+func DeleteProfileFromDefaultLocation(profileName string) error {
```
mtojek, line 147: same here

internal/profile/profile.go
```diff
@@ -0,0 +1,254 @@
...
+	KibanaConfigFile:              newKibanaConfig,
+	PackageRegistryDockerfileFile: newPackageRegistryDockerfile,
+	PackageRegistryConfigFile:     newPackageRegistryConfig,
+	SnapshotFile:                  newSnapshotFile,
+	PackageProfileMetaFile:        createProfileMetadata,
+	KibanaHealthCheckFile:         newKibanaHealthCheck,
+}
+
+// Profile manages a a given user config profile
+type Profile struct {
```
mtojek, line 41: let's put exported properties at the top

internal/profile/profile.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+		if changes {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// readProfileResources reads the associated files into the config, as opposed to writing them out.
+func (profile *Profile) readProfileResources() error {
```
mtojek, line 191: Does it have to be ptr reference?

cmd/profiles.go
```diff
@@ -0,0 +1,139 @@
...
+				return errors.Wrap(err, "error listing all profiles")
+			}
+
+			header := []string{"Name", "Date Created", "User", "elastic-package version", "Path"}
+			for _, headerVal := range header {
+				fmt.Printf("%-30s ", headerVal)
+			}
+			fmt.Printf("\n")
+			for _, iter := range profileList {
+				fmt.Printf("%-30s %-30s %-30s %-30s %-30s\n", iter.Name, iter.DateCreated.Format(time.RFC3339), iter.User, iter.Version, iter.From)
```
mtojek, line 102: nit: we already have included a dedicated library for rendering tables. Maybe you can adopt it here for consistency (up to you).

cmd/profiles.go
```diff
@@ -0,0 +1,139 @@
...
+			loc, err := locations.NewLocationManager()
+			if err != nil {
+				return errors.Wrap(err, "error fetching profile")
+			}
+			profileList, err := profile.FetchAllProfiles(loc.StackDir())
+			if err != nil {
+				return errors.Wrap(err, "error listing all profiles")
+			}
+
+			header := []string{"Name", "Date Created", "User", "elastic-package version", "Path"}
```
mtojek, line 96: Why isn't it just `version`?

cmd/profiles.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	profileCommand.AddCommand(profileNewCommand, profileDeleteCommand, profileListCommand)
+
+	return cobraext.NewCommand(profileCommand, cobraext.ContextGlobal)
+
+}
+
+func lookupEnv() string {
+	env := os.Getenv(cobraext.ProfileNameEnvVar)
```
mtojek, line 115: I would be for keeping this env var here (in `cmd`) and leave a string placeholder in `cobraext` (`%s`). Let's not mix these places if not necessary.

cmd/root.go
```diff
@@ -29,6 +29,7 @@ var commands = []*cobraext.Command{
 	setupTestCommand(),
 	setupUninstallCommand(),
 	setupVersionCommand(),
+	setupProfilesCommand(),
```
mtojek, line 32: nit: place sort these functions

internal/install/version_file.go
fearful-symmetry: Yah, agreed. Gonna remove that part.

cmd/profiles.go
fearful-symmetry: Alright, lets do that, and try and emulate the `elastic-package status` look.

mtojek:
Thanks, Alex. I will do the second round of review.

internal/profile/install_profile.go
```diff
@@ -0,0 +1,194 @@
...
+		return errors.Wrap(err, "error checking for existing profile")
+	}
+	if exists {
+		localChanges, err := profile.localFilesChanged()
+		if err != nil {
+			return errors.Wrapf(err, "error checking for changes in %s", profile.ProfilePath)
+		}
+
+		// If there's changes and we've selected CreateNew, move the old path
+		// TODO: do we want this to pe appended with some kind of version string instead?
```
mtojek, line 58: nit: pe appended

internal/profile/install_profile.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	newExists, err := newProfile.alreadyExists()
+	if err != nil {
+		return errors.Wrapf(err, "error checking profile %s", newProfile.ProfilePath)
+	}
+	if newExists {
+		return errors.Errorf("profile %s already exists", newProfile.profileName)
+	}
+	os.Mkdir(newProfile.ProfilePath, 0755)
```
mtojek, line 129: swollen error?

internal/profile/profile.go
```diff
@@ -0,0 +1,253 @@
...
+const (
+	// DefaultProfile is the name of the default profile
+	DefaultProfile = "default"
+)
+
+// ErrNotAProfile is returned in cases where we don't have a valid profile directory
+var ErrNotAProfile = errors.New("Is is not a profile")
+
+// ConfigFile is a type for for the config file names in a managed profile config
+type ConfigFile string
```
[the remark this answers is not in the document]
fearful-symmetry: `ConfigFile` is used by a handful of publicly exposed methods, so I figure it should stay exported as well.

internal/profile/install_profile.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// write the resources
+	err = profile.writeProfileResources()
+	if err != nil {
+		return errors.Wrap(err, "error writing profile file")
+	}
+	return nil
+}
+
+// updateExistingDefaultProfile migrates the old default profile to profile_old
```
mtojek, line 138: nit: this comment is outdated now

internal/profile/config_profile_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"testing"
+)
+
+const (
+	profileName = "test_profile"
+)
+
+func TestNewProfile(t *testing.T) {
+
+	elasticPackageDir, err := ioutil.TempDir("", "package")
```
mtojek, line 18: Does it need cleanup once the test is done?

internal/profile/install_profile.go
```diff
@@ -0,0 +1,202 @@
...
+	err = os.Rename(profile.ProfilePath, newFilePath)
+	if err != nil {
+		return errors.Wrap(err, "error moving default profile")
+	}
+
+	return nil
+}
+
+// createProfileFrom creates a new profile by copying over an existing profile
+func createProfileFrom(option Options) error {
```
mtojek, line 167: I have feeling that this function is similar to `createProfile`. Is it possible to unify it somehow?

internal/profile/profile.go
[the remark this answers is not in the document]
mtojek: Hm.. I tried to rename it to `configFile` and it went smoothly?

```
	modified:   internal/profile/profile.go
	modified:   internal/profile/profile_json.go
	modified:   internal/profile/static_kibana_config_yml.go
	modified:   internal/profile/static_kibana_healthcheck_sh.go
	modified:   internal/profile/static_package_registry_config_yml.go
	modified:   internal/profile/static_package_registry_dockerfile.go
	modified:   internal/profile/static_snapshot_yml.go
```

Which external functions use it? I see the usage only in `internal/profile`.

internal/profile/install_profile.go
```diff
@@ -0,0 +1,202 @@
...
+			return profiles, errors.Wrap(err, "error reading profile metadata")
+		}
+		profiles = append(profiles, metadata)
+	}
+	return profiles, nil
+}
+
+// createProfile installs a new profile at the given package path location.
+// overwriteExisting determines the behavior if a profile with the given name already exists.
+// On true, it'll overwrite the profile, on false, it'll copy the existing profile over to profileName_old
```
mtojek, line 88: this comment is a bit misleading. Maybe it could be simplified that elastic-package will backup the profile if it exists.

mtojek (commented):
I think we're close to the end of the review. Just a couple of nit-picks on my side.

mtojek:
One more thing - did you check the following use case:

The developer uses the old app, upgrades to "profiles", starts the application. Will there be a mess in `~/.elastic-package` or is it clean up?

internal/profile/install_profile.go
fearful-symmetry: Yah, good idea.

fearful-symmetry:
@mtojek my bad -- because the exported filenames of type `ConfigFile` outside of `profile` get used elsewhere, I assumed the linter or compiler would be angry if the type itself wasn't exported.

As far as "cleanup" on upgrade, that was something I was thinking about, but I wasn't sure how or if we wanted to do it.

fearful-symmetry:
Alright, so I ended up doing a more conservative migration, where it'll try to package up the existing files, and move them to a new profile.

internal/profile/profile_json.go
```diff
@@ -0,0 +1,55 @@
...
```
mtojek, line 19: I looked in the `profile.json` file:

```
{
  "Name": "default",
  "date_created": "2021-05-14T11:03:48.44578+02:00",
  "User": "marcin.tojek",
  "elastic_package_version": "8050175",
  "From": "/Users/marcin.tojek/.elastic-package/stack/default"
```

Would you mind adjusting all fields to be lower cased (consistency)?

internal/profile/profile_json.go
```diff
@@ -0,0 +1,55 @@
...
+	Name        string
+	DateCreated time.Time `json:"date_created"`
+	User        string
+	Version     string `json:"elastic_package_version"`
+	From        string
```
mtojek, line 24: The `From` field is presented in the CLI table as `PATH`. Do you think we can make it consistent here?

internal/profile/profile_json.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/elastic/elastic-package/internal/version"
+)
+
+// Metadata stores the data associated with a given profile
+type Metadata struct {
+	Name        string
+	DateCreated time.Time `json:"date_created"`
+	User        string
+	Version     string `json:"elastic_package_version"`
```
mtojek, line 23: The `elastic_package_version ` field is presented in the CLI table as `VERSION`. Do you think we can make it consistent here (just `version`)?

internal/profile/static_snapshot_yml.go
```diff
@@ -124,3 +132,15 @@ services:
       elastic-agent:
         condition: service_healthy
 `
+
+// newSnapshotFile returns a Managed Config
+func newSnapshotFile(profileName string, profilePath string) (*simpleFile, error) {
+
```
mtojek, line 138: nit: empty line

cmd/profiles.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	profileCommand.AddCommand(profileNewCommand, profileDeleteCommand, profileListCommand)
+
+	return cobraext.NewCommand(profileCommand, cobraext.ContextGlobal)
+
+}
+
+func profileToList(profiles []profile.Metadata) [][]string {
+
```
mtojek, line 135: nit: empty line

internal/install/install.go
```diff
@@ -76,11 +87,60 @@ func checkIfAlreadyInstalled(elasticPackagePath *locations.LocationManager) (boo
...
+		if err != nil {
+			return errors.Wrapf(err, "error removing config file %s", file)
+		}
+	}
+	return nil
 }
 
 func createElasticPackageDirectory(elasticPackagePath *locations.LocationManager) error {
-	err := os.RemoveAll(elasticPackagePath.RootDir()) // remove in case of potential upgrade
+
```
mtojek, line 136: nit: empty line

internal/profile/config_profile_test.go
```diff
@@ -0,0 +1,85 @@
...
+		OverwriteExisting: false,
+	}
+	err = createProfile(options)
+	assert.NoErrorf(t, err, "error creating profile %s", err)
+
+	cleanupProfile(t, elasticPackageDir)
+}
+
+func TestNewProfileFrom(t *testing.T) {
+
```
mtojek, line 37: nit: empty line

internal/profile/profile.go
```diff
@@ -0,0 +1,297 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+// Profile manages a a given user config profile
+type Profile struct {
+	profileName string
+	// ProfilePath is the absolute path to the profile
+	ProfilePath string
```
mtojek, line 21: nit: please move the exposed property to the top

internal/profile/config_profile_test.go
```diff
@@ -0,0 +1,85 @@
...
+	// actually create & check the new profile
+	option := Options{
+		PackagePath: elasticPackageDir,
+		Name:        "test_from",
+		FromProfile: profileName,
+	}
+	err = createProfileFrom(option)
+	assert.NoErrorf(t, err, "error copying updating profile %s", err)
+
+	cleanupProfile(t, elasticPackageDir)
```
mtojek, line 79: WDYT about defer-calling it just after creating the directory?

mtojek (commented):
Ok, wrapping up. I think I reviewed the entire codebase. Once they're fixed I believe it would be ready to get 👍 . Thanks for adjusting this PR!

It would be cool if you can run some basic smoke tests for the new command (profiles new/list/delete). There is `scripts` directory where you can add a shell script using them. Next, you can enable it in Makefile: https://github.com/elastic/elastic-package/blob/master/Makefile#L42 .

internal/profile/profile_json.go
fearful-symmetry: good catch!

internal/profile/config_profile_test.go
fearful-symmetry: Yah, that makes more sense.

fearful-symmetry:
Oop, I'm dumb, forgot to add a few files to the commit.

ycombinator:
I noticed that the profile folders are being created under `~/.elastic-package/stack/`. Since we've decided that profiles could go beyond just stack settings, how about creating them under something like `~/.elastic-package/profiles/`? In case a pre-profiles `elastic-package` is detected, we could move the existing configuration files from `~/.elastic-package/stack/` to `~/.elastic-package/profiles/default_migrated_NNNNNNNN/`.

mtojek:
I would vote for `~/.elastic-package/profiles/stack/`, in case we decide to introduce some other profile dependent features.

Makefile
```diff
@@ -39,6 +39,9 @@ test-stack-command:
 test-check-packages:
 	./scripts/test-check-packages.sh
 
+test-profiles-command:
+	./scripts/test-profiles-command.sh
+
 test: test-go test-stack-command test-check-packages
```
mtojek, line 45: nit: `test-profiles-command` should be part of this too.

mtojek (approved):
LGTM! It's a nifty feature. Do you have any further plan to develop this? I mean, it would be cool if there is a dedicated "profile" stored in the integration folder with custom settings, e.g. special profile settings for the system integration.

Please wait for the approval from Shaunak.

ycombinator:
> I would vote for `~/.elastic-package/profiles/stack/`, in case we decide to introduce some other profile dependent features.

If we go with this path, we would end up with profile folders like this:
```
~/.elastic-package
    profiles/
        stack/
            default/
            foo/
            bar/
            ...
```

I'm curious why we want to introduce `stack` in that path, particular since we decided in earlier conversations that profiles may contain non-stack settings in the future. Why not have paths that look like the following instead?

```
~/.elastic-package
    profiles/
        default/
        foo/
        bar/
        ...
```

mtojek:
Sorry I mistyped what I wanted to express:

```
~/.elastic-package
    profiles/
        default/
            stack/
               <stack files>
        foo/
            stack/
               <stack files>
        bar/
            stack/
               <stack files>
        ...
```

The stack directory has been introduced just in case of potential future extensions. If you don't like it, we can follow the approach you proposed.

ycombinator:
> Sorry I mistyped what I wanted to express:
> ... [correct folder structure example]

Gotcha. I'm good with your proposal as it allows for extensibility within each profile.

fearful-symmetry:
So, @mtojek / @ycombinator I'm still trying to decide how to rearrange the profiles. Right now we put development packages under `./elastic-stack/stack/development` and I'm not sure it makes sense to put that under the profile system. 

I wonder if we should put the profiles under a different directory, like just `~/.elastic-package/profiles` and we can make new subdirectories within that for any non-stack profile features. 

Or maybe just have a "global" `~/.elastic-package/stack` and a local `~/.elastic-package/profiles/[profile]/stack` ?

ycombinator:
> Right now we put development packages under `./elastic-stack/stack/development` and I'm not sure it makes sense to put that under the profile system. 

Agreed. Development packages have nothing to do with profiles (which are about configuration settings) so we should continue to keep the development packages under `~/.elastic-package/stack/development`.

> I wonder if we should put the profiles under a different directory, like just `~/.elastic-package/profiles` and we can make new subdirectories within that for any non-stack profile features. 

I think we are on the same page for this as well. I suggested this in https://github.com/elastic/elastic-package/pull/301#issuecomment-842719413 and @mtojek suggested it in https://github.com/elastic/elastic-package/pull/301#issuecomment-843845965.

mtojek (changes requested):
> Agreed. Development packages have nothing to do with profiles (which are about configuration settings) so we should continue to keep the development packages under ~/.elastic-package/stack/development.

I think we must use a single "development" directory for built packages, so it's not scattered across profiles. Currently, the "development" directory is in the same place as "Dockerfile.package-registry", so it will be a bit weird that it's somewhere else now, but I suppose we don't have a choice.

Actually that brought me to another bug (sorry about that):

1. Create a new profile  (called `first`) based on the default one.
2. Review `~/.elastic-package/stack/first/Dockerfile.package-registry`

The Dockerfile depends on the "default" profile, but shouldn't.

```
FROM docker.elastic.co/package-registry/distribution:snapshot

COPY default/package-registry.config.yml /package-registry/config.yml
COPY development/ /packages/development
```

Originally that a was variable: https://github.com/elastic/elastic-package/blob/master/internal/install/static_package_registry_dockerfile.go . Is it possible to bring it back here too?

fearful-symmetry:
Alright, this might be a little closer to what we're looking for @mtojek / @ycombinator 

We're creating a new profile directory under `~/.elastic-package/profiles`. `stack/development` remains unchanged.
Do we want to create some kind of `~/.elastic-package/profiles/[profile]/stack/` directory as well?

fearful-symmetry:
Alright, so, I converted both the docker-compose file and the Dockerfile for the package manager to use env variables instead of writing out the profile names, so that should solve the issue.

ycombinator:
Thanks for the updates @fearful-symmetry. Don't worry about the pace at which this PR is progressing. AFAIK we don't have an urgent demand for profiles (although they will solve some configuration problems rather nicely once ready) and I think it's more important to design this feature right than rush something out.

I tried out the latest set of changes and the contents under `~/.elastic-package/` look much cleaner to me now.

Moving on from the folder structure, I have a suggestion: could we rename the `elastic-package profiles new` sub-command to `elastic-package profiles create`? We have an `elastic-package create` subcommand so it would be nice to align the terminology.

ycombinator:
I created a new profile `foo`. I didn't specify any `--from` argument when creating it so it got created from the `default` profile. When I list profiles I get this:

```
elastic-package profiles list
+---------+---------------------------+---------+---------+--------------------------------------------------+
|  NAME   |       DATE CREATED        |  USER   | VERSION |                       PATH                       |
+---------+---------------------------+---------+---------+--------------------------------------------------+
| default | 2021-05-20T15:29:25-07:00 | shaunak | d5152c1 | /Users/shaunak/.elastic-package/profiles/default |
+---------+---------------------------+         +         +--------------------------------------------------+
| foo     | 2021-05-20T15:35:16-07:00 |         |         | /Users/shaunak/.elastic-package/profiles/foo     |
+---------+---------------------------+---------+---------+--------------------------------------------------+
```

It would be nice if the user and version were repeated for the `foo` profile row as well. I believe this is the same problem as https://github.com/elastic/elastic-package/issues/360.

ycombinator:
After creating a `foo` profile from the `default` profile, I now have two profiles. However, when I try to delete the `default` profile it won't let me.  I understand why this is the case: because profile-aware commands like `elastic-package stack` need to default to using the `default` profile.

However, I wonder — **for a follow up PR** — if we should introduce a notion of a **current profile**,  something that could be set via `elastic-package profiles use <profile ID>` or similar. Initially the `default` profile will be automatically set as the current profile. Creating a new profile would automatically set that one to be the current profile. It would not be possible to delete the current profile. Then, profile-aware commands like `elastic-package stack` would default to the current profile, unless a different one is explicitly specified via the `--profile` command. WDYT? If you like this idea, I'll file an enhancement issue for a follow up PR.

ycombinator:
I'm a little concerned that a profile is simply a copy of all internal configuration and script files, including files like `healthcheck.sh`. 

One one hand it gives a user an incredible amount of control to do whatever they want, including shooting themselves in their feet. I'm less concerned about users making changes that would shoot themselves in their feet because, presumably, a user could always use the `default` profile to achieve a working state while they debug their profile's changes.

On the other hand, it exposes a lot of internals of `elastic-package` implementation, which means now every file that's part of a profile is now part of the public interface of `elastic-package`. If we (`elastic-package` maintainers) wanted to change a file, say `healthcheck.sh`, that's part of a profile but the user has created a new profile on their system with this file in it, and then they upgrade `elastic-package`, would the newer version of `elastic-package` work with their profile's copy of the file?

I was more imagining file(s) under a profile to be just properties/settings file(s), much like `filebeat.yml` or a [kubeconfig](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/) file. Settings from such profile file(s) would be interpolated into actual, underlying configuration files and scripts that are part of the _implementation_ of `elastic-package`. This minimizes the surface area of the `elastic-package` public _interface_, giving  us (`elastic-package` maintainers) greater freedom to modify the internal configuration and script files as needed without having to worry too much about backwards compatibility with profiles created with older versions of `elastic-package`.

fearful-symmetry:
> if we should introduce a notion of a current profile, something that could be set via elastic-package profiles use <profile ID> or similar. 

I was actually contemplating something like this earlier today, and figured it was outside the scope of the current PR. I think you're definitely right, we should have some some kind of "current" profile.


As far as what APIs we're exposing to the user via the profile system--that's something I was wondering about as well. On one hand, you're right, and we risk giving users a bit of a foot-cannon when they upgrade the tooling and their old profile doesn't work any more. On the other hand, I've had to do some "creative" hacks in the past when testing components of the system module, so I kind of developed this from a perspective of "the user should be able to spin up the stack in a way that's very different from what we anticipate." I should _probably_ try to pull out the health check, since that seems a tad unnecessary. 

I wonder if there's a clever way to manage files under the profile system. Version tagging? Some kind of merge function?

fearful-symmetry:
Alright @ycombinator fixed the way `elastic-package profiles list` works, changed the command to `create` and moved out `healthcheck.sh`.

ycombinator:
Created https://github.com/elastic/elastic-package/issues/361 for the follow up enhancement of current profiles.

Checked out the latest changes. The table produced by `elastic-package profiles list` looks good now, much more readable IMO. Thanks for pulling out `healthcheck.sh` too.

> As far as what APIs we're exposing to the user via the profile system--that's something I was wondering about as well. On one hand, you're right, and we risk giving users a bit of a foot-cannon when they upgrade the tooling and their old profile doesn't work any more. On the other hand, I've had to do some "creative" hacks in the past when testing components of the system module, so I kind of developed this from a perspective of "the user should be able to spin up the stack in a way that's very different from what we anticipate." 

Fair point, and I'll happily defer to you on the right level of interface exposure here as you are the [most experienced user for this use case](https://github.com/elastic/elastic-package/issues/179).

> I wonder if there's a clever way to manage files under the profile system. Version tagging? Some kind of merge function? Copy on Write?

Merging sounds fraught with other potential problems — like how to recover from conflicts?

By "version tagging" I assume you mean something like "this profile is compatible with versions X through Y of `elastic-package`" or something like that?

Similarly, could you elaborate on how copy-on-write could help?

Thanks for your patience with working through these issues. I love how powerful this feature is but I'm also glad we're taking the time to think through user experience for it!

fearful-symmetry:
> Merging sounds fraught with other potential problems — like how to recover from conflicts?

@ycombinator I was mostly throwing ideas out there, most of which would require a certain bit of over-engineering. The simplest would be some kind of version tagging attached to managed files. If we push an update that requires a new `snapshot.yml` or something, we tag that file with a new major version, and if it doesn't match the version the user has, we can warn them. Could also be expanded to the `list` command, and we could perhaps tell a user that a given profile will no longer work.

mtojek:
Version conflicts might be annoying but on the hand the user might have messed up with their config so much that even the smartest handling won't solve conflicts. I'm fine with postponing this for later.

mtojek (commented):
I'm concerned about the current directory layout as it's a bit tangled:

```
/Users/marcin.tojek/.elastic-package
├── config.yml
├── deployer
│   ├── kubernetes
│   │   └── elastic-agent.yml
│   └── terraform
│       ├── Dockerfile
│       ├── run.sh
│       └── terraform-deployer.yml
├── profiles
│   └── default
│       ├── Dockerfile.package-registry
│       ├── kibana.config.yml
│       ├── package-registry.config.yml
│       ├── profile.json
│       └── snapshot.yml
├── stack
│   ├── development
│   └── healthcheck.sh
├── tmp
│   └── service_logs
└── version
```

We suggested a different layout: https://github.com/elastic/elastic-package/pull/301#issuecomment-843845965 .  

Let me bring it here as well:

```
~/.elastic-package
    profiles/
        default/
            stack/
               Dockerfile.package-registry
               kibana.config.yml
               healthcheck.sh # not sure why did you detached the healthcheck file. IMHO it's part of the stack too.
               <other stack files>
        foo/
            stack/
               <stack files as above>
        bar/
            stack/
               <stack files as above>
        ...
    stack/
         development/
             <built packages>
```

Such layout allows for extensions - profiles are not just coupled with "stack" but potentially other future features. I understand that the "stack/development" with built packages is aside, but I'm afraid we can't do anything with this one.

mtojek:
> @mtojek So we just want each profile to have it's own stack where we place the pre-existing files, then?

Correct. There would be also a global "stack" directory for built packages in "development".

> The healthcheck.sh is now in the global ~/.elastic-package/stack` based on a conversation with @ycombinator, where we decided that the healthcheck files are perhaps too closely tied to the implementation, and thus can break easily, and there's probably few reasons a user will need to change it.

Ok, it seems that I missed this discussion. I'm ok with leaving it as it's now - global healthcheck.  My counter-argument is that's a bit scattered across two places (global + profile's). It's up to you, Alex, to decide.


One last thought:

```
         default/
            stack/
               Dockerfile.package-registry
               kibana.config.yml
               healthcheck.sh # not sure why did you detached the healthcheck file. IMHO it's part of the stack too.
               <other stack files>
```

If we agree on preserving the layout above, will "profile.json" be present in the profile directory? e.g.:

```
profile_name/
    stack/
       <stack_files>
    profile.json
```

fearful-symmetry:
>  My counter-argument is that's a bit scattered across two places (global + profile's). It's up to you to decide, Alex.

It's a bit scattered, yah, but I think we should be selective about what we put under the `profile` system. As `elastic-package` grows, it's unlikely that every component of the docker orchestration will need to be tinkered with by users, and we risk increasing the number of components that might break when the user upgrades.

Either way, let me tinker with the file paths a bit more.

fearful-symmetry:
Alright, I think this looks a tad cleaner:
```
/home/alexk/.elastic-package/profiles/default
├── profile.json
└── stack
    ├── Dockerfile.package-registry
    ├── kibana.config.yml
    ├── package-registry.config.yml
    └── snapshot.yml
```

cmd/stack.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/elastic/elastic-package/internal/install"
```
mtojek, line 11: nit: please sort imports in this file

mtojek (approved):
I tested the feature and looks to be working correctly (left one nit-pick). Thanks for all adjustment, I think you can get this in.

Once it's merged - could you please update the dependency on elastic-package in the Integration repo? (`go get github.com/elastic/elastic-package@master` and `go mod tidy`). We'll make sure that it doesn't harm development of any of our integrations.

fearful-symmetry:
@ycombinator did you still want to approve this?

mtojek:
@fearful-symmetry I think you can proceed with this PR and dep change in integrations as Shaunak will not continue work on package tooling.

ycombinator:
Glad to see this PR merged! I was on PTO all last week so ++ for not waiting for my review.

PR #302, opened 2021-03-29
Adding script to generate README

ycombinator:
This PR adds a script and a Makefile target to regenerate the `README.md` file at the root of the `elastic-package` repository. The goal is to dynamically regenerate the "Commands" part of the README using the same contents used when displaying `elastic-package help <command>`, for every `<command>` supported by `elastic-package`.

### Usage

```sh
make update
```

To review the rendered version of the `README.md` generated by this PR, see https://github.com/ycombinator/elastic-package/blob/gen-readme-cmds/README.md.

Related to #75.


cmd/version.go
```diff
@@ -7,19 +7,27 @@ package cmd
 import (
 	"github.com/spf13/cobra"
 
+	"github.com/elastic/elastic-package/internal/cobraext"
+
```
mtojek, line 11: nit: empty line

tools/readme/main.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		log.Fatal(errors.Wrap(err, "loading README template failed"))
+	}
+	return readmeTmpl
+}
+
+func generateReadme(readmeTmpl *template.Template, cmdsDoc string) {
+	readme, err := os.OpenFile("../../README.md", os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		log.Fatal(errors.Wrap(err, "opening README file failed"))
```
mtojek, line 73: nit: maybe log here the log path

cmd/testrunner.go
```diff
@@ -23,32 +23,44 @@ import (
 	_ "github.com/elastic/elastic-package/internal/testrunner/runners" // register all test runners
 )
 
+func init() {
+	cobraext.CommandInfos[testCmd] = cobraext.CommandInfo{
```
mtojek, line 27: Frankly speaking, I'm not in favor of overusing `init` as it's more like static code - you never know the order of execution.

WDYT about a bit different approach:
- `cobraext` will contain a structure that extends `cobra.Command` with context (global, package) - let's say `cobraext.Command`
- leave single `init` function in `root.go` that will collect and return all `cobraext.Command`s
- no need to expose CommandInfos, which are dedicated only for README generation.

I'm happy to discuss this idea.

mtojek (commented):
The implementation seems to be correct and definitely adds value to the elastic-package. I left a comment about an alternative code design.

cmd/testrunner.go
ycombinator: I like it! I think it's a better design. I'll take a stab at it and update this PR. Thanks!

cmd/version.go
```diff
@@ -7,19 +7,27 @@ package cmd
...
 	"github.com/elastic/elastic-package/internal/version"
 )
 
-const versionLongDescription = `Use this command to print the version of elastic-package that you have installed. This is especially useful when reporting bugs.
+func init() {
+	cobraext.CommandInfos[versionCmd] = cobraext.CommandInfo{
+		Short:   "Show application version",
+		Long:    versionLongDescription,
+		Context: "global",
```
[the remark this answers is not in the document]
ycombinator: Yes, I meant to do this too but forgot. Thanks, will do!

Makefile
```diff
@@ -22,6 +22,9 @@ licenser:
 gomod:
 	go mod tidy
 
+readme:
```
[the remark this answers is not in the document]
ycombinator: Well, it's generating the README, not checking anything, so I thought it should be it's own target. 

I was also thinking we could have a more general `update` target that depended on this `readme` target OR an `update` target that directly called the README generation script. WDYT?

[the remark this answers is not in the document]
mtojek: It's rather a question about preserving consistency: is it possible to add new command and forget to regenerate README (something similar to what we have in integrations for docs).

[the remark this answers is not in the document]
ycombinator: Oh, I see what you mean. You want `check` to check if the README is up to date and return an error if it's not? I'm fine with that, but we'll also need a separate target to actually re-generate the README. That's what the `readme` target's intent was but maybe we should rename it to `update` to keep it more general (in case we have more such updates in the future)?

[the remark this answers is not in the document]
mtojek: I'm fine with both names, you can pick the one, you think, fits better :)

cmd/testrunner.go
ycombinator: I took a stab at this in c0176d7. Let me know what you think!

cmd/root.go
```diff
@@ -21,23 +38,18 @@ func RootCmd() *cobra.Command {
...
-		setupInstallCommand(),
-		setupStackCommand(),
-		setupFormatCommand(),
-		setupLintCommand(),
-		setupPromoteCommand(),
-		setupPublishCommand(),
-		setupTestCommand(),
-		setupUninstallCommand(),
-		setupVersionCommand())
+	for _, cmd := range commands {
```
mtojek, line 41: nit: won't such form work here?

`rootCmd.AddCommand(commands...)` ?

cmd/root.go
```diff
@@ -11,6 +11,23 @@ import (
 	"github.com/elastic/elastic-package/internal/logger"
 )
 
+// Make sure to preserve alphabetical order!
```
mtojek, line 14: nit: unless you introduce dedicated array `type Commands []Command` that has a method `sorted()` :)

mtojek (approved):
Perfect! I left two nit-picks, but it's up to you to decide whether they are relevant.

cmd/root.go
ycombinator: Yeah, I had done this before but decided to remove it and keep it simple since we control the addition of commands (it's not a truly open plugin-like system). So I'm going to keep it this way for now.

cmd/root.go
ycombinator: Oh wait, this is is not possible because we're not adding `commands...`. We're actually adding `cmd.Command` for each `cmd` in `commands`.

cmd/root.go
ycombinator: Actually, I added sorting in 55c1edd.

mtojek (approved):

PR #303, opened 2021-03-29
Add location library to elastic-package

fearful-symmetry:
This is my attempt at "breaking up" https://github.com/elastic/elastic-package/pull/301 into more manageable parts. This is the first part, which is breaking up `internal/install` to create a locations library.

Right now, `internal/install` is taking on the role of managing the locations of all the files that `elastic-package` cares about. In an attempt to limit the duties of `install` to just installing things, and also to prepare for the `internal/profile` package that'll come right after this, we're adding a `internal/location` library that's solely responsible for owning where things are on the filesystem.

The fact that we have _multiple_ methods for accessing some of the file paths is due to `internal/install/install.go`, which is the only library that cares about numerous file locations throughout `~/.elastic-package`, so I decided to give it a provider.

I also chose not to touch `internal/install/install.go` here, as that file is particularly tied up with the newer `profile` library, and if I update that in the next PR with `internal/profile`, I won't have an ugly merge operation on my hands.


internal/locations/locations.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+//LocationManager maintains an instance of a config path location
+type LocationManager struct {
+	StackPath string
+}
+
+// NewLocationManager returns a new manager to track the Configuration dir
+func NewLocationManager() (LocationManager, error) {
+
```
mtojek, line 40: nit: empty line

internal/locations/locations.go
```diff
@@ -0,0 +1,148 @@
...
+	terraformDeployerDir  = filepath.Join(deployerDir, "terraform")
+)
+
+//LocationManager maintains an instance of a config path location
+type LocationManager struct {
+	StackPath string
+}
+
+// NewLocationManager returns a new manager to track the Configuration dir
+func NewLocationManager() (LocationManager, error) {
```
mtojek, line 39: let's preserve consistency with the rest of codebase: `(*LocationManager, error)`. In case of an error, the * LocationManager is nil

internal/cleanup/service_logs.go
```diff
@@ -8,15 +8,15 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/elastic/elastic-package/internal/files"
-	"github.com/elastic/elastic-package/internal/install"
+	"github.com/elastic/elastic-package/internal/locations"
```
mtojek, line 11: I'm not sure if `internal/locations` is the correct place for this feature. This is a configuration that's strongly related to app configuration. I would be for placing it all together under `internal/configuration`, so we may have:
`internal/configuration/locations`
`internal/configuration/profiles`
`internal/configuration/stack`

In this case the goal would be transforming the `internal/install` into something more sophisticated that can deal with profiles and config extensions.

internal/locations/locations.go
```diff
@@ -0,0 +1,148 @@
...
+
+var (
+	serviceLogsDir        = filepath.Join(temporaryDir, "service_logs")
+	kubernetesDeployerDir = filepath.Join(deployerDir, "kubernetes")
+	terraformDeployerDir  = filepath.Join(deployerDir, "terraform")
+)
+
+//LocationManager maintains an instance of a config path location
+type LocationManager struct {
+	StackPath string
```
mtojek, line 35: Is this used somewhere? It seems to me to be a dead code or something hasn't been set.

mtojek (commented):
> I also chose not to touch internal/install/install.go here, as that file is particularly tied up with the newer profile library, and if I update that in the next PR with internal/profile, I won't have an ugly merge operation on my hands.

I understand your concerns regarding hard merge operation, but unfortunately this is the way how we should perform refactoring operations - it's not just about breaking the bigger PR into pieces. I think you should remove the refactored code as part of this PR. 
Please remember that the PR shouldn't just introduce a dead code, but it must provide an added value for the end-user or at least be used by other internals.

fearful-symmetry:
> I think you should remove the refactored code as part of this PR.

I'm not sure what you're referring to @mtojek ? Just keep `locations/locations.go` and remove the components that're swapped over to it?

internal/locations/locations.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// NewLocationManager returns a new manager to track the Configuration dir
+func NewLocationManager() (LocationManager, error) {
+	cfg, err := ConfigurationDir()
+	if err != nil {
+		return LocationManager{}, errors.Wrap(err, "error getting config dir")
+	}
+
+	return LocationManager{cfg}, nil
```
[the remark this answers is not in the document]
fearful-symmetry: Yah, I think exporting `StackPath` was probably on accident.

fearful-symmetry:
@mtojek I moved around `locations` to a new `configuration` library. However, I wasn't sure what your thinking was behind putting `stack` in `internal/configuration` ? It doesn't seem quite as related to the functionality of `profiles` and `locations`

mtojek:
Cool! Please address also other comments and let me know when it's ready for re-review.

fearful-symmetry:
Alright, that _should_ be everything.

ycombinator:
You might need to run `make format` to make CI happy.

fearful-symmetry:
Figured I was forgetting something...

fearful-symmetry:
@ycombinator can you tell if the CI issues are something I did? I'm trying to replicate it, the development files are there.

ycombinator:
@fearful-symmetry It looks like one of the system tests for the `apache.access` data set failed. You can see the failure in the Tests tab in Jenkins.

ycombinator:
I tried to reproduce the failure locally by building `elastic-package` from this PR and then running:

```
elastic-package stack up -d
$(elastic-package stack shellinit)
```
```
cd test/packages/apache
elastic-package test system
```

And the tests passed. So I'm going to tickle CI on this PR again and see what happens.

ycombinator:
jenkins, run the tests please

ycombinator:
Same test failed again. I'm going to try and dig into it more deeply locally and see if something turns up.

internal/install/application_configuration.go
```diff
@@ -66,7 +68,7 @@ func (ac *ApplicationConfiguration) StackImageRefs(version string) ImageRefs {
 
 // Configuration function returns the elastic-package configuration.
 func Configuration() (*ApplicationConfiguration, error) {
-	configPath, err := configurationDir()
+	configPath, err := locations.ConfigurationDir()
```
ycombinator, line 71: Would it make sense to use `NewLocationManager()` here and then call the `RootDir()` method on the new `LocationManager` instance? This would allow us to unexport `locations.ConfiguratoinDir()` and uniformly access configuration locations throughout the codebase exclusively via a `LocationManager` instance.

internal/configuration/locations/locations.go
```diff
@@ -0,0 +1,152 @@
...
+func ConfigurationDir() (string, error) {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", errors.Wrap(err, "reading home dir failed")
+	}
+	return filepath.Join(homeDir, elasticPackageDir), nil
+}
+
+// StackDir method returns the stack directory (see: stackDir).
+func StackDir() (string, error) {
```
ycombinator, line 109: It seems like there is a corresponding `StackDir()` method on the `LocationManager` struct as well that does essentially the same thing as this function here. The same applies to other functions below and corresponding methods in the `LocationManager` struct.

Is there some reason we couldn't remove the exported functions from this package and instead call their corresponding `LocationManager` methods from across the codebase? IIUC, the idea behind this refactoring is to use `LocationManager` everywhere?

ycombinator: I could be wrong but I think this is the same thing as what @mtojek was referring to here: https://github.com/elastic/elastic-package/pull/303#issuecomment-810380283 (sorry, I just read his comment just now).

mtojek:
Kibana complains: https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-303/11/artifact/build/elastic-stack-dump/check/logs/kibana.log

I will retry it and see if it's flaky. We've reported something similar: https://github.com/elastic/kibana/issues/95388

mtojek:
jenkins run the tests please

mtojek:
Ok, it seems that it's related to this PR, will try to reproduce it locally.

mtojek:
I think I found the problem, it seems to be related with volume's source.

Take a look at paths. The `/tmp/service_logs/` path is mounted in elastic-agent's container:

```
    streams:
      - id: logfile-apache.access-ca17e851-8464-4100-a450-0733973daf00
        data_stream:
          dataset: apache.access
          type: logs
        paths:
          - /tmp/service_logs/access.log*
        exclude_files:
          - .gz$
        processors:
          - add_fields:
              target: ''
              fields:
                ecs.version: 1.5.0
```

The elastic-agent's container has the following definition:

```
  elastic-agent:
    image: ${ELASTIC_AGENT_IMAGE_REF}
    depends_on:
      elasticsearch:
        condition: service_healthy
      kibana:
        condition: service_healthy
    healthcheck:
      test: "sh -c 'grep \"Agent is starting\" /usr/share/elastic-agent/elastic-agent.log*'"
      retries: 30
      interval: 1s
    hostname: docker-fleet-agent
    environment:
    - "FLEET_ENROLL=1"
    - "FLEET_ENROLL_INSECURE=1"
    - "FLEET_INSECURE=1"
    - "FLEET_SETUP=1"
    - "FLEET_URL=http://kibana:5601"
    - "KIBANA_HOST=http://kibana:5601"
    volumes:
    - type: bind
      source: ../tmp/service_logs/
      target: /tmp/service_logs/
```

The volume's source is relative. I suspect that this path can't be found.

fearful-symmetry:
That's odd. That path makes sense, and nothing in the code that would change it.

fearful-symmetry:
Okay, I'm still baffled by the kibana errors. The path that @mtojek pointed to seems fine, the mount and files all exist and are pointing to the correct location. I can't see anything weird in the kibana install itself.

mtojek:
I think you can ignore these Kibana errors:

Regarding failed tests - I did a short test:

```
docker exec -it <agent_container> /bin/bash
ls -la /tmp/service_logs/
```

The directory exists.

```
cd ~/go/src/github.com/elastic/integrations/packages/apache
```

The mounted directory will disappear, but it should be present and contain log files:

```
 ls -la /tmp/service_logs/
ls: cannot access /tmp/service_logs/: No such file or directory
```

fearful-symmetry:
@mtojek I can't seem to reproduce that. `elastic-package test system -v` fails to run because (I assume) all the kibana auth errors, and the directory still exists after. Could something be deleting the directory after the test?

mtojek:
> elastic-package test system -v fails to run because (I assume) all the kibana auth errors, and the directory still exists after.

I meant the directory inside the agent's container (`/tmp/service_logs`).

fearful-symmetry:
> I meant the directory inside the agent's container (/tmp/service_logs).

Yah, that's what I was referring to. I assume whatever is removing that directory isn't being run because the `test system` command is failing.

mtojek:
The system test runner fails because it expects some access logs to be present in that directory. Unfortunately the entire directory is deleted, but in fact the tool should delete only it's content. You can compare the behavior and paths in your PR and master build.

This can be a reference point: https://github.com/elastic/elastic-package/blob/master/internal/testrunner/runners/system/servicedeployer/compose.go#L55

fearful-symmetry:
Well, I'm still not sure what's up with the `/tmp/service_logs` thing. I ran the apache tests, and the logs exist inside the `elastic-package-service_apache_1` container at `/usr/local/apache2/logs`, however, they don't exist inside the `elastic-package-stack_elastic-agent_1` at `/tmp/service_logs/`

Here's what the mounts look like:
```
            {
                "Type": "bind",
                "Source": "/home/alexk/.elastic-package/tmp/service_logs",
                "Destination": "/usr/local/apache2/logs",
                "Mode": "rw",
                "RW": true,
                "Propagation": "rprivate"
            }
```

and

```
                {
                    "Type": "bind",
                    "Source": "/home/alexk/.elastic-package/tmp/service_logs",
                    "Target": "/tmp/service_logs/"
                }
```

mtojek:
So you can see that `/home/alexk/.elastic-package/tmp/service_logs` is the bridge between both containers. Could you verify that before starting the `test system` but after `stack up`, you can create a file on your host machine and it's still visible from inside of the agent container?
I assume you can do the same exercise after the `test system`.

fearful-symmetry:
Yah, something odd is going on inside that directory. Whatever I had from the docker host in `/home/alexk/.elastic-package/tmp/service_logs` doesn't appear on the container, and from inside the container I can't read or write to `/tmp/service_logs`.

I assume this is a config issue with docker-compose somehow, gonna tinker around after I get lunch.

fearful-symmetry:
@mtojek so, I did some more tinkering around, and it seems like the `test system` is breaking things. The bindmount inside the container works fine before, but once the Apache container starts up, the agent container can't seem to see anything inside the bindmount.

fearful-symmetry:
So, after the second container spins up, the link of the directory seems to get nuked:

before:
```
 67113534 drwxr-xr-x. 2 elastic-agent elastic-agent   6 Apr  2 18:07 service_logs
```
After:
```
 67113534 drwxr-xr-x. 0 elastic-agent elastic-agent   6 Apr  2 18:07 service_logs
```

fearful-symmetry:
The weird part is I can't reproduce this by manually starting up the docker container for apache:
```
docker-compose -f _dev/deploy/docker/docker-compose.yml --verbose up
```

fearful-symmetry:
Okay, think I figured it out. I was passing the wrong directory to the version check, so it was re-installing elastic-package every time.

internal/configuration/locations/locations.go
```diff
@@ -0,0 +1,106 @@
...
+func (loc LocationManager) TerraformDeployerYml() string {
+	return filepath.Join(loc.stackPath, terraformDeployerDir, terraformDeployerYmlFile)
+}
+
+// ServiceLogDir returns the log directory
+func (loc LocationManager) ServiceLogDir() string {
+	return filepath.Join(loc.stackPath, serviceLogsDir)
+}
+
+// ConfigurationDir returns the configuration directory location
```
ycombinator, line 99:
```suggestion
// configurationDir returns the configuration directory location
```

ycombinator (approved):
LGTM but please wait for @mtojek's review as well as he was actively reviewing this PR. Thanks!

mtojek (approved):
Nice! I think you can proceed with the next PR.

mtojek:
@fearful-symmetry Once this PR is merged could you please open a PR to elastic/integrations to get this change in (update dependency on elastic-package)? We need to be sure that it doesn't affect integrations.

PR #304, opened 2021-03-30
Publish: check error after cloning repository

mtojek:
This PR fixes the missing check for potential error.

Spotted in: https://github.com/elastic/integrations/pull/815

Most likely it will result in follow-ups (need to fix the root cause).


mtojek:
I'll merge this one as I need it for further debugging, self-approved.

PR #306, opened 2021-03-31
Publish: support no fork mode

mtojek:
This PR fixes issue spotted in https://github.com/elastic/integrations/pull/815. 

CI uses "elasticmachine" account which doesn't have a fork of package-storage, but has permissions to push changes to the original repository.

Sample PR: https://github.com/elastic/package-storage/pull/1083

Testing:

```
cd test/packages/apache
elastic-package publish -v --fork=false
```

### DIFF
--- a/cmd/publish.go
+++ b/cmd/publish.go
@@ -28,6 +28,9 @@ func setupPublishCommand() *cobra.Command {
 		RunE:  publishCommandAction,
 	}
 
+	// Fork flag can be a workaround for users that don't own forks of the Package Storage.
+	cmd.Flags().BoolP(cobraext.ForkFlagName, "f", true, cobraext.ForkFlagDescription)
+
 	// SkipPullRequest flag can used to verify if the "publish" command works properly (finds correct revisions),
 	// for which the operator doesn't want to immediately close just opened PRs (standard dry-run).
 	cmd.Flags().BoolP(cobraext.SkipPullRequestFlagName, "s", false, cobraext.SkipPullRequestFlagDescription)
@@ -37,6 +40,7 @@ func setupPublishCommand() *cobra.Command {
 func publishCommandAction(cmd *cobra.Command, args []string) error {
 	cmd.Println("Publish the package")
 
+	fork, _ := cmd.Flags().GetBool(cobraext.ForkFlagName)
 	skipPullRequest, _ := cmd.Flags().GetBool(cobraext.SkipPullRequestFlagName)
 
 	// Setup GitHub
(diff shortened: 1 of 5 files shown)


cmd/publish.go
```diff
@@ -37,6 +40,7 @@ func setupPublishCommand() *cobra.Command {
 func publishCommandAction(cmd *cobra.Command, args []string) error {
 	cmd.Println("Publish the package")
 
+	fork, _ := cmd.Flags().GetBool(cobraext.ForkFlagName)
```
ycombinator, line 43: Any reason not to check for error here and fail using `cobraext.FlagParsingError`?

internal/cobraext/const.go
```diff
@@ -27,6 +27,9 @@ const (
 	FailFastFlagName        = "fail-fast"
 	FailFastFlagDescription = "fail immediately if any file requires updates (do not overwrite)"
 
+	ForkFlagName        = "fork"
+	ForkFlagDescription = "use fork mode"
```
ycombinator, line 31: This description is rather vague. I'm okay with it if we only really expect CI to use this flag. But if there is any potential for human users to use it, it might be nice to explain this flag a bit more here.

internal/publish/pull_requests.go
```diff
@@ -41,12 +41,18 @@ func checkIfPullRequestAlreadyOpen(githubClient *github.Client, manifest package
 	return false, nil
 }
 
-func openPullRequest(githubClient *github.Client, githubUser, destinationBranch string, manifest packages.PackageManifest, commitHash string) error {
+func openPullRequest(githubClient *github.Client, githubUser, destinationBranch string, manifest packages.PackageManifest, commitHash string, fork bool) error {
+	var user = repositoryOwner
```
ycombinator, line 45:
```suggestion
	user := repositoryOwner
```

internal/storage/storage.go
```diff
@@ -134,6 +134,12 @@ func (prs PackageVersions) Strings() []string {
 
 // CloneRepository function clones the repository and changes branch to stage.
```
ycombinator, line 135: Now that this function depends on a fork, maybe update the godoc comment here to say that?

internal/storage/storage.go
```diff
@@ -539,6 +556,12 @@ func RemovePackages(r *git.Repository, sourceStage string, packages PackageVersi
 
 // PushChanges function pushes branches to the remote repository.
 func PushChanges(user string, r *git.Repository, stages ...string) error {
```
ycombinator, line 558: Now that this function depends on a fork, maybe update the godoc comment here to say that?

mtojek: Fixed

internal/storage/storage.go
mtojek: Fixed

internal/cobraext/const.go
mtojek: Fixed

PR #307, opened 2021-04-01
Wrong revision considered as latest

mtojek:
Another issue spotted in https://github.com/elastic/integrations/pull/815


mtojek:
self-approved

PR #308, opened 2021-04-01
Fix: adjust agent healthcheck

mtojek:
This PR modifies the elastic-package JSON log path, so we detect the moment when the agent is ready/healthy.

Issue: https://github.com/elastic/beats/issues/24894

### DIFF
--- a/internal/install/static_snapshot_yml.go
+++ b/internal/install/static_snapshot_yml.go
@@ -82,7 +82,7 @@ services:
       kibana:
         condition: service_healthy
     healthcheck:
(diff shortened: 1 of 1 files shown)


blakerouse (commented):
This is not the proper way to do a health check on the Elastic Agent docker image.

The new `status` command should be used to ensure that the running daemon is working correctly and is healthy.

https://github.com/elastic/beats/pull/24856

mtojek:
Sure, we can switch to this one once it's merged and backported.

ruflin (approved):
Lets get this change in to get the system back to green until healthcheck is in, backported and snapshot builds are available.

PR #310, opened 2021-04-02
Fix: optional source stage in publish

mtojek:
This PR fixes issues spotted in https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Fintegrations/detail/master/174/pipeline/327/

The `winlog` package is a new one, so the source stage was empty.


mtojek:
self-approved, emergency fix

PR #311, opened 2021-04-02
Fix: starting security shard

mtojek:
This PR tries to fix the issue with `.security-7` shard:

```
["org.elasticsearch.action.UnavailableShardsException: at least one primary shard for the index [.security-7] is unavailable",
```

Kibana's healthcheck will depend on the shard's state.

Issue: https://github.com/elastic/kibana/issues/95896

Comment:

I've no idea if this will fix the problem, but let's give it a try :)

### DIFF
--- a/internal/install/static_kibana_healthcheck_sh.go
+++ b/internal/install/static_kibana_healthcheck_sh.go
@@ -7,5 +7,5 @@ package install
 const kibanaHealthcheckSh = `#!/bin/bash
 
 curl -s -f http://127.0.0.1:5601/login | grep kbn-injected-metadata 2>&1 >/dev/null
-curl -s -f -u elastic:changeme http://elasticsearch:9200/_cat/shards | grep .security | grep STARTED
+curl -s -f -u elastic:changeme "http://elasticsearch:9200/_cat/indices/.security-*?h=health" | grep -v red
 `
--- a/internal/install/static_kibana_healthcheck_sh.go
+++ b/internal/install/static_kibana_healthcheck_sh.go
@@ -1,0 +1,11 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package install
+
+const kibanaHealthcheckSh = `#!/bin/bash
+
+curl -s -f http://127.0.0.1:5601/login | grep kbn-injected-metadata 2>&1 >/dev/null
+curl -s -f -u elastic:changeme http://elasticsearch:9200/_cat/shards | grep .security | grep STARTED
+`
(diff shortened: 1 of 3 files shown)


internal/install/static_kibana_healthcheck_sh.go
```diff
@@ -0,0 +1,11 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package install
+
+const kibanaHealthcheckSh = `#!/bin/bash
+
+curl -s -f http://127.0.0.1:5601/login | grep kbn-injected-metadata 2>&1 >/dev/null
+curl -s -f -u elastic:changeme http://elasticsearch:9200/_cat/shards | grep .security | grep STARTED
```
ycombinator, line 10: This checks if _any_ shard is `STARTED`. That could be potentially misleading as a replica might've started but not a primary (I know, not possible on a 1-node cluster but still). It could also be misleading in case the security index ever decides to have multiple primary shards, not just one, and some of them have not yet started.

A more robust test would be to check if _all_ primary shards have been `STARTED`, which is another way of saying that the index is not red (i.e. it is either yellow or green).

```suggestion
curl -s -f -u elastic:changeme "http://elasticsearch:9200/_cat/indices/.security-*?h=health" | grep -v red
```

mtojek: Thanks!

ycombinator (approved):
LGTM. WFG.

I'm good with this fix as a stop-gap.

Longer-term, I do think this is more of an internal Kibana concern. It should be internally performing a similar check on the Security index as done in this PR and reflecting the status via an API, allowing Fleet Server or Elastic Agent to ask the question, "is Fleet ready or not?" without having to care about exactly _why_ Fleet might not be ready yet. Details about this suggestion are in https://github.com/elastic/kibana/issues/95896#issuecomment-813659828.

internal/install/static_kibana_healthcheck_sh.go
```diff
@@ -0,0 +1,11 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package install
+
+const kibanaHealthcheckSh = `#!/bin/bash
+
+curl -s -f http://127.0.0.1:5601/login | grep kbn-injected-metadata 2>&1 >/dev/null
```
ycombinator, line 9: BTW, do we need this?

mtojek: Yes, this way we're checking when Kibana is ready. Normally the port is open and the HTTP server is exposed, but it responds with something like "not ready yet".

ycombinator: I see, thanks.

ycombinator: In this case, should we add a `set -e` at the top of this bash script?

mtojek: I forgot about this! Thanks

PR #314, opened 2021-04-06
Fix custom resource init for k8s deployer

ChrsMark:
This PR fixes a bug with custom resources not being able to be installed on k8s when using k8s service deployer. The deployer `delete`s the resources instead of `apply`ing them.


ChrsMark:
Related to elastic/integrations#822

mtojek (approved):
Ship it! Please also bump up the dependency in elastic/integrations.

PR #317, opened 2021-04-08
Add status command

### LINKED ISSUE #313: Proposal: Pending package changes/status
It would be nice to have a command that allows us to see at a glance the status of a package -- what versions are released in which environments, and which changes are still "pending" (i.e. are in the changelog/code, but have not yet been built/released by a manifest version update and PR to package-storage snapshot).

One thing that has happened in the past is that a change in one of the packages our team is responsible for got pushed to our snapshot branch, but never promoted fully to production until someone was wondering why they didn't see a package when they were wanting to try something out a week or so after we thought we had released the change. A command showing a package status would make these easier to surface.

andrewstucki:
This adds support for a command to retrieve package status information. It can execute in two contexts:

1. By passing a package command line argument, it retrieves information about a deployed package without being in the context of a package directory.
2. If you don't specify a package and are in a package directory, it retrieves the remote deployment info and also does some local checking of package manifests and pending changes.

Additionally, just to make the info easier to grok, it has colorized output which is automatically disabled if you're not using a tty (like if you're piping to another program) or can be disabled by passing a `-c`/`--no-color` flag.

Here's some screenshots:

In the package directory context:
Screen Shot 2021-04-08 at 11 40 40 AM

Specifying a package:
Screen Shot 2021-04-08 at 10 39 43 AM

Disabling color:
Screen Shot 2021-04-08 at 10 39 56 AM

Showing all deployed versions (not just latest):
Screen Shot 2021-04-08 at 11 03 36 AM

And, just for kicks, with a light terminal color scheme (not that anyone uses that):
Screen Shot 2021-04-08 at 12 40 11 PM

And here's the help text output:

```
➜  system git:(master) ✗ elastic-package status -h
Use this command to display the current deployment status of a package.

If a package name is specified, then information about that package is
returned, otherwise this command checks if the current directory is a
package directory and reports its status.

Context: package

Usage:
  elastic-package status [package] [flags]

Flags:
  -a, --all        show all deployed package versions
  -h, --help       help for status
  -c, --no-color   disable color output

Global Flags:
  -v, --verbose   verbose mode
```

Closes #313


andrewstucki:
From what I can tell amidst the build failure logs, this appears to be the culprit for the failure:

> [2021-04-08T15:20:39.795Z] Error: error running package system tests: could not complete test run: can't check enrolled agents: no agent enrolled in time

Any suggestions on how to fix it or is it a CI environment thing?

cmd/status.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return errors.Wrap(err, "parsing pending changelog entries failed")
+	}
+	bold := color.New(color.Bold)
+	red := color.New(color.FgRed, color.Bold)
+	cyan := color.New(color.FgCyan, color.Bold)
+	bold.Fprint(w, "Package: ")
+	cyan.Fprintln(w, p.Name)
+
+	environmentTable := [][]string{}
```
andrewkroh, line 179:
```suggestion
	var environmentTable [][]string
```

Doesn't make much a difference, but I always prefer this format over the empty slice format. https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices There are a two more in the changes below.

andrewkroh (commented):
Awesome! 👍

cmd/status.go
```diff
@@ -0,0 +1,308 @@
...
+
+func setupStatusCommand() *cobraext.Command {
+	cmd := &cobra.Command{
+		Use:   "status [package]",
+		Short: "Show package status",
+		Args:  cobra.MaximumNArgs(1),
+		Long:  statusLongDescription,
+		RunE:  statusCommandAction,
+	}
+	cmd.Flags().BoolP(cobraext.NoColorFlagName, "c", false, cobraext.NoColorFlagDescription)
```
mtojek, line 53: I think this logic is not required. I don't see a use case that developer would like to disable coloring. If so, there is the flag `TERM`. It would be for removing this.

cmd/status.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/elastic-package/internal/cobraext"
+	"github.com/elastic/elastic-package/internal/packages"
+)
+
+var (
+	// taken from https://github.com/fatih/color/blob/4d2835ff85a014514ee435d49f76dc8b25c9cee3/color.go#L20-L21
+	noColor = os.Getenv("TERM") == "dumb" ||
```
mtojek, line 30: Could you just use `fatih.Color` instead?

cmd/status.go
```diff
@@ -0,0 +1,308 @@
...
+)
+const statusLongDescription = `Use this command to display the current deployment status of a package.
+
+If a package name is specified, then information about that package is
+returned, otherwise this command checks if the current directory is a
+package directory and reports its status.`
+
+func setupStatusCommand() *cobraext.Command {
+	cmd := &cobra.Command{
+		Use:   "status [package]",
```
mtojek, line 47: nit: `"status"`

You can give a sample use case with a package in the long description.

internal/cobraext/const.go
```diff
@@ -53,4 +53,10 @@ const (
 
 	VerboseFlagName        = "verbose"
 	VerboseFlagDescription = "verbose mode"
+
+	NoColorFlagName        = "no-color"
+	NoColorFlagDescription = "disable color output"
+
+	ShowAllFlagName        = "all"
+	ShowAllFlagDescription = "show all deployed package versions"
```
mtojek, line 61: nit: s/package versions/package revisions/

internal/packages/packages.go
```diff
@@ -19,6 +19,9 @@ const (
 	// PackageManifestFile is the name of the package's main manifest file.
 	PackageManifestFile = "manifest.yml"
 
+	// PackageChangelogFile is the name of the package's changelog file.
+	PackageChangelogFile = "changelog.yml"
```
mtojek, line 23: Could you please put the changelog logic in a separate package, e.g. `internal/packages/changelog`?

cmd/status.go
```diff
@@ -0,0 +1,308 @@
...
+)
+
+var (
+	// taken from https://github.com/fatih/color/blob/4d2835ff85a014514ee435d49f76dc8b25c9cee3/color.go#L20-L21
+	noColor = os.Getenv("TERM") == "dumb" ||
+		(!isatty.IsTerminal(os.Stdout.Fd()) && !isatty.IsCygwinTerminal(os.Stdout.Fd()))
+)
+
+const (
+	productionURL = "https://epr.elastic.co"
```
mtojek, line 35: I have few concerns about this file. I have a feeling that it should be split into multiple units. What do you think about the following approach:
1. Leave the table rendering and package root selection in `cmd`.
2. Create new package for Package Registry client like we have ones for `elasticsearch` and `kibana`.
3. Create new package for status (e.g. `internal/packages/status`) that can provide package deployment status for the given package (based on the package root).

mtojek (commented):
Thanks for the implementation, seems to be a nice feature for the app :) I left a couple of comments to help you adjust the code in terms of SoC and general code standards we have here.

mtojek:
> Any suggestions on how to fix it or is it a CI environment thing?

Yes, there was a breaking change pushed to Kibana/Elastic Agent. I deployed an emergency fix for this, it's green now.

cmd/status.go
andrewstucki: This, as far as I know, is the only way to have cobra properly set the Usage line for a subcommand (i.e. `elastic-package status [package] [flags]`) with additional, non-flag arguments -- removing this will cause the `Usage` line to change. Is that alright?

andrewstucki: Also, see the example in the official README: https://github.com/spf13/cobra#example where it has

```go
    Use:   "print [string to print]",
```

mtojek: Thanks for bringing it up. It's a fair point. Please ignore the comment and apologize for the concern.

andrewstucki:
@mtojek let me know if there are any other changes needed

cmd/status.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+	}
+	if err := print(packageStatus, os.Stdout); err != nil {
+		return err
+	}
+	return nil
+}
+
+// data printing/formatters
+
```
mtojek, line 77: nit: empty line

cmd/status.go
```diff
@@ -0,0 +1,184 @@
...
+	var packageStatus *status.PackageStatus
+
+	showAll, err := cmd.Flags().GetBool(cobraext.ShowAllFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.ShowAllFlagName)
+	}
+
+	if len(args) > 0 {
+		packageStatus, err = status.Package(args[0], showAll)
+		if err != nil {
```
mtojek, line 54: nit: would you mind refactoring this part to get rid of code depth level (`if` in `if`)?

cmd/status.go
```diff
@@ -0,0 +1,184 @@
...
+			return err
+		}
+	} else {
+		packageRootPath, found, err := packages.FindPackageRoot()
+		if !found {
```
mtojek, line 59: nit: same here, I suppose that after refactoring, the code wrapped by `else` can be extracted.

internal/cobraext/const.go
```diff
@@ -53,4 +53,10 @@ const (
 
 	VerboseFlagName        = "verbose"
 	VerboseFlagDescription = "verbose mode"
+
+	NoColorFlagName        = "no-color"
+	NoColorFlagDescription = "disable color output"
+
+	ShowAllFlagName        = "all"
```
mtojek, line 60: nit: would you mind putting these flags in alpha num order (as others)?

internal/packages/status/status.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+	status.Changelog = changelog
+	status.Local = manifest
+	return status, nil
+}
+
+// Package returns the status of a given package
+func Package(packageName string, showAll bool) (*PackageStatus, error) {
```
mtojek, line 46: nit: this function can be renamed to `RemotePackage`, as it fetches data from the Package Registry

internal/registry/url_prefixes.go
```diff
@@ -0,0 +1,10 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package registry
+
+const (
+	// SearchAPI is the endpoint for filtering package registry packages
+	SearchAPI = "/search"
```
mtojek, line 9: Should this endpoint be exposed to other packages?

internal/registry/client.go
```diff
@@ -0,0 +1,95 @@
...
+	return c.sendRequest(http.MethodPost, resourcePath, body)
+}
+
+func (c *Client) put(resourcePath string, body []byte) (int, []byte, error) {
+	return c.sendRequest(http.MethodPut, resourcePath, body)
+}
+
+func (c *Client) delete(resourcePath string) (int, []byte, error) {
+	return c.sendRequest(http.MethodDelete, resourcePath, nil)
+}
```
mtojek, line 59: I understand why you have put all these methods, but as long as they're not used by the client, I would be for deleting them. If you prefer you can merge `get` and `sendRequest` methods (KISS).

internal/registry/client.go
```diff
@@ -0,0 +1,95 @@
...
+	Production = NewClient(productionURL)
+	// Staging is a pre-configured staging client
+	Staging = NewClient(stagingURL)
+	// Snapshot is a pre-configured snapshot client
+	Snapshot = NewClient(snapshotURL)
+)
+
+// Client is responsible for exporting dashboards from Kibana.
+type Client struct {
+	host string
```
mtojek, line 35: nit: s/host/endpoint/g

internal/packages/status/status.go
```diff
@@ -0,0 +1,85 @@
...
+	snapshotManifests, err := registry.Snapshot.Revisions(packageName, showAll)
```
mtojek, line 47: What do you think about transforming `showAll` into something more sophisticated like `ListingOptions`. The options struct can contain all relevant booleans like internal, experimental, showAll. With CLI configuration we can decide which should be enabled by default.

internal/packages/status/status.go
```diff
@@ -0,0 +1,85 @@
...
+	return &PackageStatus{
+		Name:       packageName,
+		Snapshot:   snapshotManifests,
+		Staging:    stagingManifests,
+		Production: productionManifests,
+	}, nil
+}
+
+// PendingChanges returns pending changes in a local development package folder
+func (p *PackageStatus) PendingChanges() (*changelog.Revision, error) {
```
mtojek, line 68: Could you please explain why does `PendingChanges` method is actually a separate method :) ? I'm wondering if it can be just static part of the `PackageStatus`.

cmd/status.go
andrewstucki: this was more of a comment to say, printing and formatters follow -- I'll just turn it into real godoc style comments

internal/registry/url_prefixes.go
andrewstucki: I was just following how the kibana package was structured, but I'll keep it from being exported

internal/registry/client.go
andrewstucki: this is the host -- `endpoint` generally refers to a particular API endpoint, in this case what we're capturing is the actual base url of the registry host we're going to hit.

internal/packages/status/status.go
andrewstucki: I'd be fine with that, but, at least for this use-case the other booleans should always be set, otherwise you're not going to be able to return information for most development level packages. The reason I kept it as `showAll` was due to the fact that the other options don't need to be configurable and only it does. I'm fine refactoring it or keeping it this way until someone leverages this with other options they need to set, at which point we could extract it into an options struct. Just let me know if you'd prefer me do that sooner than later.

internal/registry/client.go
mtojek: what about the protocol? I agree it's an URL, but it's host + proto :)

internal/packages/status/status.go
mtojek: If it's not a big effort to refactor, I would be go adjusting it here as we don't have many places to adjust. I understand if you prefer to postpone it and I don't want to stop you :)

internal/registry/client.go
andrewstucki: what if I just use `baseURL`?

andrewstucki: changed the name to `baseURL`

cmd/status.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return err
+	}
+	return print(packageStatus, os.Stdout)
+}
+
+func getPackageStatus(packageName string, showAll bool) (*status.PackageStatus, error) {
+	if packageName != "" {
+		return status.RemotePackage(packageName, registry.SearchOptions{
+			All:          showAll,
```
mtojek, line 64: Actually I was thinking to expose all these options with CLI and define default variables there.

mtojek (commented):
I have only this one comment and I suppose this PR should be good to go.

cmd/status.go
andrewstucki: so, when would you ever not want to set `experimental` or `internal`?

andrewstucki: So, resolution we chatted about off github was to keep `experimental`/`internal` set to true for this particular command. In the future, adding some sort of `list` or `search` command would warrant making these configurable.

internal/registry/revisions.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/elastic/elastic-package/internal/packages"
+)
+
+// SearchOptions specify the query parameters for the search API
+type SearchOptions struct {
+	Internal     bool   `url:"internal"`
+	Experimental bool   `url:"experimental"`
+	All          bool   `url:"all"`
+	Package      string `url:"package"`
```
mtojek, line 25: I just noticed that `Package` is leaking outside. WDYT about extracting `SearchOptions` (without Package) and adding a new structure `searchQuery` that combines  `SearchOptions` and `Package`.

internal/registry/revisions.go
```diff
@@ -0,0 +1,68 @@
...
+package registry
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"github.com/Masterminds/semver"
+	"github.com/pkg/errors"
+
```
mtojek, line 14: nit: empty line and sort imports

internal/registry/revisions.go
```diff
@@ -0,0 +1,68 @@
...
+// SearchOptions specify the query parameters without the package name for the search API
+type SearchOptions struct {
+	Internal     bool `url:"internal"`
+	Experimental bool `url:"experimental"`
+	All          bool `url:"all"`
+}
+
+// SearchQuery specify the package and query parameters for the search API
+type SearchQuery struct {
```
mtojek, line 28: nit: I think this struct can be unexposed

mtojek (approved):
Two nit-picks, but in general - ship it, please!

PR #318, opened 2021-04-12
Emergency fix: Use stable Docker images

mtojek:
This PR fixes the issue with bad/unstable images of Kibana/Elastic-Agent.


mtojek:
self-approved, emergency fix

PR #321, opened 2021-04-13
Add support for synthetics

mtojek:
This PR allows for "synthetics" type data streams.


dominiqueclarke (approved):
Pulled down the branch and built it, the error is indeed gone.

Issue #315, opened 2021-04-06
[K8s service deployer] Configure kube-proxy to expose metrics


ChrsMark:
`kube-proxy` will not expose metrics endpoint by default with kind setup. We need to manually edit `kube-proxy` configmap and delete the pod so as to have it recreated properly.

This is something that could be improved on `kind`'s side by having an option to enable metrics exposition by default (see https://github.com/kubernetes-sigs/kind/issues/2184), however we might can to tackle this on our side and have the deployer to edit the confgimap and delete the Pod so as to have a new one created with the proper configuration. We can first check if the endpoint is reachable so as to avoid recreating the pod without reason.

@mtojek @ycombinator what do you think about this?

ChrsMark:
Related to https://github.com/elastic/integrations/issues/822

ChrsMark:
We can define how to deploy kind cluster on our side to patch this setting as indicated in kind's issue:

```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  kubeadmConfigPatches:
  - |
    kind: KubeProxyConfiguration
    metricsBindAddress: "0.0.0.0"
```

ycombinator:
This makes sense to me.

ChrsMark:
Closed tuning integrations repo too: https://github.com/elastic/integrations/pull/908

PR #322, opened 2021-04-14
Add custom kind config so as to expose kube proxy metrics

ChrsMark:
We need to start `kind` like this:
```
kind create cluster --config - <<EOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
    kubeadmConfigPatches:
      - |
        kind: KubeProxyConfiguration
        metricsBindAddress: "0.0.0.0"
EOF
```
so as to expose kube-proxy's metrics endpoint and being able to fetch from it in the tests.

We need to tune Jenkins script of `integrations` repo too -> https://github.com/mtojek/integrations/blob/8fd6cef1125454814b24bec6d1a533717af5ac7e/.ci/Jenkinsfile#L204
@mtojek do you think there is any other place where we need to add it?

Related to https://github.com/elastic/elastic-package/issues/315


mtojek (commented):
Speaking of https://github.com/mtojek/integrations/blob/8fd6cef1125454814b24bec6d1a533717af5ac7e/.ci/Jenkinsfile#L204

I was wondering if it's possible to use a remote config, e.g.:

`kind create cluster --config hhttps://raw.githubusercontent.com/elastic/elastic-package/master/scripts/kind-config.yaml`

mtojek:
Maybe don't close the issue yet until this change is propagated to elastic/integrations.

ChrsMark:
nice suggestion @mtojek ! `kind` does not support input directly from http resources but we can pipe the `wget` output to it.

mtojek:
Hmm... something is wrong here:

```
[2021-04-14T09:02:44.433Z] ERROR: failed to create cluster: unknown apiVersion: 
```

ChrsMark:
It will fail now since it refers to a non existent file. `https://raw.githubusercontent.com/elastic/elastic-package/master/scripts/kind-config.yaml` does not actually exist before we merge so `wget` now returns nothing. I was able to reproduce it locally.  I can verify that with an existent file it works: `wget -qO- https://raw.githubusercontent.com/ChrsMark/test/main/kind-cluster.yaml | kind create cluster --config -`

mtojek:
> We can split the PR into 2 steps if we prefer to be sure:

Nah, I think we keep it in a single PR. 

Could you please review docs resources if there is anything that should be improved? https://github.com/elastic/elastic-package/blob/master/docs/howto/system_testing.md

ChrsMark:
> Could you please review docs resources if there is anything that should be improved? https://github.com/elastic/elastic-package/blob/master/docs/howto/system_testing.md

Docs look good and quite representative I think!

scripts/test-check-packages.sh
```diff
@@ -45,7 +45,7 @@ cd -
 elastic-package stack up -d -v
 
 # Boot up the kind cluster
-kind create cluster
+wget -qO-  https://raw.githubusercontent.com/elastic/elastic-package/master/scripts/kind-config.yaml | kind create cluster --config -
```
jsoriano, line 48: Umm, this makes all builds depend on the configuration in master, what will be problematic if we want to introduce and test a new version of the config in a PR and/or run builds with previous code. Is there any reason to don't take the config directly from `scripts/kind-config.yaml`? Something like this:
```suggestion
kind create cluster --config $CURRENT_DIR/kind-config.yaml
```

In the docs it may be ok to leave the reference to master as we expect this to be a working config, and the user might not have a working copy of this repo.

jsoriano: Well, it is actually being already problematic :slightly_smiling_face: https://github.com/elastic/elastic-package/pull/322#issuecomment-819373897

mtojek: In this case we'll have to keep it in both places (elastic/integrations and elastic/elastic-package). The question is which option is less disturbing :)

jsoriano: Well, if `elastic/integrations` also needs this small config, I'd rather have it duplicated there than adding a dependency to a delopment branch :slightly_smiling_face: 

Also the needings in both repos may be different, so they may need in any case a different config at some point.

mtojek: I don't have any strong preferences here. The file resource in both places is also acceptable.

scripts/test-check-packages.sh
```diff
@@ -45,7 +45,7 @@ cd -
 elastic-package stack up -d -v
 
 # Boot up the kind cluster
-kind create cluster
+kind create cluster --config $CURRENT_DIR/kind-config.yaml
```
mtojek, line 48: isn't it `pwd` instead of `CURRENT_DIR`?

mtojek: It looks like the path is still a problem. Could you try some other options like `dirname $0`, or `./scripts/kind-config.yaml`, etc.?

ChrsMark:
It sound be ok now. Thank you both for your reviews!

mtojek (approved):

scripts/test-check-packages.sh
jsoriano: Oops, yes, sorry for not being more explicit, in my suggestion I used `CURRENT_DIR` as a placeholder.

ChrsMark: No problem!

Issue #316, opened 2021-04-08
Kubernetes service deployer: wait for resources


mtojek:
Spotted in: https://github.com/elastic/integrations/pull/886 

The k8s service deployer allows for creating resources using custom YAML definitions. Unfortunately the current implementation lacks of `wait-for-healthy` mechanism. Here is the idea on how we can introduce such check:

1. Grab the output of `kubectl apply`:
    ```kubectl apply -f cluster-role-binding.yaml -f cluster-role.yaml -f deployment.yaml -f service-account.yaml -f service.yaml --output yaml```
2. Read resource names, types and namespaces.
3. Iterate over resources above with:
    ```kubectl get ClusterRoleBinding kube-state-metrics --output yaml``` or ```kubectl get deployment kube-state-metrics -n kube-system --output yaml```
4. If the `status.conditions` is not present, break here - can't verify conditions.
5. Read `status.conditions`, sort by timestamp, check the latest status if it's **"Ready"** or **"Available"**.

I assume we don't have to follow this mechanism with `kubectl delete`. If it's the case, we can introduce it in follow-up PRs.

/cc @ycombinator @ChrsMark

ChrsMark:
Thanks for working on this, it will be super useful!

I think that the idea will work (needs to prove it :)). Some ideas:
1. Maybe you can use [k8s go client](https://github.com/kubernetes/client-go) to interact with the resources and check their status etc.
2. Or you can use `jspath` to retrieve only specific parts of the resource like `kubectl -n kube-system get pod elastic-agent-8jqf6 -o jsonpath='{.status.conditions}'` which returns ```[{"lastProbeTime":null,"lastTransitionTime":"2021-04-08T08:03:54Z","status":"True","type":"Initialized"},{"lastProbeTime":null,"lastTransitionTime":"2021-04-08T08:03:56Z","status":"True","type":"Ready"},{"lastProbeTime":null,"lastTransitionTime":"2021-04-08T08:03:56Z","status":"True","type":"ContainersReady"},{"lastProbeTime":null,"lastTransitionTime":"2021-04-08T08:03:54Z","status":"True","type":"PodScheduled"}]```. I remember sth similar was useful to check which nodes are ready (see https://kubernetes.io/docs/reference/kubectl/cheatsheet/)

mtojek:
Finally I found some time to look into this.

PR #323, opened 2021-04-14
Kubernetes service deployer: Wait for ready resources

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/316


mtojek:
@ChrsMark There is another possible implementation: https://github.com/elastic/elastic-package/pull/325  , reusing Helm internals for resource handling. Let's see what CI thinks about it.

ChrsMark (approved):
Looks good to me! 
Tested locally with k8s package test and works great. I think I prefer more the Helm one since it will give us more flexibility in the future if we need something extra.

PR #326, opened 2021-04-19
Add services to cluster role so as Agent to be able to watch k8s svc

ChrsMark:
While working on https://github.com/elastic/integrations/pull/886 I hit an issue of not being able to ingest data for `state_service` metricset. The reason is that this metricset tries to add metadata of the services and for this Agent needs acces s to the respective api object.


mtojek (approved):
LGTM

mtojek:
I think you can create an issue to unify YAML definitions, so that both parties can reuse it.

PR #329, opened 2021-04-20
Use elastic-agent status as healthcheck

mtojek:
This PR modifies the healthcheck for elastic-agent and ~~fleet-server~~, so the correct one is used.

Blockers:
-  https://github.com/elastic/beats/issues/24956


mtojek:
@ruflin We managed to switch to latest Docker image snapshots. Now it's the time to enable the right healthcheck. What is the healthcheck recommendation for the fleet-server?

ruflin:
@michalpristas Instead of using `elastic-agent status`, could we also just start the agent monitoring, expose it and check there?

michalpristas:
depends on the use case, if we want to wait for agent to be healthy and apps green, status is the right approach. also if agent is not able to connect to fleet it will return unhealthy status...
if we want to check for agent capability to run server we can go that way, but this wont reflect any failures or restarts of agent and its apps.

ruflin:
I might have found a workaround here. If you add the following to the agent env variables it seems to work: 

```
- "STATE_PATH=/usr/share/elastic-agent"
```

It overwrites the default state path which is `/usr/share/elastic-agent/state`. I'm not sure about the other things this might break. The end solution should be that elastic-agent status respects the STATE_PATH set.

@mtojek Could you try if you get it running with the above env variable?

mtojek:
@ruflin it seems to work, I'll retry the CI job to double-check. Do you think we should go with this workaround or is it better to wait for the exact fix? So far, the default Docker image is broken in terms of this feature.

internal/install/static_snapshot_yml.go
```diff
@@ -102,14 +102,15 @@ services:
...
-      test: "sh -c 'grep \"Agent is starting\" -r . --include=elastic-agent-json.log'"
+      test: "elastic-agent status"
       retries: 90
       interval: 1s
     hostname: docker-fleet-agent
     environment:
     - "FLEET_ENROLL=1"
     - "FLEET_INSECURE=1"
     - "FLEET_URL=http://fleet-server:8220"
+    - "STATE_PATH=/usr/share/elastic-agent"
```
ruflin, line 113: Should we change it for both fleet-server and elastic-agent?

ruflin (commented):
I would like to get some feedback on @blakerouse on this before merging to make sure it does not have some unexpected side effects.

@mtojek It seems you just "run" the command, shouldn't there also be some check for the output itself?

mtojek:
> @mtojek It seems you just "run" the command, shouldn't there also be some check for the output itself?

I expect the "exit code" to be the source of truth.

internal/install/static_snapshot_yml.go
mtojek: I can try and change it, but previously it didn't work well (maybe it was the STATE_PATH issue). I will re-check it :)

mtojek: I updated the fleet-server to use "elastic-package status". Let's see.

blakerouse: Why are you setting the `STATE_PATH`? The `/usr/share/elastic-agent/state` is the default, why not use that? Is this a workaround for the `status` command not working.

ruflin: @blakerouse Yes, otherwise the status command does not work as it does it looks into the wrong path.

mtojek:
I will re-test it once again to see if it works correctly.

@blakerouse would you mind sharing the recommendation for merging this change?

mtojek:
Ok, something is unstable here: https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-329/9/pipeline/

I'm waiting for the timeout to grab other logs. Not sure if it's related to latest master or these changes.

mtojek:
The error is not related to this PR. Something is wrong with the latest snapshot.

mtojek:
It seems that the STATE_PATH did the trick for fleet-server as well!

mtojek:
Nope, flaky again:

```
[36melastic-agent_1              |[0m The Elastic Agent is currently in BETA and should not be used in production
[36melastic-agent_1              |[0m 
[36melastic-agent_1              |[0m Error: fail to enroll: fail to execute request to Kibana: could not decode the response, raw response: 
[36melastic-agent_1              |[0m 
[36melastic-agent_1              |[0m Error: enrollment failed: exit status 1
```

Should I switch to the previous healthcheck for the fleet-server?

ruflin:
This does not really look related to the health status. I wonder if we hit a timeout? Can you reproduce it locally?

mtojek:
I'm trying to reproduce it locally, but it always succeeds. Must be a really weird edge case. Call for improving logging around this place?

For reference:

```
untilfail ./k.sh
```

k.sh:
```
cat k.sh
#!/bin/bash

set -ex

elastic-package stack down
```

mtojek:
@ruflin @blakerouse any suggestions on how to solve this issue? Seems that the healthcheck is not ideal for the fleet server. Should I open a separate issue for this?

ruflin:
@michalpristas What would you recommend for the health check for the fleet-server. The status command or the API endpoint?

@mtojek It seems now all checks have passed? Is it still flaky?

mtojek:
> @mtojek It seems now all checks have passed? Is it still flaky?

Yes, I'm afraid so. With "api/status" it was the same problem, but I introduced a workaround - increased interval between next trials (higher chance to encounter the healthy state). I suspect it's related with restart of the agent.

ruflin (approved):
LGTM. I think we still need to work on "what is our recommended way" for docker.

Issue #3, opened 2020-06-05
Command: create new integration


mtojek:
The command bootstraps the structure for a new integration using a wizard or command line arguments.

Suggested command: `elastic-package create`

Subcommands:

```
create package
create data-stream
```

mtojek:
I'm going with this one. We need it to close the development workflow (the last missing piece).

exekias:
This is great 🤟, do you have a list of the subcommands that you plan to add? I see `package` is in progress already, I guess others will come

mtojek:
I updated the issue description for visibility. MVP includes "package" and "data-stream".

exekias:
SGTM! Other obvious assets that most devs will hit are pipelines & dashboards, I think it's fine to add them later, happy to create issues to discuss them

mtojek:
Yeah, we can research it later. Speaking of dashboards, there is the "export" command that helps you pull them from Kibana to local files.

mtojek:
I'm going with the `create data-stream` command.

mtojek:
Implementation has been merged, I'll write HOWTO doc page describing the procedure.

PR #331, opened 2021-04-20
Add "create package" command

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/3

This PR introduces a new "create" command,  which can bootstrap a new package.

TODOs:
- [x] CLI for new package
- [x] CLI: validate input parameters for new package
- [x] Flush generic package
- [x] Try to create package that already exists


mtojek:
I spent some time with `expect` (Tcl) to write a script covering all prompts, but failed due to coloring/formatting (control characters. There should be a way to cover them with https://github.com/Netflix/go-expect, but I'm not sure if we want to use unit tests for that.

I wouldn't like to end up with untested sources like we have in `mage ImportBeats`, but I'm wondering on which level should we test it.

Possible options:
1. Expose prompt values as command line options and test with Bash script.
2. Use go-expect to simulate prompt responses (go unit tests).
3. Don't test prompts and validators, just test created package directory with `check` (linting + format).

ycombinator:
@mtojek Regarding testing, I think we can actually test the expected functionality without needing something like `expect`. 

Looking at how you have implemented this, there is some nice division of responsibilities between `survey`, `surveyext`, and `archetype` packages. I think we can use this to our advantage for testing.

First off, we can put aside testing the functionality of the `survey` package itself as it is provided by a library. This implies that we can put aside testing the prompting (questions => answers) functionality itself. This would rule out options 1 and 2.

That leaves functionality offered by the `surveyext` and `archetype`  packages. I think we can write unit tests to cover their functionality. This is basically option 3.

exekias:
This is looking awesome! Any thoughts/plans on the survey questions with a single answer? I'm wondering about `package`, & `license` mainly

mtojek:
> This is looking awesome! Any thoughts/plans on the survey questions with a single answer? I'm wondering about package, & license mainly

I left them in that form to show that we don't provide/support other options and wouldn't like to silently assume a specific kind (visibility, awareness). I'm happy to change it if you suspect that the other way is better/more intuitive.

exekias:
> I left them in that form to show that we don't provide/support other options and wouldn't like to silently assume a specific kind (visibility, awareness). I'm happy to change it if you suspect that the other way is better/more intuitive.

No strong opinions here:

About package type: I think we should focus `elastic-package` to serve integration devs. Right now the two package types are `integration` and `solution`. My guess is that adding the `solution` option in the survey would probably cause confusion among integrations devs, while adding little value to the few solution packages.

I could see how other types in the future (maybe link?) could change this, though. We can always revisit this and add the question later.


About license: I'm more on the fence here, as I think it's good to show what will be the license for the package. In any case not having an alternative (and probably we don't need to open that door yet) may be confusing.

mtojek:
> Right now the two package types are `integration` and `solution`

There is one package type - integration. For reference: https://github.com/elastic/package-spec/blob/master/versions/1/manifest.spec.yml#L143

mtojek:
@ycombinator I added mentioned unit tests. I think this one is ready for review.

exekias:
> There is one package type - integration. For reference: https://github.com/elastic/package-spec/blob/master/versions/1/manifest.spec.yml#L143

You are right, I remember `solution` was a thing but things have changed. Then again, why asking the user when there is only one type of package? We can always revisit the survey when that changes

exekias (approved):

cmd/create.go
```diff
@@ -0,0 +1,168 @@
...
+			Validate: survey.Required,
+		},
+		{
+			Name: "categories",
+			Prompt: &survey.MultiSelect{
+				Message: "Categories:",
+				Options: []string{"aws", "azure", "cloud", "config_management", "containers", "crm", "custom",
+					"datastore", "elastic_stack", "google_cloud", "kubernetes", "languages", "message_queue",
+					"monitoring", "network", "notification", "os_system", "productivity", "security", "support",
+					"ticketing", "version_control", "web"},
```
ycombinator, line 98: It would be nice if some day in the future we can expose these via a golang API from the package-spec.

cmd/create.go
```diff
@@ -0,0 +1,168 @@
...
+				Default:  []string{"custom"},
+				PageSize: 50,
+			},
+			Validate: survey.Required,
+		},
+		{
+			Name: "release",
+			Prompt: &survey.Select{
+				Message: "Release:",
+				Options: []string{"experimental", "beta", "ga"},
```
ycombinator, line 108: It would be nice if some day in the future we can expose these via a golang API from the package-spec.

internal/packages/archetype/package.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	logger.Debugf("Write package changelog")
+	err = renderResourceFile(packageChangelogTemplate, &packageDescriptor, filepath.Join(baseDir, "changelog.yml"))
+	if err != nil {
+		return errors.Wrap(err, "can't render package changelog")
+	}
+
+	logger.Debugf("Write docs readme")
+	err = renderResourceFile(packageDocsReadme, &packageDescriptor, filepath.Join(baseDir, "docs", "README.md"))
```
ycombinator, line 49: Does it make sense to write the `docs/README.md` file or should we write the `_dev/build/docs/README.md` file instead... or both?

internal/surveyext/validators.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"os"
+	"regexp"
+
+	"github.com/Masterminds/semver"
+	"github.com/pkg/errors"
+)
+
+var (
+	githubOwnerRegexp = regexp.MustCompile("^(([a-zA-Z0-9-]+)|([a-zA-Z0-9-]+\\/[a-zA-Z0-9-]+))$")
```
ycombinator, line 17: Short term: it would be nice to copy this regex into https://github.com/elastic/package-spec/blob/92ee6e42f2e89456ae4ef5f66531abbc4a23ead4/versions/1/manifest.spec.yml#L270-L272.

Longer term: it would nice to expose this regex via a golang API from the package-spec. That way the package-spec can be the single source of truth for this regex.

ycombinator (commented):
This looks great!

Left one question about the README generation. 

Other than that, left some suggestions about future improvements involving the package-spec. If you agree with them, make an issue for them in the package-spec repo?

internal/packages/archetype/package_docs_readme.go
```diff
@@ -0,0 +1,11 @@
...
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package archetype
+
+const packageDocsReadme = `# {{.Manifest.Title}}
+
+This is a new integration created using the [elastic-package](https://github.com/elastic/elastic-package) tool.
+
+Consider using the README template file ` + "`_dev/docs/README.md`" + `to generate a list of exported fields or include a sample event.`
```
ycombinator, line 11:
```suggestion
Consider using the README template file ` + "`_dev/build/docs/README.md`" + `to generate a list of exported fields or include a sample event.`
```

internal/packages/archetype/package.go
mtojek: Hm... it would mean that there is a template without any placeholders, which in this case will the same as outputted one. Not sure if it doesn't introduce confusion here.

ycombinator: I see. Okay, lets leave this as-is. In any case the template file is not strictly required according to the spec but the (generated) README.md file is. So I think what you have here is good for the purposes of getting started with a new package.

ycombinator (approved):
LGTM.

Please make an issue in the package-spec repo for the future "single source of truth" improvements suggested in this PR.

internal/surveyext/validators.go
mtojek: https://github.com/elastic/package-spec/issues/163

cmd/create.go
mtojek: https://github.com/elastic/package-spec/issues/163

internal/surveyext/validators.go
mtojek: https://github.com/elastic/package-spec/issues/163

PR #332, opened 2021-04-20
Deploy agent with hostNetwork on k8s

ChrsMark:
We need to deploy Agent using `hostNetwork` option so as to have access to node's ports directly for testing endpoints like these of scheduler's and controllermanager.

### DIFF
--- a/internal/testrunner/runners/system/servicedeployer/kubernetes.go
+++ b/internal/testrunner/runners/system/servicedeployer/kubernetes.go
@@ -95,6 +95,8 @@ func (ksd KubernetesServiceDeployer) SetUp(ctxt ServiceContext) (DeployedService
 
 	ctxt.Name = kind.ControlPlaneContainerName
 	ctxt.Hostname = kind.ControlPlaneContainerName
+	// kind-control-plane is the name of the kind host where Pod is running since we use hostNetwork setting
+	// to deploy Agent Pod. Because of this, hostname inside pod will be equal to the name of the k8s host.
 	ctxt.Agent.Host.NamePrefix = "kind-control-plane"
 	return &kubernetesDeployedService{
 		ctxt:           ctxt,
--- a/internal/testrunner/runners/system/servicedeployer/kubernetes.go
+++ b/internal/testrunner/runners/system/servicedeployer/kubernetes.go
@@ -95,7 +95,7 @@ func (ksd KubernetesServiceDeployer) SetUp(ctxt ServiceContext) (DeployedService
 
 	ctxt.Name = kind.ControlPlaneContainerName
 	ctxt.Hostname = kind.ControlPlaneContainerName
-	ctxt.Agent.Host.NamePrefix = "kind-fleet-agent-"
+	ctxt.Agent.Host.NamePrefix = "kind-control-plane"
 	return &kubernetesDeployedService{
 		ctxt:           ctxt,
 		definitionsDir: ksd.definitionsDir,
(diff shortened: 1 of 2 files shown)


mtojek (approved):
WFG

mtojek:
BTW To verify features introduced in PRs you can use the `test/packages` stuff to quickly write up samples. In this case you could verify the connectivity (not required, just a hint).

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -95,7 +95,7 @@ func (ksd KubernetesServiceDeployer) SetUp(ctxt ServiceContext) (DeployedService
 
 	ctxt.Name = kind.ControlPlaneContainerName
 	ctxt.Hostname = kind.ControlPlaneContainerName
-	ctxt.Agent.Host.NamePrefix = "kind-fleet-agent-"
+	ctxt.Agent.Host.NamePrefix = "kind-control-plane"
```
mtojek, line 98: nit: missing trailing dash or is it on purpose?

I'm not sure about implications of this change. Would you mind sharing more context? Will it affect any existing integrations

mtojek, line 98: I looked into the kubectl dump for the failed build: https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-332/1/artifact/build/kubectl-dump.txt

and I see that there is a mix of `kind-control-plane` and `kind-fleet-agent-clusterscope`. I admit I'm bit lost here.

mtojek (changes requested):
I will block this PR from merging, so that we're sure that we won't merge anything breaking here.

ChrsMark:
Deploying Agent using `hostNetwork` will result into a different hostname. Pod will fallback to node's name for the hostname. This will lead into having Agent registered as `kind-control-plane` (name of the host) in fleet server while the test will be waiting for `kind-fleet-agent-*` and it will stall there forever (==timeout).

mtojek:
Would you mind adding a comment in code around NodePrefix, justifying the "kind-control-plane"? I'm a bit cautious while dealing with host network and potential implications, but if there is no other option, let's do this. I undestand that it will only impact the kubernetes integration. Do you think that it won't be a problem in the future? Let's say, we would like to build the integration with istio.

mtojek (approved):
Not blocking, thank you for the explanation.

ChrsMark:
> Would you mind adding a comment in code around NodePrefix, justifying the "kind-control-plane"? I'm a bit cautious while dealing with host network and potential implications, but if there is no other option, let's do this. I undestand that it will only impact the kubernetes integration. Do you think that it won't be a problem in the future? Let's say, we would like to build the integration with istio.

We use `hostNetwork` in our actual manifests too so it shouldn't be a problem for this testing scenario since it is actually identical to the real world's case:
https://github.com/elastic/beats/blob/188345dc96876011fd397ad7fd2f721459e102d7/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml#L457
https://github.com/elastic/beats/blob/188345dc96876011fd397ad7fd2f721459e102d7/deploy/kubernetes/metricbeat-kubernetes.yaml#L146

I cannot think of any specific issue we can hit in the future because of this. Other k8s services/apps will be available using DNS resolution of k8s (using k8s service's name like what we do for kube-state-metrics). `hostNetwork` is also useful for autodiscovery (out of package's scope) so as to enable a k8s api watcher on node's level. 

In general, I think we are on the safe side with this ("hold my beer" 😄 ).

mtojek:
@ChrsMark I think this one is good to go now.

PR #333, opened 2021-04-21
Fix: skip caching images

mtojek:
This PR ensures that latest stack images are used.


ruflin (approved):

mtojek:
I will merge this temporarily to unblock the pipeline.

PR #335, opened 2021-04-23
Use fallback GeoDB in Elasticsearch

mtojek:
This PR disables the GeoDB downloader in Elasticsearch.

It's a similar PR to the one introduced in Beats: https://github.com/elastic/beats/pull/25248


internal/install/static_snapshot_yml.go
```diff
@@ -26,6 +26,7 @@ services:
     - "script.context.ingest.cache_max_size=2000"
     - "script.context.processor_conditional.cache_max_size=2000"
     - "script.context.template.cache_max_size=2000"
+    - "ingest.geoip.downloader.enabled=false"
```
andrewkroh, line 29: Do we test with 7.12.0 anywhere? I assume this property doesn't exist in that version and this would cause a failure.

mtojek: The 7.12.0 is unsupported since we introduced changes for the Fleet Server. Unfortunately it was an acknowledged risk.

P1llus (approved):
LGTM, I saw andrew's comment was already accepted.

PR #336, opened 2021-04-26
Bump elastic stack version automation

v1v:
## What does this PR do?

Script to bump the versions:

```diff
$ ./.ci/bump-stack-version.sh 8.0.0-3beeb19e false                                                
Update stack with version 8.0.0-3beeb19e
Commit changes
Branch creation disabled.
$ git --no-pager diff                                    
diff --git a/.ci/bump-stack-version.sh b/.ci/bump-stack-version.sh
index 6402105..0b0ec68 100755
--- a/.ci/bump-stack-version.sh
+++ b/.ci/bump-stack-version.sh
@@ -33,6 +33,7 @@ if [ "$CREATE_BRANCH" = "true" ]; then
 else
 	echo "Branch creation disabled."
 fi
+exit 0
 git add internal/install/stack_version.go
 git diff --staged --quiet || git commit -m "bump stack version ${VERSION}"
 git --no-pager log -1
diff --git a/internal/install/stack_version.go b/internal/install/stack_version.go
index 3195512..74b855b 100644
--- a/internal/install/stack_version.go
+++ b/internal/install/stack_version.go
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "8.0.0-SNAPSHOT"
+	DefaultStackVersion = "8.0.0-3beeb19e-SNAPSHOT"
 )
```

This PR is one part of the required changes to enable the automation. 

## Why is it important?

Handle the stack versions on a PR basis to ensure breaking changes are handled correctly. Unfortunately we cannot use `dependabot` since those versions are handled differently.

## Issues

Similar to https://github.com/elastic/package-registry/pull/691

## Questions

- What's the backport policy for the `master` branch?

## Follow ups

- Enable automerge with `mergify`?
- This project uses a different branch strategy so there is a requirement to match the branches with the elastic stack branch strategy (7.x, 7.<minor>, master=8.0.0...)

### DIFF
--- a/.ci/bump-stack-version.sh
+++ b/.ci/bump-stack-version.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+#
+# Given the stack version this script will bump the version.
+#
+# This script is executed by the automation we are putting in place
+# and it requires the git add/commit commands.
+#
+# Parameters:
+#	$1 -> the version to be bumped. Mandatory.
+#	$2 -> whether to create a branch where to commit the changes to.
+#		  this is required when reusing an existing Pull Request.
+#		  Optional. Default true.
+#
+set -euo pipefail
+MSG="parameter missing."
+VERSION=${1:?$MSG}
+CREATE_BRANCH=${2:-true}
+
+OS=$(uname -s| tr '[:upper:]' '[:lower:]')
+
+if [ "${OS}" == "darwin" ] ; then
+	SED="sed -i .bck"
+else
+	SED="sed -i"
+fi
+
+echo "Update stack with version ${VERSION}"
+${SED} -E -e "s#(DefaultStackVersion) = \"[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1 = \"${VERSION}#g" internal/install/stack_version.go
+
+echo "Commit changes"
+if [ "$CREATE_BRANCH" = "true" ]; then
+	git checkout -b "update-stack-version-$(date "+%Y%m%d%H%M%S")"
+else
+	echo "Branch creation disabled."
+fi
+git add internal/install/stack_version.go
+git diff --staged --quiet || git commit -m "bump stack version ${VERSION}"
+git --no-pager log -1
+
+echo "You can now push and create a Pull Request"
--- a/internal/install/stack_version.go
+++ b/internal/install/stack_version.go
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "7.13.0-SNAPSHOT"
+	DefaultStackVersion = "8.0.0-SNAPSHOT"
 )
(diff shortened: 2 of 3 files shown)


internal/install/stack_version.go
```diff
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "7.13.0-SNAPSHOT"
+	DefaultStackVersion = "8.0.0-SNAPSHOT"
```
mtojek, line 9: We prefer to stick to the latest stable version, so still 7.13.0.

mtojek (commented):
You need to revert also this command: https://github.com/elastic/elastic-package/pull/333

internal/install/stack_version.go
v1v:
```suggestion
	DefaultStackVersion = "7.13.0-SNAPSHOT"
```

v1v:
> You need to revert also this command: #333

I don't understand, what's required?

mtojek (changes requested):
Ok, I think I understand now the goal of this PR, but I suppose you need to figure out a different procedure.

The intention behind the flag `DefaultStackVersion` is different. It is just a reference to the generic version tag to select a release. Users who fetched this tool once, will update local stacks with `elastic-package stack update` instead of refetching and rebuilding the source code.

I prefer to keep the 7.13.0-SNAPSHOT (or soon 7.14.0-SNAPSHOT) as this line is more stable that 8.0.0-SNAPSHOT.

It would be better to update Docker image references on the CI host, so if the tool wants to use 7.13.0-SNAPSHOT, it will look up for the correct references. 

EDIT:

Regarding 7.13.0 vs 8.0.0, I think the same policy should be applied to all repos.

mtojek:
> I don't understand, what's required?

That PR modifies Jenkins behavior to ALWAYS refetch latest Docker image from registry. It was introduced to solve issues with cached images (incompatible between each other).

.ci/bump-stack-version.sh
```diff
@@ -0,0 +1,40 @@
...
+OS=$(uname -s| tr '[:upper:]' '[:lower:]')
+
+if [ "${OS}" == "darwin" ] ; then
+	SED="sed -i .bck"
+else
+	SED="sed -i"
+fi
+
+echo "Update stack with version ${VERSION}"
+${SED} -E -e "s#(DefaultStackVersion) = \"[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1 = \"${VERSION}#g" internal/install/stack_version.go
```
v1v, line 28: What's the file with the stack version used in the tests? If any. Otherwise we don't need this automation

mtojek: It's always the current one tagged with DefaultStackVersion, e.g. 7.13.0-SNAPSHOT.

mtojek: @v1v 

Bumping up the PR, what's the decision/plan here? Would you like to discuss it first?

mtojek:
It's a bit stale. Let's close it for now.

PR #337, opened 2021-04-27
Revert: use api/status as healthcheck for fleet-server

mtojek:
This PR reverts changes introduced around fleet-server's healthcheck. Unfortunately it doesn't work well, so we have to switch back to the api/status.

### DIFF
--- a/internal/install/static_snapshot_yml.go
+++ b/internal/install/static_snapshot_yml.go
@@ -84,9 +84,9 @@ services:
       kibana:
         condition: service_healthy
     healthcheck:
-      test: "elastic-agent status"
-      retries: 60
-      interval: 1s
+      test: "curl -f http://127.0.0.1:8220/api/status | grep HEALTHY 2>&1 >/dev/null"
+      retries: 12
+      interval: 5s
     hostname: docker-fleet-server
     environment:
     - "FLEET_SERVER_ENABLE=1"


internal/install/static_snapshot_yml.go
```diff
@@ -84,9 +84,9 @@ services:
       kibana:
         condition: service_healthy
     healthcheck:
-      test: "elastic-agent status"
-      retries: 60
-      interval: 1s
+      test: "curl -f http://127.0.0.1:8220/api/status | grep HEALTHY 2>&1 >/dev/null"
+      retries: 12
+      interval: 5s
```
ycombinator, line 89: Just curious, is`/api/status` stable? If so, then I'm wondering why we switched to `elastic-agent status` in the first place?

mtojek: There was a belief that `elastic-agent status` is sufficient, but it seems that it doesn't. Quote from @blakerouse :

> probably need to add some type of stable checking to ensure that the status reported is stable and not always changing

mtojek:
It's an emergency fix, so I will merge it.

internal/install/static_snapshot_yml.go
mtojek: Yes, I'm aware of this. This is the reason why checks are not so frequent (every 5s instead of standard 1s).

PR #338, opened 2021-04-27
[Testing] Test if fleet-server policy by id is picked up

ruflin:


mtojek:
@ruflin I think you need to add more context, what's going here :D

ruflin:
I needed it to test https://github.com/elastic/beats/pull/25352 but pushed it as a PR to potentially discuss if this could be useful too.

ruflin:
I'm going to close this as we don't need this exactly in this way for elastic-package.

@mtojek The part that is interesting about the change I have in this PR is that it is possible to preconfigure a base setup without having to do many API calls. This might simplify the elastic-package setup potentially.

mtojek:
@ruflin 

When the [profiles](https://github.com/elastic/elastic-package/pull/301) feature lands in, user will be able to modify the stack setup and customize it based on needs (open ports, config options, etc.)

PR #340, opened 2021-04-28
Update connect_service_with_elastic_stack.md

ChrsMark:


docs/howto/connect_service_with_elastic_stack.md
```diff
@@ -100,6 +100,10 @@ access.log  error.log  httpd.pid
 When you need to specify the service address, keep in mind that the Elastic stack and the service operate in Docker networks,
 so they don't communicate over the `localhost`, but use dedicated domains instead, e.g.: `elastic-package-service_apache_1:80`.
 
+6. Define the service's log path in the Kibana Fleet UI.
+
+When you need to specify the log path, use the path as it is mounted in Agent's container for example `/tmp/service_logs/access.log`.
```
mtojek, line 105: nit: comma

Agent's container, for example

docs/howto/connect_service_with_elastic_stack.md
```diff
@@ -100,6 +100,10 @@ access.log  error.log  httpd.pid
 When you need to specify the service address, keep in mind that the Elastic stack and the service operate in Docker networks,
 so they don't communicate over the `localhost`, but use dedicated domains instead, e.g.: `elastic-package-service_apache_1:80`.
 
+6. Define the service's log path in the Kibana Fleet UI.
```
mtojek, line 103: I would mention here that this step is "optional" as it affects only logs based data streams. If you don't plan to test any logs based data streams, you don't need to do it.

ChrsMark: I think it is obvious that it has to do with logs related datastreams. Same goes for point 5, it would optional for someone that only wants to test logs collection.

mtojek (approved):

Issue #344, opened 2021-04-29
Remove replace directive for helm@v3.5.1 dependency


ycombinator:
In https://github.com/elastic/elastic-package/pull/345 we made `elastic-package` explicitly depend on the `helm@v3.5.1` module. This was a stop-gap measure until https://github.com/helm/helm/pull/9492 was resolved.

Once https://github.com/helm/helm/pull/9492 is resolved, we should remove the `replace` directive for the `helm@v3.5.1` module from `go.mod`.

Related: https://github.com/elastic/elastic-package/issues/342

mtojek:
I think I fixed the issue:

```
$ go version
go version go1.14.2 darwin/amd64
$ GO111MODULE=on go get github.com/elastic/elastic-package@master
go: github.com/elastic/elastic-package master => v0.0.0-20210430085756-cb436d9f36d9
$ elastic-package version
elastic-package version-hash undefined (build time: unknown)
```

PR #346, opened 2021-04-30
Try: use helm.sh/helm/v3 v3.5.1

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/344

I'm not sure, but based on @ycombinator 's observations maybe it's possible to go with helm v3.5.1.


mtojek:
self-approved, emergency fix

PR #348, opened 2021-05-05
Bump up the stack to 7.14.0-SNAPSHOT

mtojek:
This PR bumps up the stack version to 7.14.0-SNAPSHOT.

### DIFF
--- a/internal/install/stack_version.go
+++ b/internal/install/stack_version.go
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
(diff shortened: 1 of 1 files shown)


ycombinator:
Should we wait to merge this until after 7.13.0 is released? I think that's typically when we bump the snapshot versions, no?

mtojek:
As we don't need any snapshot features, I think we can wait until release (converted to draft). Anyway it's good to spot potential bugs earlier.

mtojek:
self-approved

PR #356, opened 2021-05-13
Add fleet-server to available services

ycombinator:
This will allow users to start up fleet server individually via the `--services` CLI flag.

### DIFF
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -20,9 +20,9 @@ import (
 
 var availableServices = map[string]struct{}{
 	"elasticsearch":    {},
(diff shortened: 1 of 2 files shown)


cmd/stack.go
```diff
@@ -22,6 +22,7 @@ var availableServices = map[string]struct{}{
 	"elasticsearch":    {},
 	"kibana":           {},
 	"package-registry": {},
+	"fleet-server":     {},
```
mtojek, line 25: nit: alphanum order

mtojek (approved):

Issue #305, opened 2021-03-31
Proposal: Limit promote command to one package


ruflin:
Currently with `elastic-package promote` it is possible to promote multiple packages. This was needed in the past where we bulk promoted packages. But we should switch over to a mode where each package developer / owner makes sure their packages are promote and tested. Because of this I propose that promote has to be run for a specific package and make it default to have it run for each package.

I can foresee cases where the bulk might still become useful in the future. Perhaps we can add a flag `--all` to the promote command to keep this functionality if we need it.

ycombinator:
I like this idea. To put it differently, in terms of `elastic-package` terminology, `promote` currently has "global" scope; now we would want it to be "package" scope.

One way we could achieve this is by checking if we are in a package context or not. Being in a package context means that the user is somewhere inside a package folder's hierarchy.

@ruflin Just curious: what is the use case driving this change?

ruflin:
This was triggered by PR's like these: https://github.com/elastic/package-storage/pull/1079 https://github.com/elastic/package-storage/pull/1078 The is impossible to tell from the title if one or many packages were changed. Also I'm worried that potentially packages are "accidentially" promote if it is not done one by one. Last, it is also about ownership to make sure everyone handles their packages. If we have a proper package publish API one day, I doubt bulk would exist.

ycombinator:
Thanks @ruflin. Interestingly @mtojek recently created [`elastic-package publish`](https://github.com/elastic/elastic-package/pull/292) which does the initial publishing of a package to the `snapshot` branch... and this command runs in a ["package" context](https://github.com/elastic/elastic-package#elastic-package-publish)! So I'm even more ++ to aligning `elastic-package promote` to also require "package" context.

@mtojek WDYT?

mtojek:
I can give more context for the bulk option - in the early days it was common to adjust multiples packages at once when we're introducing new fields or modifying the structure. This option is also used to promote packages parked in staging after the testing phase.

I'm ok with overall idea, but we need to focus on the CLI/UX. Keep in mind that `elastic-promote` doesn't operate on packages in Integrations, but built ones in package-storage (not all packages are in the integrations repo). I'm looking for CLI proposals :)

ruflin:
My CLI proposal would be:

```
elastic-package promote {package-name+version} {target}
```

An example here would be:

```
elastic-package promote apache-1.3.2 staging
```

[...]

PR #365, opened 2021-05-24
Add CLI flags to "promote" command

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/305

This PR introduces alternative flags to "promote" commands, so it's not required to use TUI.

Sample usage:

```
elastic-package promote  --direction snapshot-production --newest-only --packages redis-0.3.7,windows-0.8.0
```


mtojek:
self-approved

PR #367, opened 2021-05-26
Start service without the test runner 

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/190

This PR introduces the `service` command which allows for booting up the service without the test runner.


mtojek:
self-approved

Issue #181, opened 2020-11-23
Add units and metric type to fields docs


exekias:
`elastic-package` build is able to get the docs template and create the final package `README.md`.

One of the things it does is injecting `fields` blocks for each data stream. With https://github.com/elastic/package-spec/issues/42 in, it would be good to add a column for Units in all cases, and a column for Metric type for the data streams of type `metrics`.

Example PR adding these fields: https://github.com/elastic/integrations/pull/402

mtojek:
Let's try to move this forward.

PR #368, opened 2021-05-28
Render units and metric type in exported fields table

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/181

This PR modifies the builder code to render exported fields table with units and metric types (if present).


mtojek:
self-approved

Issue #289, opened 2021-03-15
Allow for default *-config.yml on pipeline tests


P1llus:
There is plenty of pipelines that has a very large amount of logfiles that they want to test with, one example is cisco and panw.

The current situation is that for pipeline tests it seems that it will only allow for configurations based on a specific config name pattern:
`LOGFILENAME-config.yml`.

Which means that we end up with the same config copied 5-10 times.

It would be great when during pipeline tests, that if it cannot find a config file for the specific logfile, it would default to trying to find something like `test-default-config.yml`, that way we could remove most of the testing config files we have.

mtojek:
Thanks for filing the issue, it's an interesting side effect. We may actually introduce a default test config which can be picked up by all pipeline test cases without specific config files. @ycombinator any opinions here?

P1llus:
While it might introduce more work, it would be nice to have a hierarchy of configurations.

Global config (with config items like adding `event.ingested` to dynamic fields, which is needed pretty much everywhere)
Package wide config, with `test-default-config.yml`.
Data-stream config, with `LOGFILENAME-config.yml` (which we have already).

ycombinator:
I'm generally in favor of this idea but I'd only start by implementing as much of it as we need at the moment with room to grow in the future. I'd also clarify exactly the hierarchy of config inheritance, especially since we have different _types_ of tests (pipeline, system, etc.).

@P1llus correct me if I'm wrong but sounds like what you are looking for _at the moment_ is a single config that defines common options for multiple pipeline test case files in a data stream? So something like this:

```
<package root>
  data_stream/
    foo/
      _dev/test/pipeline/
        LOGFILENAME_1-config.yml
        LOGFILENAME_2-config.yml
        ...
       common-config.yml    // ignore the exact naming, just want to illustrate the idea
    bar/
      _dev/test/pipeline/
        LOGFILENAME_1-config.yml
        LOGFILENAME_2-config.yml
        ...
       common-config.yml    // ignore the exact naming, just want to illustrate the idea
```

So, in the above example, settings from `<package root>/data_stream/foo/_dev/test/pipeline/common-config.yml` would be inherited by all its sibling `LOGFILENAME_*-config.yml` files. Similarly, settings from `<package root>/data_stream/bar/_dev/test/pipeline/common-config.yml` would be inherited by all its sibling `LOGFILENAME_*-config.yml`, correct?

[...]

PR #369, opened 2021-05-31
Use common test config in pipeline tests

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/289

This PR modifies the logic behind pipeline tests runner to use common config file if it exists


P1llus (approved):
LGTM, this is going to help a lot!

PR #372, opened 2021-06-02
Emit json instead of $json in README files

### LINKED ISSUE #371: Replace `$json` with `json` when generating package READMEs
See https://github.com/elastic/kibana/issues/101168 for issue details. There is a fix in Kibana to convert the incorrect `$json` language flag to `json` to avoid having to re-release all older package versions (https://github.com/elastic/kibana/pull/101187), but we should fix it in the generator going forward:

https://github.com/elastic/elastic-package/blob/3ebb7c474f589c98412960c656402dc4e66a010f/internal/docs/sample_event.go#L36

ycombinator:
This PR updates the README generation code to emit `json` as the language for sample event snippets. Previously, `$json` was being emitted but that no longer works with newer versions of Kibana. 

Note that the markdown renderer in older versions of Kibana support both `$json` and `json`, so it's safe to simply emit `json` across the board without worrying about which version of Kibana will be rendering a package's README.

Resolves #371.


jen-huang (approved):

mtojek (approved):
LGTM

ycombinator:
CI is failing because the apache status system test is failing like so:

```
one or more errors found in documents stored in metrics-apache.status-ep data stream: [0] field "apache.status.hostname" is undefined
[1] field "apache.status.total_kbytes" is undefined
```

Seems unrelated to this PR. Will check on `master` and file a separate PR if a fix is needed.

ycombinator:
Yep, the failure (see previous comment) is unrelated to this PR. It can be reproduced on `master` as well: https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/master/115/tests.

ycombinator:
CI failure is unrelated to this PR. See https://github.com/elastic/elastic-package/issues/374. Merging....

ycombinator:
Skipping the failing `apache.status` system test for now, while we investigate the failure, so as to unblock this and other PRs. See https://github.com/elastic/elastic-package/issues/374.

PR #377, opened 2021-06-09

mtojek:
Update dependency on package-spec.

Issue: https://github.com/elastic/integrations/issues/1091


mtojek:
self-approved

PR #379, opened 2021-06-15
Resolve dependency on ECS fields 

mtojek:
This PR introduces a mechanism to resolve dependency on external ECS fields and pull their definitions.

Sample files:

Build manifest - https://github.com/mtojek/elastic-package/blob/63-load-fields-from-ecs/test/packages/nginx/_dev/build/build.yml#L3
References to external fields (ECS) - https://github.com/mtojek/elastic-package/blob/63-load-fields-from-ecs/test/packages/nginx/data_stream/access/fields/ecs.yml#L6

How to test it:
1. Build the application out the PR branch with `make build`.
2. Play with the test package `nginx` and adjust ECS fields. I added a couple of external references to proof it works.
3. Try to update the dependency on the ECS repository with `elastic-package build --update -v`.

Issue: https://github.com/elastic/package-spec/issues/63

TODO:
- [x] Read `_dev/build/build.yml` manifest
- [x] Pull external ECS schema
- [x] Transform fields file into resolved form
- [x] Improve rendering of exported field table to use "built" fields 
- [x] Cache downloaded ECS schemas
- [x] Improve field validators in tests to use "built" fields
- [x] Introduce `--update` flag to update ECS dependency in the build manifest


mtojek:
@jsoriano as this PR is growing and based on the set of TODOs above it will grow more. Do you have any preference regarding the cut-off point?

jsoriano:
> @jsoriano as this PR is growing and based on the set of TODOs above it will grow more. Do you have any preference regarding the cut-off point?

As you wish, I think that the size of the PR is still ok, and it'd be better to merge the whole functionality together (unless it grows too much).

mtojek:
Ok, it's full blown, but ready for review. @jsoriano Please take a look when you're free, but no rush - I'm closing the shop for this week ;)

jsoriano:
Didn't look at the code yet, but I played with the feature a little bit, some comments:
* Having the descriptions updated automatically for ECS fields is awesome :slightly_smiling_face: 
* The `-u` flag, as is implemented now is a bit dangerous. It updates to the latest commit, this is probably not what a developer wants, this version hasn't been released yet and might have issues, breaking changes or fields that are later reverted. I would expect a developer to want one of these options:
  * To have a fixed well-known version, then the developer would directly write this version and wouldn't use `-u`. This would still allow to use an unreleased commit, but manually. For this use case it'd be great to support tags (current implementation supports them :slightly_smiling_face:, but linting fails).
  * Or if they don't care about the version, or they want to have the last one, they would like to update to the latest *released*  version, then they would use the `-u` flag, but ideally it should update to the latest released version.
* `elastic-package build -u` and `elastic-package format` format `_dev/build/build.yml` with different indentation.
* It could be nice to have a subcommand to automatically add fields found in test files to `ecs.yml`. For missing fields, maybe this command could check also the last ECS version (apart of the one defined in `build.yml`), and suggest to update if the field is there. Nothing to do in this PR in any case.

Thinking more about the update feature, maybe we want something like `elastic-package build update ecs@v1.10.0`. This would allow to support updating different things apart of ecs, and to specific versions. Without specifying the version it would update to the most recent release. Maybe we can leave this update feature out of this PR and think more about this in a follow up.

internal/fields/dependency_manager.go
```diff
@@ -0,0 +1,213 @@
...
+
+		url := fmt.Sprintf(ecsSchemaURL, gitReference, ecsSchemaFile)
+		logger.Debugf("Schema URL: %s", url)
+		resp, err := http.Get(url)
+		if err != nil {
+			return nil, errors.Wrapf(err, "can't download the online schema (URL: %s)", url)
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != 200 {
+			return nil, fmt.Errorf("unexpected HTTP status code: %d", resp.StatusCode)
```
jsoriano, line 90: Error is a bit confusing if the version specified doesn't exist:
```
Error: updating files failed: updating readme file README.md failed: rendering Readme failed: executing template failed: template: README.md:3:2: executing "README.md" at <fields "access">: error calling fields: can't create fields validator instance (path: /home/jaime/gocode/src/github.com/elastic/elastic-package/test/packages/nginx/data_stream/access): can't create field dependency manager: can't build fields schema: can't load fields: unexpected HTTP status code: 404
```

It'd be great if somehow it could directly say that the reference of ecs defined in `build.yml` doesn't exist.

internal/fields/dependency_manager.go
```diff
@@ -0,0 +1,213 @@
...
+	content, err := ioutil.ReadFile(cachedSchemaPath)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, errors.Wrapf(err, "can't read cached schema (path: %s)", cachedSchemaPath)
+	}
+	if errors.Is(err, os.ErrNotExist) {
+		logger.Debugf("Pulling ECS dependency using reference: %s", dep.Reference)
+
+		url := fmt.Sprintf(ecsSchemaURL, gitReference, ecsSchemaFile)
+		logger.Debugf("Schema URL: %s", url)
+		resp, err := http.Get(url)
```
jsoriano, line 84: :warning: here, is this the first network dependency we add to `elastic-package build`? I don't see it specially problematic, but it may be something we want to think twice.
It is probably better if we can avoid this, so developers workflow doesn't break if they are offline or if github is down.

Maybe an alternative is to have a set of subcommands to update the fields files, these commands would need connectivity, but the build process would continue being the same as now.

jsoriano (commented):
This is looking good, but some food for thought that could change the approach a little.

mtojek:
Thank you for the feedback, Jaime! I'm starting addressing your comments.

internal/fields/dependency_manager.go
mtojek: Frankly speaking we already have network requirement with the stack command - Docker images for Elastic stack are mandatory. You can go offline if you have cached them already. The same policy is applied here. If the cache is populated with fields, you can perform offline builds.

>Maybe an alternative is to have a set of subcommands to update the fields files, these commands would need connectivity, but the build process would continue being the same ...

I would be cautious with such approach, as it complicates the process (with internet connectivity, it's just `build`). It also means that we have to store these imported definitions somewhere in the repository.

I'm happy to consider other possible approach if you have any specific workload in your mind :)

mtojek:
> elastic-package build -u and elastic-package format format _dev/build/build.yml with different indentation.

Bug, fixed now. Thanks.

internal/fields/dependency_manager.go
mtojek: I adjusted the error message, but I'm not sure about passing this error up. It goes through internals of `text/template` and it helps with figuring out the execution path.

If you think that it will improve user experience, I can introduce additional code to verify dependencies first.

mtojek:
> Thinking more about the update feature, maybe we want something like elastic-package build update ecs@v1.10.0. This would allow to support updating different things apart of ecs, and to specific versions. Without specifying the version it would update to the most recent release. Maybe we can leave this update feature out of this PR and think more about this in a follow up.

Once this PR is merged we can think about follow-ups. I suppose that some of them may also affect the other fields schema we have in mind - for agent's and processors.

internal/fields/dependency_manager.go
jsoriano: Thanks, we can go on like this by now and reconsider later if needed.

internal/fields/dependency_manager.go
jsoriano: Ok, let's go on like this, I don't see it specially problematic, we might revisit this in the future.

Not directly related to offline processes, but one thing that I like to think about a build process is that it generates artifacts without modifying the source files. In this case the line between source files and a built package is a bit blurry. We may need to think more about this in the frame of https://github.com/elastic/package-registry/issues/670 and https://github.com/elastic/package-registry/issues/168.

mtojek:
I think I address the crucial issues, would you mind to re-review this, @jsoriano ? Thanks in advance.

jsoriano (approved):
Great!

PR #381, opened 2021-06-18
Print summary table after errors

jsoriano:
The table with the summary of test results is printed before the errors in case of failure. When there are many errors, this table can end up a long way scroll up, or even be unavailable depending on the buffer size of the terminal. This makes it difficult to identify the data stream whose tests failed.

Print it at the end, after the errors, so it is always available just over the prompt when tests fail.

### DIFF
--- a/internal/testrunner/reporters/formats/human.go
+++ b/internal/testrunner/reporters/formats/human.go
@@ -27,6 +27,26 @@ func reportHumanFormat(results []testrunner.TestResult) (string, error) {
 		return "No test results", nil
 	}
 
+	var report strings.Builder
+
+	headerPrinted := false
+	for _, r := range results {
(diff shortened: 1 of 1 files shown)


mtojek (approved):
LGTM, but it seems that there are some changes around dependencies (I guess tools). Would you mind to double-check with `go mod tidy` what's missing?

jsoriano:
> LGTM, but it seems that there are some changes around dependencies (I guess tools). Would you mind to double-check with `go mod tidy` what's missing?

I have opened https://github.com/elastic/elastic-package/pull/382 to solve this and prevent similar issues in the future, let me know what you think.

mtojek:
CI is happy, feel free to ship it!

PR #382, opened 2021-06-21
Don't use go get for tools in go.mod

jsoriano:
Using `go get` for packages in `go.mod` updates them, producing uncontrolled changes in unrelated PRs, and affecting reproducibility of builds.

Use `go run` for the affected commands.

`go install` could be used as a direct replacement of `go get` for this. But for commands that run with `go run` I prefer this option because it doesn't modify the tools installed in the environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,16 +8,13 @@ clean:
 	rm -rf build
 
 format:
-	go get golang.org/x/tools/cmd/goimports
-	goimports -local github.com/elastic/elastic-package/ -w .
+	go run golang.org/x/tools/cmd/goimports -local github.com/elastic/elastic-package/ -w .
 
 lint:
-	go get golang.org/x/lint/golint
-	go list ./... | xargs -n 1 golint -set_exit_status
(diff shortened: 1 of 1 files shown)


mtojek:
I remember that we added imports for tools here: https://github.com/elastic/elastic-package/blob/master/tools/tools.go

I understand that this file is still needed, right?

mtojek (approved):
A similar fix might be required for the package-spec repository.

jsoriano:
> I remember that we added imports for tools here: https://github.com/elastic/elastic-package/blob/master/tools/tools.go
> 
> I understand that this file is still needed, right?

Yes, this is complementary to this file. This `tools.go` file ensures that go mod keeps track of the dependencies, but using `go get` without any version can modify them, on the other hand `go install` and `go run` use the version in `go.mod`, without modifying it.

> A similar fix might be required for the package-spec repository.

Ok, I will check it, thanks for the pointer :+1:

jsoriano:
PR open for package spec: https://github.com/elastic/package-spec/pull/190

PR #380, opened 2021-06-17
Add option to exclude lines from sample log files in pipeline tests

jsoriano:
Needed for mysql tests https://github.com/elastic/integrations/pull/1145

PR #384, opened 2021-06-21
Avoid trying to adjust dropped events on test results

jsoriano:
Pipelines can drop events, and they appear as null values in the test result, avoid trying to convert these null events into `common.MapStr`, producing errors like:
```
ERROR: verifying test result failed: comparing test results failed: can't adjust test results: can't unmarshal event: unexpected end of JSON input
```

### DIFF
--- a/internal/testrunner/runners/pipeline/test_result.go
+++ b/internal/testrunner/runners/pipeline/test_result.go
@@ -103,10 +103,15 @@ func adjustTestResult(result *testResult, config *testConfig) (*testResult, erro
 	// Strip dynamic fields from test result
 	var stripped testResult
 	for _, event := range result.events {
+		if event == nil {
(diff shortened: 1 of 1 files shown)


mtojek (approved):
It smells like a workaround, but in general all runners (test framework) require refactoring. If it doesn't harm other cases, I'm good with this change. Maybe just drop comment on this line, why is it necessary.

mtojek:
@jsoriano Do you plan to continue working on this PR or should it be closed?

jsoriano:
Let's close it and continue droping the events on the pipeline.

jsoriano:
Oh, I closed this thinking that it was #380 facepalm reopening, this is needed for elastic/integrations#1145.

jsoriano:
Reopened again as https://github.com/elastic/elastic-package/pull/397

PR #385, opened 2021-06-22
Update dependency on elastic-package

mtojek:
This PR brings updated package-spec to the elastic-package. There are some changes around semantic validation of versions.


mtojek:
self-approved, just version bump.

PR #386, opened 2021-06-22
Add support for constant_keyword "value" in field definition

andrewkroh:
For the constant_keyword field type Elasticsearch [allows for the value](https://www.elastic.co/guide/en/elasticsearch//reference/7.12/keyword.html#constant-keyword-params) to be specified in the mapping.
If the value is not specified in the mapping then the value is set based on the first document that is
indexed. This adds support for specifying a constant_keyword "value" in a field definition.

Package data streams will be able to explicitly declare a `constant_keyword` value and will not have to
include the field with every document sent. It can be used optimize storage for an integration since 
you could omit the constant_keyword field entirely an event's `_source`.

Depends on: https://github.com/elastic/package-spec/pull/194 (this requires a spec update)

package-spec update: https://github.com/elastic/package-spec/commit/b358e974b7f9ce78f05970c4880a7a1dbdb35e59


mtojek (commented):
There are more changes required than this as we enabled fields dependency management.

It's related to the use case I mentioned in https://github.com/elastic/package-spec/pull/194/files#r656795658 (import field definition and override some properties). If we want to enable properties overriding. We have to modify the logic calling `fdm.ImportFields` (see: https://github.com/elastic/elastic-package/blob/841b4f0c70cdb91d4c85d5182637d8e000cd1caa/internal/fields/dependency_manager.go#L175) to not only import a field, but also override properties. I think there are 3 places in the codebase using it (docs, builder, fields validator). The simplest way to find them would adding a sample field definition in some test package (e.g. https://github.com/elastic/elastic-package/blob/master/test/packages/nginx/data_stream/access/fields/ecs.yml#L6).

andrewkroh:
Good point, I'll look over the places you mention and start adding support for overriding. I can see a use-case for a field like `data_stream.dataset` where you want to import from ECS, but also add `value: aws.vpcflow`.

mtojek:
@andrewkroh Do you plan to finalize this PR or split into multiple chunks? If it's ready for review, please unmark the draft mode.

internal/fields/model.go
```diff
@@ -9,6 +9,7 @@ type FieldDefinition struct {
 	Name        string            `yaml:"name"`
 	Description string            `yaml:"description"`
 	Type        string            `yaml:"type"`
+	Value       string            `yaml:"value"` // The value to associate with a constant_keyword field.
```
mtojek, line 12: As the spec is merged, you need to update the Go dependency on the package-spec.

andrewkroh: Updated to https://github.com/elastic/package-spec/commit/b358e974b7f9ce78f05970c4880a7a1dbdb35e59.

andrewkroh:
I will implement the ability to set attributes for external fields in a separate PR. The issue to track this feature is https://github.com/elastic/package-spec/commit/b358e974b7f9ce78f05970c4880a7a1dbdb35e59.

mtojek (approved):
Ship it! Please also update the dependency in the elastic/integrations.

PR #390, opened 2021-06-24
Add Debug statement for running test name

andrewstucki:
When we run system tests it's really hard to see which tests are being if a data stream contains multiple test configurations. The default behavior only shows the tests result after all of the tests for the particular test invocation have finished. In the case where you have lots of data streams with lots of tests for a package, this gets pretty painful if you're actually trying to track whether a test is failing while running the whole set of tests.

Simple solution: log the name of the test configuration right before it's run.


mtojek (approved):
LGTM

BTW Shaunak left the team.

PR #394, opened 2021-06-29
Fix defer-cleanup flag functionality for system testrunner

renekalff:
Currently the system testrunner is calling TearDown itself bypassing the defer-cleanup functionality. This MR resolves this by leaving calling TearDown to the testrunner main function.

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -162,9 +162,6 @@ func (r *runner) run() (results []testrunner.TestResult, err error) {
 		if err != nil {
 			return results, err
 		}
-		if err = r.TearDown(); err != nil {
-			return results, errors.Wrap(err, "failed to teardown runner")
-		}
 	}
 	return results, nil
 }


mtojek:
Hi @renekalff , thank you for the PR. Could you please describe manual testing steps to verify the fix?

renekalff:
> Hi @renekalff , thank you for the PR. Could you please describe manual testing steps to verify the fix?

After applying this fix run `elastic-package test system —defer-cleanup 1m -v` in a package with system testing defined like the nginx package. 

After loading the resources a debug message will be displayed that cleanup is deferred for one minute. Now manual inspection is possible, for example using kibana. 

After one minute cleanup is performed and the documents are removed from the datastream indexes. 

Setting a longer delay will allow you to develop the kibana dashboards based on the test events. These changes won’t be lost in the cleanup process nor will they be overwritten on subsequent system test executions.

mtojek (approved):
LGTM. Thanks for the fix!

PR #397, opened 2021-06-29
Avoid trying to adjust dropped events on test results

jsoriano:
If pipelines drop events during pipeline tests, they are set to `nil`, don't try to process them in any way to avoid failures.

`elastic-package test pipeline` reports this error:
```
ERROR: verifying test result failed: comparing test results failed: can't adjust test results: can't unmarshal event: unexpected end of JSON input
```

Needed for https://github.com/elastic/integrations/pull/1145


mtojek (approved):
This is an unfortunate use case where we need the "nil" value. Thank you for the fix.

jsoriano:
> This is an unfortunate use case where we need the "nil" value. Thank you for the fix.

Well, this allows to validate that certain events are dropped, but it is true that something more explicit would be better.

PR #399, opened 2021-07-01
Revert "Leave system testrunner teardown to main function (#394)"

mtojek:
This reverts commit 7ca3684a19090de1407b4f90c8fbc7bd8d61ba58.

@renekalff Unfortunately I have to revert your fix as it causes problems with system tests for traefik and zeek integrations (mentioned here: https://github.com/elastic/integrations/pull/1284).

If you have some spare time, you can root cause the interference of test results.


mtojek:
emergency fix, I have to push it now.

PR #400, opened 2021-07-01
Delete volumes when running docker compose down

jsoriano:
Remove volumes associated to scenarios when running docker-compose down.

### DIFF
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -85,6 +85,28 @@ services:
 
 Here, `SERVICE_LOGS_DIR` is a special keyword. It is something that we will need later.
 
+`elastic-package` will remove orphan volumes associated to the started services
+when they are stopped. Docker compose may not be able to find volumes defined in
+the Dockerfile for this cleanup. In these cases, override the volume definition.
+
+For example docker images for MySQL include a volume for the data directory
+`/var/lib/mysql`. In order for `elastic-package` to clean up these volumes after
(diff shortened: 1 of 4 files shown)


mtojek (approved):
LGTM

Do you think we should mention this in the `--help` for the `down` command?

jsoriano:
> LGTM
> 
> Do you think we should mention this in the `--help` for the `down` command?

Not sure, I would assume that no leftovers are left if the scenario is down.

jsoriano:
I have found that the service have their own runner, added the flag there too. @mtojek could you take another look?

mtojek (approved):
WFG 

Please bump up the elastic-package dependency in elastic/integrations too.

jsoriano:
Side note on this change, definitions like [this one](https://github.com/elastic/integrations/pull/1145/commits/a3be9183aee180fd1c924c482f2af723969ca5b6) may be needed for docker compose to be able to remove the volumes in some cases where the volume is defined in the dockerfile.

mtojek:
Good catch. Maybe you should describe it also here: https://github.com/elastic/elastic-package/blob/master/docs/howto/system_testing.md

mtojek:
I cleaned EC2 instances used by `elastic-packag test system`.

jsoriano:
> Good catch. Maybe you should describe it also here: https://github.com/elastic/elastic-package/blob/master/docs/howto/system_testing.md

Docs added for this.

mtojek (approved):

PR #403, opened 2021-07-06
Fix: defer cleanup and signal handling

mtojek:
This PR refactors the logic around the "defer cleanup" flag and "SIGINT handling".

Fixes:
* Bug: SIGINT caught multiple times, once per test config file. If the user pressed Ctrl+C for the last, third test case, signal got caught 3 times
* Bug: Defer Cleanup doesn't trigger second (extra) tear down
* SIGINT breaks only long lasting operations (waiting for data, for agent)
* Bug: there is an extra "tear down" call for system test runner, even though it's called per every single test case (in PR: it's called only per test case)
* Bug: defer cleanup happens after data is already removed :) 

Related:
- https://github.com/elastic/elastic-package/pull/394
- https://github.com/elastic/elastic-package/pull/399

### DIFF
--- a/internal/kibana/agents.go
+++ b/internal/kibana/agents.go
@@ -9,9 +9,10 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/elastic/elastic-package/internal/logger"
-
 	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/logger"
+	"github.com/elastic/elastic-package/internal/signal"
 )
 
 // Agent represents an Elastic Agent enrolled with fleet.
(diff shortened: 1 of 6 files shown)


internal/kibana/agents.go
```diff
@@ -71,6 +72,10 @@ func (c *Client) AssignPolicyToAgent(a Agent, p Policy) error {
 
 func (c *Client) waitUntilPolicyAssigned(a Agent, p Policy) error {
 	for {
+		if signal.SIGINT() {
```
jsoriano, line 75: There may be other reasons to cancel actions, and we would need to remember to add them all to all the actions that may require cancelation.

WDYT about passing these methods a context or a done channel? If it is a context it could be also passed here to `getAgent`, that could help to cancel the requests it does.

Anything that could need to handle cancelation could receive one of these contexts.

mtojek: I agree with you, but I would pair with some bigger refactoring then. Maybe a follow-up task. I think we should introduce it to all runners.

jsoriano: Ok, let's revisit this later.

jsoriano (approved):

PR #404, opened 2021-07-06
Add dependabot

v1v:
## What

Delegate the bump automation to `dependabot`. Please read below

## Why

Simplify recurrent tasks

## Next steps

`Allow` is not supported to declare the list of dependencies to be bumped, something that is supported for some other package systems. See https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#allow

The proposal is, delegate the bump automation to dependabot and for each dependency that should not be taking in consideration then:

- ignore it explicitly as stated in https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#creating-ignore-conditions-from-dependabot-ignore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - dependencies
+      - go
+  	open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
     labels:
       - dependencies
       - go
-  	open-pull-requests-limit: 10
+    open-pull-requests-limit: 10


mtojek (changes requested):
Thanks for the PR.

Could you please add another one for https://github.com/elastic/package-spec ? The dependency graph is as follows: integrations -> elastic-package -> package-spec.

If the dependabot works well for package-spec, we can enable it for elastic-package.

v1v:
https://github.com/elastic/package-spec/pull/196 has been merged, let's wait until dependabot runs and then we can move forward if we are happy with the outcome

mtojek (approved):
Looks like dependabot did a great job, let's continue this initiative.

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - dependencies
+      - go
+  	open-pull-requests-limit: 10
```
v1v, line 11:
```suggestion
    open-pull-requests-limit: 10
```

v1v:
I'll merge this PR without running a new build in the CI since it does nothing in the CI at all

Issue #94, opened 2020-09-10
[System test runner] Support use of deployment variants


ycombinator:
Enhance `elastic-package test system` to support variations of deployments for the service under test. For example, we may want to test against a specific version of the service.

The variations will be defined in the `_dev/deploy/variants.yml` file, for example:

```yml
variants:
  v1:
    SERVICE_VERSION: 1.19.1
  latest:
    SERVICE_VERSION: 2.3.0
default: latest
```

In this case the `elastic-package test system` command might take an optional flag that allows the selection of a specific variant, e.g. `elastic-package test system --deploy-variant=latest` (no strong opinions about the specific flag name). When the flag is not specified, the variant defined in the `default` property should be selected. The flag may take a special value, e.g. `__random__`, that will randomly select a variant for the test (again, no strong opinions on the value name).

When the variant is selected, all the variables defined under it must be made available to the service deployer so it may deploy the package's service(s) after interpolating these variables wherever appropriate, e.g. in the service's `docker-compose.yml` file.

Related: https://github.com/elastic/integrations/issues/781

mtojek:
Let me research the scope of changes.

mtojek:
It seems to be possible to implement it:

1. Implement logic to read available `variants` with environment variables from `variants.yml`.
2. System test runner: iterate over variants and test config files.
3. Pass variant info to `runTest`: https://github.com/elastic/elastic-package/blob/master/internal/testrunner/runners/system/runner.go#L164
4. Extend `servicedeployer.Factory()` to consider variant info: https://github.com/elastic/elastic-package/blob/d7dd1a3ac595583af20e91f3076edb095ea97bd5/internal/testrunner/runners/system/runner.go#L249
5. `elastic-package service up`: select/list variants

PR #407, opened 2021-07-07
Select service variants with "service up"

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/94

This PR extends the `elastic-package service up` command to select right service variant to boot up. Service factory (for Docker Compose) expects variant information now.

How to test:

```
make build
cd test/packages/apache
elastic-package service up -v -s v2420
ctrl+c...

elastic-package service up -v -s v2443
```

### DIFF
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -26,7 +26,7 @@ func setupServiceCommand() *cobraext.Command {
 		RunE:  upCommandAction,
 	}
 	upCommand.Flags().StringP(cobraext.DataStreamFlagName, "d", "", cobraext.DataStreamFlagDescription)
-	upCommand.Flags().StringP(cobraext.VariantFlagName, "s", "", cobraext.VariantFlagDescription)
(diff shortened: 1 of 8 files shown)


cmd/service.go
```diff
@@ -26,6 +26,7 @@ func setupServiceCommand() *cobraext.Command {
 		RunE:  upCommandAction,
 	}
 	upCommand.Flags().StringP(cobraext.DataStreamFlagName, "d", "", cobraext.DataStreamFlagDescription)
+	upCommand.Flags().StringP(cobraext.VariantFlagName, "s", "", cobraext.VariantFlagDescription)
```
jsoriano, line 29: Nit. Why `s`? I think I would prefer to don't have an abbreviated form :slightly_smiling_face:

jsoriano (approved):
Nice to have this feature!

cmd/service.go
mtojek: It's for `service variant`, it's because `-v` is reserved for `verbose`. I can drop the short form in this case :)

mtojek: Fixed

PR #420, opened 2021-07-08
Add watch of jobs resources in cluster role used by agent

MichaelKatsoulis:
Static Kubernetes manifest for creating ClusterRole for kind-fleet-agent is updated to watch for jobs in batch apiGroups


mtojek (approved):
Please wait for the green CI status.

PR #423, opened 2021-07-09
[dependabot][mergify] automate merge and add reviewers

v1v:
## What does this PR do?

- Add reviewers
- Automerge if CI is happy. Although, if no approval is needed why do we need the reviewers?

## Why is it important?

Look mum no hands!

## Issue

See https://github.com/elastic/integrations/pull/1291#issuecomment-876963997

Examples with mergify -> https://docs.mergify.io/examples/#example-rules

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
     schedule:
       interval: "daily"
     labels:
-      - dependencies
-      - go
+      - automation
+    reviewers:
+      - "mtojek"
+      - "jsoriano"
     open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,5 @@ updates:
     labels:
       - automation
     reviewers:
-      - "mtojek"
-      - "jsoriano"
+      - "elastic/integrations-developer-experience"
     open-pull-requests-limit: 10
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,11 @@
+pull_request_rules:
+  - name: automatic merge of bot 🤖
+    conditions:
+      - check-success=elastic-package/pr-merge
+      - check-success=CLA
+      - base=master
+      - author~=^dependabot(|-preview)\[bot\]$
+    actions:
+      merge:
+        method: squash
+        strict: smart+fasttrack


.github/dependabot.yml
```diff
@@ -6,6 +6,8 @@ updates:
     schedule:
       interval: "daily"
     labels:
-      - dependencies
-      - go
+      - automation
+    reviewers:
+      - "mtojek"
+      - "jsoriano"
```
jsoriano, line 12: I'd prefer to put a [github] team here so we can more easily manage notifications and we don't need need to change this if the people related to the project changes. But we can go on like this by now.

jsoriano (approved):

cachedout (approved):

.github/dependabot.yml
v1v: Sure thing.

BTW, we are running a PoC with a `oncall` github team, so the person oncall need to subscribe to the team and remove other team members. The idea is to reduce the overhead/notifications for other team members when they are not on-call, but we are still at the very beginning of this evaluation :)

jsoriano: Oh, this sounds great to ping with the person of call for issues related to builds/CI. 

> the person oncall need to subscribe to the team and remove other team members.

Would it be possible to do it automatically by checking the Pagerduty calendars?

kuisathaverat (approved):

mtojek (approved):

.github/dependabot.yml
v1v: I just answered you in the initial discussion we have opened in the obs-robots team regarding this process :)

jsoriano: Thanks!

.github/dependabot.yml
```diff
@@ -6,6 +6,8 @@ updates:
     schedule:
       interval: "daily"
     labels:
-      - dependencies
-      - go
+      - automation
+    reviewers:
+      - "mtojek"
```
v1v, line 11:
```suggestion
      - "elastic/integrations-developer-experience"
```

.github/dependabot.yml
v1v:
```suggestion
```

jsoriano (approved):

mtojek (approved):

PR #424, opened 2021-07-09
Use service variants in system tests

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/94

This PR introduces support for service variants for system tests. So far they are supported only in the `service` command.

Zrzut ekranu 2021-07-9 o 13 30 48

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -171,7 +171,7 @@ func (r *runner) run() (results []testrunner.TestResult, err error) {
 	}
 
 	for _, cfgFile := range cfgFiles {
-		for _, variantName := range anyVariants(variantsFile) {
+		for _, variantName := range r.selectVariants(variantsFile) {
 			partial, err := r.runTestPerVariant(result, locationManager, cfgFile, dataStreamPath, variantName)
 			results = append(results, partial...)
 			if err != nil {
@@ -660,13 +660,16 @@ func validateFields(docs []common.MapStr, fieldsValidator *fields.Validator, dat
 	return nil
 }
 
-func anyVariants(variantsFile *servicedeployer.VariantsFile) []string {
+func (r *runner) selectVariants(variantsFile *servicedeployer.VariantsFile) []string {
 	if variantsFile == nil || variantsFile.Variants == nil {
 		return []string{""} // empty variants file switches to no-variant mode
 	}
 
 	var variantNames []string
 	for k := range variantsFile.Variants {
+		if r.options.ServiceVariant != "" && r.options.ServiceVariant != k {
+			continue
+		}
 		variantNames = append(variantNames, k)
 	}
 	return variantNames
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -172,8 +172,6 @@ func (r *runner) run() (results []testrunner.TestResult, err error) {
 
 	for _, cfgFile := range cfgFiles {
 		for _, variantName := range anyVariants(variantsFile) {
-			logger.Debugf("Use service variant: %s", variantName)
-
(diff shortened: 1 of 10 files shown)


internal/testrunner/runners/system/runner.go
```diff
@@ -143,40 +144,70 @@ func (r *runner) run() (results []testrunner.TestResult, err error) {
...
+			}
+
+			results = append(results, partial...)
+			tdErr := r.tearDownTest()
+			if err != nil {
+				return results, err
+			}
+			if tdErr != nil {
+				return results, errors.Wrap(tdErr, "failed to tear down runner")
+			}
```
jsoriano, line 210: Nit. Move the code in this loop to run the test on a variant to its own method?

test/packages/apache/_dev/deploy/variants.yml
```diff
@@ -1,6 +1,6 @@
 variants:
   v2420:
     SERVICE_VERSION: 2.4.20
-  v2443:
-    SERVICE_VERSION: 2.4.43
+  v2423:
+    SERVICE_VERSION: 2.4.23
```
jsoriano, line 5: Have we considered to add something to skip variants? This may be needed in the future.

internal/testrunner/runners/system/runner.go
```diff
@@ -143,40 +144,70 @@ func (r *runner) run() (results []testrunner.TestResult, err error) {
...
+		return result.WithError(errors.Wrap(err, "_dev/deploy directory not found"))
+	}
+
+	variantsFile, err := servicedeployer.ReadVariantsFile(devDeployPath)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return result.WithError(errors.Wrap(err, "can't read service variant"))
+	}
+
+	for _, cfgFile := range cfgFiles {
+		for _, variantName := range anyVariants(variantsFile) {
```
jsoriano, line 174: There should be some way to run with a single variant.

jsoriano (commented):
Nice! LGTM, but could we run the CI job with all the integrations, or at least with mysql? In https://github.com/elastic/integrations/pull/1145 I added the variants defined in Beats but I didn't test them :grimacing:

test/packages/apache/_dev/deploy/variants.yml
mtojek: We have the "skip" option for the entire test config not for particular variant. Maybe a follow up issue?

mtojek:
> could we run the CI job with all the integrations, or at least with mysql? In elastic/integrations#1145 I added the variants defined in Beats but I didn't test them 😬

I can copy the mysql package to the test packages here, so we can test it. I will create a separate PR.

test/packages/apache/_dev/deploy/variants.yml
jsoriano: :+1:

jsoriano:
> > could we run the CI job with all the integrations, or at least with mysql? In [elastic/integrations#1145](https://github.com/elastic/integrations/pull/1145) I added the variants defined in Beats but I didn't test them grimacing
> 
> I can copy the mysql package to the test packages here, so we can test it. I will create a separate PR.

Oh, forget it, I mixed things when I saw the apache tests passing :slightly_smiling_face:

internal/testrunner/runners/system/runner.go
mtojek: Good idea, fixed.

internal/testrunner/runners/system/runner.go
mtojek: I added CLI option for this:

```elastic-package test system --variant```

jsoriano: Great! Thanks.

jsoriano (approved):

Issue #406, opened 2021-07-07
Wait for healthy services in system tests


jsoriano:
Wait for healthy services when starting them using the compose service deployer used on system tests. This is important to avoid errors when collecting metrics.

Similar to https://github.com/elastic/elastic-package/issues/316, but for compose.

mtojek:
Let me try to implement this feature.

PR #425, opened 2021-07-13
Implement wait routine for Docker Compose

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/406

This PR introduces a wait routine to Docker Compose service deployer (and also Terraform which bases on the Compose). 

Testing:

I used `mysql` integration with removed [mysql_is_ready](https://github.com/elastic/integrations/blob/master/packages/mysql/_dev/deploy/docker/docker-compose.yml#L10-L14) service. Frankly speaking I'm not convinced if I prefer `tianon/true` solution more :)

### DIFF
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -12,11 +12,14 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
 
+	"github.com/elastic/elastic-package/internal/docker"
 	"github.com/elastic/elastic-package/internal/logger"
+	"github.com/elastic/elastic-package/internal/signal"
 )
 
 // Project represents a Docker Compose project.
@@ -245,6 +248,58 @@ func (p *Project) Logs(opts CommandOptions) ([]byte, error) {
 		return nil, err
 	}
 	return b.Bytes(), nil
+}
+
+// WaitForHealthy method waits until all containers are healthy.
+func (p *Project) WaitForHealthy(opts CommandOptions) error {
+	// Read container IDs
+	args := p.baseArgs()
+	args = append(args, "ps")
+	args = append(args, "-q")
+
+	var b bytes.Buffer
+	if err := p.runDockerComposeCmd(dockerComposeOptions{args: args, env: opts.Env, stdout: &b}); err != nil {
+		return err
+	}
+
+	containerIDs := strings.Split(strings.TrimSpace(b.String()), "\n")
+	for {
+		if signal.SIGINT() {
+			return errors.New("SIGINT: cancel waiting for policy assigned")
+		}
+
+		healthy := true
+
+		logger.Debugf("Wait for healthy containers: %s", strings.Join(containerIDs, ","))
+		descriptions, err := docker.InspectContainers(containerIDs...)
+		if err != nil {
+			return err
+		}
+
+		for _, containerDescription := range descriptions {
+			logger.Debugf("Container status: %s", containerDescription.String())
+
+			// Service is up and running and it's healthy
+			if containerDescription.State.Status == "running" && containerDescription.State.Health.Status == "healthy" {
+				continue
+			}
+
+			// Container started and finished with exit code 0
+			if containerDescription.State.Status == "exited" && containerDescription.State.ExitCode == 0 {
+				continue
+			}
+
+			// Any different status is considered unhealthy
+			healthy = false
+		}
+
+		if healthy {
+			break
+		}
+
+		time.Sleep(time.Second)
+	}
+	return nil
 }
 
 func (p *Project) baseArgs() []string {
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -279,6 +279,11 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 		for _, containerDescription := range descriptions {
 			logger.Debugf("Container status: %s", containerDescription.String())
 
+			// No healthcheck defined for service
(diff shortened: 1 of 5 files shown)


internal/compose/compose.go
```diff
@@ -247,6 +250,63 @@ func (p *Project) Logs(opts CommandOptions) ([]byte, error) {
 	return b.Bytes(), nil
 }
 
+// WaitForHealthy method waits until all containers are healthy.
+func (p *Project) WaitForHealthy(opts CommandOptions) error {
+	// Read container IDs
+	args := p.baseArgs()
+	args = append(args, "ps")
+	args = append(args, "-q")
```
jsoriano, line 258: One problem of `docker-compose ps` is that it doesn't show services that are not even started. So this method may end up succesfully when there is no container running.

This may be ok though if we only want to wait for the started services and not for others.

internal/compose/compose.go
```diff
@@ -247,6 +250,63 @@ func (p *Project) Logs(opts CommandOptions) ([]byte, error) {
...
+	var b bytes.Buffer
+	if err := p.runDockerComposeCmd(dockerComposeOptions{args: args, env: opts.Env, stdout: &b}); err != nil {
+		return err
+	}
+
+	containerIDs := strings.Split(strings.TrimSpace(b.String()), "\n")
+	for {
+		if signal.SIGINT() {
+			return errors.New("SIGINT: cancel waiting for policy assigned")
+		}
```
jsoriano, line 269: Other place where we may want some context or channel for cancelation :slightly_smiling_face:

internal/docker/docker.go
```diff
@@ -90,3 +118,26 @@ func ConnectToNetwork(containerID, network string) error {
 	}
 	return nil
 }
+
+// InspectContainers function inspects selected Docker containers.
+func InspectContainers(containerIDs ...string) ([]ContainerDescription, error) {
+	args := []string{"inspect"}
+	args = append(args, containerIDs...)
+	cmd := exec.Command("docker", args...)
```
jsoriano, line 126: We could use the [docker API](https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerInspect) directly, so we don't have to define and parse the docker API structs here. Or you are explicitly avoiding adding this new dependency?

jsoriano (commented):
LGTM, but some ideas before merging if you would like to explore other options.

> Frankly speaking I'm not convinced if I prefer tianon/true solution more :)

Actually the good thing of relying in `docker-compose up` only would be that we don't have to do other tricky things as parsing responses of docker or docker compose to check the state of the containers.

Maybe an alternative is to continue using `tianon/true`, but automating it so we don't need to add this to all integrations.

It should be possible to add `-f services_are_ready.yml` to all docker compose calls, where `services_are_ready.yml` is a generated file with a service that depends on all the services defined in the main file :slightly_smiling_face: Something like this pseudo-template-code.
```
services:
  wait_for_others:
    image: tianon/true
    depends_on:
    {{ for service in `docker-compose config --services` }}
       {{ service }}:
         condition: service_healthy
    {{ endfor }}
```

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -78,8 +78,14 @@ func (d *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
...
-	if err := p.Up(opts); err != nil {
-		return nil, errors.Wrap(err, "could not boot up service using docker compose")
+	err = p.Up(opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not boot up service using Docker Compose")
+	}
+
+	err = p.WaitForHealthy(opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "service is unhealthy")
```
jsoriano, line 88: Could we add as tests an scenario with system tests that would fail consistently if they are run without waiting for healthy services?

internal/compose/compose.go
mtojek: That's why I have doubts once I opened this PR. Maybe we should agree on using `tianon/true` as standard? It seems to be more trustful. I have also noticed that we have services without healthcheck defined. It's a bit like eventual consistency - the policy is already setup, agent is ready, but there are no data. Eventually the service will boot up and metrics data will flow.

internal/compose/compose.go
mtojek: Agree, another place to be covered while refactoring this runner.

internal/docker/docker.go
mtojek: I think we decided to go with "docker" as it was quicker. We didn't have to struggle with different Docker APIs (different versions) and it's consistent with "docker-compose". I'm fine with creating an issue to replace all "docker" calls with a library and APIs.

internal/testrunner/runners/system/servicedeployer/compose.go
mtojek: Hm... I'm wondering how we can approach this as in most cases metrics will *eventually* flow. We would need a way to collect them immediately.

mtojek:
> Maybe an alternative is to continue using tianon/true, but automating it so we don't need to add this to all integrations.

I would vote for this 👍 .

> It should be possible to add -f services_are_ready.yml to all docker compose calls, where services_are_ready.yml is a generated file with a service that depends on all the services defined in the main file 🙂 Something like this pseudo-template-code.

Personally I prefer the WYSIWIG approach, so we don't add anything in background or transform Dockerfiles.

jsoriano (approved):
Ok, I think we can go on with this. There are still some open questions and things we could improve, but let's think more about this in the future.

I am also ok with not merging this, and wait a bit to see if with `tianon/true` we are good enough. Maybe having to add this is not so bad after all :slightly_smiling_face:

PR #429, opened 2021-07-15
document completion command

### LINKED ISSUE #402: Shell completions
With the Cobra v1.2.0 [release](https://github.com/spf13/cobra/releases/tag/v1.2.0), shell completions are automatically supported. We should upgrade the `cobra` dependency and then add `elastic-package completion` to the docs.

endorama:
Provide documentation for `completion` command added with cobra 1.2.1.

Closes #402

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,16 @@ _Context: global_
 Use this command to get a listing of all commands available under `elastic-package` and a brief
 description of what each command does.
 
+### `elastic-package completion`
+
+_Context: global_
+
+Use this command to output shell completion information.
+
+The command output shell completions information (for `bash`, `zsh`, `fish` and `powershell`). The output can be sourced in the shell to enable command completion.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,6 +56,16 @@ _Context: global_
 Use this command to get a listing of all commands available under `elastic-package` and a brief
 description of what each command does.
 
+### `elastic-package completion`
```
mtojek, line 59: Could you please put this command in alphanum. order?

endorama: I did so, but it got overwritten by running `make update-readme` (which I had to run because the CI was red).

As `completion` is not a command we implemented, I moved it near `help` for this reason.

How may I overcome this issue?

mtojek: Ah, you're right. We generate all these command infos.

mtojek (approved):
Let's leave it as is. Thanks for working on this!

Issue #32, opened 2020-07-30
Automatically check if the tool is up to date


exekias:
We are currently recommending people to install this tool by using `go get`. I think this is the best approach until we get to some degree of stability in the features it provides (we are not close to this by any means).

It would be nice to have an automatic check when the tools is called that warns you if it's not up to date, and gives you the command to run if you need to update.

This would ensure that developers don't lag behind on tooling, with the confusion that that may cause (ie my integration works in local but fails in CI...).

To avoid degrading the user experience, I think we should only check for new versions from time to time (maybe store the last successful check timestamp into a file) and have a good silent timeout if things are too slow.

We could use the [following API](https://docs.github.com/en/rest/reference/git#get-a-reference) to retrieve master branch ref:
```
turing.local :: integrations/packages/nginx ‹master› » time curl https://api.github.com/repos/elastic/elastic-package/git/ref/heads/master
{
  "ref": "refs/heads/master",
  "node_id": "MDM6UmVmMjY5NjEyNzUzOnJlZnMvaGVhZHMvbWFzdGVy",
  "url": "https://api.github.com/repos/elastic/elastic-package/git/refs/heads/master",
  "object": {
    "sha": "9e6eefab29ae9533a7d387cc62247a77a9ec07ff",
    "type": "commit",
    "url": "https://api.github.com/repos/elastic/elastic-package/git/commits/9e6eefab29ae9533a7d387cc62247a77a9ec07ff"
  }
}
curl https://api.github.com/repos/elastic/elastic-package/git/ref/heads/maste  0.02s user 0.01s system 6% cpu 0.333 total
```

exekias:
There is a small issue we would need to overcome: Retrieving the local commit hash

If the tool was installed with `go get` it should be present under Go's path: `src/github.com/elastic/elastic-package`. We should treat the case where we cannot really discover the local hash (ie if the user installed it by other means)

mtojek:
Let's review possible options.

**1. Git tag only**

Just tag next commits with versions whenever is good to release a change. With mergify and dependabot in place,
all updates will be automatically pulled into elastic/integrations repository.

Nothing will change for end-users, unless we decide to show notifications in the application that there is a newer
version available. They will have to refresh and rebuild sources as they do now.

**2. Use goreleaser to release Linux/Windows/Mac binaries**

We can use goreleaser to push new binaries to the Github Releases and store them there. The solution will base
[...]

PR #431, opened 2021-07-16
Enable goreleaser to release binaries

mtojek:
This PR enables support for GoReleaser to release elastic-package distributions.

How to trigger a release:

```bash
git tag -a v0.1.0 -m "First release"
git push origin v0.1.0
```
(source: [Quick start](https://goreleaser.com/quick-start/))

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -70,7 +70,7 @@ pipeline {
     }
     stage('Release') {
       when {
-        buildingTag()
+        tag pattern: '(v)?\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
       }
       steps {
         withCredentials([string(credentialsId: "${GITHUB_TOKEN_CREDENTIALS}", variable: 'GITHUB_TOKEN')]) {
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -7,6 +7,7 @@ pipeline {
   environment {
     BASE_DIR="src/github.com/elastic/elastic-package"
     JOB_GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
+    GITHUB_TOKEN_CREDENTIALS = "2a9602aa-ab9f-4e52-baf3-b71ca88469c7"
     PIPELINE_LOG_LEVEL='INFO'
     AWS_ACCOUNT_SECRET = 'secret/observability-team/ci/elastic-observability-aws-account-auth'
     HOME = "${env.WORKSPACE}"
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -67,6 +68,16 @@ pipeline {
         }
       }
     }
+    stage('Release') {
+      when {
+        buildingTag()
```
v1v, line 73: What's the tag strategy? In this case, it will happen for each tag, so in case you'd like to distinguish with some regex/pattern

mtojek: There is semver limitation: https://goreleaser.com/limitations/semver/

v1v: I'd say to apply the same pattern in the when condition, something like:

```suggestion
        tag pattern: '(v)?\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
```

v1.2.3 or 1.2.3 are supported

v1v (approved):
NIT comment

Regarding the goreleaser command to be executed, that's something to be done in a follow up I assume

.ci/Jenkinsfile
mtojek: Thanks for suggestion!

mtojek:
I will experiment with Goreleaser once I merge this PR.

.ci/Jenkinsfile
```diff
@@ -67,6 +68,16 @@ pipeline {
         }
       }
     }
+    stage('Release') {
+      when {
+        tag pattern: '(v)?\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
```
jsoriano, line 73: @mtojek have we thought about the versioning pattern or release cycles that `elastic-package` is going to follow?

I don't think it should follow semantic versioning. Any version of `elastic-package` should work with any supported package. If we introduce a breaking change in the spec, the spec should have some versioning so `elastic-package` can support both versions (as `docker-compose` does for example). Breaking changes in `elastic-package` should be considered as bugs. And I would prefer to don't have different release branches.

In any case, I shouldn't assume this pattern is for elastic versioning, it can also be `v<year>.<month>.<revision>`, I would like something like this for `elastic-package`. :slightly_smiling_face:

mtojek: Frankly speaking, I like semantic versioning for every use case, as it allows for introducing potentially dangerous (breaking changes). We have dependabot, mergify, no goreleaser, which means that every new release can be automatically pulled in into elastic/integrations.

> I don't think it should follow semantic versioning. Any version of elastic-package should work with any supported package. If we introduce a breaking change in the spec, the spec should have some versioning so elastic-package can support both versions (as docker-compose does for example).

I admit that we never followed this rule. I'm not sure if it's possible in real life without bumping version in package manifests (so far every package is 1.0.0). Many times there were cases in which we had to harden the spec validation (by introducing additional semantic validators) which obviously caused failures for packages. I agree we should bump up the spec version then, but I'm not sure about consequences across the entire platform (when format version changes - I know it shouldn't cause any problems). 

> In any case, I shouldn't assume this pattern is for elastic versioning, it can also be v<year>.<month>.<revision>, I would like something like this for elastic-package. 🙂

I see we can mimic it with semver, although I admit I'm too much addicted to the natural semver form (major, minor, patch).

I'm happy to zoom about this next week if you prefer!

jsoriano: Ok, we can think more about this, let's talk next week!

PR #433, opened 2021-07-16
Fix goreleaser: populate ldflags

mtojek:
This PR fixes missing build info while user runs `elastic-package version`.

Tested:

```
$ goreleaser build --skip-validate --single-target
$ ./dist/elastic-package_darwin_amd64/elastic-package version
elastic-package version-hash e59a09c (build time: 2021-07-16T13:11:19+02:00)
```

@jsoriano Thank you for suggestions in https://github.com/elastic/elastic-package/pull/431#discussion_r671170415 . I suggest to review and merge this PR, and then we can review the entire working solution in terms of further adjustments. This distribution method is not advertised yet, so we can disable it if we decide it doesn't meet our requirements.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,4 @@
+builds:
+  - ldflags:
+      - -X github.com/elastic/elastic-package/internal/version.CommitHash={{.ShortCommit}}
+      - -X github.com/elastic/elastic-package/internal/version.BuildTime={{.Timestamp}}


.goreleaser.yml
```diff
@@ -0,0 +1,4 @@
+builds:
+  - ldflags:
+      - -X github.com/elastic/elastic-package/internal/version.CommitHash={{.ShortCommit}}
+      - -X github.com/elastic/elastic-package/internal/version.BuildTime={{.Timestamp}}
```
jsoriano, line 4: Users are going to be probably more interested about the version itself :slightly_smiling_face: Add the tag to the output of `elastic-package version` too, so it displays something like this?
```
$ goreleaser build --skip-validate --single-target
$ ./dist/elastic-package_darwin_amd64/elastic-package version
elastic-package v0.9.1 version-hash e59a09c (build time: 2021-07-16T13:11:19+02:00)
```
```suggestion
      - -X github.com/elastic/elastic-package/internal/version.BuildTime={{.Timestamp}}
      - -X github.com/elastic/elastic-package/internal/version.Tag={{.Tag}}
```

jsoriano (commented):
Nice to see how goreleaser can be customized.

I think we should add the version itself to the output of `elastic-package version`.

.goreleaser.yml
mtojek: I want to do this in the next follow-up (just after this one :) ) as it requires a bit of coding. With this PR I want to fix first what's still broken.

If you prefer I can try to modify version's logic in this PR (both options work for me).

jsoriano: SGTM to do it in a follow up.

jsoriano (approved):

PR #439, opened 2021-07-19
Check if newer version is available

mtojek:
This PR introduces logic to check using Github API Releases if newer version is available.

### DIFF
--- a/internal/version/check_update.go
+++ b/internal/version/check_update.go
@@ -1,0 +1,56 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package version
+
+import (
+	"context"
+
+	"github.com/Masterminds/semver"
+
+	"github.com/elastic/elastic-package/internal/github"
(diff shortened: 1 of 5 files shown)


internal/version/check_update.go
```diff
@@ -0,0 +1,56 @@
...
+// CheckUpdate function checks using Github Release API if newer version is available.
+func CheckUpdate() {
+	if Tag == "" {
+		logger.Debugf("Distribution built without a version tag, can't determine release chronology. Please consider using official releases at " +
+			"https://github.com/elastic/elastic-package/releases")
+		return
+	}
+
+	githubClient := github.UnauthorizedClient()
+	release, _, err := githubClient.Repositories.GetLatestRelease(context.Background(), repositoryOwner, repositoryName)
```
jsoriano, line 30: Nit. Use `context.TODO()` in case at some moment we pass a context here?
```suggestion
	release, _, err := githubClient.Repositories.GetLatestRelease(context.TODO(), repositoryOwner, repositoryName)
```

jsoriano (approved):
:+1:

internal/version/check_update.go
mtojek: Good idea, committed, thanks!

PR #441, opened 2021-07-21
Generate Cobertura report

mtojek:
Issue: https://github.com/elastic/integrations/issues/755

This PR introduces logic to generate Cobertura test coverage reports.

Zrzut ekranu 2021-07-21 o 16 56 03

Sample reports:
https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-441/7/cobertura/
https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-441/9/cobertura/


internal/testrunner/coverage_output.go
```diff
@@ -0,0 +1,248 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
jsoriano, line 1: Nit. Underscores shouldn't be used in go file names, unless they mean something for `go build`. This file could be named `coverageoutput.go`.

internal/testrunner/coverage_output.go
```diff
@@ -0,0 +1,248 @@
...
+		if len(testCases) == 0 {
+			methods = append(methods, &coberturaMethod{
+				Name:  "no-test",
+				Lines: []*coberturaLine{{Number: 1, Hits: 0}},
+			})
+		} else {
+			for i, tc := range testCases {
+				methods = append(methods, &coberturaMethod{
+					Name:  tc,
+					Lines: []*coberturaLine{{Number: i + 1, Hits: 1}},
```
jsoriano, line 186: Thoughts for a possible future improvement. Coverage is at the data stream level now, right? Maybe in the future we can do it at least at the file level, e.g. saying that pipeline tests add coverage for pipelines, but not for the manifest or other config files, and system tests add coverage for all the files.

cmd/testrunner.go
```diff
@@ -116,6 +117,11 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 			return cobraext.FlagParsingError(err, cobraext.ReportOutputFlagName)
 		}
 
+		testCoverage, err := cmd.Flags().GetBool(cobraext.TestCoverageFlagName)
```
jsoriano, line 120: Should this flag only be allowed if `--report-output file` is also used?

jsoriano (approved):
Nice! added only some non-blocking comments.

internal/builder/packages.go
```diff
@@ -15,8 +15,22 @@ import (
 	"github.com/pkg/errors"
 )
 
-// FindBuildDirectory locates the target build directory.
-func FindBuildDirectory() (string, bool, error) {
+// MustFindBuildDirectory function locates the target build directory. If the directory doesn't exist, it will create it.
+func MustFindBuildDirectory() (string, error) {
```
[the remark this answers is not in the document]
mtojek: Ah yes, I can rename it to just `BuildDirectory()`. This will fit too! Thanks for this feedback!

internal/testrunner/coverage_output.go
mtojek: Ah I know what you mean... We shouldn't mark a missing pipeline test if there are no pipelines at all. Sounds like a follow up.

cmd/testrunner.go
mtojek: The `--report-output` refers to test results, so I assumed it's something else. I don't see a value in printing the coverage to STDOUT, so I didn't bind these flags together.

Do you think we should introduce additional flag for `--coverage-output file` (or stdout) or is there low gain?

internal/testrunner/coverage_output.go
mtojek: I'm afraid we're overusing them here. I will adjust this file unit.

Do you know if there is any official recommendation written down? I must have missed reading it.

jsoriano: I think there is not, but for packages (and then for directories), there is a recommendation to don't use underscores. And for files, some suffixes have special meanings for go build, as the ones for architectures or operating systems, so in general I think that is a good idea to avoid them.

internal/testrunner/coverage_output.go
jsoriano: Yes, this would be for the future for sure, no need to do anything else by now :slightly_smiling_face:

cmd/testrunner.go
jsoriano: I was thinking that it may be weird to have files created even when there is only stdout output by default. But improving this, if needed, sounds like something for the future. For example we might show coverage info in the normal test reports, this could be nice, then `--report-output` would control the output of both features.

internal/testrunner/coverage_output.go
mtojek: Thanks for the suggestion, fixed!

internal/builder/packages.go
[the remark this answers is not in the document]
mtojek: Fixed

cmd/testrunner.go
mtojek: I see, let's keep it for further improvement if there is feature need for this feature.

internal/testrunner/coverage_output.go
mtojek: https://github.com/elastic/elastic-package/issues/442

jsoriano (approved):

PR #447, opened 2021-07-27
Adjust test coverage calculation

mtojek:
This PR adjusts test coverage calculation:

- [x] verify if test is actually expected (no pipeline defined = no need for pipeline tests)
- [x] don't include package context tests in calculation (artificially increases coverage, e.g. installed asset = test unit, so larger packages have better coverage)
- [x] consider multiple test cases as one (just check if tests are present or not) 

Testing:

Before - https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/master/168/cobertura/apache/
After - https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-447/1/cobertura/apache/

### DIFF
--- a/internal/testrunner/coverageoutput.go
+++ b/internal/testrunner/coverageoutput.go
@@ -159,8 +159,16 @@ func findDataStreamsWithoutTests(packageRootPath string, testType TestType) ([]s
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


jsoriano (approved):
:+1:

PR #451, opened 2021-07-29
Escape special characters in docs

mtojek:
Issue: https://github.com/elastic/integrations/issues/1366

This PR updates elastic-package to escape `* { } < >` characters while rendering docs.

### DIFF
--- a/internal/docs/exported_fields.go
+++ b/internal/docs/exported_fields.go
@@ -70,7 +70,10 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord) 
 	builder.WriteString("\n")
 	for _, c := range collected {
 		description := strings.TrimSpace(strings.ReplaceAll(c.description, "\n", " "))
-		builder.WriteString(fmt.Sprintf("| %s | %s | %s |", c.name, description, c.aType))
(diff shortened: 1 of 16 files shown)


internal/docs/exported_fields.go
```diff
@@ -70,7 +70,10 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord)
 	builder.WriteString("\n")
 	for _, c := range collected {
 		description := strings.TrimSpace(strings.ReplaceAll(c.description, "\n", " "))
-		builder.WriteString(fmt.Sprintf("| %s | %s | %s |", c.name, description, c.aType))
+		builder.WriteString(fmt.Sprintf("| %s | %s | %s |",
+			strings.ReplaceAll(c.name, "*", "\\*"),
+			strings.ReplaceAll(description, "*", "\\*"),
```
jsoriano, line 75: Nit. Consider using a [Replacer](https://pkg.go.dev/strings#NewReplacer), so we only need to keep a list of replacements.

jsoriano (approved):
:+1:

internal/docs/exported_fields.go
mtojek: Fixed

Issue #328, opened 2021-04-20
Use upstream's manifest to deploy agent on k8s for tests


ChrsMark:
Using upstream's manifests to deploy Agent on k8s for testing will keep testing process up to date with the proposed way of deploying agent on k8s. We will avoid hitting issues like https://github.com/elastic/elastic-package/pull/326.

Manifests: https://github.com/elastic/beats/blob/master/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml

Note: The above manifests are for standalone version so version for far so we need a way to use them with managed version too for testing.

MichaelKatsoulis:
After merging of https://github.com/elastic/beats/pull/26756 we have proposed manifests for fleet managed agent.
It would be nice for testing with elastic package to support multi node kubernetes cluster (maybe have it as default kind config [link](https://kind.sigs.k8s.io/docs/user/quick-start/#multinode-clusters)) and deploy agent as Daemonset . This approach resembles more to real world scenarios and helps test PRs like https://github.com/elastic/integrations/pull/1321

mtojek:
👍 

But we need to make sure that all required ENVs are present. See: https://github.com/elastic/elastic-package/blob/master/internal/install/static_kubernetes_elastic_agent_yml.go#L39

PR #452, opened 2021-07-29
Get elastic-agent-managed-daemonset.yaml from upstream 7.x instead of using a local static file

MichaelKatsoulis:
This PR aims to close https://github.com/elastic/elastic-package/issues/328.

There is no need anymore for maintaining a static go file for elastic-agent-managed deployment. 
Instead the upstream manifest (which creates a Daemonset) can be used. [manifest](https://raw.githubusercontent.com/elastic/beats/7.x/deploy/kubernetes/elastic-agent-managed-kubernetes.yaml
)

It is still in draft mode due to one dependancy.


internal/install/install.go
```diff
@@ -238,3 +246,20 @@ func createServiceLogsDir(elasticPackagePath *locations.LocationManager) error {
 	}
 	return nil
 }
+
+// DownloadFile will download a url from a path and return the response body as a string.
+func DownloadFile(filepath string, url string) (string, error) {
```
mtojek, line 251: nit: does it have to be exposed?

internal/install/install.go
```diff
@@ -184,9 +187,14 @@ func writeKubernetesDeployerResources(elasticPackagePath *locations.LocationMana
 		return errors.Wrap(err, "can't read application configuration")
 	}
 
+	elasticAgentManagedYaml, err := DownloadFile("elastic-agent-managed-kubernetes.yaml", elasticAgentManagedYamlUrl)
```
mtojek, line 190: I don't think that `filepath` is needed here.

internal/install/install.go
```diff
@@ -184,9 +187,14 @@ func writeKubernetesDeployerResources(elasticPackagePath *locations.LocationMana
...
+	if err != nil {
+		return errors.Wrap(err, "downloading elastic-agent-managed-kubernetes.yaml failed")
```
mtojek, line 192: I would replace the filename with something abstract in case the filename changes.

internal/install/install.go
```diff
@@ -238,3 +246,20 @@ func createServiceLogsDir(elasticPackagePath *locations.LocationManager) error {
 	}
 	return nil
 }
+
+// DownloadFile will download a url from a path and return the response body as a string.
+func DownloadFile(filepath string, url string) (string, error) {
+	// Get the data
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", err
```
mtojek, line 255: Please use the `errors.Wrap` to stick some context here.

internal/install/install.go
```diff
@@ -238,3 +246,20 @@ func createServiceLogsDir(elasticPackagePath *locations.LocationManager) error {
...
+	}
+	defer resp.Body.Close()
+
+	// Convert to string
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
```
mtojek, line 262: same here

internal/install/install.go
mtojek, line 190: WDYT to replace this function with less generic one like `downloadElasticAgentManagedYAML()`? I don't think we need generic `DownloadFile` function here.

mtojek (commented):
While I was reviewing this draft, I found a potential issue. This file will be downloaded during the installation procedure and be left there forever. If the remote file changes (new commits in 7.x), the elastic-package won't pick it up.
Also, the installation procedure doesn't require internet access so far, so this would be blocking. I think we should consider an alternative approach:

1. Remove all local `elastic-agent.yaml` definitions.
2. Replace the [logic](https://github.com/elastic/elastic-package/blob/master/internal/testrunner/runners/system/servicedeployer/kubernetes.go#L144) which uses local file with one downloading from remote location every time (you can also use the cache directory in `~/.elastic-package/cache`).

ChrsMark:
Great to see this happening! It will help us eliminate issues when changes are happening on Beats/Agent regarding the deployment model/roles etc.

> While I was reviewing this draft, I found a potential issue. This file will be downloaded during the installation procedure and be left there forever. If the remote file changes (new commits in 7.x), the elastic-package won't pick it up.
> Also, the installation procedure doesn't require internet access so far, so this would be blocking. I think we should consider an alternative approach:
> 

Why the file will stay there forever? I would expect a "download&replace" approach to work here. Or to be more consistent before downloading the deployer can try to `k delete -f` any old Agents and then download the new one and apply again. In case we want to save time and bandwidth we can download first and check the diff and if there is any do the update otherwise skip the delete+deploy steps.

internal/install/install.go
MichaelKatsoulis: no!

internal/install/install.go
MichaelKatsoulis: 😔 it is not

mtojek:
> Why the file will stay there forever? I would expect a "download&replace" approach to work here. 

No, it's a different pattern. Logic behind `EnsureInstalled` runs an embedded "installer" which create required files and directories in the `~/.elastic-package`. It's executed *only once* per application revision (if newer application is installed, then it will replace the directory), *on startup*.

> Or to be more consistent before downloading the deployer can try to k delete -f any old Agents and then download the new one and apply again. In case we want to save time and bandwidth we can download first and check the diff and if there is any do the update otherwise skip the delete+deploy steps.

That's why I suggested to [post](https://github.com/elastic/elastic-package/pull/452#pullrequestreview-718166736). We can download the YAML file in runtime and cache it locally if necessary.

test/packages/kubernetes/data_stream/pod/fields/ecs.yml
```diff
@@ -7,3 +7,7 @@
 - name: service.type
   type: keyword
   description: Service type
+- name: orchestrator.cluster.name
```
mtojek, line 10: To enable dependency management here, you need to include [build.yaml](https://github.com/elastic/integrations/blob/master/packages/mysql/_dev/build/build.yml) file.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -144,14 +150,54 @@ func findKubernetesDefinitions(definitionsDir string) ([]string, error) {
...
+	if err != nil {
+		return "", errors.Wrap(err, "can't read application configuration")
+	}
+	elasticAgentManagedYaml, err := downloadElasticAgentManagedYAML(elasticAgentManagedYamlURL)
+	if err != nil {
+		return "", errors.Wrapf(err, "downloading failed for file from source  %s", elasticAgentManagedYamlURL)
+	}
+	// Replace fleet url
+	elasticAgentManagedYaml = strings.ReplaceAll(elasticAgentManagedYaml,
+		"https://fleet-server:8220",
```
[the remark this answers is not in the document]
MichaelKatsoulis: This is a valid concern. I will change it

test/packages/kubernetes/data_stream/pod/fields/ecs.yml
MichaelKatsoulis: So I will just add this file

mtojek: You need to format it with `elastic-package format`:

> [2021-07-30T13:08:28.902Z] Error: checking package failed: formatting the integration failed (path: /var/lib/jenkins/workspace/t-manager_elastic-package_PR-452/src/github.com/elastic/elastic-package/test/packages/kubernetes, failFast: true): walking through the integration files failed: formatting file failed (path: /var/lib/jenkins/workspace/t-manager_elastic-package_PR-452/src/github.com/elastic/elastic-package/test/packages/kubernetes/_dev/build/build.yml): file is not formatted (path: /var/lib/jenkins/workspace/t-manager_elastic-package_PR-452/src/github.com/elastic/elastic-package/test/packages/kubernetes/_dev/build/build.yml)

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -144,14 +149,56 @@ func findKubernetesDefinitions(definitionsDir string) ([]string, error) {
...
+}
+
+// getElasticAgentYaml retrieves elastic-agent-managed.yaml from upstream and modifies the file as needed
+// to run locally.
+func getElasticAgentYaml() (string, error) {
+	appConfig, err := install.Configuration()
+	if err != nil {
+		return "", errors.Wrap(err, "can't read application configuration")
+	}
+	elasticAgentManagedYaml, err := downloadElasticAgentManagedYAML(elasticAgentManagedYamlURL)
```
mtojek, line 188: nit: YAML comes as bytes. Did you check if we can skip converting to string? It seems that byte-body will fit well here.

mtojek, line 188: nit: please add a logger debug indicating the download from upstream

internal/kubectl/kubectl_apply.go
```diff
@@ -84,6 +84,22 @@ func Apply(definitionPaths ...string) error {
 	return nil
 }
 
+// ApplyStdin function adds resources to the Kubernetes cluster based on provided stdin.
+func ApplyStdin(input string) error {
+	logger.Debugf("Apply Kubernetes stdin")
+	out, err := applyKubernetesResourcesStdin(input)
+	if err != nil {
+		return errors.Wrap(err, "can't modify Kubernetes resources (apply from stdin)")
```
mtojek, line 92: nit: I think you're right in L89, "apply stdin"

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -144,14 +149,56 @@ func findKubernetesDefinitions(definitionsDir string) ([]string, error) {
...
 	logger.Debug("install Elastic Agent in the Kubernetes cluster")
 
-	locationManager, err := locations.NewLocationManager()
+	elasticAgentManagedYaml, err := getElasticAgentYaml()
 	if err != nil {
-		return errors.Wrap(err, "can't locate Kubernetes file for Elastic Agent in ")
+		return errors.Wrap(err, "can't retrieve Kubernetes file for Elastic Agent")
 	}
 
-	err = kubectl.Apply(locationManager.KubernetesDeployerAgentYml())
```
mtojek, line 152: nit: logger debug: downloaded N bytes

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -144,14 +149,56 @@ func findKubernetesDefinitions(definitionsDir string) ([]string, error) {
...
+func downloadElasticAgentManagedYAML(url string) (string, error) {
+	// Get the data
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get file from url %s", url)
+	}
+	defer resp.Body.Close()
+
+	// Convert to string
+	b, err := io.ReadAll(resp.Body)
```
mtojek, line 174: is it necessary to go with string?

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -144,14 +149,56 @@ func findKubernetesDefinitions(definitionsDir string) ([]string, error) {
...
 	}
 	return nil
 }
+
+// downloadElasticAgentManagedYAML will download a url from a path and return the response body as a string.
+func downloadElasticAgentManagedYAML(url string) (string, error) {
+	// Get the data
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get file from url %s", url)
```
mtojek, line 169: nit: from URL

internal/kubectl/kubectl.go
```diff
@@ -49,3 +50,21 @@ func modifyKubernetesResources(action string, definitionPaths ...string) ([]byte
 	}
 	return output, nil
 }
+
+// applyKubernetesResourcesStdin applies a kubernetes manifest provided as stdin.
```
mtojek, line 54: nit: a Kubernetes manifest

internal/testrunner/runners/system/servicedeployer/kubernetes.go
MichaelKatsoulis: It is not necessary. I did that in order to use it later in Regexp `ReplaceAllString`. 
But I can instead use `ReplaceAll` that takes a slice of bytes as input.

mtojek: > But I can instead use ReplaceAll that takes a slice of bytes as input.

Sounds good to me

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -144,14 +149,58 @@ func findKubernetesDefinitions(definitionsDir string) ([]string, error) {
...
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read response body")
+	}
+	return b, nil
+}
+
+// getElasticAgentYaml retrieves elastic-agent-managed.yaml from upstream and modifies the file as needed
+// to run locally.
+func getElasticAgentYaml() ([]byte, error) {
```
mtojek, line 183: nit: `...YAML()`

mtojek (approved):

mtojek:
@MichaelKatsoulis Ship it, please.

PR #455, opened 2021-08-02
Use Go 1.16

mtojek:
This PR modifies `go.mod` to use Go 1.16 to build the application.

Spotted in: https://github.com/elastic/elastic-package/pull/456

PR #457, opened 2021-08-02
Fix: use mage env with goreleaser

mtojek:
This PR fixes the issue introduced in https://github.com/elastic/elastic-package/pull/455. Most likely it uses default `go` binary which is 1.15.

I was wrong with the assumption: https://github.com/elastic/elastic-package/pull/455#issuecomment-891071523 .


.ci/Jenkinsfile
```diff
@@ -74,9 +74,11 @@ pipeline {
         tag pattern: '(v)?\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
       }
       steps {
-        withCredentials([string(credentialsId: "${GITHUB_TOKEN_CREDENTIALS}", variable: 'GITHUB_TOKEN')]) {
-          dir("${BASE_DIR}") {
-            sh 'curl -sL https://git.io/goreleaser | bash'
+        withMageEnv(){
```
jsoriano, line 77: We should also set `GO_VERSION` environment variable, or add a `.go-version` file to pin the version of Go, if not we depend on the [default in the library](https://github.com/elastic/apm-pipeline-library/blob/683617e72b45b1fec13e9ac0f92fc059fc7d417b/vars/goDefaultVersion.groovy#L41).

Or we can also set a version here, as `withMageEnv(version: ...)`.

mtojek: Thanks for suggestion, added .go-version file.

jsoriano (approved):

mtojek:
I will merge this one to see if it improves the situation.

PR #467, opened 2021-08-05
[Discuss] How to support staging builds

ruflin:
Trying to test a staging build (not SNAPSHOT) so far does not seem to be possible with elastic-package. The reason is that the docker image url is slightly different. With the change I applied, for example the following works:

```
elastic-package stack update --version=8.0.0-alpha1-5d2ed18b -v
```

I wonder how we should support this best. Should we detect that if there is an additional hash that staging builds are used with a different path? Are there other builds with a hash? Additional flag?


ruflin:
@mtojek @jsoriano Would be great to get your input on the above.

internal/install/application_configuration_yml.go
```diff
@@ -5,9 +5,9 @@
 package install
 
 const (
-	elasticAgentImageName  = "docker.elastic.co/beats/elastic-agent"
-	elasticsearchImageName = "docker.elastic.co/elasticsearch/elasticsearch"
-	kibanaImageName        = "docker.elastic.co/kibana/kibana"
+	elasticAgentImageName  = "docker.elastic.co/staging/elastic-agent"
```
mtojek, line 8: I'm ok with this if only all these images are publicly available and published with the same cycle.

Otherwise I guess we need an additional command line (`--namespace`?) to override it with staging.

jsoriano: +1 to use an explicit flag for this.

ruflin: TBH it is kind of add that the "default" namespace is product specific but if it is staging it is a shared namespace. I'm sure there are historical reasons.

Images are public AFAIK.

mtojek: Do you know if `staging` also contains daily snapshots (ones we used to work with)?

ruflin: I don't think, only build candidates. But the build candidates are pretty important to test with ;-)

mtojek: I see. In this case we need a flag to change the namespace and support it internally too (Elastic Agent + Kubernetes service deployer). I suggest to convert this into an issue, so we can work on this (I guess that this draft is to show what is it about).

jsoriano: Wdyt about adding an explicit `--staging` flag to test with release candidates? This flag could change the namespaces or apply  any other logic that is needed to work with these versions. Without this flag current behaviour and image names would be maintained.

Other option may be to discover the images using the artifacts api instead of the docker registry, e.g. https://artifacts-api.elastic.co/v1/search/8.0.0-alpha1 has links to alpha1 archived docker images, using the same format as https://artifacts-api.elastic.co/v1/search/7.14.0 has for a released version. The good thing of this approach is that it would work the same for any version (replacing docker pull with a download + docker load), the bad thing is that I don't know if with this API we can discover specific snapshot tags (as `8.0.0-alpha1-5d2ed18b`), what is not nice for reproducibility.

In any case I am +1 with opening an issue for further discussion.

jsoriano: > the bad thing is that I don't know if with this API we can discover specific snapshot tags (as `8.0.0-alpha1-5d2ed18b`), what is not nice for reproducibility.

It is possible to discover and access specific builds with these endpoints:
* https://artifacts-api.elastic.co/v1/versions/8.0.0-alpha1
* https://artifacts-api.elastic.co/v1/versions/8.0.0-alpha1/builds/8.0.0-alpha1-3693e96c
* https://artifacts-api.elastic.co/v1/versions/8.0.0-alpha1/builds/8.0.0-alpha1-5d2ed18b

ruflin: I opened the issue here: https://github.com/elastic/package-spec/issues/205 I'll leave it to the two of you to decide if the PR should be closed or if you will build up on top of it.

ruflin:
@mtojek @jsoriano Do you remember by chance where we landed here? I'm closing the PR for now.

Issue #453, opened 2021-07-30
export dashboards: skip certificate verification


mtojek:
The goal of this issue is to skip certificate verification (command line option) when user runs `export dashboards`.

cc @akshay-saraswat

PR #468, opened 2021-08-05
export dashboards: add option to skip TLS verification

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/453

FYI @akshay-saraswat


internal/cobraext/const.go
```diff
@@ -63,6 +63,9 @@ const (
 	SkipPullRequestFlagName        = "skip-pull-request"
 	SkipPullRequestFlagDescription = "skip opening a new pull request"
 
+	SkipTLSVerifyFlagName        = "skip-tls-verify"
```
jsoriano, line 66: Nit. Use `tls` as prefix of the flag in case there are more tls flags in the future.
```suggestion
	SkipTLSVerifyFlagName        = "tls-skip-verify"
```

jsoriano (approved):

internal/cobraext/const.go
mtojek: Brain fart on my side, you're absolutely right about this. Working on this...

jsoriano (approved):

PR #474, opened 2021-08-09
Strip data streams from test Kubernetes

mtojek:
This PR strips test Kubernetes package from data streams redundant in terms of functionality (features to test).


MichaelKatsoulis (approved):
I think that some custom definitions like `cronjob, job, ss, rq, pv, pvc`  are not needed. All the rest look perfect

PR #476, opened 2021-08-10
Fix: geo_points can be imported from ECS

mtojek:
Reported on slack by @leehinman (thanks!) :

> I'm trying out the new generated ecs fields in elastic-package and I'm getting an odd error:
Build the package
Error: updating files failed: updating readme file README.md failed: rendering Readme failed: executing template failed: template: README.md:17:2: executing "README.md" at <fields "access">: error calling fields: collecting fields files failed: visiting fields failed: can't import field: field definition not found in schema (name: source.geo.location)
in the cached fields.ecs.yml source.geo.location is defined, and the others like source.geo.city_name work fine.
in ecs.ymlthe definition is:
> - external: ecs
  name: source.geo.location
which is the same as source.geo.city_name.  Any ideas on what might be wrong?

which is obviously a bug as all `geo.location` (type: geo_point) fields should be imported without problems. I hope this is only about adjusting the regexp.


jsoriano (approved):
Should we add a test for this case?

mtojek:
Converting to draft as I found some problems with validation (added static tests for this).

jsoriano (approved):
:+1:

PR #480, opened 2021-08-11
Use package-spec v1.0.0

mtojek:
This PR bumps up dependency on package-spec to v1.0.0.

Comment:

I tried to use `go get github.com/elastic/package-spec@v1.0.0` (and `go get github.com/elastic/package-spec/code/go@v1.0.0`) and `go mod tidy` to add v1.0.0 dependency to `go.mod`, but it always gets resolved to `v0.0.0-...`. I guess this is because of the `code/go/go.mod` directory. If you have an idea how to fix this issue (`go.mod` in subdirectory?`), I would be grateful.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/elastic/go-elasticsearch/v7 v7.14.0
 	github.com/elastic/go-licenser v0.3.1
 	github.com/elastic/go-ucfg v0.8.3
-	github.com/elastic/package-spec/code/go v0.0.0-20210623152222-b358e974b7f9
+	github.com/elastic/package-spec/code/go v0.0.0-20210811110254-a7173524bc76
 	github.com/fatih/color v1.12.0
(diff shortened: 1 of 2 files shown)


jsoriano:
> I tried to use `go get github.com/elastic/package-spec@v1.0.0` (and `go get github.com/elastic/package-spec/code/go@v1.0.0`) and `go mod tidy` to add v1.0.0 dependency to `go.mod`, but it always gets resolved to `v0.0.0-...`. I guess this is because of the `code/go/go.mod` directory. If you have an idea how to fix this issue (`go.mod` in subdirectory?`), I would be grateful.

I think it is not possible to serve tagged versions of packages under subdirectories of git repositories (see https://github.com/golang/go/issues/34055). I guess the only option would be to move the Go library to its own repository, or to remove the go.mod in the subdirectory, but then dependant projects would bring more dependencies.

PR #484, opened 2021-08-17
chore: decouple pipeline concerns

mdelapenya:
## What does this PR do?
It separates the concerns for running static analys checks from the execution of tests.

For that we are providing a new Make goal `make check-static`, which will run `build format lint licenser gomod update check-git-clean` goals. In the same manner, `check` will keep its former behaviour, as it will call `check-static` first, keeping build expectations.

Finally, we are adding a new stage in the pipeline, separating static checks from tests.

## Why is it important?
Separation of concerns: when the pipeline fails, we want to know in a first look what caused it.

## Follow-ups
We'd like to decouple the pipeline a little bit more right after the team approves/discusses this PR. Some ideas that we have in mind are:

- running unit tests (`make test-go`) in parallel with the static analysis.
- running stack-command, check-packages and profiles-command in separated stages, ideally in parallel.

But we'd like to find an agreement with the team on that first.

## Screenshots
**Before**:
Screenshot 2021-08-17 at 13 23 39

**After**:
Screenshot 2021-08-17 at 13 22 21

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -33,6 +33,7 @@ pipeline {
      */
     stage('Checkout') {
       steps {
+        pipelineManager([ cancelPreviousRunningBuilds: [ when: 'PR' ] ])
         deleteDir()
         gitCheckout(basedir: "${BASE_DIR}")
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -61,7 +61,7 @@ pipeline {
           withKubernetes() {
             withCloudTestEnv() {
               dir("${BASE_DIR}"){
-                sh(label: 'Check',script: 'make test')
+                sh(label: 'Check',script: 'make test check-git-clean')
               }
             }
           }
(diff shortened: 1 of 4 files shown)


.ci/Jenkinsfile
```diff
@@ -33,6 +33,7 @@ pipeline {
      */
     stage('Checkout') {
       steps {
+        pipelineManager([ cancelPreviousRunningBuilds: [ when: 'PR' ] ])
```
mdelapenya, line 36: This line will kill previous executions for a same PR, avoid consuming CI resources unnecessarily

.ci/Jenkinsfile
```diff
@@ -42,13 +43,26 @@ pipeline {
...
+        cleanup()
+        withMageEnv(){
+          dir("${BASE_DIR}"){
+            sh(label: 'Check',script: 'make check-static')
+          }
+        }
+      }
+    }
+    /**
+     Check the source code.
```
mdelapenya, line 56:
```suggestion
     Run the test suite.
```

.ci/Jenkinsfile
mdelapenya: @mtojek because we added this behaviour in a separate commit, we can rever it if we don't like it

jsoriano (approved):

PR #485, opened 2021-08-17
chore: decouple test concerns

mdelapenya:
## What does this PR do?
This PR is adding 4 things:

1. a specific Make goal for running Go tests on CI: `make test-go-ci`, which invokes the original `test-go` and generates an xUnit output. For that it installs two go libraries, which will be managed as project's tools, to parse test results and convert the test results into the XML formats accepted by Jenkins.
2. adds code coverage to the `make test-go` command
3. splits the `Tests` stage into two sequential stages:
   1. unit tests, running the new `make test-go-ci` goal
   2. integration tests, which at the same time spreads the tests in three parallel branches
      1. command tests
      2. package tests
      3. profile commands tests
4. A post-stage to collect the code coverage results, publishing to Jenkins just once. We are forced to use an intermediate upload to a GCP bucket to solve https://github.com/elastic/elastic-package/pull/485#issuecomment-901082403.

## Why is it important?
Separation of concerns: if the unit tests fail, we want to stop the soonest but we also want to know it super fast in the pipeline representation.

In 3) we are making explicit what artifacts are being published from each test type, which is also important for this separation of concerns we propose.

## Screenshots
Screenshot 2021-08-19 at 09 26 38

## Follow-ups
We could possibly extract the profile commands from the parallel branch, as it's super short, and move it right after the unit tests

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -152,19 +152,6 @@ pipeline {
           }
         }
       }
-      post {
-        always {
-          dir("${BASE_DIR}") {
-            archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/test-results/*.xml')
-            archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/kubectl-dump.txt')
-            archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/elastic-stack-dump/check/logs/*.log')
-            junit(allowEmptyResults: false,
-                keepLongStdio: true,
-                testResults: "build/test-results/*.xml")
-            coverageReport('build/test-coverage')
-          }
-        }
-      }
     }
     stage('Release') {
       when {
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -147,7 +147,7 @@ def generateTestCommandStage(Map args = [:]){
           archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/test-results/*.xml')
           archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/kubectl-dump.txt')
           archiveArtifacts(allowEmptyArchive: true, artifacts: 'build/elastic-stack-dump/**/*.log')
-          junit(allowEmptyResults: false,
+          junit(allowEmptyResults: true,
               keepLongStdio: true,
               testResults: "build/test-results/*.xml")
           coverageReport('build/test-coverage')
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -87,32 +87,11 @@ pipeline {
               cleanup()
               dir("${BASE_DIR}") {
                 script {
-                  def artifacts = [:]
-                  def testCommands = ['test-stack-command', 'test-check-packages', 'test-profiles-command']
-
-                  artifacts['test-stack-command'] = [
-                    'build/elastic-stack-dump/**/*.log'
-                  ]
-                  artifacts['test-check-packages'] = [
-                    'build/test-results/*.xml', 'build/kubectl-dump.txt', 'build/elastic-stack-dump/**/*.log'
-                  ]
-                  artifacts['test-profiles-command'] = []
-
-                  def parallelTasks = [:]
-
-                  testCommands.each { testCommand ->
-                    def junitArtifacts = false
-                    def publishCoverage = false
-
-                    if (testCommand == 'test-check-packages') {
-                      junitArtifacts = true
-                      publishCoverage = true
-                    }
-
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,3 +1,5 @@
+CODE_COVERAGE_REPORT_NAME_UNIT = $(PWD)/build/test-coverage/coverage-unit-report
```
mdelapenya, line 1: File name comes from https://github.com/elastic/apm-pipeline-library/blob/f460c26aea47055f07550cc56ae9e16572f9fe2d/vars/coverageReport.groovy#L41

.ci/Jenkinsfile
```diff
@@ -28,70 +28,107 @@ pipeline {
...
+                  def parallelTasks = [:]
+
+                  testCommands.each { testCommand ->
+                    def junitArtifacts = false
+                    def publishCoverage = false
+
+                    if (testCommand == 'test-check-packages') {
+                      junitArtifacts = true
+                      publishCoverage = true
+                    }
```
mdelapenya, line 110: Making explicit what each command is doing in terms of publishing CI artifacts. I was tempted to create a separate file, but I think here the code is simpler to read/maintain.

.ci/Jenkinsfile
```diff
@@ -28,70 +28,107 @@ pipeline {
...
-            dir("${BASE_DIR}") {
-              sh 'curl -sL https://git.io/goreleaser | bash'
+        stage('Integration Tests') {
+          failFast true
+          options { skipDefaultCheckout() }
+          steps {
+            withGithubNotify(context: 'Integration Tests', tab: 'tests') {
+              cleanup()
+              dir("${BASE_DIR}") {
+                script {
```
jsoriano, line 89: I struggled a bit to see what is this script executing at the end. Could this be moved to a function that receives the `testCommands` as parameter?

jsoriano: Or seeing https://github.com/elastic/elastic-package/pull/485/files#r690968480, could this be moved to a function for each command? So something like this is done here to chose the different options for each command:
```
parallel([
  'stack-command': integrationTest(command: 'test-stack-command'),
  'check-packages': integrationTest(command: 'test-check-packages`, junitArtifacts: true, publishCoverage: true),
  'profiles-command': integrationTest(command: 'test-profiles-command'),
])
```

jsoriano (changes requested):

.ci/Jenkinsfile
mdelapenya: Done in 58c4f66

Makefile
```diff
@@ -28,7 +30,18 @@ test-go:
...
 	# does not invalidate the cache so having the -count=1 to invalidate the test cache is useful.
-	go test -v -count 1 ./...
+	go test -v -count 1 -coverprofile=$(CODE_COVERAGE_REPORT_NAME_UNIT).out ./...
+
+# Prepare junit build context
+test-go-ci-pre:
+	mkdir -p $(PWD)/build/test-results
+	mkdir -p $(PWD)/build/test-coverage
+	GO111MODULE=off go get github.com/tebeka/go2xunit
+	GO111MODULE=off go get github.com/boumenot/gocover-cobertura
```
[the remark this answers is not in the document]
mdelapenya: Done in 0fff407

jsoriano (approved):
:+1:

mdelapenya:
Before merging this, I noticed that the code coverage reports are wrong: because we publish the unit tests report and then the integration tests one, it appears twice. I saw a  possible bug on Jenkins related to this: https://issues.jenkins.io/browse/JENKINS-59308

I'd send a new commit removing the publish step for the unit tests (as in the past), unless anybody in the @elastic/observablt-robots team knows if this bug has a workaround (apart from pushing the reports just once).

.ci/Jenkinsfile
```diff
@@ -160,14 +166,42 @@ def generateTestCommandStage(Map args = [:]){
...
+  stage('Publish coverage reports') {
+    dir("${BASE_DIR}") {
+      def bucketUri = getCoverageBucketURI() + "*.xml"
+      googleStorageDownload(bucketUri: bucketUri, credentialsId: "${JOB_GCS_CREDENTIALS}", localDirectory: 'build/test-coverage', pathPrefix: getCoveragePathPrefix())
+      coverageReport('build/test-coverage')
+    }
+  }
+}
+
+def stashCoverageReport() {
```
mdelapenya, line 195: Pushing to a GCP allows to publish the code coverage report just once.

@v1v, I believe this set of helpers can be wrapped into a new step, as I hard-copied them from the integrations pipeline.

PR #487, opened 2021-08-18
Filter directory listings by extension

adriansr:
This updates elastic-agent internals to filter for the expected extensions when listing directories in various processes, instead of processing all the files in a given directory.

Changed:
- data_streams/*/fields/: Only .yml files are loaded.
- _dev/build/docs/: Only .md files are considered.
- data_streams/*/elasticsearch/ingest_pipelines/: .yml and .json are
  loaded.

This prevents cryptic errors while developing packages, such as:

> ERROR: creating fields validator for data stream failed: can't load fields for data stream (path: /Users/adrian/go/src/github.com/elastic/integrations/packages/carbonback_edr/data_stream/log): unmarshalling field body failed: yaml: control characters are not allowed

In this case this all was caused by a `.swp` file being present in the directory as one of it's files was open in an editor.

### DIFF
--- a/internal/builder/external_fields.go
+++ b/internal/builder/external_fields.go
@@ -37,7 +37,7 @@ func resolveExternalFields(packageRoot, destinationDir string) error {
 		return errors.Wrap(err, "can't create field dependency manager")
 	}
 
-	fieldsFile, err := filepath.Glob(filepath.Join(destinationDir, "data_stream", "*", "fields", "*"))
+	fieldsFile, err := filepath.Glob(filepath.Join(destinationDir, "data_stream", "*", "fields", "*.yml"))
 	if err != nil {
 		return err
 	}
--- a/internal/packages/assets.go
+++ b/internal/packages/assets.go
@@ -122,8 +122,17 @@ func loadElasticsearchAssets(pkgRootPath string) ([]Asset, error) {
 
 		if dsManifest.Type == dataStreamTypeLogs || dsManifest.Type == dataStreamTypeTraces {
 			elasticsearchDirPath := filepath.Join(filepath.Dir(dsManifestPath), "elasticsearch", "ingest_pipeline")
-			pipelineFiles, _ := os.ReadDir(elasticsearchDirPath)
-			if pipelineFiles == nil || len(pipelineFiles) == 0 {
+
+			var pipelineFiles []string
+			for _, pattern := range []string{"*.json", "*.yml"} {
+				files, err := filepath.Glob(filepath.Join(elasticsearchDirPath, pattern))
+				if err != nil {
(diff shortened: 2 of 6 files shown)


jsoriano:
> In this case this all was caused by a `.swp` file being present in the directory as one of it's files was open in an editor.

Oh, this totally happens to me :D

internal/builder/external_fields.go
```diff
@@ -37,10 +37,20 @@ func resolveExternalFields(packageRoot, destinationDir string) error {
 		return errors.Wrap(err, "can't create field dependency manager")
 	}
 
-	fieldsFile, err := filepath.Glob(filepath.Join(destinationDir, "data_stream", "*", "fields", "*"))
+	fieldsPaths, err := filepath.Glob(filepath.Join(destinationDir, "data_stream", "*", "fields"))
```
jsoriano, line 40: Umm, wdyt about leveraging Glob directly here?
```suggestion
	fieldsPaths, err := filepath.Glob(filepath.Join(destinationDir, "data_stream", "*", "fields", "*.yml"))
```

internal/packages/assets.go
```diff
@@ -122,7 +123,7 @@ func loadElasticsearchAssets(pkgRootPath string) ([]Asset, error) {
 
 		if dsManifest.Type == dataStreamTypeLogs || dsManifest.Type == dataStreamTypeTraces {
 			elasticsearchDirPath := filepath.Join(filepath.Dir(dsManifestPath), "elasticsearch", "ingest_pipeline")
-			pipelineFiles, _ := os.ReadDir(elasticsearchDirPath)
+			pipelineFiles, _ := common.ListFilesByExtension(elasticsearchDirPath, ".yml", ".json")
```
jsoriano, line 126: For cases with multiple extensions this could be done like this:
```suggestion
			pipelineFiles, err := filepath.Glob(filepath.Join(elasticsearchDirPath, "?(.yml|.json)")
			...
```

This kind of glob syntax works in bash, I haven't tried it in go.

jsoriano (commented):
Looks good, but I wonder if we can directly leverage `filepath.Glob` instead of defining our own filter by extension.

I was worried about this filtering producing unexpected results if we add new files in the spec for these directories. But actually the code is safer with this filtering in this case, so let's do it.

internal/builder/external_fields.go
adriansr: That makes sense. I originally expected having to filter directories for multiple extensions, but turned out that's not the case most of the time. The helper function is unnecessary.

internal/packages/assets.go
adriansr: Unfortunately Go doesn't support this syntax, but merging the output of two Globs is trivial.

jsoriano (approved):

PR #502, opened 2021-09-02
Embed static files using embed package

jsoriano:
Several files were included as strings in Go source files. Leverage the `embed` package to move them to single files.

Also, include a manifest for elastic-agent in the kubernetes test runner, so it doesn't depend on Beats repository for that.


jsoriano:
Tests failing because of https://github.com/golang/go/issues/44354, opening PR to change linter (https://github.com/elastic/elastic-package/pull/503).

MichaelKatsoulis:
> Also, include a manifest for elastic-agent in the kubernetes test runner, so it doesn't depend on Beats repository for that.

@jsoriano there was an [issue](https://github.com/elastic/elastic-package/issues/328) to do exactly the opposite as long as we had a fleet-managed agent manifest in upstream beats repo. The point was not to maintain both at the same time. 
I am not sure why we want to change that. We just need to patch the manifest a bit to be fit for testing.

jsoriano:
> > Also, include a manifest for elastic-agent in the kubernetes test runner, so it doesn't depend on Beats repository for that.
> 
> @jsoriano there was an [issue](https://github.com/elastic/elastic-package/issues/328) to do exactly the opposite as long as we had a fleet-managed agent manifest in upstream beats repo. The point was not to maintain both at the same time.
> I am not sure why we want to change that. We just need to patch the manifest a bit to be fit for testing.

Downloading the manifest from the beats/elastic-agent repository adds a dependency into `elastic-package` that may unexpectedly affect the result of tests in integrations. Any change in this file can unexpectedly break `elastic-package` or the reproducibility of integrations tests. It also adds cognitive load to both, `elastic-package` and `elastic-agent` maintainers, that must be aware that any change in this file may end up affecting many other teams. It is also inconsistent with what we do with other runners such as docker compose, where the definition is already part of `elastic-package`.

`elastic-package` is meant to be the reference tool for developers of Elastic packages, inside and outside of Elastic. It should be as standalone as possible, and as a test runner, it should give consistent and reproducible results (same result for a given version of elastic-package and the package tested). For that I think that eventually we have to end-up using specific released versions of everything in `elastic-package` (no more 7.xs or SNAPSHOTs at least in the default behaviour), and introduce all changes in its code, in a controlled way.

A middle-ground solution for this could be to download an specific version from the beats repository (instead of one of a branch as is being done now) and continue doing the transformations here. But then we would miss the advantages of https://github.com/elastic/elastic-package/issues/328, and then I would prefer to directly have the manifest here.

We can explore additional ways of customizing this `elastic-agent` deployment in kubernetes from the package side. For example to solve the issue that originated https://github.com/elastic/elastic-package/issues/328, maybe another option would have been to add an additional `ClusterRole` to the manifests included for the integration, giving the additional required permission.

If we want to test if the latest manifest in the beats repo works well with the latest kubernetes integration, I think that we should do this in an [e2e-testing](https://github.com/elastic/e2e-testing) suite.

internal/profile/static.go
```diff
@@ -0,0 +1,78 @@
...
+
+// PackageRegistryConfigFile is the config file for the Elastic Package registry
+const PackageRegistryConfigFile configFile = "package-registry.config.yml"
+
+//go:embed _static/package_registry.yml
+var packageRegistryConfigYml string
+
+// newPackageRegistryConfig returns a Managed Config
+func newPackageRegistryConfig(_ string, profilePath string) (*simpleFile, error) {
+
```
mtojek, line 51: nit: empty line

internal/profile/static.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+)
+
+// SnapshotFile is the docker-compose snapshot.yml file name
+const SnapshotFile configFile = "snapshot.yml"
+
+//go:embed _static/docker-compose-stack.yml
+var snapshotYml string
+
+// newSnapshotFile returns a Managed Config
```
mtojek, line 19: nit: this comment is a bit misleading (not a Managed Config)

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -157,75 +153,25 @@ func installElasticAgentInCluster() error {
 	return nil
 }
 
-// downloadElasticAgentManagedYAML will download a url from a path and return the response body.
-func downloadElasticAgentManagedYAML(url string) ([]byte, error) {
```
mtojek, line 161: To be honest I'm on the fence with this change as it isn't inlined with idea we had before - have a single place where we store k8s definitions, so that the correct manifest is picked up without any updates in elastic-package's codebase.

Maybe we should discuss possible options here?

mtojek (approved):
We discussed and agreed offline to keep a customized copy of Kubernetes manifest here. Ship it!

Issue #475, opened 2021-08-10
Collect internal logs from elastic-agent container


mtojek:
The goal of this task is to collect internal logs from elastic-agent container (and replace the [workaround](https://github.com/elastic/integrations/pull/1445)):
* <any>beat's application log
* fleet server log

This [issue](https://github.com/elastic/kibana/issues/107514) confirmed that internal logs can be useful in debugging or root-causing hard to reproduces issues. Unfortunately such logs may contain confidential data (AWS credentials, env variables, passwords, infrastructure description), which can't be simply published online (as Jenkins artifacts).

I can see few options to address this issue:
1. Push internal logs to the dedicated bucket (internal only).
2. Censor internal logs and store together with other logs publicly.
3. Encrypt internal logs with internally known key and store  together with other logs publicly.

Any thoughts on this @elastic/observablt-robots @jsoriano ?

cachedout:
I think @v1v will probably want to weigh in here but I'd prefer option 1 then and option 3 as a second choice.

I do not believe we should pursue option 2.

mtojek:
There is already `dump` command, which doesn't pull log files from inside of containers:

```
➜  package-spec git:(prepare-for-1.0.1) elastic-package stack dump --help
Dump stack data for debug purposes

Usage:
  elastic-package stack dump [flags]

Flags:
  -h, --help            help for dump
      --output string   output location for the stack dump (default "elastic-stack-dump")

Global Flags:
  -p, --profile string   select a profile to use for the stack configuration. Can also be set with ELASTIC_PACKAGE_PROFILE (default "default")
  -v, --verbose          verbose mode
```

I suppose it can pull these files too and with Jenkins we can push them to a safe place.

mtojek:
I will see what I can do here.

mtojek:
I will follow up this issue in the context of https://github.com/elastic/elastic-package/pull/505#discussion_r705090295 .

mtojek:
Status update:

I managed to mount appropriate directories, but agent instances complain about required permissions? 

```
2021-09-13T15:55:01.056Z	INFO	log/reporter.go:40	2021-09-13T15:55:01Z - message: Application: filebeat--7.15.0-SNAPSHOT--36643631373035623733363936343635[665040cd-952021-09-13T15:54:42.228Z        INFO    cmd/enroll_cmd.go:543   Spawning Elastic Agent daemon as a subprocess to complete bootstrap process.
[...]

PR #505, opened 2021-09-08
Collect internal logs of fleet-server

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/475

This PR extends `dump` command functionality with support for internal logs. Let's first start with fleet-server logs.


internal/stack/logs.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, false, nil // we need to pull internal logs only from the Fleet Server container
+	}
+
+	p, err := compose.NewProject(DockerComposeProjectName)
+	if err != nil {
+		return nil, false, errors.Wrap(err, "could not create docker compose project")
+	}
+
+	serviceContainer := p.ContainerName(serviceName)
+	out, err := docker.Exec(serviceContainer, "sh", "-c", `find data/logs/default/fleet-server-json* -printf '%p\n' -exec cat {} \;`)
```
jsoriano, line 42: Would it be an option to mount the log directories as a volume, and get the logs directly from there?

jsoriano (approved):
Added a question/suggestion, but it may be more complicated, current implementation LGTM.

internal/stack/logs.go
mtojek: I would have to research stability of this. In general we had issues with volume mounts if directories are recreated in runtime or mounts got lost. But I like this idea, do you think we should do it here or in a followup?

There are also other files in this directory (filebeat and metricbeat logs), but we will have to store them in internal location (like GCP bucket) as they may contain confidential data.

jsoriano: Sure, we can do it as a follow up.

PR #506, opened 2021-09-13
Update dependency on elastic-package

mtojek:
This PR updates the dependency on package-spec to depend on proper Git tag.


mtojek:
self-approved, just dependency update

Issue #510, opened 2021-09-15
Add support for elastic-agent-complete to Elastic Package


dominiqueclarke:
The `elastic-package stack up` command currently enrolls a standard agent using the mainline elastic-agent image. 

We should configure the command to also enroll a complete agent using the `elastic-agent-complete` image, or provide a flag to determine which type of agent to enroll.

mtojek:
@dominiqueclarke Thank you for the reference PR, but we need to rethink how to support it. Adding an extra container with agent might be a performance overkill here, so maybe it's valid to replace/extend the original elastic-agent image instead.

The "elastic-agent-complete" means that it includes "heartbeat", right? Could you please share image details? How much is that image heavier than standard one. Maybe we can just switch it to complete.

dominiqueclarke:
Thanks @mtojek. The elastic-agent-complete image contains all the binaries necessary to run beats, plus additional dependencies required to run the synthetics agent via heartbeat. This includes node, chromium, playwright, the synthetics agent itself and more. 

Reference ticket: https://github.com/elastic/beats/issues/26398
Agent image (~400MB bigger than the original): https://container-library.elastic.co/r/beats/elastic-agent-complete?show_snapshots=true

The complete image will indeed cover the scope of existing capabilities of the mainline Elastic Agent, so it would make sense to switch the agent over to complete.

mtojek:
Version bump for elastic/integrations: https://github.com/elastic/integrations/pull/1748

mtojek:
Ok, I see that we need a small workaround for stacks < 7.15.0, where `elastic-agent-complete` wasn't available.

mtojek:
I guess we can close this issue as it's deployed in elastic/integrations.

PR #520, opened 2021-09-22
Use elastic-agent-complete Docker image

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/510


mtojek:
I did some manual testing and it went well.

PR #523, opened 2021-09-27
Update stack command

jlind23:
Update stack command in order to add a warn in case of low docker memory settings threshold.


mtojek (commented):
Let me explain what's behind the scenes.

The [readme tool](https://github.com/elastic/elastic-package/blob/master/tools/readme/main.go) collects usages for every command of the app and generates the README file.

To update the `stack up`, you need to touch this file: 
https://github.com/elastic/elastic-package/blob/4c8b1fff98cc69a1814d7e40c27a0831c5a76647/cmd/stack.go#L33

and then `make update`. 

In general it's for consistency between README file and application help.

PR #524, opened 2021-09-27
Update stack.go

jlind23:
Add a more explicit message for common error handling


mtojek (commented):
Could you please check if `make update` doesn't regenerate the README file? and submit it if the file is different.

mtojek (approved):

PR #530, opened 2021-10-04
Import external references in the right order

marc-gr:
External fields were imported after being used for the first time, causing some checks ie: `isNumericKeyword` to fail.


mtojek (approved):
Valid concern, thanks for opening the PR!

PR #531, opened 2021-10-04
Fix check to stop looking for build path

marc-gr:
While executing `elastic-package stack up` or others, a search for the `build` dir is done. Prior to this change, in windows platforms this ended up in an infinite loop since it was matching against `/` which was never found. This change builds the root path using `VolumeName` and `PathSeparator` to build the root path based on the current platform.

### DIFF
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -5,6 +5,7 @@
 package builder
 
 import (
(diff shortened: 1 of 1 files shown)


mtojek (approved):
LGTM. Thanks @marc-gr !

PR #532, opened 2021-10-05
feat: support using a custom kibana ref when running stack up/update commands

mdelapenya:
## What does this PR do?
It adds the ability to override kibana ref reading from  the `KIBANA_IMAGE_REF_OVERRIDE` environment variable.

## Why is it important?
As an elastic-package consumer, we have a use case where we need to run the stack with custom Kibana images, basically generated from a Kibana PR, so that we are able to run the e2e-testing framework against that custom kibana. With this change now it's possible to run the stack with a custom kibana image.

In particular, our custom pipeline to run the e2e tests for a Kibana PR is broken until we support this use case.

## How to test this locally?

```shell
KIBANA_IMAGE_REF_OVERRIDE=docker.elastic.co/observability-ci/kibana:pr105419 go run main.go stack up -v
```

The tool will UP the stack using this image for Kibana.

## Follow-up & other concerns
AS this is my first contribution to this tool, feel free to correct me and/or point me to other use cases that I'm not covering in this PR :pray:

### DIFF
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -48,7 +48,7 @@ func (s stack) ImageRefOverridesForVersion(version string) ImageRefs {
 
 	elasticAgentRefOverride := checkImageRefOverride("ELASTIC_AGENT_IMAGE_REF_OVERRIDE", DefaultStackVersion)
 	elasticsearchRefOverride := checkImageRefOverride("ELASTICSEARCH_IMAGE_REF_OVERRIDE", DefaultStackVersion)
-	kibanaRefOverride := checkImageRefOverride("KIBANA_IMAGE_REF_OVERRIDE", DefaultKibanaRef)
+	kibanaRefOverride := checkImageRefOverride("KIBANA_IMAGE_REF_OVERRIDE", fmt.Sprintf("%s:%s", kibanaImageName, DefaultStackVersion))
 
 	if !ok {
 		return ImageRefs{
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -92,15 +92,6 @@ func (ac *ApplicationConfiguration) StackImageRefs(version string) ImageRefs {
 	return refs
 }
 
-// StackImageOverrideRefs function selects the appropriate set of Docker image references for the given stack version, including overriden versions.
-func (ac *ApplicationConfiguration) StackImageOverrideRefs(version string, kibanaRef string) ImageRefs {
-	refs := ac.c.Stack.ImageRefOverridesForVersion(version)
-	refs.ElasticAgent = stringOrDefault(refs.ElasticAgent, fmt.Sprintf("%s:%s", selectElasticAgentImageName(version), version))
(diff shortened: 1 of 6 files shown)


internal/stack/options.go
```diff
@@ -11,6 +11,9 @@ type Options struct {
 	DaemonMode   bool
 	StackVersion string
 
+	// ref overrides
+	KibanaRefOverride string
```
mtojek, line 15: I suggest not exposing additional flag for Kibana as we will for sure need more like this in the feature.

Actually there might be a simpler way to do this, via `stack.ImageRefOverridesForVersion`. You can try to modify **only** this function to read special ENV vars: `ELASTIC_AGENT_IMAGE_REF_OVERRIDE`, `KIBANA_IMAGE_REF_OVERRIDE`, etc. and give it priority to `image_ref_overrides` from `config.yml`.

BTW this `config.yml` is yet another way to globally overide image references. It's used to replace faulty images in case of emergency)

mtojek (changes requested):

internal/stack/options.go
mdelapenya: I removed the flag and used an env var to get the custom kibana ref. Please let me know your thoughts, thanks!

mdelapenya: @mtojek wait, I did not read it properly. Will work on a new commit to modify ONLY `ImageRefOverridesForVersion `. Sorry!

mdelapenya: I think it's now ready, thanks for the review!

internal/install/application_configuration.go
```diff
@@ -34,11 +34,31 @@ type stack struct {
...
 
+func checkImageRefOverride(envVar string, fallback string) string {
+	logger.Infof("Checking %s, fallback: %s", envVar, fallback)
+	refOverride := os.Getenv(envVar)
+	if refOverride == "" {
+		return fallback
+	}
+
+	logger.Infof("Setting %s %s", envVar, refOverride)
+
```
mdelapenya, line 45:
```suggestion
```

internal/install/application_configuration.go
```diff
@@ -34,11 +34,28 @@ type stack struct {
...
+		return fallback
+	}
+
+	return refOverride
+}
+
 func (s stack) ImageRefOverridesForVersion(version string) ImageRefs {
 	refs, ok := s.ImageRefOverrides[version]
+
+	kibanaRefOverride := checkImageRefOverride("KIBANA_IMAGE_REF_OVERRIDE", DefaultKibanaRef)
```
mtojek, line 49: I guess you don't have to define to `DefaultKibanaRef` as it's already there: https://github.com/elastic/elastic-package/blob/4c8b1fff98cc69a1814d7e40c27a0831c5a76647/internal/install/application_configuration_yml.go#L11

mtojek, line 49: Is there a chance that you will also need other Refs? for Elasticsearch and Agent? Maybe it's valid to introduce them as well.

mtojek (commented):
Much, much cleaner implementation. Good job, Manu! I left only minor comments.

internal/install/application_configuration.go
mdelapenya: I have them in my local branch 😅

Issue #519, opened 2021-09-22
Collect unsafe logs of service containers


mtojek:
Follow-up on https://github.com/elastic/elastic-package/issues/475 

The `elastic-package test system` runs service containers that produce some logs, which might contain sensible data. So far we ignored these logs, but as long as we're pushing other logs to a safe place, we can do the same for service containers.

mtojek:
This is the first candidate to find the real root cause, why is it flaky: https://github.com/elastic/integrations/issues/1761

mtojek:
I will consider this a part-time job (in progress) as it can help with operations, especially debugging flaky tests.

mtojek:
All PRs merged, feature is enabled. Resolving.

PR #534, opened 2021-10-06
Collect service container logs

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/519

This PR extends the Docker Compose service deployer to collect service logs. Once this is merged and released, I will adjust elastic/integrations to collect these logs using safe log collection.

### DIFF
--- a/internal/testrunner/runners/system/servicedeployer/compose.go
+++ b/internal/testrunner/runners/system/servicedeployer/compose.go
@@ -6,9 +6,14 @@ package servicedeployer
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -151,15 +156,19 @@ func (s *dockerComposeDeployedService) TearDown() error {
 		return errors.Wrap(err, "could not create Docker Compose project for service")
 	}
 
-	downOptions := compose.CommandOptions{
+	containerLogs, err := p.Logs(compose.CommandOptions{
```
jsoriano, line 159: Having the whole logs in memory can be a lot in some cases, would it be possible to pipe them and receive here a reader?

jsoriano (approved):
Looks good, optimization proposed but could be done later or if this is an issue at some point.

PR #538, opened 2021-10-12
ci: use withMageEnv after the dir

v1v:
### What

Run `withMageEnv` within the BASE_DIR folder in order to read the `.go-version` file. Otherwise it honours the default go version set in the shared library step.

Add the tools to the PATH when running the k8s/kind stuff.

### Before

image

### After

image

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -215,7 +215,9 @@ def withKubernetes(Closure body) {
        chmod +x ${HOME}/bin/kubectl
        kubectl version --client
        ''') }
-    body()
+    withEnv(["PATH+TOOLS=${HOME}/bin"]) {
+      body()
+    }
 }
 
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -144,10 +144,10 @@ def generateTestCommandStage(Map args = [:]){
     withNode(labels: "ubuntu-18 && immutable", sleepMax: 20, forceWorkspace: true) {
       cleanup()
       try {
-        withMageEnv(){
+        dir("${BASE_DIR}"){
```
mtojek, line 147: It looks like changing the order of execution in this stub, changed path visibility?

v1v: Aha, it uses the worker go version maybe, let me put a fix and run `dir` in conjunction with `withMageEnv`

PR #549, opened 2021-10-20
[pre-commit] enable it

v1v:
🚧 

Just a proposal! based on the same set of rules defined in https://github.com/elastic/e2e-testing/blob/master/.pre-commit-config.yaml


mtojek:
It's a bit stale. Let's close it for now.

PR #571, opened 2021-11-04
Add guidelines about the supported version of the stack

jsoriano:
With the upcoming 8.0 release, we have to decide what kibana versions constrains to use for packages compatible with 8.x. We have had some discussions about this, let's formalize it a bit more.

Drafting a document with guidelines for package developers.

### DIFF
--- a/docs/howto/stack_version_support.md
+++ b/docs/howto/stack_version_support.md
@@ -1,0 +1,76 @@
+Intro...
+
+TBD: When to reconsider the version constraint.
+
+
+Support for all versions since the release of the package
+----
+Example: `>=7.16.0`.
+
+With this approach, the released package can be used in any version since the
+release of the package.
+
+Pros:
+* No need to keep multiple development branches, no need for backports.
+
+Cons:
+* False premise, difficult to ensure compatibility with future majors.
+
+Recommendation: Not recommended.
+
+Support for multiple majors
+----
+Example: `^7.14.0 || ^8.0.0`.
+
+With this approach, the released package can be used in any tested major version
+since the release of the package.
+
+Pros:
+* No need to keep multiple development branches, no need for backports.
+* Less disruptive for users.
+
+Cons:
+* Testing should be done with more versions.
+* May require compatibility code. 
+
+Recommendation: Good for simple packages monitoring services that don't need
+many special features, and have low coupling with specific versions of the stack.
+
+Support for current major
+----
+Example: `^8.0.0`
+
+With this approach, the released package can be used only in the specified
+major.
+
+Pros:
+* Testing need to be done with less versions.
+* More open to use newer features.
+
+Cons:
+* May require multiple development branches and backport to support multiple
+  majors.
+
+Recommendation: Good option for packages that use features not available in
+older majors, or that introduce breaking changes.
+
+Support for current minor, or to specific versions
+----
+Example: `~8.1.0` or `8.1.0`
+
+With this approach, the released package can be used only in the specified
+minor.
+
+Pros:
+* Testing is limited to more specific versions.
+* Even more open to the use of the latest features.
+
+Cons:
+* Requires one development branch per version of the stack, couples package
+  lifecycle to the stack, more prone to need backports.
+* Releases are needed for each version of the stack.
+* Stack updates may be affected by updates of these packages, difficult to
+  provide compatibility during upgrades.
+
+Recommendation: Good option for packages that are highly coupled to specific
+versions of the stack, or that may be even bundled with it.
--- a/docs/howto/stack_version_support.md
+++ b/docs/howto/stack_version_support.md
@@ -3,24 +3,6 @@ Intro...
 TBD: When to reconsider the version constraint.
 
 
-Support for all versions since the release of the package
-----
-Example: `>=7.16.0`.
-
-With this approach, the released package can be used in any version since the
-release of the package.
-
-Pros:
-* No need to keep multiple development branches, no need for backports.
-
-Cons:
-* False premise, difficult to ensure compatibility with future majors.
-* In case of broken compatibility, we have to break contracts and
-  remove/replace package in the storage.
-
-Recommendation: Not recommended. This caused problems in the past in some
-packages and was replaced by `^`.
-
 Support for multiple majors
 ----
 Examples: `^7.14.0 || ^8.0.0`, `^7.14.0 || ^8.0.0 || ^9.0.0`
@@ -36,11 +18,38 @@ Pros:
 Cons:
 * Testing should be done with more versions.
 * May require compatibility code.
+* Deprecation path needed if package removed before next major.
 * There are differences between major versions that have influence on test
   results (for example: different GeoIP databases).
 
 Recommendation: Good for simple packages monitoring services that don't need
 many special features, and have low coupling with specific versions of the stack.
+
+
+Support for current major
+----
+Examples: `^8.0.0`, `^8.2.0`
+
+With this approach, the released package can be used only in the specified
+major.
+
+Pros:
+* Testing need to be done with less versions.
+* Changes don't need to take into account older versions.
+* More open to use newer features of the stack.
+
+Cons:
+* May require multiple development branches and backport to support multiple
+  majors. This is specially a problem in repositories with multiple packages.
+* Deprecation path needed if package removed before next major.
+* Stack updates may be affected by updates of these packages, difficult to
+  provide compatibility during upgrades, though frictions may be more acceptable
+  in upgrades between majors.
+
+Recommendation: Good option for packages that use features not available in older
+majors, or that introduce breaking changes. Also recommended if the maintainer
+decides to don't provide bugfixes for previous major.
+
 
 Support for last minors of previous major and first minors of the next
 ----
@@ -67,28 +76,6 @@ Recommendation: Good for cases where package developers need or want to take
 a defensive stand on the supported versions and have the capacity to follow
 the release train of the stack.
 
-Support for current major
-----
-Examples: `^8.0.0`, `^8.2.0`
-
-With this approach, the released package can be used only in the specified
-major.
-
-Pros:
-* Testing need to be done with less versions.
-* Changes don't need to take into account older versions.
-* More open to use newer features of the stack.
-
-Cons:
-* May require multiple development branches and backport to support multiple
-  majors. This is specially a problem in repositories with multiple packages.
-* Stack updates may be affected by updates of these packages, difficult to
-  provide compatibility during upgrades, though frictions may be more acceptable
-  in upgrades between majors.
-
-Recommendation: Good option for packages that use features not available in
-older majors, or that introduce breaking changes. Or if the maintainer decides
-to don't provide bugfixes for previous major.
 
 Support for current minor, or to specific versions
 ----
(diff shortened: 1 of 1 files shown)


jsoriano:
FTR, current packages tested with 8.0 snapshot: https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Fintegrations/detail/master/1123/pipeline.
There are problems in ~40% of packages, this means that in principle 60% of current packages could be already compatible with 8.0.
Of this 40%, a big percentage has failures because of the different geoip values (8.0 uses a different geoip database than 7.x).

mtojek:
> There are problems in ~40% of packages, this means that in principle 60% of current packages could be already compatible with 8.0.
Of this 40%, a big percentage has failures because of the different geoip values (8.0 uses a different geoip database than 7.x).

If this is the only problem, maybe we should pick the low hanging fruit and use the embedded database everywhere, not just in 8.x.

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
+Intro...
+
+TBD: When to reconsider the version constraint.
+
+
+Support for all versions since the release of the package
+----
+Example: `>=7.16.0`.
```
mtojek, line 8: If you described this case, then you should also cover simple `^7.16.0` (packages won't be present in 8.0.0)

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+
+With this approach, the released package can be used in any version since the
+release of the package.
+
+Pros:
+* No need to keep multiple development branches, no need for backports.
+
+Cons:
+* False premise, difficult to ensure compatibility with future majors.
```
mtojek, line 17: Maybe another bulletpoint:
In case of broken compatibility, we have to break contracts and remove/replace package in the storage.

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+
+Recommendation: Not recommended.
```
mtojek, line 19: Not recommended. We replaced this constraint some time ago.

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+With this approach, the released package can be used in any tested major version
+since the release of the package.
+
+Pros:
+* No need to keep multiple development branches, no need for backports, same
+  package for 7.x and 8.x.
+* Less disruptive for users.
+
+Cons:
+* Testing should be done with more versions.
```
mtojek, line 34: nit: manual testing

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+Pros:
+* Testing need to be done with less versions.
+* Changes don't need to take into account older versions.
+* More open to use newer features of the stack.
+
+Cons:
+* May require multiple development branches and backport to support multiple
+  majors. This is specially a problem in repositories with multiple packages.
+
+Recommendation: Good option for packages that use features not available in
```
mtojek, line 58: nit: ... or you agree to not provide any bugfixes to 7.x.

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+* Even more open to the use of the latest features.
+
+Cons:
+* Requires one development branch per version of the stack, couples package
+  lifecycle to the stack, more prone to need backports.
+* Releases are needed for each version of the stack.
+* Stack updates may be affected by updates of these packages, difficult to
+  provide compatibility during upgrades.
+
+Recommendation: Good option for packages that are highly coupled to specific
```
mtojek, line 79: Good option for early alpha, 0.0.1/experimental packages, PoCs, ones that will be possibly replaced with other packages in the future.

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+* No need to keep multiple development branches, no need for backports.
+
+Cons:
+* False premise, difficult to ensure compatibility with future majors.
+
+Recommendation: Not recommended.
+
+Support for multiple majors
+----
+Example: `^7.14.0 || ^8.0.0`.
```
mtojek, line 23: I would add a general note to this one:

* support matrix can be extended in the future with: `^7.14.0 || ^8.0.0 || ^9.0.0`

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+
+With this approach, the released package can be used in any tested major version
+since the release of the package.
+
+Pros:
+* No need to keep multiple development branches, no need for backports, same
+  package for 7.x and 8.x.
+* Less disruptive for users.
+
+Cons:
```
mtojek, line 33: * you may end up with a train: `^7.16.0 || ^8.0.0 || ^9.0.0 || ^10.0.0 || ^11.0.0 || ^12.0.0`

docs/howto/stack_version_support.md
ruflin: ++

docs/howto/stack_version_support.md
```diff
@@ -0,0 +1,80 @@
...
+With this approach, the released package can be used only in the specified
+major.
+
+Pros:
+* Testing need to be done with less versions.
+* Changes don't need to take into account older versions.
+* More open to use newer features of the stack.
+
+Cons:
+* May require multiple development branches and backport to support multiple
```
ruflin, line 55: I wonder if there is also an issue here with upgrade. Ideally a user could upgrade to an 8.0 compatible package already in 7.16. If we don't have packages compatible with both versions it would in theory mean for some moments, there is no compatible package installed as packages are only upgraded after the stack. This is also why Elastic Agent must be compatible with 8.0 for example, as ES is upgraded first.

ruflin (commented):
Left some comments but also some high level thoughts. We need to decide a short and long term solution. Potentially these are the same but don't have to.

To make sure the user has always a valid package installed, I think we need packages compatible with 7.16 and 8.0 (or beyond). This could be only a single release of the package which has this but ideally it should be there.

^8.0.0 sounds to me like the intuitive choice. The main reason I propose 8.0.x is to buy us some time to make sure we make the right decision. So shipping ^7.16.0 || 8.0.x would be for a single version of the package and not necessarily what we decide long term.

docs/howto/stack_version_support.md
jsoriano: Adding it.

docs/howto/stack_version_support.md
jsoriano: Adding this.

docs/howto/stack_version_support.md
jsoriano: Why would this be a con? If a package developer needs to maintain a package for all these versions, wouldn't it be better if they can do it with a single package instead of with many of them?

docs/howto/stack_version_support.md
jsoriano: I refer to testing in general in all appearances of the term here. In general we should be able to run CI with multiple versions, independently of them being 7.16.0 and 8.1.0, or 8.0.0 and 8.2.0.
The difference here is that the same package is supporting more versions than the alternative where there would be a package for 7.x versions and another one for 8.x. Though at the end the difference may be not so big.

docs/howto/stack_version_support.md
jsoriano: Ok, I will add a con about this, though in majors I would accept a major level of friction than in minors.

docs/howto/stack_version_support.md
jsoriano: Added as additional example.

docs/howto/stack_version_support.md
jsoriano: Added, not sure if this is an option for us for supported/ga packages, but it can be an option for unstable/beta versions, or for community packages in the future.

docs/howto/stack_version_support.md
jsoriano: Good point, added :+1:

jsoriano (commented):
Thanks @mtojek and @ruflin for your comments!

jsoriano:
@ruflin I have added a section about the `^7.16.0 ||  8.0.x` case. The main difference I see with other cases is that it allows us to delay long-term decisions, and that new releases are needed on each minor till longer-term decisions are taken. For teams that are able to dedicate efforts on each release of the stack this may be the safest option, but it can cause problems to users of less maintained packages.

jsoriano:
> We need to decide a short and long term solution. Potentially these are the same but don't have to.

And I would say more, we don't need to take the same decision for all packages. For example the situation of APM and Apache packages is very different, and some teams may have more resources to dedicate to releases (to just bump versions) than others.

docs/howto/stack_version_support.md
mtojek: It's easy to forget about extending the version range - new person, new hire, person unaware of such constraint. It's my personal opinion, but "exposing the package until revoked/withdrawn" seems to be natural to me. I can't achieve that with just ">= 7.16.0" constraint.

docs/howto/stack_version_support.md
mtojek: I think you can add an issue to the backlog to implement the support matrix which will cover next versions. I suppose it should be executing in the background, not blocking any code pushes, just to prevent few hours CI runs.

mtojek:
> We need to decide a short and long term solution. Potentially these are the same but don't have to.

We can't give orders to package owners. They are mature and can take decisions based on customer needs, operational requirements or team capacity.

Please consider the message below as 100% opinionated, personal:

Regarding the `^7.16.0 || 8.0.x` or `^7.16.0 || ^8.0.0` - I'm not fan of this solution as I see this as a short term plan, just to buy more time to decide what to do in 8.x.x. It doesn't really solve the problem with backporting, because sooner or later a developer will come with a new package, targeting the `^8.4.0` release, and will make backporting impossible (officially). As it was noticed before there is still operational activity required to 

I guess we should evaluate for how long we allow for backporting and what is its real influence on customers. If stack upgrades are doable and not so complex, maybe we can advise them just to update it?

ruflin:
After a discussion with @mtojek and @jsoriano my thinking is currently as following:

* There must be at least 1 package that is compatible with 7.16 and 8.0 to make sure during an upgrade there is still a valid package around. The exception here can be packages like apm that are upgraded with the stack and are in sync. 
* We don't know today how long a package will be compatible / working or should be not supported anymore. For this we need a more generic solution (see further below)
* In theory packages should be compatible with the current major but we don't know if it is the case for the next major. I think if someone did not test a package with the next major, it should not automatically work.

With the above, my suggestion for most packages at the moment would be to use `^7.16.0 || ^8.0.0`. This also allows any package developer to still ship packages to 7.16 and ensures it is not automatically compatible with 9.0 (see last point above).

What about if a package in 8.4 is deprecated and should not be shown in 8.5 again. There is a thread started here around this: https://github.com/elastic/package-spec/issues/227 One of the ideas is that even if a package was put in with the compatibility of `^8.0.0` but a newer package is available that states `<=8.4.x` + a deprecation / removal message, in 8.5 the package will not be shown anymore by the registry. It is the registry that has the magic inside to stop publishing packages to newer versions of Kibana. The old Kibana versions will still see the package.

docs/howto/stack_version_support.md
jsoriano: Yeah, good point, added: https://github.com/elastic/integrations/issues/2134

docs/howto/stack_version_support.md
jsoriano: Umm, so this is really a con for all cases except `>=`, right? In all the other cases you need to modify the constraints every time a new version needs to be supported.

mtojek: Yes, that's unfortunate. I don't see a good alternative solution to address this problem, so let's go with the best one we see now.

jsoriano:
Added intro with guidelines about what conditions to use and when to update them. Opening for review.

andresrc:
Adding my thoughts here. Mostly aligned with https://github.com/elastic/elastic-package/pull/571#issuecomment-961750162:
* I would like to minimize the need to perform releases just to change version. Having said this there are several integrations that should be aligned with the stack (APM, etc). Probably elastic-agent and fleet-server should fall into this group as well.
* With "Make it Minor", unless versioning scheme changes (it seems it is still an open point) `^7.16.0 || ^8.0.0` is equivalent to `>=7.16.0`.
* The main issue is when we need to deprecate, supersede a package. For that, I agree with https://github.com/elastic/elastic-package/pull/571#issuecomment-961750162, though we need to acknowledge that users will be able to install the deprecated version manually, given that deprecation information comes "from a future version". If there's an active connection to the registry, we might be able to flag these situations (at some point in the future).

jlind23 (approved):

mtojek (approved):

PR #572, opened 2021-11-04
Sign cryptographically zipped packages

mtojek:
Issue: https://github.com/elastic/package-registry/issues/728

The goal of this PR is to extend the `build` action to sign cryptographically zipped packages. It uses the [GopenPGP](https://gopenpgp.org/) library.

TODO:
- [x] verify if all signer's environment variables are present
- [x] sign a package using [gopenpgp](https://github.com/ProtonMail/gopenpgp)
- [x] add test scripts for signing
- [ ] store the signature in a safe place accessible for the CI


internal/files/sign.go
```diff
@@ -0,0 +1,110 @@
...
+	if signerPrivateKeyfile == "" {
+		return fmt.Errorf("signer private keyfile is required. Please define it with environment variable %s", signerPrivateKeyfileEnv)
+	}
+
+	_, err := os.Stat(signerPrivateKeyfile)
+	if err != nil {
+		return errors.Wrap(err, "can't access the signer private keyfile")
+	}
+
+	signerPassphrase := os.Getenv(signerPassphraseEnv)
```
jsoriano, line 44: Nit. Not sure if an empty passphrase is a thing, you may want to use `os.LookupEnv` to distinguish unset variable from variable set with an empty string.

internal/files/sign.go
```diff
@@ -0,0 +1,110 @@
...
+		return errors.Wrap(err, "signingKey.Unlock failed")
+	}
+	defer unlockedKey.ClearPrivateParams()
+
+	keyRing, err := crypto.NewKeyRing(unlockedKey)
+	if err != nil {
+		return errors.Wrap(err, "crypto.NewKeyRing failed")
+	}
+
+	message := crypto.NewPlainMessage(data)
```
jsoriano, line 85: Is it possible to build the message from a reader to avoid having the whole package in memory?

internal/files/sign.go
jsoriano, line 44: When supporting community contributors we may need other sources of passphrases, as the OS keyrings or asking them in the CLI. It can be risky to keep the password in an environment variable for long.
We may also consider supporting Vault, so keys are obtained directly from there.
But we are good by now with this method.

jsoriano (approved):
Looks good, added some comments but nothing blocking.

> store the signature in a safe place accessible for the CI

This makes me think again about having a separate command to handle signatures. This would allow to build the package and to sign it in two different stages, maybe in two different processes in two different machines.

Another possible reason to have an specific subcommand for signatures is to validate them, so giving a package and a signature you can use `elastic-package` to check that the signature is correct. Though not sure if this feature belongs to `elastic-package`.

internal/files/sign.go
mtojek: I can experiment with streaming APIs as there are some here: https://github.com/ProtonMail/gopenpgp/blob/master/crypto/

internal/files/sign.go
mtojek: Good ideas but as candidates for further improvement. Let's deliver the first few iterations of the solution.

mtojek: It's rather a question about the "no password" option. Currently, the implementation doesn't allow for this mode (key with the password required). WDYT?

mtojek:
> This makes me think again about having a separate command to handle signatures. This would allow to build the package and to sign it in two different stages, maybe in two different processes in two different machines.

I will prepare a draft presenting different options (pipeline vs web service vs "in place " build and sign).

> Another possible reason to have an specific subcommand for signatures is to validate them, so giving a package and a signature you can use elastic-package to check that the signature is correct. Though not sure if this feature belongs to elastic-package.

It seems to be more for development and testing purposes than package delivery. I can create a follow-up issue for this. At the moment, it's crucial to implement the validation logic in Kibana (MVP).

internal/files/sign.go
jsoriano: I think is fine, we can revisit the decision later if we find the use case.

internal/files/sign.go
jsoriano: Oh, this would be great, but we can do this as a follow up.

jsoriano:
Please go on merging this, I didn't want to block it :slightly_smiling_face: 

> I will prepare a draft presenting different options (pipeline vs web service vs "in place " build and sign).

Ok, in any case, as it is implemented here, we could later add a sign subcommand even if the build command is also able to sign with the `--sign` flag.

> At the moment, it's crucial to implement the validation logic in Kibana (MVP).

Absolutely :+1:

internal/files/sign.go
mtojek: https://github.com/elastic/elastic-package/issues/574

Issue #392, opened 2021-06-28
Allow adding attributes to external field definitions


andrewkroh:
When importing a field definition using the external field feature I would like to be able to add an attribute to the field's defintion. The use case is to import an external field that is of `constant_keyword` type and then specify what the `value` of the field should be for all events.

Here's an example:

```
- name: data_stream.dataset
  external: ecs
  value: nginx
```

PR #575, opened 2021-11-09
Allow to override attributes from external fields

jsoriano:
Attributes defined in the fields definition are combined with the imported field, with some exceptions:
* `external` is removed, as it doesn't make sense in the final definition.
* `type` is always used as defined in the imported field, to avoid having ECS fields with different types.

Some use cases:
* Give an integration-specific description to a field.
* Set a value for a `constant_keyword` (https://github.com/elastic/elastic-package/issues/392).
* Use an external field as dimension (https://github.com/elastic/package-spec/issues/247).

Fixes https://github.com/elastic/elastic-package/issues/392
Fixes https://github.com/elastic/package-spec/issues/247

### DIFF
--- a/internal/common/mapstr.go
+++ b/internal/common/mapstr.go
@@ -56,6 +56,64 @@ func (m MapStr) Put(key string, value interface{}) (interface{}, error) {
 	return old, nil
 }
 
+// DeepUpdate recursively copies the key-value pairs from d to this map.
+// If the key is present and a map as well, the sub-map will be updated recursively
+// via DeepUpdate.
+// DeepUpdateNoOverwrite is a version of this function that does not
+// overwrite existing values.
+func (m MapStr) DeepUpdate(d MapStr) {
+	m.deepUpdateMap(d, true)
+}
+
+// DeepUpdateNoOverwrite recursively copies the key-value pairs from d to this map.
+// If a key is already present it will not be overwritten.
+// DeepUpdate is a version of this function that overwrites existing values.
+func (m MapStr) DeepUpdateNoOverwrite(d MapStr) {
+	m.deepUpdateMap(d, false)
+}
+
+func (m MapStr) deepUpdateMap(d MapStr, overwrite bool) {
+	for k, v := range d {
+		switch val := v.(type) {
+		case map[string]interface{}:
+			m[k] = deepUpdateValue(m[k], MapStr(val), overwrite)
+		case MapStr:
+			m[k] = deepUpdateValue(m[k], val, overwrite)
+		default:
+			if overwrite {
+				m[k] = v
+			} else if _, exists := m[k]; !exists {
+				m[k] = v
+			}
+		}
+	}
+}
+
+func deepUpdateValue(old interface{}, val MapStr, overwrite bool) interface{} {
+	switch sub := old.(type) {
+	case MapStr:
+		if sub == nil {
+			return val
+		}
+
+		sub.deepUpdateMap(val, overwrite)
+		return sub
+	case map[string]interface{}:
+		if sub == nil {
+			return val
+		}
+
+		tmp := MapStr(sub)
+		tmp.deepUpdateMap(val, overwrite)
+		return tmp
+	default:
+		// We reach the default branch if old is no map or if old == nil.
+		// In either case we return `val`, such that the old value is completely
+		// replaced when merging.
+		return val
+	}
+}
+
 // Delete deletes the given key from the map.
 func (m MapStr) Delete(key string) error {
 	k, d, _, found, err := mapFind(key, m, false)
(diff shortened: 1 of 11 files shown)


internal/fields/dependency_manager.go
```diff
@@ -145,8 +145,12 @@ func (dm *DependencyManager) injectFieldsWithRoot(root string, defs []common.Map
 			}
 
 			transformed := transformImportedField(imported)
```
mtojek, line 147: I'm wondering if it isn't possible to sketch the implementation without involving the `DeepUpdate` method (less error-prone code).

For example:

```
base := def
base.Delete("external")
transformed := transformImportedField(base, imported)
```

Function `transformImportedField` can create new `base` struct if it's nil.

internal/docs/exported_fields.go
```diff
@@ -127,11 +127,20 @@ func visitFields(namePrefix string, f fields.FieldDefinition, records []fieldsTa
 
 	if len(f.Fields) == 0 && f.Type != "group" {
 		if f.External != "" {
-			var err error
-			f, err = fdm.ImportField(f.External, name)
+			imported, err := fdm.ImportField(f.External, name)
 			if err != nil {
 				return nil, errors.Wrap(err, "can't import field")
 			}
+			f.Type = imported.Type
```
mtojek, line 134: nit: WDYT about introducing a new method to extract these if-conditions? For example: `f.Update(imported)`

internal/common/mapstr.go
```diff
@@ -56,6 +56,64 @@ func (m MapStr) Put(key string, value interface{}) (interface{}, error) {
 	return old, nil
 }
 
+// DeepUpdate recursively copies the key-value pairs from d to this map.
+// If the key is present and a map as well, the sub-map will be updated recursively
+// via DeepUpdate.
+// DeepUpdateNoOverwrite is a version of this function that does not
+// overwrite existing values.
+func (m MapStr) DeepUpdate(d MapStr) {
```
mtojek, line 64: I wish the `DeepUpdate` isn't mandatory in this implementation :)

internal/fields/dependency_manager.go
jsoriano: I will give a try, but this code would modify the definition, and I wanted to avoid this. For that we need a way to clone it and use as the base struct, or to deep update the final map.
Would you prefer clone over deep update?

internal/docs/exported_fields.go
jsoriano: Good idea :+1:

internal/fields/dependency_manager.go
mtojek: I just wouldn't like to add more potentially error-prone methods to `mapStr`, but if you feel confident the function maybe it's worth leaving it as is.

internal/docs/exported_fields.go
jsoriano: Done in https://github.com/elastic/elastic-package/pull/575/commits/095d0ca37007747a338b502c0e94ab0b36a283bd.

Though doing it as a method of the `FieldDefinition` implies to do it for all its attributes, including the sub-fields. I have implemented this part by updating sub-fields with the same name, and appending the rest.

An alternative could have been to extract these if-conditions in a method here, not as a method of the struct, and then I could only update the fields relevant here.

Let me know what you think.

internal/fields/model.go
```diff
@@ -16,3 +16,52 @@ type FieldDefinition struct {
 	External    string            `yaml:"external"`
 	Fields      []FieldDefinition `yaml:"fields"`
 }
+
+func (orig *FieldDefinition) Update(fd FieldDefinition) {
+	if fd.Name != "" {
```
mtojek, line 21: Ok, so just have to remember about updating these conditions when adding a new property to the `FieldDefinition`.

I'm curious if it isn't shorter to use the YAML marshaller to update fields.

internal/fields/dependency_manager.go
jsoriano: Well, I copied it from libbeat where it has served well for years :slightly_smiling_face:

Maybe in the future we can refactor these methods to use `FieldDefinition`s instead of `mapStr`s.

internal/fields/model.go
jsoriano: > Ok, so just have to remember about updating these conditions when adding a new property to the `FieldDefinition`.
> 

Yes, but well, having it close to the struct I guess that this is easier.

> I'm curious if it isn't shorter to use the YAML marshaller to update fields.

I can give a try to this if prefered. Not sure if we will have flexibility on the nested fields, but we could keep this part as is.

internal/fields/dependency_manager.go
mtojek: I remember that Stefen strongly recommended not to use `mapStr` anymore :)

Anyway, I see that you covered this part with tests, so it shouldn't be a big unknown. Let's not elongate the discussion.

mtojek (approved):
LGTM if the CI is happy.

internal/fields/model.go
mtojek: I don't want to slow you down with this as this is a working implementation. We can come back to this in the future.

PR #440, opened 2021-07-20
Try: use Cobertura to report code coverage

mtojek:
Do not merge this PR.

Issue: https://github.com/elastic/integrations/issues/755

This draft PR is intended for experiments with Cobertura to see code coverage for different packages.

PR #585, opened 2021-11-15
Report ingest pipeline processor coverage

adriansr:
This PR updates the pipeline tests to generate code-coverage (at the processor level) for ingest pipelines:

image

image

Screenshots taken using [Report Generator.](https://github.com/danielpalme/ReportGenerator)

### DIFF
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -149,10 +149,6 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				if err != nil {
 					return cobraext.FlagParsingError(err, cobraext.DataStreamsFlagName)
 				}
-
-				if testCoverage && len(dataStreams) > 0 {
-					return cobraext.FlagParsingError(errors.New("test coverage can be calculated only if all data streams are selected"), cobraext.DataStreamsFlagName)
-				}
 			}
 
 			if runner.TestFolderRequired() {
@@ -203,6 +199,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				API:                esClient.API,
 				DeferCleanup:       deferCleanup,
 				ServiceVariant:     variantFlag,
+				WithCoverage:       testCoverage,
 			})
 
 			results = append(results, r...)
--- a/internal/elasticsearch/node_stats/stats_api.go
+++ b/internal/elasticsearch/node_stats/stats_api.go
@@ -1,0 +1,144 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package node_stats
+
+import (
+	"encoding/json"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/elasticsearch"
+	"github.com/elastic/elastic-package/internal/elasticsearch/pipeline"
+)
+
+// StatsRecord contains stats for a measurable entity (pipeline, processor, etc.)
+type StatsRecord struct {
+	Count, Current, Failed int64
+	TimeInMillis           int64 `json:"time_in_millis"`
+}
+
+// ProcessorStats contains a processor's stats and some metadata.
+type ProcessorStats struct {
+	Type, Extra string
+	Conditional bool
+	Stats       StatsRecord
+}
+
+// PipelineStats contains stats for a pipeline.
+type PipelineStats struct {
+	StatsRecord
+	Processors []ProcessorStats
+}
+
+// PipelineStatsMap holds the stats for a set of pipelines.
+type PipelineStatsMap map[string]*PipelineStats
+
+type wrappedProcessor map[string]ProcessorStats
+
+func (p wrappedProcessor) extract() (stats ProcessorStats, err error) {
+	if len(p) != 1 {
+		return stats, errors.Errorf("can't extract processor stats. Need a single processor, got %d: %+v", len(p), p)
+	}
+	for k, v := range p {
+		stats = v
+		if off := strings.Index(k, ":"); off != -1 {
+			stats.Extra = k[off+1:]
+			k = k[:off]
+		}
+		switch v.Type {
+		case "conditional":
+			stats.Conditional = true
+		case k:
+		default:
+			return stats, errors.Errorf("can't understand processor identifier '%s' in %+v", k, p)
+		}
+		stats.Type = k
+	}
+	return stats, nil
+}
+
+type pipelineStatsRecord struct {
+	StatsRecord
+	Processors []wrappedProcessor
+}
+
+type pipelineStatsRecordMap map[string]pipelineStatsRecord
+
+func (r pipelineStatsRecord) extract() (stats *PipelineStats, err error) {
+	stats = &PipelineStats{
+		StatsRecord: r.StatsRecord,
+		Processors:  make([]ProcessorStats, len(r.Processors)),
+	}
+	for idx, wrapped := range r.Processors {
+		if stats.Processors[idx], err = wrapped.extract(); err != nil {
+			return stats, errors.Wrapf(err, "converting processor %d", idx)
+		}
+	}
+	return stats, nil
+}
+
+type pipelinesStatsNode struct {
+	Ingest struct {
+		Pipelines pipelineStatsRecordMap
+	}
+}
+
+type pipelinesStatsResponse struct {
+	Nodes map[string]pipelinesStatsNode
+}
+
+func GetPipelineStats(esClient *elasticsearch.API, pipelines []pipeline.Resource) (stats PipelineStatsMap, err error) {
+	statsReq := esClient.Nodes.Stats.WithFilterPath("nodes.*.ingest.pipelines")
+	resp, err := esClient.Nodes.Stats(statsReq)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Node Stats API call failed")
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read Stats API response body")
+	}
+
+	if resp.StatusCode != 200 {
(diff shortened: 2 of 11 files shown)


adriansr:
The coverage reports in CI are not including the pipeline source code:
image

Looks like by the time the report is generated, the source code is not available. @mdelapenya @mtojek can you help?

cmd/testrunner.go
```diff
@@ -149,10 +149,6 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				if err != nil {
 					return cobraext.FlagParsingError(err, cobraext.DataStreamsFlagName)
 				}
-
-				if testCoverage && len(dataStreams) > 0 {
```
mtojek, line 153: This condition was placed here to prevent coverage miscalculation. The runner is not aware of data stream presence (or being skipped) so it always assumes that all data streams have been selected.

internal/testrunner/coverageoutput.go
```diff
@@ -57,48 +64,53 @@ type coberturaCoverage struct {
...
-type coberturaClass struct {
+// CoberturaClass represents a class in a Cobertura XML report.
+type CoberturaClass struct {
 	Name       string             `xml:"name,attr"`
 	Filename   string             `xml:"filename,attr"`
 	LineRate   float32            `xml:"line-rate,attr"`
 	BranchRate float32            `xml:"branch-rate,attr"`
 	Complexity float32            `xml:"complexity,attr"`
-	Methods    []*coberturaMethod `xml:"methods>method"`
+	Methods    []*CoberturaMethod `xml:"methods>method"`
```
mtojek, line 92: You will have to update the docs with new mapping information: https://github.com/elastic/integrations/blob/master/docs/testing_and_validation.md#review-test-coverage

mtojek (commented):
Cool feature, Adrian!

> Looks like by the time the report is generated, the source code is not available. @mdelapenya @mtojek can you help?

First guess: I think this might be caused by having paths as names and you can't refer to these paths from the repository root.

I'm afraid that you need to analyze the Cobertura XML report manually and adjust the implementation later on. This is the path I went last time: https://github.com/elastic/integrations/issues/755#issuecomment-883364183 and static files in https://github.com/elastic/elastic-package/pull/440 .

> Screenshot taken from Report Generator.

If you plan to merge this feature, we'll kindly ask you to prepare a doc page describing how to use it.

cmd/testrunner.go
adriansr: I understand that. But given that (afaik) the CI never tests a single datastream, it would be very useful for pipeline developers to be able to run tests for a single datastream locally and get coverage results.

adriansr:
@mtojek I'm not sure how it works during CI, but my understanding is that the coverage results are uploaded to GCloud storage (in https://github.com/elastic/elastic-package/blob/490e70dfcdf898450087b87fb9301b608785835b/.ci/Jenkinsfile#L205) and then downloaded before generating the Coverage report (in https://github.com/elastic/elastic-package/blob/490e70dfcdf898450087b87fb9301b608785835b/.ci/Jenkinsfile#L192). 

The generated coverage file contains absolute paths to the code. For example in `zoom` package:
```
  <sources>
    <source>/Users/adrian/go/src/github.com/elastic/integrations/packages/zoom</source>
  </sources>
```

and then:
```
        <class name="account" filename="data_stream/webhook/elasticsearch/ingest_pipeline/account.yml" line-rate="0" branch-rate="0" complexity="0">
```

My understanding is that when the report is generated, the source code is not available at this path, but I don't understand in what context the `coverageReport` function is executed (can't find a definition for it). Maybe @mdelapenya could chime in as I see that he wrote the coverage handling in `Jenkinsfile`.

jsoriano:
`coverageReport` is defined here: https://github.com/elastic/apm-pipeline-library/blob/4a58b62664d47cb32e6c2924baf75627bf8346c9/vars/coverageReport.groovy

internal/testrunner/coverageoutput.go
```diff
@@ -42,11 +43,17 @@ func (tcd *testCoverageDetails) withTestResults(results []TestResult) *testCover
 			tcd.dataStreams[result.DataStream] = []string{}
 		}
 		tcd.dataStreams[result.DataStream] = append(tcd.dataStreams[result.DataStream], result.Name)
+		if tcd.cobertura != nil && result.Coverage != nil {
+			tcd.cobertura.Merge(result.Coverage)
+		} else {
+			tcd.cobertura = result.Coverage
```
mdelapenya, line 49: What happens if `result.Coverage` is nil and `tcd.cobertura` is not? In that line we will be setting the latter to nil. Is that correct?

internal/testrunner/coverageoutput.go
```diff
@@ -113,6 +125,101 @@ func (c *coberturaCoverage) bytes() ([]byte, error) {
 	return buffer.Bytes(), nil
 }
 
+// Merge merges two coverage reports for a given class.
+func (c *CoberturaClass) Merge(b *CoberturaClass) {
```
mdelapenya, line 129: Do you see feasible adding unit tests for this method? I see a lot of calculations that would benefit for the tests

internal/elasticsearch/pipeline/processors.go
```diff
@@ -0,0 +1,186 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+type Processor struct {
```
mdelapenya, line 16: Do you see feasible adding unit tests for this method? I see a lot of calculations that would benefit for the tests

internal/testrunner/coverageoutput.go
adriansr: Fixed, thanks

internal/testrunner/coverageoutput.go
adriansr: Done

internal/elasticsearch/pipeline/processors.go
adriansr: Done

internal/elasticsearch/node_stats/stats_api.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+	"encoding/json"
+	"io"
+	"strings"
+
+	"github.com/elastic/go-elasticsearch/v7"
+	"github.com/pkg/errors"
+
+	es "github.com/elastic/elastic-package/internal/elasticsearch"
```
jsoriano, line 15: Trying to reduce the need of these `es` aliases in https://github.com/elastic/elastic-package/pull/596.

internal/elasticsearch/pipeline/processors.go
```diff
@@ -0,0 +1,194 @@
...
+			return nil, errors.Errorf("processor#%d is not a single key map (kind:%v Content:%d)", idx, entry.Kind, len(entry.Content))
+		}
+		var proc Processor
+		if err := entry.Content[1].Decode(&proc); err != nil {
+			return nil, errors.Wrapf(err, "error decoding processor#%d configuration", idx)
+		}
+		if err := entry.Content[0].Decode(&proc.Type); err != nil {
+			return nil, errors.Wrapf(err, "error decoding processor#%d type", idx)
+		}
+		proc.Line = entry.Line
```
jsoriano, line 55: Nit. Could the `Line` of the last `Content` be used as the last line of the processor so we have per-line coverage instead of per-processor coverage? (and Cobertura paints red/green the whole processor :slightly_smiling_face:)

jsoriano (commented):
Thanks @adriansr, this is great!

Is there something else keeping this as draft?

adriansr:
This is currently waiting for the required CI fixes to have source code displayed in Jenkins coverage reports.

Once that is sorted out I'll make any required adjustments to the format (for now only tested it with a different coverage tool) and prepare the new documentation.

internal/elasticsearch/pipeline/processors.go
adriansr: Makes sense, done.

Originally I decided against it because it will mark the inner processors of a compound processor as covered (inner processor of `foreach`, and `on_failure` processors), while ES doesn't provide coverage information for those.

internal/elasticsearch/node_stats/stats_api.go
adriansr: done

adriansr:
This is ready to review pending adjustments to CI in order to make the new source code reports available.

Updated docs at https://github.com/elastic/integrations/pull/2371

mtojek:
We collect Cobertura reports also for elastic-package. @adriansr Could you please double check if this new feature is enabled also here?

internal/elasticsearch/node_stats/stats_api.go
```diff
@@ -0,0 +1,144 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package node_stats
```
mtojek, line 5: Please rename it to `nodestats`.

internal/elasticsearch/pipeline/coverage/coverage.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/elasticsearch/node_stats"
+	"github.com/elastic/elastic-package/internal/elasticsearch/pipeline"
+	"github.com/elastic/elastic-package/internal/packages"
+	"github.com/elastic/elastic-package/internal/testrunner"
+)
+
+// Get returns a coverage report for the provided set of ingest pipelines.
+func Get(options testrunner.TestOptions, pipelines []pipeline.Resource) (*testrunner.CoberturaCoverage, error) {
```
mtojek, line 21: Here is a bit of spaghetti code that has to be untangled first. Code coverage isn't the responsibility of Elasticsearch client or Pipeline, but test runners and it should stay there.

I suggest refactoring (moving around) this dependency first, then we can continue reviewing as this PR is relatively big.

mtojek (changes requested):
To be honest, this PR is massive. Do you think you can split it into smaller, reviewable chunks? It would be easier to fully understand what's going on behind the scenes.

PR #590, opened 2021-11-17
Add flag to select the kibana version in elastic-package status

jsoriano:
With this flag elastic-package status can be used to check the versions
available of a package for a given version of the stack.

For example:
```
$ elastic-package status apache --version 8.0.0
Package: apache
Package Versions:
+-------------+---------+---------+--------------------+--------------------------------+
| ENVIRONMENT | VERSION | RELEASE |       TITLE        |          DESCRIPTION           |
+-------------+---------+---------+--------------------+--------------------------------+
| Snapshot    | 1.3.0   | ga      | Apache HTTP Server | Collect logs and metrics from  |
|             |         |         |                    | Apache servers with Elastic    |
|             |         |         |                    | Agent.                         |
+-------------+---------+---------+--------------------+--------------------------------+
| Staging     | -       | -       | -                  | -                              |
+-------------+---------+---------+--------------------+--------------------------------+
| Production  | -       | -       | -                  | -                              |
+-------------+---------+---------+--------------------+--------------------------------+
```

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -37,7 +37,7 @@ func setupStatusCommand() *cobraext.Command {
 		RunE:  statusCommandAction,
 	}
 	cmd.Flags().BoolP(cobraext.ShowAllFlagName, "a", false, cobraext.ShowAllFlagDescription)
-	cmd.Flags().String(cobraext.StackVersionFlagName, "", cobraext.StackVersionFlagDescription)
+	cmd.Flags().String(cobraext.StatusKibanaVersionFlagName, "", cobraext.StatusKibanaVersionFlagDescription)
 
 	return cobraext.NewCommand(cmd, cobraext.ContextPackage)
 }
@@ -53,9 +53,9 @@ func statusCommandAction(cmd *cobra.Command, args []string) error {
 		packageName = args[0]
 	}
 
-	kibanaVersion, err := cmd.Flags().GetString(cobraext.StackVersionFlagName)
+	kibanaVersion, err := cmd.Flags().GetString(cobraext.StatusKibanaVersionFlagName)
 	if err != nil {
-		return cobraext.FlagParsingError(err, cobraext.StackVersionFlagName)
+		return cobraext.FlagParsingError(err, cobraext.StatusKibanaVersionFlagName)
 	}
 	options := registry.SearchOptions{
 		All:           showAll,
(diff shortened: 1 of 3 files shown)


ruflin:
Nice. Later this could also be expanded that `elastic-package status apache --info=kibana.version` and it adds a column with the `kibana.version` value. This could be used for other fields too which might be interesting.

cmd/status.go
```diff
@@ -37,6 +37,7 @@ func setupStatusCommand() *cobraext.Command {
 		RunE:  statusCommandAction,
 	}
 	cmd.Flags().BoolP(cobraext.ShowAllFlagName, "a", false, cobraext.ShowAllFlagDescription)
+	cmd.Flags().String(cobraext.StackVersionFlagName, "", cobraext.StackVersionFlagDescription)
```
mtojek, line 40: This flag might be misleading here. In `elastic-package status apache --version 8.0.0` the version flag may suggest that this is Apache 8.0.0, not Kibana.

What do you think about replacing it with `--kibana-version`?

ruflin: Or `kibana.version`? Thinking of it as the filters you can apply on epr, for example https://epr.elastic.co/search?experimental=true&kibana.version=8.0.0 It would become a more generic feature ...

mtojek: I think that `--kibana-version` will look more like a command option. I haven't seen options with dots, I guess it's a standard? 

But calling for a query is also an option: `--query="kibana.version=8.0.0"`.

mtojek: BTW do we really need the experimental flag? It's always enabled...

jsoriano: I used `--version` because it is the flag used to specify the stack version in the stack subcomand, but I see how it can be confusing here.

I have opted for `--kibana-version`, I also prefer it to the version with the dot. How the kibana version is finally encoded in the query is an implementation detail.

> But calling for a query is also an option: `--query="kibana.version=8.0.0"`.

`--kibana-version` covers a very specific use case, so I prefer to have a specific flag for that. Also, package developers don't need to know the registry API to use this command, so I don't think we need generic options by now.

Humans needing to write url query parameters is a call for problems.

> I haven't seen options with dots, I guess it's a standard?

This is not very common, but I am pretty sure it is possible to have options with dots, take a look to `metricbeat --help` :slightly_smiling_face:

> BTW do we really need the experimental flag? It's always enabled...

I hope we can get rid of this soon https://github.com/elastic/package-spec/issues/225 :fire:

mtojek: Thanks for adjusting this!

mtojek (approved):

PR #591, opened 2021-11-17
Remove references to internal parameter

jsoriano:
Remove references to the internal parameter, that is not available anymore in the registry, see https://github.com/elastic/package-registry/pull/765.


mtojek (commented):
Ship it

mtojek (approved):

PR #596, opened 2021-11-19
Remove ambiguous elasticsearch imports

jsoriano:
Along the code there are several places where we are importing these packages:
* `github.com/elastic/elastic-package/internal/elasticsearch`
* `github.com/elastic/go-elasticsearch/v7`
* `github.com/elastic/go-elasticsearch/v7/esapi`

The `elasticsearch` ones are ambiguous and need an alias, we are using `es`, but not consistently, sometimes for the internal package, and sometimes for `go-elasticsearch`.

This PR adds aliases of the required types of `go-elasticsearch` into the internal package, so other packages only need to import this one.

Also, elastic-package only needs the `API` part of the `Client`s, so this is the only part exposed now. We could use interfaces also for this but these objects are not easy to interface because functions are members and members of embedded objects.

### DIFF
--- a/internal/testrunner/runners/asset/runner.go
+++ b/internal/testrunner/runners/asset/runner.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"strings"
 
-	es "github.com/elastic/go-elasticsearch/v7"
 	"github.com/pkg/errors"
 
 	"github.com/elastic/elastic-package/internal/logger"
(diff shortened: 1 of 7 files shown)


internal/testrunner/runners/asset/runner.go
```diff
@@ -29,7 +28,6 @@ const (
 type runner struct {
 	testFolder      testrunner.TestFolder
 	packageRootPath string
-	esClient        *es.Client
```
jsoriano, line 32: Not used. :shrug:

mtojek (approved):

PR #599, opened 2021-11-24
Add warning message to docker boot 

fearful-symmetry:
So I learned (or forgot, and then remembered) the hard way that if you don't run `elastic-stack` up in an integrations directory, it won't add any custom-built integrations.

This just adds a little warning message if it can't find the development directory. However, considering the opportunity for weird silent failures here, I wonder if we should fail out if `found == false`, perhaps with a user override.

### DIFF
--- a/internal/stack/boot.go
+++ b/internal/stack/boot.go
@@ -41,6 +41,8 @@ func BootUp(options Options) error {
 		if err != nil {
 			return errors.Wrap(err, "copying package contents failed")
 		}
+	} else {
+		fmt.Println("WARNING: no custom build packages directory found. Any locally built integrations will not be installed.")
 	}
 
 	fmt.Println("Packages from the following directories will be loaded into the package-registry:")


internal/stack/boot.go
```diff
@@ -41,6 +41,8 @@ func BootUp(options Options) error {
 		if err != nil {
 			return errors.Wrap(err, "copying package contents failed")
 		}
+	} else {
+		fmt.Println("WARNING: no custom build packages directory found. Any locally built integrations will not be installed.")
```
mtojek, line 45: There is a message logged in L39 informing about used package directories. Do you see this as insufficient/less informative/improvement needed?

fearful-symmetry: Yah, I think we need something that prints a message in the negative case and actually reports when something isn't working as expected.

jsoriano (commented):
Oh, I had commented but forgot to submit :facepalm:

internal/stack/boot.go
mtojek: It is working as expected as the process wasn't started in the integrations repo. Maybe we should rephrase this warning into something like:

`logger.Info("Local packages won't be included as the elastic-package didn't find any built packages.")`.

WDYT?

mtojek:
> And yet another thing we can do to help on these scenarios is to make the registry automatically able to reindex packages without needing to be restarted.

I can give some historical context:

1. I tried to mount built packages into Docker container so that when the end-user rebuilds packages, auto-discovery will reindex all files. It was problematic as it was reindexing really slowly when the directory was mounted (a few seconds vs a few minutes).
2. I decided to rebuild the Docker image for the Package Registry and restart it. It went well until Kibana enabled package caching, which means that if Kibana precaches the developed package, you have to restart also the Kibana :) or artificially change the package version.
3. I do believe that there are improvements possible, but TBH nothing would be better than installing packages directly in Kibana instead and simply forgetting about running the Registry here. It would really simplify this workflow. 

I'm waiting for option 3. to be delivered :)

jsoriano:
> I'm waiting for option 3. to be delivered :)

++

Everything else are workarounds :D

fearful-symmetry:
So, my concern/paranoia here is mostly about silently bypassing something in the local integrations build, i.e
1) Make a change in an integration
2) Build it, start up the stack outside the integration directory
3) Test it without noticing that you're not running the version that you were working on.

This is kind of a subtle failure mode, and it leaves more leeway for user oversight than I'd like.

Is there a reason why the `local build dir -> .elastic-package/ build dir` step happens during the docker bootup? If we moved that step to `elastic-package build` it would render this kind of a moot issue @jsoriano / @mtojek

mtojek:
> Is there a reason why the local build dir -> .elastic-package/ build dir copy step happens during the docker bootup? If we moved that step to elastic-package build it would render this kind of a moot issue @jsoriano / @mtojek

`elastic-package build` is a package-oriented operation, which shouldn't trigger any docker builds as these aren't related. In the nearest future, we plan to get rid of rebuilding the Docker image for Package Registry, so it will simplify the workflow.

Considering your suggestion, If you want to build 100 packages, does it mean you have to rebuild the Docker image 100 times? Another analogy, when you are building a binary out of a Go code using `go build`, or a program in C with `gcc`, it doesn't build any Docker image as it's unaware of the distribution form.

fearful-symmetry:
> elastic-package build is a package-oriented operation, which shouldn't trigger any docker builds as these aren't related.

I might be mixing stuff up, but the part of `elastic-package stack up` that copies the build dir doesn't need to trigger any builds or rebuilds? We can do _just the copying_ in the package-build stage without triggering any kind of docker operation?

mtojek:
> I might be mixing stuff up, but the part of elastic-package stack up that copies the build dir doesn't need to trigger any builds or rebuilds? 

No, it does trigger the docker build operation for a custom Package Storage image (see [code ref](https://github.com/elastic/elastic-package/blob/master/internal/stack/boot.go#L53-L58)).

We considered alternative approaches in the past. For example, mounting a volume resulted in a slow start of the Package Registry, and using a filewatcher to refresh the Package Registry's internal state wasn't stable. As I already said, the flow will be simplified if Kibana allows for installing packages directly without any interaction with Package Registry.

fearful-symmetry:
@mtojek What I'm referring to is the exact copy operation here: https://github.com/elastic/elastic-package/blob/6215ed4b3d1e95b71bc903db64aacf9b30b641d3/internal/stack/boot.go#L38-L44

Is there a reason that needs to be in the `BootUp` phase? Couldn't that simply be copied to the `BuildPackage()` function, and then we'd just build the container normally during `BootUp` ?

mtojek:
Then, it will be misleading for users who want to build a package, but prefers to boot a clean stack. I don't see an intuitive workaround for this. Do you have any other ideas, @jsoriano?

fearful-symmetry:
I feel like this behavior should be explicit rather than implicit. If a user wants to boot a clean stack or development stack, there should be a CLI flag that controls that behavior, rather than depending on something more subtle like the development directory. Also keep in mind that if someone runs `elastic-package stack up -v`, the verbose log messages will quickly swallow some of these initialization messages.

mtojek:
It's implicit as it's natural to stay in the package directory while developing it. If you need to add files to the Git index, you need to be in the working directory.

> Also keep in mind that if someone runs elastic-package stack up -v, the verbose log messages will quickly swallow some of these initialization messages.

Would a special logging tag help here? For example: `[development stack]` or `[clean stack]`?

jsoriano:
I don't think that we should add more implicit behaviours to `elastic-package build` or `elastic-package stack`.

I agree with Alex that the current behaviour is confusing: you need to remember to run `elastic-package stack up` from the directory where the integration lives and you need to run it again after build so the registry includes the package. This is a usual problem new package developers find.

> If you need to add files to the Git index, you need to be in the working directory.

This is partially true. There may be many reasons why the terminal where you run `elastic-package stack` is not in the proper directory. You may be modifying some files in a package, then you open a new terminal to start the stack to try it, and this terminal may be opened in `$HOME`, and then you start a stack without the package you are trying to test. Or you may be using an IDE for everything, including managing Git.

Log messages (even big red warnings) are still a partial solution. They may help sometimes, but you can easily overlook them.

As mentioned the ideal solution would be to have https://github.com/elastic/kibana/issues/70582, and when we have it, we can probably make `elastic-package install` to explicitly install the current package. In the meantime any other solution is going to be a compromise.

Maybe we should think ahead in the UX based on `elastic-package install`. It could be implemented this way now:
* It stops the registry started in the stack maintained with `elastic-package stack` (if no stack is started it could fail explicitly mentioning that).
* It starts a registry that includes the package to be installed.
* It installs the package.

Then it doesn't matter how the stack was started, the installation is going to install what is expected by the developer.

We would replace the step of running `elastic-package stack up` to restart the registry with the new package and manually installing the package with `elastic-package install`, that is more explicit for the operation that the developer is trying to do.

Once https://github.com/elastic/kibana/issues/70582 is available, we use this API instead of restarting the registry, but the developer UX is the same.

fearful-symmetry:
+1 on @jsoriano 's idea, an `install` command like that would be relatively explicit, and make the develop/test workflow a tad easier too.

mtojek:
Keep in mind that it will be still confusing as there is a package cache in Kibana.

Consider the following scenario:

0. User is developing the package `foo-1.2.3`.
1. User builds the package with `elastic-package build`.
2. User boots the Elastic stack, the stack contains the new package. User orders Kibana to install it, Kibana caches the package.

Unfortunately, there is a bug and user needs to build the package again (same version).

1. User applies a bugfix and builds the package.
2. User runs `elastic-package install`, which rebuilds the Package Registry, installs the package in Kibana, BUT....

... Kibana installs the same old, cached package and the user is confused now.

jsoriano:
> ... Kibana installs the same old, cached package and the user is confused now.

I guess that this problem also exists with current implementation?

mtojek:
Yes. In such cases, it's recommended to take down the stack or artificially bump up the package version.

My point is the solution you proposed won't fix the entire problem.

mtojek:
The PR is stale, let's close it for now.

PR #602, opened 2021-12-01
[mergify] use merge queue

v1v:
### What

Migrate from the `merge action` to the `queue action`.

See https://docs.mergify.io/actions/queue/#queue-rules to know more about how the `queue` command works.

### Why

We use the merge action to automerge automated PRs, so we cannot use this approach from January 2022

```
The configuration uses the deprecated strict mode of the merge action.
A brownout is planned for the whole December 6th, 2021 day.
This option will be removed on January 10th, 2022.
For more information: https://blog.mergify.io/strict-mode-deprecation/
```

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,3 +1,8 @@
+queue_rules:
+  - name: default
+    conditions:
+      - check-success=elastic-package/pr-merge
+
 pull_request_rules:
   - name: automatic merge of bot 🤖
     conditions:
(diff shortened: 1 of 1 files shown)


cachedout:
@v1v This looks good, but is there meta-issue that tracks these changes? IMHO we should have something filed so we can see what needs to be migrated, and what the migration status is.

cachedout (approved):

kuisathaverat (approved):

PR #604, opened 2021-12-02
Support overriding configuration folder

endorama:
Configuration folder path was hardcoded to `user home folder/.elastic-package`. For testing purposes or for development this created some issues, as any run of the binary would run `EnsureInstalled`, eventually failing or overriding data.

This commit introduce a new environment variable, `ELASTIC_PACKAGE_DATA_HOME`, that can be used to override the data folder where this CLI unpack it's configuration and data files.

The suffix `DATA_HOME` is taken from XDG spec and leaves room for splitting configuration and data in the future without introducing a breaking change.

Tests have been added to check `configurationDir` behaviour.

### DIFF
--- a/internal/configuration/locations/locations.go
+++ b/internal/configuration/locations/locations.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	// elasticDataHomeEnv is the name of the environment variable used to override data folder for elastic-package
-	elasticDataHomeEnv = "ELASTIC_PACKAGE_DATA_HOME"
+	elasticPackageDataHome = "ELASTIC_PACKAGE_DATA_HOME"
 
 	elasticPackageDir = ".elastic-package"
 	stackDir          = "stack"
--- a/internal/configuration/locations/locations.go
+++ b/internal/configuration/locations/locations.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	// elasticDataHomeEnv is the name of the environment variable used to override data folder for elastic-package
+	// elasticPackageDataHome is the name of the environment variable used to override data folder for elastic-package
 	elasticPackageDataHome = "ELASTIC_PACKAGE_DATA_HOME"
 
 	elasticPackageDir = ".elastic-package"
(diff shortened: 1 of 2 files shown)


internal/configuration/locations/locations.go
```diff
@@ -13,6 +13,9 @@ import (
 )
 
 const (
+	// elasticDataHomeEnv is the name of the environment variable used to override data folder for elastic-package
+	elasticDataHomeEnv = "ELASTIC_PACKAGE_DATA_HOME"
```
mtojek, line 17: I like this separation of concerns, but I'm afraid that we'll violate it this way. The XDG spec assumes:

```
Write user-specific data in $XDG_DATA_HOME
Write configuration files in $XDG_CONFIG_HOME
Write cache files in $XDG_CACHE_HOME
```

I'm afraid we store everything together:
* user-specific data - stack profiles
* configuration - Elastic stack definition
* cached - cached ECS fields

I wouldn't focus now on splitting the `~/.elastic-package` into multiple configs, especially that we aren't solving any issue there.

Let me know what do you think.

endorama: I agree that today this is not the case and I would not focus on splitting folder content. The goal was to keep the door open for a smooth transition if we want to go down that path in the future.

Moreover, I think this naming has some advantages even without separation of concern:
- is not very generic (like "home" is), identifies that it contains data and reduce confusion
- you can consider configuration and cache special types of "data" for an application
- splitting in the future will require mainly to separate configuration and cache from the current set of files (i.e. all templates are data and would be left there) 

Based on this consideration I considered calling it `DATA_HOME` today as most effective based on it's content and ease of future expansion.

mtojek, line 17: nit: elasticPackageDataHome?

endorama:
```suggestion
	elasticPackageDataHome = "ELASTIC_PACKAGE_DATA_HOME"
```

mtojek (approved):
Thanks for adding test cases!

PR #605, opened 2021-12-02
Add allowed ip check

marc-gr:
To identify any IP in the testing value that is not in the supported Maxmind's allowed set (found here https://github.com/elastic/elastic-package/blob/master/docs/howto/ingest_geoip.md) a validation check is added at test time.

Will output something similar to 

```
0] parsing field value failed: the IP "107.170.139.210" is not one of the allowed test IPs
[1] parsing field value failed: the IP "96.241.146.97" is not one of the allowed test IPs
```


mtojek (commented):
It looks like a good PR, but I'm concerned about system tests. See the test output:

```
one or more errors found in documents stored in metrics-aws.ec2_metrics-ep data stream: [0] parsing field value failed: the IP "3.84.55.9" is not one of the allowed test IPs
```

internal/testrunner/runners/system/runner.go
```diff
@@ -438,7 +438,9 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 
 	// Validate fields in docs
 	fieldsValidator, err := fields.CreateValidatorForDataStream(serviceOptions.DataStreamRootPath,
-		fields.WithNumericKeywordFields(config.NumericKeywordFields))
+		fields.WithNumericKeywordFields(config.NumericKeywordFields),
+		fields.WithDisabledAllowedIPCheck(),
```
mtojek, line 442: nit: I would leave a comment here explaining why is it disabled for system tests.

internal/fields/validate.go
```diff
@@ -99,6 +115,20 @@ func CreateValidatorForDataStream(dataStreamRootPath string, opts ...ValidatorOp
...
 }
 
+//go:embed _static/allowed_geo_ips.txt
+var allowedGeoIPs string
+
+func initializeAllowedIPsList() map[string]struct{} {
+	m := map[string]struct{}{
+		"0.0.0.0": {}, "255.255.255.255": {},
+	}
+	for _, ip := range strings.Split(allowedGeoIPs, "\n") {
```
mtojek, line 125: nit: you may want to trim the accidental empty line at the file end.

mtojek (commented):
I think the only one left thing is adjusting pipeline tests?

mtojek:
I think you have to update pipeline tests for the AWS test package.

mtojek (approved):
LGTM

internal/fields/validate.go
```diff
@@ -316,6 +351,20 @@ func (v *Validator) parseElementValue(key string, definition FieldDefinition, va
...
+		if !valid {
+			break
+		}
+
+		if err := ensurePatternMatches(key, valStr, definition.Pattern); err != nil {
+			return err
+		}
+
+		if v.enabledAllowedIPCheck && !v.isAllowedIPValue(valStr) {
+			return fmt.Errorf("the IP %q is not one of the allowed test IPs", valStr)
```
efd6, line 366: It would be nice if this emitted a link to the list of acceptable IPs.

PR #606, opened 2021-12-06
Add note about removing macOS quarantine after release download.

matschaffer:
Thought it'd be handy to have this in the readme, at least until we're able to sign releases.


mtojek (approved):
Definitely a good idea, thanks @matschaffer! I have also created a relevant issue: https://github.com/elastic/elastic-package/issues/607

PR #619, opened 2021-12-13
Force GeoIP db in all stack versions

marc-gr:
Forces to use the same GeoIP database for all stack versions to avoid inconsistencies in test outputs.

Since we were using a specific GeoIP database for 8.x, tests expected files will have different output if the geoip processor is used depending on which stack is used. Currently PRs in integrations are testing against 8.0.0, but tests on the 7.16.0 stack also run in master and in elastic-package PRs, generating errors for pipeline tests eg:

```
-                "as": {
-                    "number": 35908
-                },
-                "geo": {
-                    "continent_name": "Asia",
-                    "country_iso_code": "BT",
-                    "country_name": "Bhutan",
-                    "location": {
-                        "lat": 27.5,
-                        "lon": 90.5
-                    }
-                },
                 "ip": "67.43.156.13",
                 "port": 80
```

To avoid this, we force the same GeoIP database for any version of the stack, to provide consistency in tests output.


mtojek (approved):
Let's add some context in the issue description why do we need this.

Let's make sure that CI status is green for test packages.

PR #625, opened 2021-12-15
Parallelize tests to shorten total time

mtojek:
This PR modifies the Jenkinsfile to parallelize test execution.

Please take a look at the pipeline graph.


mtojek:
The total time decreased from 54 min to 31m. We can go lower if we remove few data streams from packages, as it was before. WDYT? @marc-gr

marc-gr:
> The total time decreased from 54 min to 31m. We can go lower if we remove few data streams from packages, as it was before. WDYT? @marc-gr

Makes sense, I guess we just need to keep one for each of the packages?

mtojek:
Yes, one representative data stream.

marc-gr (approved):

PR #633, opened 2021-12-22
Fix: use container name consistent with Docker Compose API

r00tu53r:
* Docker compose v1 (python) and v2 (Go) use
  different service naming conventions. Set
  service name based on the version.

Confirmed with docker compose source and tests:

https://github.com/docker/compose/blob/v2/pkg/compose/compose.go#L36

### DIFF
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -28,7 +28,7 @@ type Project struct {
 	name             string
 	composeFilePaths []string
 
-	dockerComposeV1  bool
+	dockerComposeV1 bool
 }
 
 // Config represents a Docker Compose configuration file.
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -351,7 +351,6 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
 }
 
 func (p *Project) dockerComposeVersion() (*semver.Version, error) {
-
 	var b bytes.Buffer
 
 	args := []string{
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
 
@@ -22,14 +23,10 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
-const (
-	DockerComposeV1 = "v1"
-	DockerComposeV2 = "v2"
-)
-
 // Project represents a Docker Compose project.
 type Project struct {
 	name             string
+	dcMajorVersion   string
 	composeFilePaths []string
 }
 
@@ -146,8 +143,9 @@ func NewProject(name string, paths ...string) (*Project, error) {
 	}
 
 	c := Project{
-		name,
-		paths,
+		name:             name,
+		dcMajorVersion:   "",
+		composeFilePaths: paths,
 	}
 
 	return &c, nil
@@ -345,28 +343,37 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
 	return cmd.Run()
 }
 
-func (p *Project) DockerComposeVersion() (string, error) {
+func (p *Project) dockerComposeVersion() (*semver.Version, error) {
 
-	const (
-		v1marker = "docker-compose version 1"
-	)
 	var b bytes.Buffer
 
-	args := []string{"version"}
+	args := []string{
+		"version",
+		"--short",
+	}
 	if err := p.runDockerComposeCmd(dockerComposeOptions{args: args, stdout: &b}); err != nil {
-		return "", errors.Wrap(err, "running Docker Compose version command failed")
+		return nil, errors.Wrap(err, "running Docker Compose version command failed")
 	}
-	versionStr := string(b.Bytes())
-	if strings.Contains(versionStr, v1marker) {
-		return DockerComposeV1, nil
+	dcVersion := b.String()
+	ver, err := semver.NewVersion(dcVersion)
+	if err != nil {
+		return nil, errors.Wrapf(err, "docker compose version not a valid semver %s", dcVersion)
 	}
-	// v2marker: Docker Compose version v2.x.y
-	return DockerComposeV2, nil
+	return ver, nil
 }
 
 // ContainerName method the container name for the service.
-func (p *Project) ContainerName(dcver, serviceName string) string {
-	if dcver == DockerComposeV1 {
+func (p *Project) ContainerName(serviceName string) string {
+	if p.dcMajorVersion == "" {
+		ver, err := p.dockerComposeVersion()
+		if err != nil {
+			logger.Errorf("Unable to determine docker-compose version: %v. Defaulting to 1.x", err)
+			p.dcMajorVersion = "1"
+		} else {
+			p.dcMajorVersion = strconv.FormatInt(ver.Major(), 10)
+		}
+	}
+	if p.dcMajorVersion == "1" {
 		return fmt.Sprintf("%s_%s_1", p.name, serviceName)
 	}
 	return fmt.Sprintf("%s-%s-1", p.name, serviceName)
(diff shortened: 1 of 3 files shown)


internal/compose/compose.go
```diff
@@ -340,7 +345,29 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
 	return cmd.Run()
 }
 
+func (p *Project) DockerComposeVersion() (string, error) {
```
mtojek, line 348: As this method is used only by `ContainerName`, do you think we could make it package-visible and access it only in `ContainerName`?

This way you would prevent adjusting all calling code. If we need this condition in other places, we can refactor it then.

internal/compose/compose.go
```diff
@@ -22,6 +22,11 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
+const (
+	DockerComposeV1 = "v1"
+	DockerComposeV2 = "v2"
```
mtojek, line 27: Do we need to expose these consts?

internal/compose/compose.go
```diff
@@ -340,7 +345,29 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
 	return cmd.Run()
 }
 
+func (p *Project) DockerComposeVersion() (string, error) {
+
+	const (
+		v1marker = "docker-compose version 1"
```
mtojek, line 351: Please move it to the file const section.

Could you please also add a comment with a full dump of `docker-compose version`, so it's clear for the next developer what's going on and why.

internal/compose/compose.go
```diff
@@ -340,7 +345,29 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
...
+	)
+	var b bytes.Buffer
+
+	args := []string{"version"}
```
mtojek, line 355: I think there is a switch `--short`, that can strip all unnecessary information and even use semver library to parse the value.

mtojek (commented):
Thanks for your contribution, Sai! I left a few comments, so let me know what do you think :)

mtojek:
BTW CI reported some problems with this PR.

internal/compose/compose.go
r00tu53r: Awesome thanks !!

internal/compose/compose.go
mtojek: Hey, why did you resolve this comment? It isn't fixed yet.

internal/compose/compose.go
mtojek: Same story: it isn't resolved.

internal/compose/compose.go
mtojek: Unresolved

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -93,16 +93,20 @@ func (d *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
 		return nil, errors.Wrap(err, "service is unhealthy")
 	}
 
+	dcver, err := p.DockerComposeVersion()
```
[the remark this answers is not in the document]
mtojek: Unresolved

mtojek (commented):
Hey @r00tu53r, did you forget to push your branch? I don't see any changes applied, but comments were marked as resolved.

r00tu53r:
@mtojek sorry I forgot to push the branch. I've resolved all of the above issues.

internal/compose/compose.go
r00tu53r: Sounds good. I have refactored this to make it package-visible. To avoid calling `docker-compose --version` each time `ContainerName` is invoked I've stored the value in a project variable.

internal/compose/compose.go
```diff
@@ -340,7 +343,38 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
 	return cmd.Run()
 }
 
+func (p *Project) dockerComposeVersion() (*semver.Version, error) {
+
```
mtojek, line 347: nit: empty line (cane be removed)

internal/compose/compose.go
```diff
@@ -340,7 +343,38 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "docker compose version not a valid semver %s", dcVersion)
+	}
+	return ver, nil
+}
+
 // ContainerName method the container name for the service.
 func (p *Project) ContainerName(serviceName string) string {
-	return fmt.Sprintf("%s_%s_1", p.name, serviceName)
+	if p.dcMajorVersion == "" {
```
mtojek, line 367: It might be cleaner to keep the `ContainerName` function as simple as possible:

```
if p.dockerComposeV1 {
   return fmt.Sprintf("%s_%s_1", p.name, serviceName)
}
return fmt.Sprintf("%s-%s-1", p.name, serviceName)
```

Then, the logic determining `p.dockerComposeV1` can be moved to the `NewProject`.

`dockerComposeV1` can be bool, as the [Compose V2](https://docs.docker.com/compose/cli-command/) seems to deprecate V1 in the future.

internal/compose/compose.go
```diff
@@ -340,7 +343,38 @@ func (p *Project) runDockerComposeCmd(opts dockerComposeOptions) error {
...
+		"version",
+		"--short",
+	}
+	if err := p.runDockerComposeCmd(dockerComposeOptions{args: args, stdout: &b}); err != nil {
+		return nil, errors.Wrap(err, "running Docker Compose version command failed")
+	}
+	dcVersion := b.String()
+	ver, err := semver.NewVersion(dcVersion)
+	if err != nil {
+		return nil, errors.Wrapf(err, "docker compose version not a valid semver %s", dcVersion)
```
mtojek, line 360:
```suggestion
		return nil, errors.Wrapf(err, "docker compose version is not a valid semver (value: %s)", dcVersion)
```

internal/compose/compose.go
```diff
@@ -141,8 +143,9 @@ func NewProject(name string, paths ...string) (*Project, error) {
 	}
 
 	c := Project{
-		name,
-		paths,
+		name:             name,
+		dcMajorVersion:   "",
```
mtojek, line 147: As suggested in the other comment, `dcMajorVersion` can be replaced with `dockerComposeV1` (bool).

Then in the `NewProject`, you can evaluate its value. 

BTW I hope that the `Project` isn't created somewhere just with `{}`.

mtojek (commented):
I added a few suggestions, let me know what do you think.

internal/compose/compose.go
```diff
@@ -141,10 +143,18 @@ func NewProject(name string, paths ...string) (*Project, error) {
 	}
 
 	c := Project{
-		name,
-		paths,
+		name:             name,
+		dockerComposeV1:  true,
```
mtojek, line 147: nit: You could also skip this property here and set it only when `ve.Major` == 1 or undefined. But it would the same outcome :)

mtojek (approved):
Two small nit-picks are left, but it looks fine now!

Let me know what do you think about them and then I will merge this PR.

internal/compose/compose.go
r00tu53r: Sorry I didn't quite get what you meant. Did you mean we could use the semver.Version (`*Version`) inside the `Project` and use `ver.Major()` in `ContainerName` instead ?

mtojek: Currently, you're setting the property `dockerComposeV1:  true` by default. But you could adjust it to be the other way around, so  just assume:

```go
c := Project{
   name: name,
   composeFilePaths: paths
}
```

and then just enable it for two cases:

```go
if ver.Major() == 1 {
   c.dockerComposeV1 = true
}
```

```go
	if err != nil {
		logger.Errorf("Unable to determine docker-compose version: %v. Defaulting to 1.x", err)
                 c.dockerComposeV1 = true
		return &c, nil
	}
```

mtojek: Let's see if I have permissions to push it here, so that we don't need to bother you anymore :)

mtojek: Done

mtojek:
Error seems to be unrelated. I will merge this PR.

PR #635, opened 2021-12-22
Flaky: increase boot timeout

mtojek:
The [main build job](https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/master/295/) just failed because of slow agent startup. This PR increases the boot timeout.


mtojek:
self-approved.

PR #637, opened 2021-12-24
Emergency fix: remove username/password in Elastic Agent

mtojek:
Related: https://github.com/elastic/beats/issues/29603


mtojek:
emergency fix, self-approved

PR #648, opened 2022-01-16
internal/version: make version available when installed via go install github.com/elastic/elastic-package

efd6:
This makes to output of elastic-package version informative when the go.mod-based install within integrations is used. It does not prevent the "CommitHash is undefined" warning since that is based on build-time var substitution. The logic required to properly handle update detection is reasonably involved, so I have not added this (see [here](https://github.com/kortschak/ugbt/blob/18ee59e607e8117ba0bf85621cf69203df511891/cmd.go#L658) for the general case — not all of that would be required here).

The output of `elastic-package version` when it has been installed using `go install github.com/elastic/elastic-package@latest` or by `go install github.com/elastic/elastic-package` from within a module with it written as a require would look something like this

```
elastic-package v0.43.1 version-hash 032f343 (build time: 2052-01-18T19:38:06+4:30)
```

Please take a look.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -5,6 +5,7 @@
 package version
 
 import (
+	"runtime/debug"
 	"strconv"
 	"time"
 )
@@ -19,6 +20,17 @@ var (
 	// Tag describes the semver version of the application (set externally with ldflags).
 	Tag string
 )
+
+// Set Tag to version stored in modinfo if it is not available from the builder.
+func init() {
+	if Tag != "" {
+		return
+	}
+	info, ok := debug.ReadBuildInfo()
+	if ok && info.Main.Version != "(devel)" {
+		Tag = info.Main.Version
+	}
+}
 
 // BuildTimeFormatted method returns the build time preserving the RFC3339 format.
 func BuildTimeFormatted() string {


efd6:
Failure looks to be unrelated.

mtojek (commented):
Hey @efd6! thanks for the PR. Regarding failing tests, please merge the latest mainline as we pushed a fix there.

Could you please add to the PR description the sample output of the `version` command?

efd6:
The output of the version command will remain the same as previously when the Tag is set using -ldflags=-X. I can't demonstrate until there is a merged commit with this code in, but I will construct a hand drawn facsimile.

mtojek (approved):
I'm a bit concerned about the `profiles` logic, but in the worst case, we can revert this change.

If the CI is happy, I'm fine too.

efd6:
@mtojek With this change in and installing this particular commit I get this:
```
$ go install github.com/elastic/elastic-package@65f8e5
go: downloading github.com/elastic/elastic-package v0.33.2-0.20220118093250-65f8e5e840b5
$ elastic-package version
elastic-package v0.33.2-0.20220118093250-65f8e5e840b5 version-hash undefined (build time: unknown)
```
If it's installed with a tagged version the pseudoversion suffix will be absent.

What are your concerns about the profiles logic?

mtojek:
@efd6 I'm curious what would be the content of the `profile.json` file if you create a new profile:

`cat ~/.elastic-package/profiles/default/profile.json`

In my case it's now:

```json
{
  "name": "default",
  "date_created": "2022-01-11T16:50:19.230744+01:00",
  "user": "marcin.tojek",
  "version": "30faef5",
  "path": "/Users/marcin.tojek/.elastic-package/profiles/default"
}
```

version contains a hash.

> elastic-package v0.33.2-0.20220118093250-65f8e5e840b5 version-hash undefined (build time: unknown)

It still looks strange, so maybe we should fully switch to the `debug.ReadBuildInfo`? We just need to make sure that the logic uses it correctly to check the `~/.elastic-package` dir in terms of stack defs and profiles. WDYT?

efd6:
In the case with the install I just did, I see
```
 cat ~/.elastic-package/profiles/default/profile.json
{
  "name": "default",
  "date_created": "2022-01-18T20:46:06.107629+10:30",
  "user": "xxx",
  "version": "undefined",
  "path": "/Users/xxx/.elastic-package/profiles/default"
}
```

Using only the embedded version information should be fine, in fact the hash gives no ordering to versions as is noted in output when there is no version in the [current code](https://github.com/elastic/elastic-package/blob/65f8e5e840b5c210ab0ed41f80cf2ff787b81b76/internal/version/check_update.go#L23-L27). The work that's needed is much less than is done in `ugbt`. If you'd like to go that route I'm happy to discuss.

mtojek:
If there is a way to modify the code to properly return `version` for both `make build` and `go get` and `go get @master`, then I'm more than happy to look into a draft PR :)

PR #651, opened 2022-01-18
Set version var during make build

efd6:
This is a follow up from #648, @mtojek please take a look.

The change bakes in the current version tag during `make build` if the repo is checked out at a tagged version.

Prior to this a user building `elastic-package` using `make` would be given an executable that is unaware of its version. Now `make` will leave an executable that renders `elastic-package version` similar to this when the source tree is checked out at a tagged version:
```
elastic-package v0.42.2 version-hash 3478ac7 (build time: 2022-01-20T09:11:07+4:30
```
If not at a tagged version, no semver version is output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,12 @@
 CODE_COVERAGE_REPORT_FOLDER = $(PWD)/build/test-coverage
 CODE_COVERAGE_REPORT_NAME_UNIT = $(CODE_COVERAGE_REPORT_FOLDER)/coverage-unit-report
+VERSION_IMPORT_PATH = github.com/elastic/elastic-package/internal/version
 
 .PHONY: build
 
 build:
-	go get -ldflags "-X github.com/elastic/elastic-package/internal/version.CommitHash=`git describe --always --long --dirty` -X github.com/elastic/elastic-package/internal/version.BuildTime=`date +%s`" \
+	go install -ldflags \
+	    "-X $(VERSION_IMPORT_PATH).CommitHash=`git describe --always --long --dirty` -X $(VERSION_IMPORT_PATH).BuildTime=`date +%s` -X $(VERSION_IMPORT_PATH).Tag=`(git describe --exact-match --tags 2>/dev/null || echo '') | tr -d '\n'`" \
 	    github.com/elastic/elastic-package
 
 clean:


Makefile
```diff
@@ -1,10 +1,12 @@
 CODE_COVERAGE_REPORT_FOLDER = $(PWD)/build/test-coverage
 CODE_COVERAGE_REPORT_NAME_UNIT = $(CODE_COVERAGE_REPORT_FOLDER)/coverage-unit-report
+VER_PKG = github.com/elastic/elastic-package/internal/version
```
mtojek, line 3: Let's rename it to the:`VERSION_IMPORT_PATH`

mtojek (commented):
@efd6 

Could you please add to the PR description, what is the test case the PR fixes?

Makefile
efd6: The choice was intentional to reduce line length. It's not far from its use, so the context is still strong. If you would still like me to make that change given this, let me know.

mtojek: Yes, I suggest extending it as VER_PKG might be misleading.

Makefile
```diff
@@ -1,10 +1,12 @@
...
 
 .PHONY: build
 
 build:
-	go get -ldflags "-X github.com/elastic/elastic-package/internal/version.CommitHash=`git describe --always --long --dirty` -X github.com/elastic/elastic-package/internal/version.BuildTime=`date +%s`" \
+	go install -ldflags \
+	    "-X $(VER_PKG).CommitHash=`git describe --always --long --dirty` -X $(VER_PKG).BuildTime=`date +%s` -X $(VER_PKG).Tag=`(git describe --exact-match --tags 2>/dev/null || echo '') | tr -d '\n'`" \
```
jsoriano, line 9: Maybe this line can be split in multiple lines for easier readability.

efd6: Unfortunately this is not possible due to the interaction of make string handling and `-ldflags` behaviour.

Makefile
efd6: It's not clear to me how, but done.

mtojek (approved):

Makefile
mtojek: My first guess was PKG_CONFIG as it is a Makefile.

Thanks for adjusting it.

Makefile
mtojek: My first guess was PKG_CONFIG as it is a Makefile.

Thanks for adjusting it.

PR #656, opened 2022-01-19
Extract processors from an ingest.Pipeline

adriansr:
This adds a _Processors()_ method to the ingest.Pipeline type to retrieve a list of processors from the pipeline source code.


internal/elasticsearch/ingest/processors.go
```diff
@@ -0,0 +1,245 @@
...
+			return
+		}
+	}
+}
+
+func (s *tokenStack) Top() json.Token {
+	if n := len(*s); n > 0 {
+		return (*s)[n-1]
+	}
+	return io.EOF
```
efd6, line 110: This is sort of surprising. Maybe return nil when empty?

internal/elasticsearch/ingest/processors.go
```diff
@@ -0,0 +1,245 @@
...
+	var names []string
+	// start and end offsets for each processor.
+	var startPos, endPos []int
+	var stack tokenStack
+	decoder := json.NewDecoder(bytes.NewReader(content))
+
+	for {
+		// Read the next token and it's offset.
+		tk, err := decoder.Token()
+		off := int(decoder.InputOffset())
```
efd6, line 146: After the err check? `off` is not used when `err` is not nil.

internal/elasticsearch/ingest/processors.go
```diff
@@ -0,0 +1,245 @@
...
+func processorsFromYAML(content []byte) (procs []Processor, err error) {
+	var p struct {
+		Processors []yaml.Node
+	}
+	if err = yaml.Unmarshal(content, &p); err != nil {
+		return nil, err
+	}
+	for idx, entry := range p.Processors {
+		if entry.Kind != yaml.MappingNode || len(entry.Content) != 2 {
+			return nil, errors.Errorf("processor#%d is not a single-key map (kind:%v Content:%d)", idx, entry.Kind, len(entry.Content))
```
mtojek, line 52: nit: `content:%d`

internal/elasticsearch/ingest/processors.go
```diff
@@ -0,0 +1,245 @@
...
+// extract a list of processors from a pipeline definition in YAML format.
+func processorsFromYAML(content []byte) (procs []Processor, err error) {
+	var p struct {
+		Processors []yaml.Node
+	}
+	if err = yaml.Unmarshal(content, &p); err != nil {
+		return nil, err
+	}
+	for idx, entry := range p.Processors {
+		if entry.Kind != yaml.MappingNode || len(entry.Content) != 2 {
```
mtojek, line 51: Is it always 2 or there is slight chance that the `Content` will contain something extra?

internal/elasticsearch/ingest/processors.go
```diff
@@ -0,0 +1,245 @@
...
+			last = line
+		}
+	}
+	return last
+}
+
+// tokenStack contains the current state of parsing a JSON document.
+type tokenStack []json.Token
+
+func (s *tokenStack) Push(t json.Token) {
```
mtojek, line 85: Do we need to expose all these methods?

internal/elasticsearch/ingest/processors.go
```diff
@@ -0,0 +1,245 @@
...
+	LastLine int `yaml:"-"`
+}
+
+// Processors return the list of processors in an ingest pipeline.
+func (p Pipeline) Processors() (procs []Processor, err error) {
+	switch p.Format {
+	case "yaml", "yml":
+		procs, err = processorsFromYAML(p.Content)
+	case "json":
+		procs, err = processorsFromJSON(p.Content)
```
mtojek, line 35: I see that YAML processing has much simpler logic than JSON processing. Do both formats differ much from each other in terms of properties? JSON is a subset of YAML. If there are differences, maybe it's valid to drop a comment here, justifying why there are two separate functions.

internal/elasticsearch/ingest/processors.go
adriansr: done

internal/elasticsearch/ingest/processors.go
adriansr: My understanding its that is always 2 for MappingNode, but wanted to check just in case

internal/elasticsearch/ingest/processors.go
adriansr: Removed

efd6 (approved):

internal/elasticsearch/ingest/processors.go
mtojek: It's much simpler now, tons of code removed, good job!

mtojek (approved):
Ship it!

jsoriano (approved):
Nice! Will this be used in the coverage reports?

Issue #655, opened 2022-01-19
Predefine agent policies in the Kibana config


mtojek:
Kibana Fleet team is going to deprecate and remove default agent policies in https://github.com/elastic/kibana/issues/108456.

It means that the `elastic-package stack up` won't boot up eventually if the ~8.0.0-SNAPSHOT~ 8.1.0-SNAPSHOT image is selected. The goal of this issue is to predefine policies in the `kibana.config.yml` (or a file close to it) and use them instead ones created by Elastic Agents.

@juliaElastic provided us also sample policies: [here](https://github.com/elastic/kibana/issues/108456#issuecomment-1016502179) (thanks!)

mtojek:
It looks like we're safe to merge the [PR](https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-676/6/pipeline/). The flag `is_default` worked well.

PR #662, opened 2022-01-24
Use GCP ADC for auth in Terraform Deployer

endorama:
Terraform Google Provider allow to pass credentials file content in `GOOGLE_CREDENTIALS` env var.

More about this in the Terraform docs: https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#full-reference

Unfortunately this does not seem to work other GCP Cloud SDK components,
as I've not been able to find references to that environment variable.

As Terraform provider and components support ADC (Application Defatul
Credentials) this commit changes how Terraform Deployer setup
credentials to use ADC instead of Terraform dedicated variable.


internal/install/_static/terraform_deployer_run.sh
```diff
@@ -16,6 +16,15 @@ cleanup() {
 }
 trap cleanup EXIT INT TERM
 
+# Save GCP credentials on disk and perform authentication
+# NOTE: this is required for bq (and maybe other gcloud related tools) to authenticate
+export "GOOGLE_APPLICATION_CREDENTIALS=/root/.config/gcloud/application_default_credentials.json"
+echo "$GOOGLE_CREDENTIALS" > "$GOOGLE_APPLICATION_CREDENTIALS"
+gcloud auth login --cred-file "$GOOGLE_APPLICATION_CREDENTIALS"
+# NOTE: Terraform support authentication through GOOGLE_CREDENTIALS and usual gcloud ADC but other
+# tools (like bq) don't support the first, so we always rely on gcloud ADC.
```
mtojek, line 25: Could you please point me to the integration/system test, where you intend to use `bq`? I'm wondering if something changed since that time and maybe we're fine to use a [terraform module](https://github.com/terraform-google-modules/terraform-google-bigquery).

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -16,6 +16,15 @@ cleanup() {
 }
 trap cleanup EXIT INT TERM
 
+# Save GCP credentials on disk and perform authentication
+# NOTE: this is required for bq (and maybe other gcloud related tools) to authenticate
+export "GOOGLE_APPLICATION_CREDENTIALS=/root/.config/gcloud/application_default_credentials.json"
+echo "$GOOGLE_CREDENTIALS" > "$GOOGLE_APPLICATION_CREDENTIALS"
+gcloud auth login --cred-file "$GOOGLE_APPLICATION_CREDENTIALS"
```
mtojek, line 23: If this is required by `bq`, maybe we can execute it just before running `bq`?

mtojek, line 23: Is there any default location, so that `cred-file` is not needed to be defined?

mtojek (changes requested):
If we decide to extend the terraform deployer with such a change, we have to add a test package here and make sure the flow works.

Let's decide first if this is the way we'd like to follow.

@endorama Could you please clarify (describe) your use case? Is there any PR we can look at?

internal/install/_static/terraform_deployer_run.sh
endorama: Is being used in this PR https://github.com/elastic/integrations/pull/2312

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -16,6 +16,15 @@ cleanup() {
 }
 trap cleanup EXIT INT TERM
 
+# Save GCP credentials on disk and perform authentication
```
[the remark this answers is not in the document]
endorama: Right, sorry for the bad implementation :) This is an easy fix, adding a guard clause for presence of the `GOOGLE_CREDENTIALS` variable. Before adding it I'd like to discuss the other points.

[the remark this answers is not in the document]
mtojek: It isn't a matter of adding a guard, but is it really necessary to put it here?  see: https://github.com/elastic/elastic-package/pull/662#discussion_r791557079

internal/install/_static/terraform_deployer_run.sh
endorama: I did not find any documentation allowing to use a default location, the argument is required as is like providing username/password.

endorama: In case we are going to use other tools, something possible given the new Docker image, I think it make sense to unify authentication so all tools leverage the same credentials and should not do it multiple times (with possible concurrency risks). 
Unifying authentication also prevents that Terraform and tools run with different credentials, something that would make the CI harder to debug.

internal/install/_static/terraform_deployer_run.sh
mtojek: >This approach has multiple disadvantages: it requires more infrastructure to be created which increase complexity

What do you mean by "more infrastructure"? It sounds easier to load data on S3, than refactoring service deployer :) 

> it prevent use of emulator and there is no support to waiting for the load job completion.

This is really unfortunate and without a proper waiter it won't be possible to perform it. 

Thanks for the explainaton.

internal/install/_static/terraform_deployer_run.sh
mtojek: Ok, so it looks like we need to confirm that it works as expected. It would be great if you prepare a similar test package like the special [aws](https://github.com/elastic/elastic-package/tree/main/test/packages/parallel/aws). It's the AWS integration with stripped data streams, only one left with a system test using Terraform service deployer to create the EC2 machine.

Here is my proposal on the plan:

You need to do a similar exercise with GCP. You can use this PR to push the sample package, but please make sure it's stripped and CI discovers it. Then, please you can execute the system test, but I'm afraid that it may fail due to configured environment variables. It's about modifying the Jenkinsfile to enable those envs. Then, we should observe some issues around `bq` and terraform service deployer, and eliminate them.

internal/install/_static/terraform_deployer_run.sh
endorama: >  It sounds easier to load data on S3, than refactoring service deployer :)

😅 totally true. 

If `google_bigquery_job` was working as expected I would have chosen that way, because at least is predictable and all within terraform, but I think we could only have something fragile at the moment.

internal/install/_static/terraform_deployer_run.sh
endorama: > Ok, so it looks like we need to confirm that it works as expected.

May you clarify what should be confirmed? 

What's the goal for this test package? Is to test if tests are being able to be executed correctly through terraform deployer?

mtojek: Yes, that's the goal - to verify if the terraform service deployer works correctly with Google Cloud: auth, bq, Jenkins changes.

endorama: I reviewed the pro and cons of moving authentication where `bq` is used. The main cons I see is around how this affects authentication globally. `gcloud auth` state is globally shared across all instances of `gcloud` (and related tools). So doing that within terraform code would create side effects that may be unintuitive or difficult to debug.

What is the downside of the current approach? (given that it does not trigger always but only when it makes sense)

mtojek: > What is the downside of the current approach? (given that it does not trigger always but only when it makes sense)

I think you included the answer here. If there is a way of not introducing the coupling on auth, why should we do this? Please remember that it isn't a dedicated GCP service.

> gcloud auth state is globally shared across all instances of gcloud (and related tools). So doing that within terraform code would create side effects that may be unintuitive or difficult to debug.

It means that you can call it this way:

````
gcloud auth login
bq
gcloud auth revoke
```

endorama: @mtojek I added the test package, the pipeline is green but is it really running some tests? https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-662/8/pipeline/279

endorama: You are at risk of race conditions if multiple resources are using this commands. Terraform runs in parallel and this command affects a global state.

> If there is a way of not introducing the coupling on auth, why should we do this?

My reasoning about this is that coupling is already present as in "terraform requires credentials to run", so if I use it with GCP resources is coupled with GCP auth, if I use it with AWS is coupled with AWS auth. How could we avoid this? I agree that ideally credentials would be passed in env vars and we should not rely on other tools.

> Please remember that it isn't a dedicated GCP service.

Could we consider having a dedicated deployer for gcp? Would this help untangling this?

mtojek:
@endorama Please pull the latest main branch.

endorama:
The CI is failing because the GCP package does not pass the lint check. This is the shown error:
```
Error: checking package failed: linting package failed: found 4 validation errors:
   1. item [.gitignore] is not allowed in folder [.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]
   2. item [.terraform.lock.hcl] is not allowed in folder [.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]
   3. item [billing-schema.json] is not allowed in folder [.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]
   4. item [test-data.ndjson.tftpl] is not allowed in folder **[.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]**
```
File 3 and 4 are mandatory.
File 2 has been added because [Terraform suggest to do it for dependency management](https://www.terraform.io/language/files/dependency-lock#dependency-installation-behavior). In my opinion is a great idea as it makes the build reproducible.
File 1 has been added to prevent local folder or state files to be included.

@mtojek I'll open an issue in https://github.com/elastic/package-spec to discuss this further.

endorama:
I updated `github.com/elastic/package-spec`  to `v1.4.1` for testing waiting for https://github.com/elastic/elastic-package/pull/693

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -2,25 +2,41 @@
...
+    export "GOOGLE_APPLICATION_CREDENTIALS=/root/.config/gcloud/application_default_credentials.json"
+    printenv "GOOGLE_CREDENTIALS" > "$GOOGLE_APPLICATION_CREDENTIALS"
+    gcloud auth login --cred-file "$GOOGLE_APPLICATION_CREDENTIALS"
+    # NOTE: Terraform support authentication through GOOGLE_CREDENTIALS and usual gcloud ADC but other
+    # tools (like bq) don't support the first, so we always rely on gcloud ADC.
+    unset "GOOGLE_CREDENTIALS"
+  fi
+}
+
+if [[ "${BASH_SOURCE[0]}" = "$0" ]]; then
```
mtojek, line 26: Why is this condition required here?

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -2,25 +2,41 @@
```
mtojek, line 2: I don't see any container logs for the GCP test package in `beats-ci-temp-internal/Ingest-manager/elastic-package/PR-662-16/insecure-logs/gcp`. Is it intended?

internal/install/_static/terraform_deployer_run.sh
endorama: It is not required, but it avoid running the code in case the file is `source`d and identify a "main" like are, like the `main` function in go or C. From that point onward all code will be executed only when the file is explicitly executed (`bash file.sh` or `./file.sh`). It helps grouping the relevant "main" code in the same area, preventing arbitrary code being added between functions.

Consider it my personal way of writing BASH files, I think it aids reading them (like `__file__ == __main__` does for Python)

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -2,25 +2,41 @@
...
+  gcp_auth
+
+  terraform init
+  terraform plan
+  terraform apply -auto-approve && touch /tmp/tf-applied
 
-echo "Terraform definitions applied."
+  echo "Terraform definitions applied."
 
-set +x
```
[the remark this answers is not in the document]
mtojek: In #655 we tried different approaches to fix the bug with `bash/sh`. If it isn't a problem, I would preserve `set -x` and `set +x` in this PR, but please make sure (locally) it doesn't leak any credentials.

internal/install/_static/terraform_deployer_run.sh
mtojek: I understand your point of view, but please keep it simple as it was before.

internal/install/_static/terraform_deployer_run.sh
endorama: No is not, [from the pipeline logs](https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-662/16/pipeline/280#step-1190-log-244) it seems it not running system tests (command start at line 244, hidden by the task header)

I set up the `env.yml` file in [`data_stream/billing/_dev/deploy/tf`](https://github.com/elastic/elastic-package/pull/662/files#diff-344f9471070f52db55ee64eb5349532ba4f52bcdf2a2418bfa983add481ef6d6). Is this not enough to run system tests?

mtojek: I checked your branch and you haven't configured any test policies. Test policies are used by elastic-package during system tests. Otherwise the elastic-package won't know what you re trying to test. It's covered in [our manual](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md#test-case-definition).

Please take a look at the AWS test package and [ec2_metrics](https://github.com/endorama/elastic-package/tree/terraform-deployer-fix-gcp-auth/test/packages/parallel/aws/data_stream/ec2_metrics/_dev/test/system) tests.

mtojek:
@endorama I converted this PR to draft as it's still in progress (missing tests, unaddressed comments).

internal/install/_static/terraform_deployer_run.sh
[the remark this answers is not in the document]
endorama: Addressed in #699

internal/install/_static/terraform_deployer_run.sh
endorama: Test package with first test case (not requiring changes in this PR) added in https://github.com/elastic/elastic-package/pull/701

internal/install/_static/terraform_deployer_run.sh
endorama: Is run by container engine but is still read by developers :) The point was to simplifying reading the code flow. Anyway I removed it (will push once rebased onto main)

mtojek:
@endorama Are you going to rebase this PR as we merged the other one?

mtojek:
Closing it for now as it's a stale PR.

PR #665, opened 2022-01-26
Reduce terraform deployer build verbosity

endorama:
Reduce output from gcloud SDK installation process, by silencing the various steps.

This is needed to reduce image build log verbosity; gcloud SDK installation output is also difficult to read on terminal that do not support colours.

### DIFF
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -18,6 +18,7 @@ RUN echo "deb [arch=amd64] https://apt.releases.hashicorp.com focal main" | tee 
 HEALTHCHECK --timeout=3s CMD sh -c "[ -f /tmp/tf-applied ]"
 
 ENV TF_IN_AUTOMATION=true
+ENV TF_CLI_ARGS="-no-color"
 ADD run.sh /
 WORKDIR /workspace
 
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -1,14 +1,14 @@
 FROM hashicorp/terraform:light as terraform
 
 FROM ubuntu:20.04
-
-RUN apt-get -q update && apt-get install -yq curl apt-transport-https ca-certificates gnupg && apt-get clean
-
 ENV GCLOUD_SDK_VERSION 370.0.0-0
-RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
+
+RUN apt-get -q update \
+  && apt-get install -yq curl apt-transport-https ca-certificates gnupg \
+  && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
   && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - \
-  && apt-get update -y \
-  && apt-get install google-cloud-sdk=${GCLOUD_SDK_VERSION} -y
+  && apt-get update -yq \
+  && apt-get install google-cloud-sdk=${GCLOUD_SDK_VERSION} -yq
 
 HEALTHCHECK --timeout=3s CMD sh -c "[ -f /tmp/tf-applied ]"
 
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -3,11 +3,12 @@ FROM hashicorp/terraform:light as terraform
 FROM ubuntu:20.04
 ENV GCLOUD_SDK_VERSION 370.0.0-0
 
-RUN apt-get -q update \
-  && apt-get install -yq curl apt-transport-https ca-certificates gnupg \
-  && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
+RUN apt-get -qq update \
+  && apt-get install -yq curl apt-transport-https ca-certificates gnupg
+
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
   && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - \
-  && apt-get update -yq \
+  && apt-get update -qq \
   && apt-get install google-cloud-sdk=${GCLOUD_SDK_VERSION} -yq
 
 HEALTHCHECK --timeout=3s CMD sh -c "[ -f /tmp/tf-applied ]"
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -10,7 +10,7 @@ ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
 RUN curl --silent "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$GCLOUD_SDK_VERSION-linux-x86_64.tar.gz" > /tmp/google-cloud-sdk.tar.gz \
   && mkdir -p /usr/local/gcloud \
   && tar -C /usr/local/gcloud -xf /tmp/google-cloud-sdk.tar.gz \
-  && /usr/local/gcloud/google-cloud-sdk/install.sh -q --override-components="bq" --disable-prompts \
+  && /usr/local/gcloud/google-cloud-sdk/install.sh -q --override-components="bq" &>/dev/null \
   && rm /tmp/google-cloud-sdk.tar.gz
 
 HEALTHCHECK --timeout=3s CMD sh -c "[ -f /tmp/tf-applied ]"
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -22,4 +22,4 @@ ENV TF_CLI_ARGS="-no-color"
 ADD run.sh /
 WORKDIR /workspace
 
-ENTRYPOINT sh /run.sh
+ENTRYPOINT bash /run.sh
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -1,17 +1,14 @@
 FROM hashicorp/terraform:light as terraform
 
-FROM python:3-alpine
+FROM ubuntu:20.04
 
-RUN apk add --no-cache git openssh curl
+RUN apt-get -q update && apt-get install -yq curl apt-transport-https ca-certificates gnupg && apt-get clean
 
-# Announcements for new releases: https://groups.google.com/g/google-cloud-sdk-announce
-ENV GCLOUD_SDK_VERSION 369.0.0
-ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
-RUN curl --silent "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$GCLOUD_SDK_VERSION-linux-x86_64.tar.gz" > /tmp/google-cloud-sdk.tar.gz \
-  && mkdir -p /usr/local/gcloud \
-  && tar -C /usr/local/gcloud -xf /tmp/google-cloud-sdk.tar.gz \
-  && /usr/local/gcloud/google-cloud-sdk/install.sh -q --override-components="bq" &>/dev/null \
-  && rm /tmp/google-cloud-sdk.tar.gz
+ENV GCLOUD_SDK_VERSION 370.0.0-0
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
+  && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - \
+  && apt-get update -y \
+  && apt-get install google-cloud-sdk=${GCLOUD_SDK_VERSION} -y
 
 HEALTHCHECK --timeout=3s CMD sh -c "[ -f /tmp/tf-applied ]"
 
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -1,7 +1,6 @@
-FROM hashicorp/terraform:light as terraform
-
 FROM ubuntu:20.04
 ENV GCLOUD_SDK_VERSION 370.0.0-0
+ENV TERRAFORM_VERSION 1.1.4
 
 RUN apt-get -qq update \
   && apt-get install -yq curl apt-transport-https ca-certificates gnupg
@@ -11,9 +10,12 @@ RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.c
   && apt-get update -qq \
   && apt-get install google-cloud-sdk=${GCLOUD_SDK_VERSION} -yq
 
-HEALTHCHECK --timeout=3s CMD sh -c "[ -f /tmp/tf-applied ]"
+RUN echo "deb [arch=amd64] https://apt.releases.hashicorp.com focal main" | tee -a /etc/apt/sources.list.d/hashicorp.list \
+  && curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add - \
+  && apt-get -qq update \
+  && apt-get install -yq terraform=${TERRAFORM_VERSION}
 
-COPY --from=terraform /bin/terraform /usr/bin/terraform
+HEALTHCHECK --timeout=3s CMD sh -c "[ -f /tmp/tf-applied ]"
 
 ENV TF_IN_AUTOMATION=true
 ADD run.sh /
--- a/internal/install/_static/Dockerfile.terraform_deployer
+++ b/internal/install/_static/Dockerfile.terraform_deployer
@@ -2,15 +2,15 @@ FROM hashicorp/terraform:light as terraform
 
 FROM python:3-alpine
 
-# required by gcloud SDK
 RUN apk add --no-cache git openssh curl
 
+# Announcements for new releases: https://groups.google.com/g/google-cloud-sdk-announce
 ENV GCLOUD_SDK_VERSION 369.0.0
 ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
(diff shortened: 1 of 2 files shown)


internal/install/_static/Dockerfile.terraform_deployer
```diff
@@ -2,15 +2,15 @@ FROM hashicorp/terraform:light as terraform
 
 FROM python:3-alpine
 
-# required by gcloud SDK
 RUN apk add --no-cache git openssh curl
 
+# Announcements for new releases: https://groups.google.com/g/google-cloud-sdk-announce
```
mtojek, line 7: 👍

internal/install/_static/Dockerfile.terraform_deployer
```diff
@@ -2,15 +2,15 @@ FROM hashicorp/terraform:light as terraform
...
 ENV GCLOUD_SDK_VERSION 369.0.0
 ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
-RUN curl "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$GCLOUD_SDK_VERSION-linux-x86_64.tar.gz" > /tmp/google-cloud-sdk.tar.gz \
+RUN curl --silent "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$GCLOUD_SDK_VERSION-linux-x86_64.tar.gz" > /tmp/google-cloud-sdk.tar.gz \
```
mtojek, line 10: `curl -s <url> -o /tmp/google-cloud-sdk.tar.gz`

internal/install/_static/Dockerfile.terraform_deployer
```diff
@@ -2,15 +2,15 @@ FROM hashicorp/terraform:light as terraform
...
   && mkdir -p /usr/local/gcloud \
-  && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
-  && /usr/local/gcloud/google-cloud-sdk/install.sh -q --override-components="bq" \
+  && tar -C /usr/local/gcloud -xf /tmp/google-cloud-sdk.tar.gz \
+  && /usr/local/gcloud/google-cloud-sdk/install.sh -q --override-components="bq" &>/dev/null \
```
mtojek, line 13: According to the public [docs](https://cloud.google.com/sdk/docs/downloads-interactive#silent), there is `--disable-prompts` flag. We can give it a try and let the installer decide what's rubbish and what isn't :)

endorama: updated. Where can I look at the logs? I've not been able to find the pipeline where this log is displayed :(

mtojek: Do you mean [this](https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/PR-665/2/pipeline)?

```
[2022-01-26T14:00:31.542Z]  ---> Running in f3dbb078bd82
[2022-01-26T14:00:41.632Z] Welcome to the Google Cloud SDK!
[2022-01-26T14:00:41.632Z] WARNING: You appear to be running this script as root. This may cause 
[2022-01-26T14:00:41.632Z] the installation to be inaccessible to users other than the root user.
[2022-01-26T14:00:43.561Z] usage: install.py [-h] [--usage-reporting USAGE_REPORTING]
[2022-01-26T14:00:43.561Z]                   [--screen-reader SCREEN_READER] [--rc-path RC_PATH]
[2022-01-26T14:00:43.561Z]                   [--command-completion COMMAND_COMPLETION]
[2022-01-26T14:00:43.561Z]                   [--path-update PATH_UPDATE] [--disable-installation-options]
[2022-01-26T14:00:43.561Z]                   [--override-components [OVERRIDE_COMPONENTS [OVERRIDE_COMPONENTS ...]]]
[2022-01-26T14:00:43.561Z]                   [--additional-components ADDITIONAL_COMPONENTS [ADDITIONAL_COMPONENTS ...]]
[2022-01-26T14:00:43.561Z]                   [--quiet] [--install-python INSTALL_PYTHON]
[2022-01-26T14:00:43.561Z] install.py: error: unrecognized arguments: --disable-prompts
[2022-01-26T14:00:47.792Z] Service 'terraform' failed to build: The command '/bin/sh -c curl --silent "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$GCLOUD_SDK_VERSION-linux-x86_64.tar.gz" > /tmp/google-cloud-sdk.tar.gz   && mkdir -p /usr/local/gcloud   && tar -C /usr/local/gcloud -xf /tmp/google-cloud-sdk.tar.gz   && /usr/local/gcloud/google-cloud-sdk/install.sh -q --override-components="bq" --disable-prompts   && rm /tmp/google-cloud-sdk.tar.gz' returned a non-zero code: 2
```

mtojek: It looks there are different options available, sorry for linking wrong docs. Could you please adjust the command?

endorama: `--disable-prompts` is not available with the installation method in use, the tests for https://github.com/elastic/elastic-package/pull/665/commits/8e505f2353f21fa23ef6ce0e4e989675540ce553 broke with:

```
[2022-01-26T14:00:41.632Z] Welcome to the Google Cloud SDK!
[2022-01-26T14:00:41.632Z] WARNING: You appear to be running this script as root. This may cause 
[2022-01-26T14:00:41.632Z] the installation to be inaccessible to users other than the root user.
[2022-01-26T14:00:43.561Z] usage: install.py [-h] [--usage-reporting USAGE_REPORTING]
[2022-01-26T14:00:43.561Z]                   [--screen-reader SCREEN_READER] [--rc-path RC_PATH]
[2022-01-26T14:00:43.561Z]                   [--command-completion COMMAND_COMPLETION]
[2022-01-26T14:00:43.561Z]                   [--path-update PATH_UPDATE] [--disable-installation-options]
[2022-01-26T14:00:43.561Z]                   [--override-components [OVERRIDE_COMPONENTS [OVERRIDE_COMPONENTS ...]]]
[2022-01-26T14:00:43.561Z]                   [--additional-components ADDITIONAL_COMPONENTS [ADDITIONAL_COMPONENTS ...]]
[2022-01-26T14:00:43.561Z]                   [--quiet] [--install-python INSTALL_PYTHON]
[2022-01-26T14:00:43.561Z] install.py: error: unrecognized arguments: --disable-prompts
```

Reverting the commit

mtojek: Look at the usage. It seems that you can try using other flags: `--quiet`, `--disable-installation-options`.

endorama: `--quiet` was already used. `--disable-installation-options` does not change the output. 

I've not been able to find a better alternative than redirecting to `/dev/null`. Also part of the installation output is sent to `stderr` (the colored lines), so the only way to suppress output is to redirect both `stdout` and `stderr` to `/dev/null`.

mtojek: I see, thanks for checking this. In this case, I would be for not hiding any possible errors by flushing them to `/dev/null`. Let's remove output redirection, but keep the "quiet mode"

mtojek: @endorama I looked at [docs](https://cloud.google.com/sdk/docs/install#other_installation_options) once again and it recommends  not executing `install.sh` for CI setups:

> Using Cloud SDK with scripts or Continuous Integration/Deployment? Download a versioned archive for a non-interactive installation of a specific version of Cloud SDK.

Step 4 of the [instruction](https://cloud.google.com/sdk/docs/downloads-versioned-archives#installation_instructions) says that `install.sh` just sets the PATH accordingly and installs bash-completion, which means we really don't need to do this here.

Could you please try again without calling `install.sh` at all? I suppose you need to unpack the archive in the correct place instead of `/tmp`.

PS We will definitely need a package system test to verify if it works as expected.

jsoriano: Sorry, arriving late to the party. Now that this image is having dependencies with more complex installation methods, wdyt about changing the base image to an ubuntu one and [use apt](https://cloud.google.com/sdk/docs/install#deb) (with the `--quiet` flag :)) to install it?

mtojek: > wdyt about changing the base image to an ubuntu one and use apt

I don't have a preference here as long as we keep it working. Personally, I prefer smaller images (like alpine), Ubuntu has always seemed full-blown to me.

jsoriano: > I prefer smaller images (like alpine), Ubuntu has always seemed full-blown to me.

Regarding size, I think that the base image is the lesser of the problems once we include the gcloud sdk :slightly_smiling_face: 
```
eb91706519b5   5 minutes ago   /bin/sh -c curl "https://dl.google.com/dl/cl…   1.05GB    
```
Apart of that, the ubuntu base image is quite small (72MB, ~100MB with python 3), and you probably already have it because is the base image of current versions of the stack.

jsoriano: Looks more readable, and seems to take less space (tough maybe more components need to be installed with the optional packages).
```
FROM ubuntu:20.04

RUN apt-get -q update && apt-get install -yq curl apt-transport-https ca-certificates gnupg && apt-get clean

ADD google-cloud-sdk.list /etc/apt/sources.list.d/google-cloud-sdk.list
RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -

ENV GCLOUD_SDK_VERSION 369.0.0-0

RUN apt-get -q update && \
    apt-get -yq install google-cloud-sdk=${GCLOUD_SDK_VERSION} && \
    apt-get clean
```
```
da5b98a5dec3   About a minute ago   /bin/sh -c apt-get -q update &&     apt-get …   620MB     
```

mtojek:
```
eb91706519b5   5 minutes ago   /bin/sh -c curl "https://dl.google.com/dl/cl…   1.05GB    
```

Yeah, that's something I wanted originally to omit, but it looks like it may not be so easy unless we modify the service deployer image to pull it in runtime?

endorama: @mtojek I'm already using the versioned archive installation. `install.sh` triggers a python script that takes care of installing other components so I don't think can be safely skipped.

> Could you please try again without calling install.sh at all? I suppose you need to unpack the archive in the correct place instead of /tmp.

I tried it but the result is not good. This is what happens when running `gcloud` in the docker container without running install:

```
/workspace # gcloud
ERROR: (gcloud) Command name argument expected.
ERROR: gcloud failed to load (gcloud.interactive): Problem loading gcloud.interactive: cannot import name 'Mapping' from 'collections' (/usr/local/lib/python3.10/collections/__init__.py).

This usually indicates corruption in your gcloud installation or problems with your Python interpreter.

[...]

> PS We will definitely need a package system test to verify if it works as expected.

This is in the works!

I'm moving forward to using Ubuntu as base image.

Comment (endorama) [reply internal/install/_static/Dockerfile.terraform_deployer]: I updated the image to use Ubuntu.

These are the dimensions for the two images:
```
ubuntu        978MB
alpine        594MB
```

@jsoriano the gcloud SDK when installed from a versioned archive is a lot less that 1.05GB, as is possible to specify which components to install thus reducing the overall size. 

I'm happy with both, let me know what do you think of the current change :)

Comment (mtojek) [reply internal/install/_static/Dockerfile.terraform_deployer]: Much better now and managed in a normal way :)

I will drop few nit-picks.

File: internal/install/_static/Dockerfile.terraform_deployer [hunk 4]
```diff
@@ -1,17 +1,14 @@
...
+RUN apt-get -q update && apt-get install -yq curl apt-transport-https ca-certificates gnupg && apt-get clean
 
-ENV GCLOUD_SDK_VERSION 369.0.0
-ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
-RUN curl "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-$GCLOUD_SDK_VERSION-linux-x86_64.tar.gz" > /tmp/google-cloud-sdk.tar.gz \
-  && mkdir -p /usr/local/gcloud \
-  && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
-  && /usr/local/gcloud/google-cloud-sdk/install.sh -q --override-components="bq" \
-  && rm /tmp/google-cloud-sdk.tar.gz
+ENV GCLOUD_SDK_VERSION 370.0.0-0
```
mtojek, line 7: Please move this to the top of the file (after FROM)

internal/install/_static/Dockerfile.terraform_deployer
```diff
@@ -1,17 +1,14 @@
...
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
```
mtojek, line 8: I think you mixed all commands a bit.

1. Install GPG key and cloud-sdk repo
2. apt-get update -yq && apt-get install -yq curl apt-transport-https.... - google-cloud-sdk=${GCLOUD_SDK_VERSION} &&  apt-get clean

mtojek (commented):
I think we have an agreement. It's more like a final polishing of the PR.

internal/install/_static/Dockerfile.terraform_deployer
endorama: `curl apt-transport-https gnupg` are required to add the deb repo and install the repo key, so two steps are required unfortunately. 
Would you prefer to split GPG key and repo addition from the actual package installation?

mtojek: You can try something like this:

```
apt-get update -yq && \ 
apt-get install apt-transport-https gnupg && \
deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && \
apt-get update -yq && \
apt-get install -yq google-cloud-sdk=${GCLOUD_SDK_VERSION} && \
apt-get clean
```

The aim is to consider this as an atomic piece to prevent any APT repo update from breaking the container build.

internal/install/_static/Dockerfile.terraform_deployer
endorama: done

internal/install/_static/Dockerfile.terraform_deployer
endorama: I updated it with the suggestion. Just consider that if in the future another repo will be required I think would be best to extract the dependency installation so there is no implicit dependency.

mtojek (commented):
Something is fishy with the terraform as it [doesn't start](https://beats-ci.elastic.co/blue/rest/organizations/jenkins/pipelines/Ingest-manager/pipelines/elastic-package/branches/PR-665/runs/5/log/?start=0). Could you please investigate it?

internal/install/_static/Dockerfile.terraform_deployer
```diff
@@ -1,17 +1,14 @@
 FROM hashicorp/terraform:light as terraform
```
jsoriano, line 1: Now that we have apt maybe we can also install terraform using it :) https://www.terraform.io/downloads

And we could also pin the version.

This is unrelated to this PR in any case.

mtojek: Actually it may help if this is a PATH issue or library issue :)

mtojek:
@endorama Please pull the latest main branch.

internal/install/_static/Dockerfile.terraform_deployer
endorama: @mtojek I extracted it again, as Hashicorp has it's own repo that has the same requirements. It happened sooner than expected :D

internal/install/_static/Dockerfile.terraform_deployer
endorama: Added installation through deb thanks for the suggestion!

internal/install/_static/Dockerfile.terraform_deployer
```diff
@@ -1,23 +1,24 @@
...
-FROM python:3-alpine
+RUN apt-get -qq update \
+  && apt-get install -yq curl apt-transport-https ca-certificates gnupg
 
-# required by gcloud SDK
-RUN apk add --no-cache git openssh curl
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
+  && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - \
+  && apt-get update -qq \
+  && apt-get install google-cloud-sdk=${GCLOUD_SDK_VERSION} -yq
```
mtojek, line 11: @endorama did you try it locally? it takes ages on my setup unless there is something wrong with it.

endorama: Takes around 1 minute on my machine. :thinking: It may be something related to Docker Machine?

mtojek: Confirmed, the problem with my machine. It took 4 minutes to load it.

mtojek (commented):
I fetch this code locally, run it, and checked logs for the terraform service deployer:

```
$ docker logs 49fee7edacf7
/run.sh: 3: set: Illegal option -o pipefail
```

You have to update the `terraform_deployer_run.sh` to `/bin/bash`. `sh` most likely points to a different shell.

internal/compose/compose.go
```diff
@@ -275,49 +275,57 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
+			logger.Debugf("Wait for healthy containers: %s", strings.Join(containerIDs, ","))
+			descriptions, err := docker.InspectContainers(containerIDs...)
+			if err != nil {
+				return err
 			}
-
-			// Service is up and running and it's healthy
-			if containerDescription.State.Status == "running" && containerDescription.State.Health.Status == "healthy" {
-				continue
+			for _, containerDescription := range descriptions {
```
mtojek, line 298: Let's extract this change around ticker to a separate PR and focus here on improving the terraform Docker image.

mtojek (changes requested):

internal/compose/compose.go
endorama: Done https://github.com/elastic/elastic-package/pull/687

mtojek: So, could you please remove it from this PR?

mtojek (changes requested):

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -1,7 +1,9 @@
-#!sh
+#!/usr/bin/env bash
 
 set -euxo pipefail
```
mtojek, line 3: Let's drop `-x` to have a clean nice output in container logs. This script doesn't contain tangled logic which may cause issues.

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -1,7 +1,9 @@
-#!sh
+#!/usr/bin/env bash
```
mtojek, line 1: The reason why the CI is failing is because of the Dockerfile ENTRYPOINT set to `sh`: [here](https://github.com/elastic/elastic-package/blob/main/internal/install/_static/Dockerfile.terraform_deployer#L24). 

Let's quickly refactor it to `/bin/bash` instead and verify.

mtojek (changes requested):

internal/install/_static/terraform_deployer_run.sh
```diff
@@ -1,6 +1,8 @@
...
 
-set -euxo pipefail
+set -euo pipefail
+
+echo "BASH_VERSION: $BASH_VERSION"
```
mtojek, line 5: Could you please remove this line?

mtojek (approved):
LGTM

jsoriano (approved):

endorama:
🎉🎉 thank you!

PR #670, opened 2022-01-26
Export installed objects for a given package

### LINKED ISSUE #600: Expand `elastic-package export` to include Elasticsearch objects
Currently, the `export` command only supports Kibana Dashboards.  Our team would like support for `elastic-package export` to include Elasticsearch assets from the integrations, specifically:

* ILM Policies*
* Composable Templates
* Index Template Settings/Mappings Components
* Ingest Pipelines
* Transforms

*This is not currently in the [package spec](https://github.com/elastic/package-spec#supported-assets), but if in actuality it is or becomes available in a package, we would want to grab it.

Also: a nice-to-have as a result of expanding the types available to `export` would be the specification of a target dir for the output artifacts.

jsoriano:
Add the `elastic-package dump installed-objects` to dump the objects installed for a package in Elasticsearch.

This command is intended to be used as an exploratory tool, to be able to analyze the resources associated to a given package.

For testing, a helper is added that starts a client for a mocked server with recorded responses.

To try it:
```
make build
./elastic-package stack up
eval "$(./elastic-package stack shellinit)"
./elastic-package dump installed-objects --package system
```

Fixes #600


jsoriano:
So far, it exports the following objects when run for the system package:
```
output/
output/ilm_policies
output/ilm_policies/metrics.json
output/component_templates
output/component_templates/logs-system.system@custom.json
output/component_templates/metrics-system.process.summary@settings.json
output/component_templates/.fleet_component_template-1.json
output/component_templates/metrics-system.load@custom.json
output/component_templates/metrics-system.network@custom.json
output/component_templates/metrics-system.fsstat@settings.json
output/component_templates/logs-system.syslog@custom.json
output/component_templates/metrics-system.process@custom.json
output/component_templates/logs-system.security@settings.json
output/component_templates/metrics-system.filesystem@settings.json
output/component_templates/logs-system.system@settings.json
output/component_templates/metrics-system.network@settings.json
output/component_templates/metrics-system.uptime@settings.json
output/component_templates/metrics-system.socket_summary@custom.json
output/component_templates/metrics-system.process@settings.json
output/component_templates/metrics-system.diskio@custom.json
output/component_templates/logs-system.application@custom.json
output/component_templates/logs-system.auth@settings.json
output/component_templates/metrics-system.load@settings.json
output/component_templates/metrics-system.memory@settings.json
output/component_templates/metrics-system.filesystem@custom.json
output/component_templates/logs-system.auth@custom.json
output/component_templates/metrics-system.uptime@custom.json
output/component_templates/metrics-system.fsstat@custom.json
output/component_templates/logs-system.application@settings.json
output/component_templates/metrics-system.diskio@settings.json
output/component_templates/metrics-system.socket_summary@settings.json
output/component_templates/metrics-system.cpu@settings.json
output/component_templates/metrics-system.memory@custom.json
output/component_templates/logs-system.syslog@settings.json
output/component_templates/metrics-system.process.summary@custom.json
output/component_templates/metrics-system.cpu@custom.json
output/component_templates/logs-system.security@custom.json
output/index_templates
output/index_templates/metrics-system.network.json
output/index_templates/metrics-system.fsstat.json
output/index_templates/metrics-system.memory.json
output/index_templates/logs-system.system.json
output/index_templates/metrics-system.process.summary.json
output/index_templates/metrics-system.diskio.json
output/index_templates/logs-system.auth.json
output/index_templates/logs-system.syslog.json
output/index_templates/logs-system.security.json
output/index_templates/metrics-system.filesystem.json
output/index_templates/metrics-system.process.json
output/index_templates/metrics-system.socket_summary.json
output/index_templates/metrics-system.load.json
output/index_templates/logs-system.application.json
output/index_templates/metrics-system.cpu.json
output/index_templates/metrics-system.uptime.json
output/ingest_pipelines
output/ingest_pipelines/logs-system.auth-1.8.0.json
output/ingest_pipelines/logs-system.security-1.8.0.json
output/ingest_pipelines/logs-system.application-1.8.0.json
output/ingest_pipelines/logs-system.syslog-1.8.0.json
output/ingest_pipelines/logs-system.system-1.8.0.json
```

jsoriano:
Getting the object starting from the data streams is not reliable, they may not exist if no data has been ingested. Refactored to start discovering from the index templates. For the apache package these objects are exported:
```
output/component_templates/.fleet_component_template-1.json
output/component_templates/logs-apache.access@custom.json
output/component_templates/logs-apache.access@settings.json
output/component_templates/logs-apache.error@custom.json
output/component_templates/logs-apache.error@settings.json
output/component_templates/metrics-apache.status@custom.json
output/component_templates/metrics-apache.status@settings.json
output/ilm_policies/logs.json
output/ilm_policies/metrics.json
output/index_templates/logs-apache.access.json
output/index_templates/logs-apache.error.json
output/index_templates/metrics-apache.status.json
output/ingest_pipelines/logs-apache.access-1.3.4.json
output/ingest_pipelines/logs-apache.error-1.3.4.json
```

jsoriano:
Opening for review.

@DJRickyB @tomcallahan it'd be nice if you could give it a try.

You can find sample files exported by this new subcommand in https://github.com/elastic/elastic-package/tree/bbbb70a54495b8b32f42846797733ccdc74ed54d/internal/export/testdata/apache-export-all

cmd/export.go
```diff
@@ -33,12 +38,25 @@ func setupExportCommand() *cobraext.Command {
...
 
+	exportInstalledObjectsCmd := &cobra.Command{
+		Use:   "installed-objects",
+		Short: "Export installed Elasticsearch objects",
+		Long:  exportInstalledObjectsLongDescription,
+		RunE:  exportInstalledObjectsCmd,
+	}
+	exportInstalledObjectsCmd.Flags().StringP(cobraext.ExportPackageFlagName, "p", "", cobraext.ExportPackageFlagDescription)
+	exportInstalledObjectsCmd.MarkFlagRequired(cobraext.ExportPackageFlagName)
+	exportInstalledObjectsCmd.Flags().StringP(cobraext.ExportOutputFlagName, "o", "output", cobraext.ExportOutputFlagDescription)
```
mtojek, line 49: I'm wondering if we can map it on the existing package structure on disk instead of `output` directory, similarly to other `export`'s subcommand.

mtojek (commented):
First review round, mostly focused on API.

cmd/export.go
```diff
@@ -33,12 +38,25 @@ func setupExportCommand() *cobraext.Command {
 	exportDashboardCmd.Flags().StringSliceP(cobraext.DashboardIDsFlagName, "d", nil, cobraext.DashboardIDsFlagDescription)
 	exportDashboardCmd.Flags().Bool(cobraext.TLSSkipVerifyFlagName, false, cobraext.TLSSkipVerifyFlagDescription)
 
+	exportInstalledObjectsCmd := &cobra.Command{
+		Use:   "installed-objects",
+		Short: "Export installed Elasticsearch objects",
```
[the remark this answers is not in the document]
jsoriano: The command only exports the installed Elasticsearch objects now, as they are the relevant ones for performance testing.

We could consider at least index patterns and dashboards as things to export too. I haven't thought a lot about what would we do if we also want to export them in the same fashion. One option is to just add it here and rephrase this. On the other hand we already have a command to export dashboards, and maybe we could have another one to export index patterns, or add a flag to export dashboards to export the used pattern too.

cmd/export.go
```diff
@@ -33,12 +38,25 @@ func setupExportCommand() *cobraext.Command {
...
+		Long:  exportInstalledObjectsLongDescription,
+		RunE:  exportInstalledObjectsCmd,
+	}
+	exportInstalledObjectsCmd.Flags().StringP(cobraext.ExportPackageFlagName, "p", "", cobraext.ExportPackageFlagDescription)
+	exportInstalledObjectsCmd.MarkFlagRequired(cobraext.ExportPackageFlagName)
```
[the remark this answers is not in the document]
jsoriano: With the approach here you can export the objects without needing the source of the package, I think this will be useful for the current use case, and it actually doesn't need anything about the current package context.

If we want to make all `export` subcommands to be consistent on this I think I would prefer to add both options to both subcommands (i.e. both work with the package context and with the package flag) :)

cmd/export.go
```diff
@@ -117,3 +135,36 @@ func promptDashboardIDs(kibanaClient *kibana.Client) ([]string, error) {
...
+
+	outputPath, err := cmd.Flags().GetString(cobraext.ExportOutputFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.ExportOutputFlagName)
+	}
+
+	tlsSkipVerify, _ := cmd.Flags().GetBool(cobraext.TLSSkipVerifyFlagName)
+
+	clientOptions := elasticsearch.DefaultClientOptionsFromEnv()
+	clientOptions.SkipTLSVerify = tlsSkipVerify
```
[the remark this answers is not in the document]
jsoriano: I think that in Go I prefer to avoid method chaining for initialization, but I will give it a try here as it seems we are using this approach in other initializations.

cmd/export.go
jsoriano: Unlike dashboards, these objects are not going to be part of a package, so there would be little reason to do this. This, along with other differences may be a reason to don't put this new subcommand under `export`, but not sure where to put it then.

cmd/export.go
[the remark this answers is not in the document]
mtojek: > If we want to make all export subcommands to be consistent on this I think I would prefer to add both options to both subcommands (i.e. both work with the package context and with the package flag) :)

Yes, I was looking at low hanging fruits :) I will leave you to make the decision.

cmd/export.go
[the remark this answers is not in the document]
mtojek: Same story, if this is a low-hanging fruit, then please consider it.

cmd/export.go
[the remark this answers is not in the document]
mtojek: I guess it's a sort of a personal preference. I think it's more important to keep it consistent, but you can consider this comment as a nit-pick

internal/export/installedobjects.go
```diff
@@ -0,0 +1,280 @@
...
+// NewInstalledObjectsExporter creates an InstalledObjectsExporter for a given package.
+func NewInstalledObjectsExporter(client *elasticsearch.API, packageName string) *InstalledObjectsExporter {
+	return &InstalledObjectsExporter{
+		packageName: packageName,
+		client:      client,
+	}
+}
+
+// ExportAll discovers and exports all known resources as files in the given directory.
+func (e *InstalledObjectsExporter) ExportAll(ctx context.Context, dir string) (count int, err error) {
```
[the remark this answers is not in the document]
mtojek: We know about:
[index_patterns](https://github.com/elastic/package-spec/blob/main/versions/1/kibana/spec.yml#L59)
[ingest_pipelines](https://github.com/elastic/package-spec/blob/main/versions/1/elasticsearch/spec.yml#L15)
[ilm_policies](https://github.com/elastic/package-spec/blob/main/versions/1/data_stream/spec.yml#L39)

On the other hand, the ingest pipeline can appear in different places (`elasticsearch` or `data_stream/elasticsearch`) and it might be hard to figure out where to dump it.

We don't know know what to do with  "component-templates". Hm...

I'm afraid that I can't give a simple answer on what to do. We could introduce a new command like `elastic-package dump installed-objects` and it will create a new directory with dumped resources. This way we can leave export and import to be "organized" load/store actions, which preserve file structure.

mtojek (approved):
LGTM implementation looks good, not sure if we arguments to discuss the API.

internal/export/installedobjects.go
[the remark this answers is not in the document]
jsoriano: Interesting, I didn't know the spec also has supports for custom ILM policies and index patterns, though I see this is barely used. For ingest pipelines we would probably want to export them as YAML to make them useful for package developers.

Adding a `dump` command makes sense to me. So we keep `export` for commands that export things in formats more suited to be included in packages. I will do that.

cmd/export.go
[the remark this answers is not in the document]
jsoriano: I have added a test so this continues being a low hanging fruit in the future, but I am not taking it by now :)

cmd/export.go
jsoriano: Moving the subcommand to a new `dump` command.

cmd/export.go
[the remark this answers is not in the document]
jsoriano: Not such a low hanging fruit as the source of data would be a different API, and not needed at the moment, I would leave it for a future improvement.

jsoriano:
@mtojek moved to a new `dump` command, let me know what you think.

I would be tempted to make this the only action of this `dump` subcommand, but I wonder if we may have more things to dump in the future, so I would play safe here and leave it as a subcommand under `dump`.

internal/common/helpers.go
```diff
@@ -12,3 +12,13 @@ func TrimStringSlice(slice []string) {
 		slice[iterator] = strings.TrimSpace(item)
 	}
 }
+
+// StringSliceContains checks if the slice contains the given string.
+func StringSliceContains(slice []string, s string) bool {
```
mtojek, line 17: nit: I'm wondering to how many places in the source code it can applied :)

cmd/dump.go
```diff
@@ -0,0 +1,78 @@
...
+		return errors.Wrap(err, "failed to initialize Elasticsearch client")
+	}
+
+	dumper := dump.NewInstalledObjectsDumper(client.API, packageName)
+	n, err := dumper.DumpAll(cmd.Context(), outputPath)
+	if err != nil {
+		return errors.Wrap(err, "dump failed")
+	}
+	if n == 0 {
+		cmd.Printf("No objects were dumped for package %s, is it installed?", packageName)
```
mtojek, line 73: nit: do you think we can check if the package is installed apriori? I hope there is a fleet API to make sure it is.

cmd/dump.go
```diff
@@ -0,0 +1,78 @@
...
+Use this command as a exploratory tool to dump objects as they are installed by Fleet when installing a package. Dumped objects are stored in files as they are returned by APIs of the stack, without any processing.`
+
+func setupDumpCommand() *cobraext.Command {
+	dumpInstalledObjectsCmd := &cobra.Command{
+		Use:   "installed-objects",
+		Short: "Dump objects installed in the stack",
+		Long:  dumpInstalledObjectsLongDescription,
+		RunE:  dumpInstalledObjectsCmd,
+	}
+	dumpInstalledObjectsCmd.Flags().StringP(cobraext.DumpPackageFlagName, "p", "", cobraext.DumpPackageFlagDescription)
```
mtojek, line 29: Shouldn't `-p` and `-o` be a `command` level instead of `subcommand`?

mtojek, line 29: I hope it isn't possible to install multiple package versions at once :)

cmd/dump.go
```diff
@@ -0,0 +1,78 @@
...
+	dumpInstalledObjectsCmd.MarkFlagRequired(cobraext.DumpPackageFlagName)
+	dumpInstalledObjectsCmd.Flags().StringP(cobraext.DumpOutputFlagName, "o", "output", cobraext.DumpOutputFlagDescription)
```
mtojek, line 31: Let's discuss possible `dump` use cases:

```
elastic-package dump installed-objects -p hello-world 
elastic-package dump assets -p hello-world
elastic-package dump assets -p hello-world 
elastic-package dump indices -p hello-world
elastic-package dump sample-event -p hello-world
```

WDYT about storing data in a structured way, for example `output`-dir can be replaced with `package-dump`, which may contain `installed-objects`, `indices`, etc.

internal/common/helpers.go
jsoriano: We can look for them in a follow up, and perhaps if we wait to https://tip.golang.org/doc/go1.18#generics we can make it generic and apply it even more extensively :)

cmd/dump.go
jsoriano: Maybe, but this way this tool can also discover letfovers after uninstalling a package :) Also adding a check against the fleet API adds an additional dependency. As the command is now, it only needs access to Elasticsearch.

cmd/dump.go
jsoriano: Well, if some day this is possible, I guess that the worst thing that can happen with current implementation is that the objects of all installed versions will be collected, what I would find fine. We can add an optional version filter if this day arrives.

jsoriano:
I have found a couple of places where there can be nested pipelines, adding some logic to collect them too...

cmd/dump.go
mtojek: > Maybe, but this way this tool can also discover letfovers after uninstalling a package

It's a discussion whether it's a bug or feature :) We can keep it as is for now.

mtojek (approved):
LGTM

jsoriano:
@DJRickyB @tomcallahan did you have the chance of trying this new command to dump ES objects related to packages?

DJRickyB:
Hello! I'll get into this today

DJRickyB:
This works great, thank you very much. Interesting, the departure from the existing `export` command to this `dump` command, given the similarity of function but as we're only interested in the ES objects anyway, this is great

jsoriano:
> This works great, thank you very much. Interesting, the departure from the existing `export` command to this `dump` command, given the similarity of function but as we're only interested in the ES objects anyway, this is great

@DJRickyB thanks for trying it!

Yep, both commands are similar, but we decided to leave the `export` command as a tool for package developers to export resources in a way that can be included directly in packages, for example the `export dashboards` subcommand does some transformations and stores the dashboards and its elements in the proper paths inside the current package. These commands are expected to be run in the working directory of a package.

On the other hand `dump` commands will be used to dump resources as they are, without any transformation, or without expectations on the structure of the directory where they are dumped.

DJRickyB:
Thank you for the context, makes total sense

PR #673, opened 2022-01-27
Fix: use correct index template name for hidden data streams

aleksmaus:
Fix elastic-package where it was not correctly deriving the index template name for the ```hidden``` datastreams 
and failing in CI:

For example PR https://github.com/elastic/integrations/pull/2563

resulting in error:
```
asset test: index_template logs-osquery_manager.actions is loaded – osquery_manager.actions
no error details
Expand to view the stacktrace
 could not find expected asset: could not find index_template asset "logs-osquery_manager.actions". Assets loaded:
- .logs-osquery_manager.action_results (type: index_template)
- .logs-osquery_manager.action_results@mappings (type: component_template)
- .logs-osquery_manager.action_results@settings (type: component_template)
- .logs-osquery_manager.action_results@custom (type: component_template)
- .logs-osquery_manager.actions (type: index_template)
- .logs-osquery_manager.actions@mappings (type: component_template)
- .logs-osquery_manager.actions@settings (type: component_template)
- .logs-osquery_manager.actions@custom (type: component_template)
- logs-osquery_manager.result (type: index_template)
- logs-osquery_manager.result@mappings (type: component_template)
- logs-osquery_manager.result@settings (type: component_template)
- logs-osquery_manager.result@custom (type: component_template)
```

Screenshot after this change, showing it passes the tests now:

Screen Shot 2022-01-27 at 2 29 29 PM

### DIFF
--- a/internal/packages/packages.go
+++ b/internal/packages/packages.go
@@ -242,6 +242,8 @@ func (dsm *DataStreamManifest) GetPipelineNameOrDefault() string {
 
 // IndexTemplateName returns the name of the Elasticsearch index template that would be installed
 // for this data stream.
+// The template name starts with dot "." if the datastream is hidden which is consistent with kibana implementation
+// https://github.com/elastic/kibana/blob/3955d0dc819fec03f68cd1d931f64da8472e34b2/x-pack/plugins/fleet/server/services/epm/elasticsearch/index.ts#L14
 func (dsm *DataStreamManifest) IndexTemplateName(pkgName string) string {
 	if dsm.Dataset == "" {
 		return dsm.IndexTemplateNamePrefix() + fmt.Sprintf("%s-%s.%s", dsm.Type, pkgName, dsm.Name)
--- a/internal/packages/packages.go
+++ b/internal/packages/packages.go
@@ -246,13 +246,13 @@ func (dsm *DataStreamManifest) GetPipelineNameOrDefault() string {
 // https://github.com/elastic/kibana/blob/3955d0dc819fec03f68cd1d931f64da8472e34b2/x-pack/plugins/fleet/server/services/epm/elasticsearch/index.ts#L14
 func (dsm *DataStreamManifest) IndexTemplateName(pkgName string) string {
 	if dsm.Dataset == "" {
-		return dsm.IndexTemplateNamePrefix() + fmt.Sprintf("%s-%s.%s", dsm.Type, pkgName, dsm.Name)
+		return fmt.Sprintf("%s%s-%s.%s", dsm.indexTemplateNamePrefix(), dsm.Type, pkgName, dsm.Name)
 	}
 
-	return dsm.IndexTemplateNamePrefix() + fmt.Sprintf("%s-%s", dsm.Type, dsm.Dataset)
+	return fmt.Sprintf("%s%s-%s", dsm.indexTemplateNamePrefix(), dsm.Type, dsm.Dataset)
 }
 
-func (dsm *DataStreamManifest) IndexTemplateNamePrefix() string {
+func (dsm *DataStreamManifest) indexTemplateNamePrefix() string {
 	if dsm.Hidden {
 		return "."
 	}
--- a/internal/packages/packages.go
+++ b/internal/packages/packages.go
@@ -115,6 +115,7 @@ type DataStreamManifest struct {
 	Title         string `config:"title" json:"title" yaml:"title"`
 	Type          string `config:"type" json:"type" yaml:"type"`
 	Dataset       string `config:"dataset" json:"dataset" yaml:"dataset"`
+	Hidden        bool   `config:"hidden" json:"hidden" yaml:"hidden"`
 	Release       string `config:"release" json:"release" yaml:"release"`
 	Elasticsearch *struct {
 		IngestPipeline *struct {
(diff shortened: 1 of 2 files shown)


internal/packages/packages.go
```diff
@@ -243,10 +244,17 @@ func (dsm *DataStreamManifest) GetPipelineNameOrDefault() string {
 // for this data stream.
 func (dsm *DataStreamManifest) IndexTemplateName(pkgName string) string {
```
mtojek, line 245: As this isn't a standard, straightforward logic, would you mind adding some unit tests?

internal/packages/packages.go
```diff
@@ -243,10 +244,17 @@ func (dsm *DataStreamManifest) GetPipelineNameOrDefault() string {
...
 	if dsm.Dataset == "" {
-		return fmt.Sprintf("%s-%s.%s", dsm.Type, pkgName, dsm.Name)
+		return dsm.IndexTemplateNamePrefix() + fmt.Sprintf("%s-%s.%s", dsm.Type, pkgName, dsm.Name)
```
mtojek, line 247: What would be the value for a hidden data stream without dataset defined? `.logs-aPackage.aDataStream`?

aleksmaus: Not 100% sure what it suppose to be, but currently the kibana behavour is different from the elatic-package validation. Kibana creates templates with "." dot prefix for the package with hidden datastreams.

internal/packages/packages.go
aleksmaus: Sure will do

aleksmaus: Done

internal/packages/packages.go
mtojek: > Not 100% sure what it suppose to be,

So do I :) I'm not if it's a bug or works as intended. @ruflin @joshdover Could you please confirm the index template name ("hidden" case)?

joshdover: So this pre-dates my time on Fleet, but this definitely appears to be intentional. We have this function in Kibana which is used for all asset naming:

https://github.com/elastic/kibana/blob/0ee514b1fe044cdaca253f8f14988fc17635fed3/x-pack/plugins/fleet/server/services/epm/elasticsearch/index.ts#L10-L18

It appears it was added as part of https://github.com/elastic/kibana/pull/86277 over a year ago, though I don't see any linked package-spec changes. Maybe @kevinlog can provide more context.

mtojek: Thanks for diving into code, Josh! @aleksmaus I think you can adjust function implementation to look similar (`baseType` vs `.baseType`) and please hard-link to Kibana in the comment.

mtojek (commented):
Left one comment

internal/packages/packages.go
aleksmaus: Thanks for looking into this. Will update and add comment. Endpoint didn't have issues passing this validation in CI because it lives in a separate repo, maybe skipping this test.

aleksmaus: Took a look at kibana code, it looks consistent with this change. Added a comment. Let me know.

internal/packages/packages.go
```diff
@@ -241,12 +242,21 @@ func (dsm *DataStreamManifest) GetPipelineNameOrDefault() string {
...
 	}
 
-	return fmt.Sprintf("%s-%s", dsm.Type, dsm.Dataset)
+	return dsm.IndexTemplateNamePrefix() + fmt.Sprintf("%s-%s", dsm.Type, dsm.Dataset)
+}
+
+func (dsm *DataStreamManifest) IndexTemplateNamePrefix() string {
```
mtojek, line 255: nit: does it have to be exposed?

internal/packages/packages.go
```diff
@@ -241,12 +242,21 @@ func (dsm *DataStreamManifest) GetPipelineNameOrDefault() string {
...
+// The template name starts with dot "." if the datastream is hidden which is consistent with kibana implementation
+// https://github.com/elastic/kibana/blob/3955d0dc819fec03f68cd1d931f64da8472e34b2/x-pack/plugins/fleet/server/services/epm/elasticsearch/index.ts#L14
 func (dsm *DataStreamManifest) IndexTemplateName(pkgName string) string {
 	if dsm.Dataset == "" {
-		return fmt.Sprintf("%s-%s.%s", dsm.Type, pkgName, dsm.Name)
+		return dsm.IndexTemplateNamePrefix() + fmt.Sprintf("%s-%s.%s", dsm.Type, pkgName, dsm.Name)
 	}
 
-	return fmt.Sprintf("%s-%s", dsm.Type, dsm.Dataset)
+	return dsm.IndexTemplateNamePrefix() + fmt.Sprintf("%s-%s", dsm.Type, dsm.Dataset)
```
mtojek, line 252: I think you can refactor it into:`fmt.Sprintf("%s%s-%s", dsm.optionalIndexTemplateNamePrefix(), dsm.Type, dsm.Dataset)` to prevent concatenation and fmt.Sprintf separately.

or maybe:

`dsm.withIndexTemplateNamePrefix(fmt.Sprintf(...))`

internal/packages/packages.go
aleksmaus: valid point. doesn't have to.

mtojek (approved):

PR #674, opened 2022-01-28
README: suggest `go install` instead of `go get`

afharo:
When following the instructions in the README, I noticed this deprecation message:

```
go get: installing executables with 'go get' in module mode is deprecated.
        Use 'go install pkg@version' instead.
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.
```

This PR suggests `go install` instead.


mtojek (approved):

efd6:
It may be worth noting that if `go install github.com/elastic/elastic-package` is used without the `@latest` within the integrations repo, that the correct version will be installed for its use.

PR #684, opened 2022-02-03
Try with kibana

mtojek:


mtojek:
Unfortunately it doesn't have access to staging, will try to boot it manually.

PR #687, opened 2022-02-04
Improve Docker Compose container execution

endorama:
This PR has been extracted from https://github.com/elastic/elastic-package/pull/665.

When `elastic-package` runs docker containers it relies on [`WaitForHealthy()`](https://github.com/endorama/elastic-package/blob/89161065b946c1531132f0e20c03b4f48210f0b7/internal/compose/compose.go#L267) to ensure expected containers are running.

This functions has a missing feature: timing out if the condition is not met.

This PR adds this feature, adding a 10 minute timeout after which an error is returned.

To prevent waiting time waiting for a condition that can never be met a check is added for unrecoverable termination condition (container is reported as exited and with an exit code greater than 0) to immediately exit with an error, skipping waiting until timeout expiration.

### DIFF
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -23,11 +23,12 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
-// waitForHealthyTimeout is the maximum duration for WaitForHealthy().
-var waitForHealthyTimeout = time.Duration(10 * time.Minute)
-
-// waitForHealthyInterval is the check interval for WaitForHealthy().
-const waitForHealthyInterval = time.Duration(1 * time.Second)
+const (
+	// waitForHealthyTimeout is the maximum duration for WaitForHealthy().
+	waitForHealthyTimeout = 10 * time.Minute
+	// waitForHealthyInterval is the check interval for WaitForHealthy().
+	waitForHealthyInterval = 1 * time.Second
+)
 
 // Project represents a Docker Compose project.
 type Project struct {
@@ -285,7 +286,11 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 	timeout := startTime.Add(waitForHealthyTimeout)
 
 	containerIDs := strings.Split(strings.TrimSpace(b.String()), "\n")
-	for time.Now().Before(timeout) {
+	for {
+		if time.Now().After(timeout) {
+			return errors.New("timeout waiting for healthy container")
+		}
+
 		if signal.SIGINT() {
 			return errors.New("SIGINT: cancel waiting for policy assigned")
 		}
@@ -298,6 +303,7 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 		if err != nil {
 			return err
 		}
+
 		for _, containerDescription := range descriptions {
 			logger.Debugf("Container status: %s", containerDescription.String())
 
@@ -332,10 +338,6 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 
 		// NOTE: using sleep does not guarantee interval but it's ok for this use case
 		time.Sleep(waitForHealthyInterval)
-	}
-
-	if time.Now().After(timeout) {
-		return errors.New("timeout waiting for healthy container")
 	}
 
 	return nil
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -23,6 +23,8 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
+const waitForHealthyTimeout = 10 * time.Minute
+
 // Project represents a Docker Compose project.
 type Project struct {
 	name             string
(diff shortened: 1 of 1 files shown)


internal/compose/compose.go
```diff
@@ -275,49 +275,57 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
+				if containerDescription.State.Status == "running" && containerDescription.State.Health.Status == "healthy" {
+					continue
+				}
+
+				// Container started and finished with exit code 0
+				if containerDescription.State.Status == "exited" && containerDescription.State.ExitCode == 0 {
+					continue
+				}
+
+				if containerDescription.State.Status == "exited" && containerDescription.State.ExitCode > 0 {
```
endorama, line 316: The same line have been added in https://github.com/elastic/elastic-package/pull/688/commits/8563a1c8a7fec6375d82520d423ce33abf331b8f#diff-8019656c843609e3cb15ca57ab96dc41fdadae8215d0ff22d9b989f795aef4a4 so this PR will need to be rebased on top of `main` once https://github.com/elastic/elastic-package/pull/688 is merged

endorama: removed

internal/compose/compose.go
```diff
@@ -275,54 +275,58 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 		return err
 	}
 
+	timeout := time.After(10 * time.Minute)
+	ticker := time.NewTicker(1 * time.Second)
```
jsoriano, line 279:
```suggestion
	ticker := time.NewTicker(1 * time.Second)
	defer ticker.Stop()
```

internal/compose/compose.go
```diff
@@ -275,54 +275,58 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
-
-			// Any different status is considered unhealthy
-			healthy = false
-		}
-
-		if healthy {
-			break
 		}
-
-		time.Sleep(time.Second)
```
jsoriano, line 323: :+1:

internal/compose/compose.go
```diff
@@ -275,54 +275,58 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 		return err
 	}
 
+	timeout := time.After(10 * time.Minute)
```
jsoriano, line 278: I was going to suggest to use `context.WithTimeout()`, so we can also interrupt calls to docker if they hang, but leveraging this context would require more changes in `internal/docker`.

internal/compose/compose.go
```diff
@@ -275,54 +275,58 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
-		logger.Debugf("Wait for healthy containers: %s", strings.Join(containerIDs, ","))
-		descriptions, err := docker.InspectContainers(containerIDs...)
-		if err != nil {
-			return err
-		}
+		select {
+		case <-timeout:
+			return errors.New("time out waiting for healthy container")
+		case <-ticker.C:
+			healthy := true
```
jsoriano, line 291: Nit. Do nothing and continue in this `case`, to have the main part of the loop in the upper level.
```suggestion
		case <-ticker.C:
		}
		healthy := true
```

internal/compose/compose.go
```diff
@@ -275,54 +275,58 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
+				return nil
 			}
-
-			// Container exited with code > 0
-			if containerDescription.State.Status == "exited" && containerDescription.State.ExitCode > 0 {
-				return fmt.Errorf("container (ID: %s) exited with code %d", containerDescription.ID, containerDescription.State.ExitCode)
-			}
-
-			// Any different status is considered unhealthy
-			healthy = false
```
jsoriano, line 316: :+1:

jsoriano (commented):
Ticker needs to be stopped, for the rest it LGTM :+1:

internal/compose/compose.go
mtojek, line 278: Please move this timeout to const at the top of the file and name it accordingly.

internal/compose/compose.go
```diff
@@ -275,54 +275,58 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
-		healthy := true
-
-		logger.Debugf("Wait for healthy containers: %s", strings.Join(containerIDs, ","))
-		descriptions, err := docker.InspectContainers(containerIDs...)
-		if err != nil {
-			return err
-		}
+		select {
+		case <-timeout:
+			return errors.New("time out waiting for healthy container")
```
mtojek, line 289: nit: timeout waiting...

internal/compose/compose.go
```diff
@@ -275,54 +275,58 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
 		}
 
-		healthy := true
-
-		logger.Debugf("Wait for healthy containers: %s", strings.Join(containerIDs, ","))
-		descriptions, err := docker.InspectContainers(containerIDs...)
-		if err != nil {
-			return err
-		}
+		select {
```
mtojek, line 287: Let me challenge this and consider another approach.

We already have "for" loop with a 1-sec delay per iteration. The function can store `time.Now()` and just compare it if the `startTime.Add(timeout)` is greater than `time.Now`. We don't need to add channels/select/ticker. WDYT?

endorama: I spent some time searching differences between `time.Sleep` and `time.Tick`, and the only thing I found is that a tick tries to guarantee the interval, while sleep is unaware of execution times. There seems to be no difference in performance for the 2 implementations.

In general I prefer the ticker approach as it's behaviour is explicit, but in this case I don't see particular benefits as guaranteeing the interval is not a concern.

I'll rework it with your approach.

jsoriano: I think I prefer the ticker and select approach too. It also works better than time arithmetic when you have multiple source of events, as may happen if/when we have contexts here.

jsoriano: But we can wait to have contexts here, not a strong opinion :slightly_smiling_face:

mtojek: I'm a bit confused here as adding a for-select block will introduce another nesting level and make the function hard to read. TBH I would follow the KISS rule, so if it's possible to solve with a single if-clause, I would go for it.

In general, we can consider refactoring of this area, as there is a similar logic in the [SIGINT](https://github.com/elastic/elastic-package/blob/8d9732d0c4364d28ba1c34f2ad7fe91dfc7d2785/internal/signal/sigint.go#L24) function. It would be great, If we can refactor all parts (SIGINT, timeout, waiter logic, contexts). You have my sword :)

endorama: By running a quick search I found there are 5 `for {` loops in the codebase. I think it would make a lot of sense to generalize the approach, so interval/timeout/signals are handled consistently. 

Are you ok to keep it for a future PR?  I already have a ticker-based `Waiter` struct implementation to use as a baseline, but I'm blocked on other work streams so I would prefer to push this refactor to the back burner for a bit.

internal/compose/compose.go
```diff
@@ -316,12 +321,19 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
 		if healthy {
 			break
 		}
 
-		time.Sleep(time.Second)
+		// NOTE: using sleep does not guarantee interval but it's ok for this use case
+		time.Sleep(interval)
 	}
+
+	if !healthy {
```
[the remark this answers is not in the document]
endorama: the bug is obvious, I misinterpreted the logic. `healthy` must be reinitialized at each loop, otherwise once it become `false` the logic does not change it's value anymore. I tried to use it to detect the timeout, but I stumbled upon the bug. The current implementation is clearer and uses times instead of a side effect on the `healthy` value.

internal/compose/compose.go
```diff
@@ -275,20 +281,23 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
...
 
+		// NOTE: healthy must be reinitialized at each iteration
 		healthy := true
 
 		logger.Debugf("Wait for healthy containers: %s", strings.Join(containerIDs, ","))
 		descriptions, err := docker.InspectContainers(containerIDs...)
 		if err != nil {
 			return err
 		}
-
```
mtojek, line 291: nit: let's leave this empty line

internal/compose/compose.go
```diff
@@ -23,6 +23,12 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
+// waitForHealthyTimeout is the maximum duration for WaitForHealthy().
+var waitForHealthyTimeout = time.Duration(10 * time.Minute)
+
+// waitForHealthyInterval is the check interval for WaitForHealthy().
+const waitForHealthyInterval = time.Duration(1 * time.Second)
```
mtojek, line 30: nit: I think you can make those consistent:

```
const (
   waitForHealthyTimeout = ...
   waitForHealthyInterval = ...
)
```

internal/compose/compose.go
```diff
@@ -23,6 +23,12 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
+// waitForHealthyTimeout is the maximum duration for WaitForHealthy().
+var waitForHealthyTimeout = time.Duration(10 * time.Minute)
```
jsoriano, line 27: Nit. No need to convert to `time.Duration`, a number multiplying a duration is already a duration.

internal/compose/compose.go
```diff
@@ -275,20 +281,23 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 		return err
 	}
 
+	startTime := time.Now()
+	timeout := startTime.Add(waitForHealthyTimeout)
+
 	containerIDs := strings.Split(strings.TrimSpace(b.String()), "\n")
-	for {
+	for time.Now().Before(timeout) {
```
jsoriano, line 288: Nit. Break the loop directly with a return so we don't have to remember to check for the timeout again after the loop? (And we are consistent on how we handle other end conditions as `SIGINT`).
```suggestion
	for {
		if time.Now().After(timeout) {
			return errors.New("timeout waiting for healthy container")
		}
```

jsoriano (commented):
I don't have anything against `for {`s I am ok with leaving them if they make sense :slightly_smiling_face:

mtojek (approved):
LGTM

jsoriano (approved):

PR #699, opened 2022-02-14
Re-add -x to Terraform deployer script

endorama:
In #665 we removed `-x` to debug an issue with script shebang, but the
intention wasn't to remove it entirely from the script.

This PR sets `-x` for the whole the script.


mtojek (approved):
LGTM

PR #701, opened 2022-02-15
Add GCP test package

endorama:
Adds `gcp` test package under `test/packages/parallel` with a basic test.

This PR splits from #662, where we deemed necessary proper testing to ensure correct functionality. To reduce complexity of #662 this PR only focus in having a single test running successfully in the CI.

Change:
- add gcp test package
- bump gcp test package version to 999.999.999
- add GCP account supports in Jenkins pipeline
- add gcp.compute datastream tests

@mtojek I don't know how to get GCP credentials usable by the CI, but I edited the Jenkinsfile to support GCP too.


.ci/Jenkinsfile
```diff
@@ -371,9 +371,15 @@ def withCloudTestEnv(Closure body) {
   if (!aws.containsKey('secret_key')) {
     error("${AWS_ACCOUNT_SECRET} doesn't contain 'secret_key'")
   }
+  // GCP
```
mtojek, line 374: Unfortunately, it's not so simple in this case, but no worries, we'll sort it out :)

Could you please try something similar to this:

```groovy
withCredentials([string(credentialsId: "${env.JOB_GCS_BUCKET_INTERNAL}", variable: 'GOOGLE_CREDENTIALS')]) {
  maskedVars.add([var: 'GOOGLE_CREDENTIALS', password: env.GOOGLE_CREDENTIALS]);
}
```

The idea is to pull the env from credentials and place it in the masked vars. Be careful not to leak the confidential data with some "print" :)

endorama: well, I tried :D I updated the code with your suggestion, still I do not understand who is going to set `GOOGLE_CREDENTIALS` variable in the GCP project :) (If there is any docs about this, happy to read them)

mtojek: Elastic-package uses the env.yml file to pass selected env variables to the Terraform docker container.

I see that you also use GCP_PROJECT_ID. Is it required? If so, then we need to set it here too (hardcode?)

endorama: `GCP_PROJECT_ID` is required, and is the project where resources will be created into, and where the credentials passed are valid. Should it be a masked var as well?

mtojek: > Should it be a masked var as well? 

I don't think it makes any difference. If it's simpler to implement, keep it masked.

> Which project should I use?

`elastic-observability`, that's the one that contains an internal bucket with logs. If you need a numeric ID, log in to the GCP control panel and copy it from there.

mtojek: I noticed that you mistyped it. It should be:

```groovy
withCredentials([string(credentialsId: "${env.JOB_GCS_EXT_CREDENTIALS}", variable: 'GOOGLE_CREDENTIALS')]) {
```

mtojek: @v1v We're having a hard time figuring out why the Jenkins pipeline fails here. Apparently, there isn't any error logged. We want to read the credentials for the `elastic-observability` account and store them in GOOGLE_CREDENTIALS. 

Would you mind taking a look?

endorama: Fixed and I added `GCP_PROJECT_ID`, but CI is still red without errors. I'm not sure why there are no errors in the CI tasks.

v1v: A bit of clarity about debugging errors:

1) Genuine errors are reported correctly in the Blue Ocean UI
2) Pipeline errors are not reported correctly in the Blue Ocean UI

The problem is related to the credentials type:

```
11:09:05  ERROR: Credentials 'beats-ci-gcs-plugin-file-credentials' is of type 'Secret file' where 'org.jenkinsci.plugins.plaincredentials.StringCredentials' was expected

```


The below command helped to debug the error:

```
curl -s https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-701/5/consoleText | grep -i error
```

I'll raise an issue to be able to show errors related to `2)` since the ones for `1)` are reported in the GitHub comment, but the ones for `2)` are not (I don't know if I'll be success)

v1v: BTW, sorry that you had some issues while debugging these errors. It's not a great user experience!

The traditional jenkins UI is the last resort to debug errors that are not easy to catch

mtojek: Thank you, @v1v for bringing it up!

What we're trying to achieve here is to extract GOOGLE_CREDENTIALS for the `elastic-observability` account and pass it to the system tests runner (elastic-package [Terraform service deployer](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md#terraform-service-deployer)). We have already done with AWS credentials (access key, secret access key) and it works well.

Now, it's time to fetch credentials for GCP. The `GOOGLE_CREDENTIALS` env will be passed to the Terraform service deployer (Docker container) to execute TF scenario and create GCP resources.

PS. Probably same story for `GCP_PROJECT_ID`.

mtojek:
I got this error now:

```

[36mterraform_1  |[0m 
[36mterraform_1  |[0m Plan: 1 to add, 0 to change, 0 to destroy.
[36mterraform_1  |[0m google_compute_instance.default: Creating...
[36mterraform_1  |[0m 
[36mterraform_1  |[0m Error: Error loading zone 'us-central1-a': googleapi: Error 403: Required 'compute.zones.get' permission for 'projects/elastic-obs-integrations-dev/zones/us-central1-a', forbidden
[36mterraform_1  |[0m 
[36mterraform_1  |[0m   with google_compute_instance.default,
[36mterraform_1  |[0m   on compute.tf line 7, in resource "google_compute_instance" "default":
[36mterraform_1  |[0m    7: resource "google_compute_instance" "default" {
[36mterraform_1  |[0m 
[36mterraform_1  |[0m + echo 'Terraform definitions applied.'
[36mterraform_1  |[0m Terraform definitions applied.
[36mterraform_1  |[0m + set +x
```

I'm wondering which credentials is the TF using.

.ci/Jenkinsfile
```diff
@@ -375,6 +378,11 @@ def withCloudTestEnv(Closure body) {
     [var: "AWS_ACCESS_KEY_ID", password: aws.access_key],
     [var: "AWS_SECRET_ACCESS_KEY", password: aws.secret_key],
   ])
+  // GCP
+  withCredentials([file(credentialsId: env.JOB_GCS_EXT_CREDENTIALS, variable: 'FILE_CREDENTIAL')]) {
+    maskedVars.add([var: 'GOOGLE_CREDENTIALS', password: readFile(file: env.FILE_CREDENTIAL)]);
+    maskedVars.add([var: 'GCP_PROJECT_ID', password: env.GCP_PROJECT_ID])
+  }
```
v1v, line 385: You don't need this section but a shared library step to wrap the command that requires the GCP connection:

- https://github.com/elastic/apm-pipeline-library/blob/main/vars/withGCPEnv.txt

This step allows to use the credentials stored in vault, this is much more dynamic. You can see an example in:
- https://github.com/elastic/golang-crossbuild/blob/f89836691dfbc6784d096150ac1acb5458ed63c2/Jenkinsfile#L162-L170

mtojek: I can experiment with `withGCPEnv`, but all we need is to sniff the GOOGLE_CREDENTIALS and pass it to the TF container (only this ENV, maybe also GCP_PROJECT_ID). We don't need full initialization of `withGCPEnv`. Once we pull the credentials, we can disable the `withGCPEnv`'s context.

Let me know what you think.

mtojek: Just in case: I adapted the PR to use `withGCP`

mtojek:
I fixed the issue with Google Cloud credentials. Not the problem is:

```

**Comment by mtojek** [comment]
Error is most likely related to: https://github.com/elastic/kibana/issues/125625

**Comment by mtojek** [comment]
@endorama Please change the default version temporarily to 8.0.1-SNAPSHOT to get rid of the error. It's the same change as in [here](https://github.com/elastic/elastic-package/pull/708). We expected the release v8.0.1 to happen last week, but it isn't still there?

File: internal/install/stack_version.go [hunk 3]
```diff
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "8.0.0"
+	DefaultStackVersion = "8.0.1"
```
mtojek, line 9: 8.0.1-SNAPSHOT

mtojek:
@endorama CI failed due to a different reason than the misconfigured GCP terraform deployer. I suggest to keep digging in internal filebeat/metricbeat logs or policy.

endorama:
Question: given the permission issue we are observing, I'm wondering the benefits of running the CI worker in a project and creating test resources in a different project. Is what we usually do? Should we continue trying to do this or should we just use the same GCP project?

/cc @v1v

mtojek:
@endorama I'm working (with Victor's help) on setting up a different service account. This way we can control permissions. I don't have preferences here, but I bet that we don't need to create a new project. Let's focus on merging/closing this PR first. Then, we can talk about improvements.

test/packages/parallel/gcp/data_stream/compute/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,6 @@
+# wait_for_data_timeout: 10m
+vars:
+  project_id: elastic-obs-integrations-dev
```
mtojek, line 3: Shouldn't it be inherited, for example `{{GCP_PROJECT_ID}}`?

endorama:
Updated permission for test SA. Triggering tests

/test

mtojek:
Nice, next level:

```
{"log.level":"error","@timestamp":"2022-03-02T17:05:08.372Z","log.logger":"centralmgmt","log.origin":{"file.name":"cfgfile/list.go","file.line":99},"message":"Error creating runner from config: 1 error: string value is not set accessing 'project_id'","service.name":"metricbeat","ecs.version":"1.6.0"}
{"log.level":"error","@timestamp":"2022-03-02T17:05:08.372Z","log.logger":"centralmgmt.fleet","log.origin":{"file.name":"management/manager.go","file.line":307},"message":"1 error: Error creating runner from config: 1 error: string value is not set accessing 'project_id'","service.name":"metricbeat","ecs.version":"1.6.0"}
```

test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/env.yml
```diff
@@ -2,5 +2,7 @@ version: '2.3'
 services:
   terraform:
     environment:
+      # terraform dedicated variable
       - TF_VAR_GCP_PROJECT_ID=${GCP_PROJECT_ID}
+      - GCP_PROJECT_ID=${GCP_PROJECT_ID}
```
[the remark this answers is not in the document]
endorama: `TF_VAR` is Terraform specific (as is the way Terraform accepts variables through environment variables); to prevent a dependency in `test-default-config` I think is better to have defined the pass through variable `GCP_PROJECT_ID` (which is a convention and how you would expect it to be called) and the `TF_VAR` format.

Even if it's true that the value is the same I would not consider this duplication as the variables have different usage.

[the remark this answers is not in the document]
endorama: (I also wanted to quickly check if this was the root cause, so open to discussion)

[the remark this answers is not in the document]
mtojek: Question: is this ENV (`GCP_PROJECT_ID `) required by any part of Terraform code? If not, let's follow KISS.

.ci/Jenkinsfile
```diff
@@ -375,6 +379,12 @@ def withCloudTestEnv(Closure body) {
     [var: "AWS_ACCESS_KEY_ID", password: aws.access_key],
     [var: "AWS_SECRET_ACCESS_KEY", password: aws.secret_key],
   ])
+  // GCP
+  withGCPEnv(secret: env.ELASTIC_PACKAGE_GCP_SECRET) {
+    maskedVars.add([var: 'GOOGLE_CREDENTIALS', password: readFile(file: env.GOOGLE_APPLICATION_CREDENTIALS)]);
+    maskedVars.add([var: 'GCP_PROJECT_ID', password: env.ELASTIC_OBSERVABILITY_PROJECT_ID])
+  }
+  // Masking
```
v1v, line 387: by doing so, aren't you adding this for all the parallel stages regardless whether they interact with GCP or no? I wonder whether this should be done only when needed, and it's feasible?

endorama:
Current error related to `ListMetricDescriptor` request:

```
ERROR | Could not make ListMetricDescriptors request for metric type firewall/dropped_bytes_count: rpc error: code = Unauthenticated desc = transport: metadata: GCE metadata "instance/service-accounts/default/token?scopes=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring.read%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring.write" not defined
ERROR | Error creating runner from config: 1 error: error calling metricDescriptor function: Could not make ListMetricDescriptors request for metric type firewall/dropped_bytes_count: rpc error: code = Unauthenticated desc = transport: metadata: GCE metadata "instance/service-accounts/default/token?scopes=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring.read%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring.write" not defined
ERROR | 1 error: Error creating runner from config: 1 error: error calling metricDescriptor function: Could not make ListMetricDescriptors request for metric type firewall/dropped_bytes_count: rpc error: code = Unauthenticated desc = transport: metadata: GCE metadata "instance/service-accounts/default/token?scopes=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring.read%2Chttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fmonitoring.write" not defined
```

I added `monitoring.viewer` role to SA.

/test

.ci/Jenkinsfile
mtojek: Yes, that's a side effect of refactoring done some time ago. Most likely it can be improved, but let's keep it aside from this PR. It's consistent with AWS at the moment.

endorama:
Searching for [gce metadata instance/service-accounts/default/token not defined](https://www.google.com/search?q=gce+metadata+instance%2Fservice-accounts%2Fdefault%2Ftoken+not+defined&client=ubuntu&hs=3Fq&channel=fs&sxsrf=APq-WBuiYG-9QenpPYUSkg4OrqqNHmKJ4g%3A1646317490821&ei=ss8gYvnHMYLakwX1m4dg&oq=%22gce+metadata+not+defined%22&gs_lcp=Cgdnd3Mtd2l6EAMYATIGCAAQFhAeMgYIABAWEB46BwgAEEcQsAM6BQgAEIAEOgYIABAIEB5KBAhBGABKBAhGGABQ0whYywtg1x1oAXABeACAAWSIAa8BkgEDMS4xmAEAoAEByAEIwAEB&sclient=gws-wiz) brings very few results, but hey resolve around permission issues.

Given that `elastic-package-gcp` GCP SA has correct Compute and Monitoring permissions (the same I use in a test SA to run test locally) I'm wondering which SA is being used to run data collection. Could it be is not the new one? :thinking:

endorama:
I'm also wondering if this is an effect of this issue https://github.com/elastic/integrations/pull/2712 that I tried fixing with https://github.com/elastic/elastic-package/pull/701/commits/25e89a4e6570a3ab318f8d5354085a5833028b53

@gpop63 pointed out that his test to solve this have been unsuccessful so far.

test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/env.yml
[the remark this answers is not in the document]
mtojek: So it means that you don't need it in this file.

test/packages/parallel/gcp/data_stream/billing/agent/stream/stream.yml.hbs
```diff
@@ -0,0 +1,12 @@
+metricsets: ["billing"]
+period: {{period}}
+project_id: {{project_id}}
+{{#if credentials_file}}
+credentials_file_path: {{credentials_file}}
+{{/if}}
+{{#if credentials_json}}
+credentials_json: {{credentials_json}}
```
andrewkroh, line 8: Should this be single quoted?

test/packages/parallel/gcp/data_stream/compute/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,6 @@
+# wait_for_data_timeout: 10m
+vars:
+  project_id: "{{GCP_PROJECT_ID}}"
+  credentials_json: "{{GOOGLE_CREDENTIALS}}"
```
andrewkroh, line 4:
```suggestion
  credentials_json: '{{{GOOGLE_CREDENTIALS}}}'
```

Because of HTML escaping from https://github.com/aymerick/raymond#html-escaping.

test/packages/parallel/gcp/data_stream/billing/agent/stream/stream.yml.hbs
endorama: we are adding test case for `compute` data stream, so `billing` has not been updated because is not tested. (I wouldn't work anyway as it requires other changes too). I'll remove it.

mtojek:
I paired with @gpop63 and checked all the wires. It looks like we need at least 8.1.0 to use `credentials_json`, I will check that.

endorama:
Luckily `8.1.0` release date is today. I'm going to rework this PR to clean it up, but as soon as `elastic-package` CI starts using `8.1.0` this is can to be merged.

endorama:
Tests are failing due to an empty `project_id` field.

`GCP_PROJECT_ID` is set in `Jenkinsfile` as a masked value (but I don't think this has any implication) and `GCP_PROJECT_ID` env is used in `test-default-config.yml` to set `project_id`. I removed the `GCP_PROJECT_ID` env from `tf/env.yml` but as per [this conversation](https://github.com/elastic/elastic-package/pull/701#discussion_r818841665) shouldn't be a factor.

What am I missing?

mtojek:
Well, the Terraform runner reads the right project ID and it isn't hardcoded there, so Jenkins sets it correctly. We need to trace it down to the Metricbeat.

mtojek:
@endorama Please do not use the force-push mode as it erases the history. We won't be able to figure out why it was working (collecting metrics) before and now. Please use rebase/pull instead.

test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/env.yml
```diff
@@ -0,0 +1,7 @@
+version: '2.3'
+services:
+  terraform:
+    environment:
+      # pass project id to Terraform
+      - TF_VAR_gcp_project_id=${GCP_PROJECT_ID}
```
mtojek, line 6: Shouldn't we preserve casing consistency (upper-case)?

mtojek (commented):
I'm concerned now as we managed to make it running a few days ago. Did you try to copy over files and figure out the diff?

test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/env.yml
endorama: This format is consistent with Terraform [documentation about setting input variables from environment variables](https://www.terraform.io/language/values/variables#environment-variables).

I think this consistency helps Terraform practitioners (both beginners and experts alike) more.

I also know we are using `TEST_RUN_ID`, something I think should be changed for the same reason, but I didn't want to address that in this PR.

mtojek: My bad looks like it's aligned with the official docs. Probably it's something we can mention in our docs. Anyway, nothing to adjusted.

test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/compute.tf
```diff
@@ -0,0 +1,33 @@
+data "google_compute_image" "default" {
+  # https://cloud.google.com/compute/docs/images
+  family  = "ubuntu-minimal-2004-lts"
+  project = "ubuntu-os-cloud"
+}
+
+resource "google_compute_instance" "default" {
+  name = "test-${var.TEST_RUN_ID}"
```
mtojek, line 8: Let's add a prefix identifying the instance with elastic-package. The `test` prefix may be too vague.

test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/vars.tf
```diff
@@ -0,0 +1,16 @@
...
+  default = "detached"
+}
+
+variable "gcp_project_id" {
+  type = string
+}
+
+variable "zone" {
+  type = string
+  // NOTE: if you change this value you **must** change it also for test
```
mtojek, line 11: if it possible to read it from the `env.yml` config?

mtojek (commented):
I'm glad that the issue is fixed. I left few observations around the test configuration, mostly nit-picks/improvements.

endorama:
@mtojek May you please help us understand why https://github.com/elastic/elastic-package/pull/701/commits/ce21b001e2de5b4704a3865140371e05f4fccda6 is necessary? I expected https://github.com/elastic/elastic-package/pull/701/commits/141fc96ec3d073d558ae76bedbb6c7e41c53cbd9 to be enough, as `GCP_PROJECT_ID` was not set but `TF_VAR_gcp_project_id` was.

test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/vars.tf
endorama: We can pass it like the project id with `TF_VAR_zone`, but I would prefer having a default value so if someone runs it locally the value is already present.

mtojek: You can define default variables like [this](https://github.com/elastic/elastic-package/blob/main/test/packages/parallel/aws/data_stream/ec2_metrics/_dev/deploy/tf/env.yml#L9).

mtojek:
It was designed to treat `TF_VAR_` variables as internal, which should be used within the Terraform. We wouldn't like to pollute agent policies with `TF_VAR_`s, which is more like an implementation detail, hence we skip them [here](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/servicedeployer/terraform_env.go#L44).

Analyzing your case:

```yaml
services:
  terraform:
    environment:
      - GCP_PROJECT_ID=${GCP_PROJECT_ID}
      # pass project id to Terraform
      - TF_VAR_gcp_project_id=${GCP_PROJECT_ID}
      - GOOGLE_CREDENTIALS=${GOOGLE_CREDENTIALS}
```

From Terraform's perspective, it should be ok to set only the `TF_VAR_gcp_project_id`, but you need the other line:

```yaml
      - GCP_PROJECT_ID=${GCP_PROJECT_ID}
```

to read the `GCP_PROJECT_ID` in the system test config. That's how `GCP_PROJECT_ID` and `GOOGLE_CREDENTIALS` magically appear in the agent policy.

endorama:
Great that makes perfect sense! So the current code is correct, may I ask for the final review then?

mtojek:
I guess you have to run the formatter now?

mtojek (approved):
Thanks for working on this!

endorama:
First try (not)

:laughing:

Happy to see this land and THANK YOU everyone for all the help!

PR #704, opened 2022-02-17
Generate a cobertura report for processors in pipeline tests

adriansr:
This PR updates the pipeline tests to generate code-coverage (at the processor level) for ingest pipelines:

image
image

Changes in CI will be needed so that coverage reports are visible in CI.

For local testing, reports are available under `build/test-coverage` and can be visualized with a tool like [Report Generator.](https://github.com/danielpalme/ReportGenerator)

### DIFF
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -149,10 +149,6 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				if err != nil {
 					return cobraext.FlagParsingError(err, cobraext.DataStreamsFlagName)
 				}
-
-				if testCoverage && len(dataStreams) > 0 {
-					return cobraext.FlagParsingError(errors.New("test coverage can be calculated only if all data streams are selected"), cobraext.DataStreamsFlagName)
-				}
 			}
 
 			if runner.TestFolderRequired() {
@@ -203,6 +199,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				API:                esClient.API,
 				DeferCleanup:       deferCleanup,
 				ServiceVariant:     variantFlag,
+				WithCoverage:       testCoverage,
 			})
 
 			results = append(results, r...)
--- a/internal/testrunner/coverageoutput.go
+++ b/internal/testrunner/coverageoutput.go
@@ -49,7 +49,7 @@ func (tcd *testCoverageDetails) withTestResults(results []TestResult) *testCover
 		}
 		tcd.dataStreams[result.DataStream] = append(tcd.dataStreams[result.DataStream], result.Name)
 		if tcd.cobertura != nil && result.Coverage != nil {
-			if err := tcd.cobertura.Merge(result.Coverage); err != nil {
+			if err := tcd.cobertura.merge(result.Coverage); err != nil {
 				tcd.errors = append(tcd.errors, errors.Wrapf(err, "can't merge Cobertura coverage for test `%s`", result.Name))
 			}
 		} else if tcd.cobertura == nil {
@@ -132,8 +132,8 @@ func (c *CoberturaCoverage) bytes() ([]byte, error) {
 	return buffer.Bytes(), nil
 }
 
-// Merge merges two coverage reports for a given class.
-func (c *CoberturaClass) Merge(b *CoberturaClass) error {
+// merge merges two coverage reports for a given class.
+func (c *CoberturaClass) merge(b *CoberturaClass) error {
 	// Check preconditions: classes should be the same.
 	equal := c.Name == b.Name &&
 		c.Filename == b.Filename &&
@@ -164,8 +164,8 @@ func (c *CoberturaClass) Merge(b *CoberturaClass) error {
 	return nil
 }
 
-// Merge merges two coverage reports for a given package.
-func (p *CoberturaPackage) Merge(b *CoberturaPackage) error {
+// merge merges two coverage reports for a given package.
+func (p *CoberturaPackage) merge(b *CoberturaPackage) error {
 	// Merge classes
 	for _, class := range b.Classes {
 		var target *CoberturaClass
@@ -176,7 +176,7 @@ func (p *CoberturaPackage) Merge(b *CoberturaPackage) error {
 			}
 		}
 		if target != nil {
-			if err := target.Merge(class); err != nil {
+			if err := target.merge(class); err != nil {
 				return err
 			}
 		} else {
@@ -186,8 +186,8 @@ func (p *CoberturaPackage) Merge(b *CoberturaPackage) error {
 	return nil
 }
 
-// Merge merges two coverage reports.
-func (c *CoberturaCoverage) Merge(b *CoberturaCoverage) error {
+// merge merges two coverage reports.
+func (c *CoberturaCoverage) merge(b *CoberturaCoverage) error {
 	// Merge source paths
 	for _, path := range b.Sources {
 		found := false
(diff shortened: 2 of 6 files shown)


internal/testrunner/coverageoutput.go
```diff
@@ -42,11 +48,19 @@ func (tcd *testCoverageDetails) withTestResults(results []TestResult) *testCover
...
+		} else if tcd.cobertura == nil {
+			tcd.cobertura = result.Coverage
+		}
 	}
 	return tcd
 }
 
-type coberturaCoverage struct {
+// CoberturaCoverage is the root element for a Cobertura XML report.
+type CoberturaCoverage struct {
```
mtojek, line 63: Is there any reason why these properties are exposed? Is it for the consistency with `CoberturaCoverage`?

internal/testrunner/coverageoutput.go
```diff
@@ -113,6 +132,109 @@ func (c *coberturaCoverage) bytes() ([]byte, error) {
...
+			}
+		} else {
+			p.Classes = append(p.Classes, class)
+		}
+	}
+	return nil
+}
+
+// Merge merges two coverage reports.
+func (c *CoberturaCoverage) Merge(b *CoberturaCoverage) error {
```
mtojek, line 190: Do you think that this method can be package-only visible?

cmd/testrunner.go
```diff
@@ -203,6 +199,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				API:                esClient.API,
 				DeferCleanup:       deferCleanup,
 				ServiceVariant:     variantFlag,
+				WithCoverage:       testCoverage,
```
mtojek, line 202: Does this property apply to all test runners (system tests, static tests, etc.)? Maybe we should add a WARN informing if we can't calculate the coverage properly.

mtojek (commented):
Could you please describe how will the coverage calculation change once this PR is merged?

Consider the following cases:
1. Run `test` - cover all data streams of a package
2. Run `test --data-stream foo` - cover only the `foo` data stream
3. Run `test pipeline` - cover all data streams, but only pipeline tests
4. Run `test` - but no tests present

Sorry for adding more tasks to do, but I'd like to double-check if there are any risky side effects.

Also, what kind of changes do we need on the CI side?

internal/testrunner/coverageoutput.go
adriansr: I think I got mixed here a little bit. The original idea was for the test runners to have the option to generate their own coverage in Cobertura format (via the new TestResult.Coverage), that's why these types are exposed.

To align with this idea, I think it makes more sense to put GetPipelineCoverage into the pipeline runner instead of the global testrunner package as is now.

internal/testrunner/coverageoutput.go
adriansr: done

cmd/testrunner.go
adriansr: The way I see it we have the old default coverage, which only tells whether we have a particular kind of test for the package, and the opt-in detailed coverage added by this PR. Each individual test runner can err if they wanted to calculate detailed coverage but failed, as the pipeline tests will do after this PR.

internal/testrunner/coverageoutput.go
mtojek: > To align with this idea, I think it makes more sense to put GetPipelineCoverage into the pipeline runner instead of the global testrunner package as is now.

Yes, I have a similar feeling about this :)

mtojek:
@adriansr Please re-request the review when this PR is ready for another round. It would great to research [these cases](https://github.com/elastic/elastic-package/pull/704#pullrequestreview-888389045).

internal/testrunner/coverageoutput_test.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+func TestCoberturaCoverage_Merge(t *testing.T) {
+	tests := []struct {
+		name               string
+		rhs, lhs, expected CoberturaCoverage
+		wantErr            bool
+	}{
+		{
+			name: "merge sources",
```
mtojek, line 20: nit: I'm thinking now if it isn't simpler and easier to read to load these tests from a `testdata` directory. WDYT?

internal/testrunner/runners/pipeline/coverage.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/elasticsearch/ingest"
+	"github.com/elastic/elastic-package/internal/packages"
+	"github.com/elastic/elastic-package/internal/testrunner"
+)
+
+// GetPipelineCoverage returns a coverage report for the provided set of ingest pipelines.
+func GetPipelineCoverage(options testrunner.TestOptions, pipelines []ingest.Pipeline) (*testrunner.CoberturaCoverage, error) {
+	packagePath, err := packages.MustFindPackageRoot()
```
mtojek, line 21: Isn't the package root present in `TestOptions`?

internal/testrunner/runners/pipeline/coverage.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// Tests install pipelines as `filename-<nonce>` (without original extension).
+	// Use the filename part for the report.
+	pipelineName := pipeline.Name
+	if nameEnd := strings.LastIndexByte(pipelineName, '-'); nameEnd != -1 {
+		pipelineName = pipelineName[:nameEnd]
+	}
+
+	// File path has to be relative to the packagePath added to the cobertura Sources list
```
mtojek, line 101: Let's discuss the safe deployment of this feature.

If we merge this PR and release a new elastic-package, will it break the CI tests for Elastic Integrations? If it breaks, then we need to go the other way round and adjust the Jenkinsfile first and make sure that pipeline sources are in the right directory (relative to the runner?).

mtojek (commented):
Thanks for providing more information, Adrian. I see only concerns around integrating this change with our CI. I left a comment in a relevant place in the source code.

Apart from this, I think it's 👍 .

internal/testrunner/runners/pipeline/coverage.go
adriansr: Turns out it is.

internal/testrunner/runners/pipeline/coverage.go
adriansr: There's no reason to believe it will break any tests, I've tested it multiple times with all the existing integrations.

I think you're misunderstanding the CI issues that will need to be addressed. I explained them in this comment: https://github.com/elastic/elastic-package/pull/585#issuecomment-970186919

Those issues will not cause a test failure. Just the coverage report displayed in CI will not show annotated source code.

mtojek: Yes, that's what I mean. Let me share my concerns, Jenkins uses [coverageReport](https://github.com/elastic/integrations/blob/main/.ci/Jenkinsfile#L280) function to collect the coverage stats. I don't know what will happen if that function files (due to missing pipeline sources). Will it not show the coverage split by source code or fail the entire CI pipeline?

adriansr: It is not causing a CI pipeline failure in this PR, just the source code is reported missing. See:

https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-704/4/cobertura/apache_access/apache_data_stream_access_elasticsearch_ingest_pipeline_default_yml/

mtojek: Understood, thanks for clarifying this.

mtojek (approved):

PR #711, opened 2022-03-01
ci: support for "skip-ci" github label in order to reduce what to run

v1v:
## What

`skip-ci` label will only run the linting stage and then will stop doing anything else.

## Why

See the below use case:

> As a developer, I don't want to run CI because I'm sure is going to fail and will just waste CI cycles

## Issue

Similar to https://github.com/elastic/beats/pull/21377

## UX

See the below comment that points to the reason for the failure:

image

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -73,10 +73,6 @@ pipeline {
                 sh(label: 'Check',script: 'make check-static')
               }
             }
-            // IMPORTANT: notify a build status in order to avoid GitHub checks with a wrong status.
-            whenTrue(matchesPrLabel(label: 'skip-ci')) {
-              error 'Pull Request has been configured to be disabled when there is a skip-ci label match'
-            }
           }
         }
         /**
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):

.ci/Jenkinsfile
```diff
@@ -145,6 +153,10 @@ pipeline {
   }
   post {
     always {
+      // IMPORTANT: notify a build status in order to avoid GitHub checks with a wrong status.
+      whenTrue(matchesPrLabel(label: 'skip-ci')) {
+        error 'Pull Request has been configured to be disabled when there is a skip-ci label match'
```
endorama, line 158: Out of curiosity, is not possible to report it as skipped instead of errord?

endorama:
cc @mtojek

.ci/Jenkinsfile
v1v: If so, then GitHub checks will be passed

image

and as consequence, there is a possibility the PR could be accidentally merged without running everything in the CI.

I'm inclined to be explicit with an artificial failure to avoid the above if it makes any sense.

v1v:
@mtojek , the  `why` section in the description contains the reason, I kindly raised this based on the feedback from @endorama , you can freely close it if it's not needed :)

mtojek:
I see now, thanks for pointing it out, Victor. I believe that in the case of the [Add GCP test package](https://github.com/elastic/elastic-package/pull/701) the problem is that we're waiting for the release v8.0.1, which should happen really soon. I wouldn't like to consider it as a justification to disable all tests, including successful ones.

If this is the only reason why we need this PR, I would be for not merging it.

PR #714, opened 2022-03-02
internal/testrunner/runners: don't just say no, say why

efd6:
* improve user-facing message when an error.message cannot be unmarshaled.
* be more liberal in accepting types to report and warn about unexpected
  types.

Please take a look.

### DIFF
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -337,20 +337,24 @@ func verifyFieldsInTestResult(result *testResult, fieldsValidator *fields.Valida
 }
 
 func checkErrorMessage(event json.RawMessage) error {
-	var pipelineError = struct {
+	var pipelineError struct {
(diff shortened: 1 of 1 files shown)


internal/testrunner/runners/pipeline/runner.go
```diff
@@ -337,20 +337,24 @@ func verifyFieldsInTestResult(result *testResult, fieldsValidator *fields.Valida
 }
 
 func checkErrorMessage(event json.RawMessage) error {
-	var pipelineError = struct {
+	var pipelineError struct {
 		Error struct {
-			Message string
+			Message interface{}
```
jsoriano, line 342: I think this comes from the simulate API, can we expect different types here?

jsoriano (commented):
LGTM, added a question to better understand the problem this PR is solving.

jsoriano (approved):
Thanks!

PR #718, opened 2022-03-04
internal/testrunner/runners/pipeline: use diff with limited context lines when rendering failures

efd6:
This implements a lightweight diff -Un function to use when rendering differences
reducing the amount of text needed to read through to get to informative differences.

Please take a look.

Example output:
```
FAILURE DETAILS:
panw/panos test-panw-panos-globalprotect-sample.log:
--- want
+++ got
@@ -983 +983 @@
              },
              "panw": {
                  "panos": {
-                     "act ion_flags": "true",
+                     "action_flags": "true",
                      "client_ver": "5.2.8",
                      "description": "Pre-tunnel latency: 35ms, Post-tunnel latency: 16ms",
                      "error_code": "0",
                      "repeat_count": 1,
                      "sequence_number": 1041590,
-                     "seri al_number": "SERIALNR",
+                     "serial_number": "SERIALNR",
                      "source": {
                          "nat": {
                              "ip": "fc00::1234"
@@ -1008 +1008 @@
                  ],
                  "ip": [
                      "fc00::abcd",
-                     "fc 00::1234"
+                     "fc00::1234"
                  ],
                  "user": [
-                     "us er"
+                     "user"
                  ]
              },
              "source": {
-                 "ip": "fc 00::abcd",
+                 "ip": "fc00::abcd",
                  "nat": {
-                     "ip": "fc00 ::1234"
+                     "ip": "fc00::1234"
                  },
                  "user": {
-                     "name": "us er"
+                     "name": "user"
                  }
              },
              "tags": [
```

### DIFF
--- a/internal/testrunner/runners/pipeline/test_result.go
+++ b/internal/testrunner/runners/pipeline/test_result.go
@@ -5,10 +5,12 @@
 package pipeline
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/kylelemons/godebug/diff"
 	"github.com/pkg/errors"
@@ -63,7 +65,7 @@ func compareResults(testCasePath string, config *testConfig, result *testResult)
 		return errors.Wrap(err, "marshalling expected test results failed")
 	}
 
-	report := diff.Diff(string(expected), string(actual))
+	report := diffUlite(string(expected), string(actual), 3)
 	if report != "" {
 		return testrunner.ErrTestCaseFailed{
 			Reason:  "Expected results are different from actual ones",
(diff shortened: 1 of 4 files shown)


internal/testrunner/runners/pipeline/test_result.go
```diff
@@ -73,6 +75,81 @@ func compareResults(testCasePath string, config *testConfig, result *testResult)
 	return nil
 }
 
+// diffUlite implements a unified diff-like rendering with u lines of context.
+// It differs from a complete unified diff in that it does not provide the length
+// of the chunk, so it cannot be used to generate patches, but can be used for
+// human inspection.
+func diffUlite(a, b string, u int) string {
```
mtojek, line 82: Could you please make sure that there isn't any Go library, that can deliver the same functionality? I would rather follow that approach than implementing it on your own.

Maybe [go-diff](https://github.com/sergi/go-diff)?

internal/testrunner/runners/pipeline/test_result.go
```diff
@@ -63,7 +65,7 @@ func compareResults(testCasePath string, config *testConfig, result *testResult)
 		return errors.Wrap(err, "marshalling expected test results failed")
 	}
 
-	report := diff.Diff(string(expected), string(actual))
+	report := diffUlite(string(expected), string(actual), 3)
```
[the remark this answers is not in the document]
efd6: The intention was to not change this; it's provided because the defaul for git is 3, but I will do this.

internal/testrunner/runners/pipeline/test_result.go
efd6: I did a survey first. go-diff is much more complicated than is needed here. My initial intention was to use go-cmp, but it is too focused on Go use and though this is a Go program, integrations should not assume users have any Go coding experience. The second package I looked at was indeed go-diff, but it is massively over`engineered for what is needed with an overcomplicated API.

mtojek (approved):

PR #720, opened 2022-03-04
Validate built packages too

jsoriano:
Built packages can have content that was not in their source files, such
as imported ECS fields. Validate final packages to ensure that content
included during build is also valid.

For example while testing `ValidateFromZip` I found that some packages
contained ECS fields of type `match_text_only`, that was not supported
in the spec (fixed in https://github.com/elastic/package-spec/pull/284).

@mtojek not sure if this is the best place to do this validation, but I think
we should have something like this safeguard to ensure that built packages
still comply with the spec.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -24,7 +24,17 @@ func setupLintCommand() *cobraext.Command {
 		Use:   "lint",
 		Short: "Lint the package",
 		Long:  lintLongDescription,
-		RunE:  lintCommandAction,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := cobraext.ComposeCommandActions(cmd, args,
+				lintCommandAction,
+				validateSourceCommandAction,
+			)
+			if err != nil {
+				return err
+			}
+			cmd.Println("Done")
+			return nil
+		},
 	}
 
 	return cobraext.NewCommand(cmd, cobraext.ContextPackage)
@@ -32,14 +42,6 @@ func setupLintCommand() *cobraext.Command {
 
 func lintCommandAction(cmd *cobra.Command, args []string) error {
 	cmd.Println("Lint the package")
-
-	packageRootPath, found, err := packages.FindPackageRoot()
-	if !found {
-		return errors.New("package root not found")
-	}
-	if err != nil {
-		return errors.Wrap(err, "locating package root failed")
-	}
 
 	readmeFiles, err := docs.AreReadmesUpToDate()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/builder/packages.go
```diff
@@ -181,24 +191,36 @@ func BuildPackage(options BuildOptions) (string, error) {
 		return "", errors.Wrapf(err, "can't compress the built package (compressed file path: %s)", zippedPackagePath)
 	}
 
-	if !options.SignPackage {
-		return zippedPackagePath, nil
+	err = validator.ValidateFromZip(zippedPackagePath)
```
mtojek, line 194: I fully agree with the assumption that we have a gap in validation, but I'm wondering if we shouldn't package the `build` flow instead of this bugfix.

```
1. Resolve dependencies.
2. 1st validation.
3. Build the package (.zip or dir).
4. 2nd validation (.zip or dir).
```

If we enable validate as part of the `build` action, then, I suppose, we don't need to call "validate" in the "check" flow. WDYT?

One thing regarding this feature. The built package will also differ in terms of Kibana JSON files as some properties are encoded. Maybe it's worth leaving a comment here describing differences between 1st and 2nd validation we know about.

internal/builder/packages.go
```diff
@@ -9,11 +9,13 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/pkg/errors"
+
```
mtojek, line 13: nit: empty line

internal/builder/packages.go
jsoriano: > If we enable validate as part of the `build` action, then, I suppose, we don't need to call "validate" in the "check" flow. WDYT?

Ah, yes, you are right.

> One thing regarding this feature. The built package will also differ in terms of Kibana JSON files as some properties are encoded. Maybe it's worth leaving a comment here describing differences between 1st and 2nd validation we know about.

I wonder if we need two different specs :grimacing: One for elastic-package source packages, and another one for built packages.

We have already some differences:
* Source packages can have external fields, that don't need validation in check time.
* Built packages shouldn't have any external fields, but all their fields should be validated.
* Kibana objects are decoded in source packages, but encoded in built packages. 

Or instead of two different specs, we could have something like validation profiles, that allow to enable or disable checks. There could be a predefined profile for source packages, and another one for built packages. This could be also used to disable checks on demand.

Or well, maybe we don't need validation of source packages, specially if we move validation to a later step after building.

mtojek: > Or instead of two different specs, we could have something like validation profiles, that allow to enable or disable checks. There could be a predefined profile for source packages, and another one for built packages. This could be also used to disable checks on demand.

You got my attention with it as it's in line with what I had in mind some time ago, but rather than officially defining profiles as a support matrix with flags, use annotations to disable some of them. This way would prevent opening doors to introduce 3rd and 4th profile, but rather stay with two. The only gap I see here is the fact that there are no annotations in Golang like in Java or Python :) Maybe we should define profiles as separate packages and register checks as plugins?

mtojek (approved):
LGTM! 

nit: if you haven't done it, you can double-check if command usage messages and "README" are up-to-date in terms of building&linting.

jsoriano:
> nit: if you haven't done it, you can double-check if command usage messages and "README" are up-to-date in terms of building&linting.

General behaviour doesn't change, so I guess we are fine there.

Issue #691, opened 2022-02-08
Update the url used to pull elastic agent image


jlind23:
From 8.2 on, Elastic Agent team will ship their docker images to a new namespace following the format below:
`docker.elastic.co/elastic-agent/elastic-agent:version`

We should then update elastic-package accordingly.

PR #723, opened 2022-03-09
Use different namespace for elastic-agent

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/691

This PR updates the logic to select the right Docker image for Elastic Agent. Starting with the ~v8.3.0~v8.2.0 release, the `elastic-package` should use a different namespace for Elastic Agent.

### DIFF
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -19,12 +19,12 @@ import (
 
 const (
 	stackVersion715 = "7.15.0-SNAPSHOT"
-	stackVersion830 = "8.3.0-SNAPSHOT"
+	stackVersion820 = "8.2.0-SNAPSHOT"
 )
 
 var (
 	elasticAgentCompleteFirstSupportedVersion = semver.MustParse(stackVersion715)
-	elasticAgentCompleteOwnNamespaceVersion   = semver.MustParse(stackVersion830)
+	elasticAgentCompleteOwnNamespaceVersion   = semver.MustParse(stackVersion820)
 )
 
 // ApplicationConfiguration represents the configuration of the elastic-package.
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -17,9 +17,15 @@ import (
 	"github.com/elastic/elastic-package/internal/logger"
 )
 
-const stackVersion715 = "7.15.0-SNAPSHOT"
+const (
+	stackVersion715 = "7.15.0-SNAPSHOT"
+	stackVersion830 = "8.3.0-SNAPSHOT"
+)
 
-var elasticAgentCompleteFirstSupportedVersion = semver.MustParse(stackVersion715)
+var (
+	elasticAgentCompleteFirstSupportedVersion = semver.MustParse(stackVersion715)
+	elasticAgentCompleteOwnNamespaceVersion   = semver.MustParse(stackVersion830)
+)
 
 // ApplicationConfiguration represents the configuration of the elastic-package.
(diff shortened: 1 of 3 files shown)


internal/install/application_configuration_yml.go
```diff
@@ -5,10 +5,11 @@
 package install
 
 const (
-	elasticAgentImageName         = "docker.elastic.co/beats/elastic-agent"
-	elasticAgentCompleteImageName = "docker.elastic.co/beats/elastic-agent-complete"
-	elasticsearchImageName        = "docker.elastic.co/elasticsearch/elasticsearch"
-	kibanaImageName               = "docker.elastic.co/kibana/kibana"
+	elasticAgentImageName               = "docker.elastic.co/beats/elastic-agent"
+	elasticAgentCompleteLegacyImageName = "docker.elastic.co/beats/elastic-agent-complete"
+	elasticAgentCompleteImageName       = "docker.elastic.co/elastic-agent/elastic-agent-complete"
```
mtojek, line 10: @jlind23 I understand that it also applies to `elastic-agent-complete` as we're using the full image version.

jsoriano (approved):
LGTM, but images are not available yet in the new namespace, right?

internal/install/application_configuration.go
```diff
@@ -83,7 +89,9 @@ func selectElasticAgentImageName(version string) string {
 	v, err := semver.NewVersion(version)
 	if err != nil {
 		logger.Errorf("stack version not in semver format (value: %s): %v", v, err)
-	} else if !v.LessThan(elasticAgentCompleteFirstSupportedVersion) {
+	} else if !v.LessThan(elasticAgentCompleteFirstSupportedVersion) && v.LessThan(elasticAgentCompleteOwnNamespaceVersion) {
+		return elasticAgentCompleteLegacyImageName
+	} else if !v.LessThan(elasticAgentCompleteOwnNamespaceVersion) {
 		return elasticAgentCompleteImageName
```
jsoriano, line 95: Nit. Simplify these ifs:
```
	if err != nil {
		return elasticAgentImageName
	}
	if !v.LessThan(elasticAgentCompleteOwnNamespaceVersion) {
		return elasticAgentCompleteImageName
	}
	if !v.LessThan(elasticAgentCompleteFirstSupportedVersion) {
		return elasticAgentCompleteLegacyImageName
	}
	return elasticAgentImageName
```

This would make more explicit what image we return in case of error when parsing the version. I wonder why we don't return the newest one (`elasticAgentCompleteImageName`).

mtojek: Fixed.

> This would make more explicit what image we return in case of error when parsing the version. I wonder why we don't return the newest one (elasticAgentCompleteImageName).

It was implemented this way to ensure it's working also for the older stack versions, when the "complete" image wasn't available.

jsoriano:
Umm, `docker pull docker.elastic.co/elastic-agent/elastic-agent:8.2.0-SNAPSHOT` already works for me, has it been added in the last hour or something else is pending?

mtojek:
We need `elastic-agent-complete:8.2-SNAPSHOT` as it contains heartbeat and other beats.

jsoriano:
Ah ok, this one doesn't work yet :)

jsoriano:
Ah no, sorry, it isn't :man_facepalming:

mtojek:
@mergifyio rebase

mtojek:
Everything looks correct now! Please take a look again and confirm that it looks good.

jlind23 (approved):

PR #728, opened 2022-03-14
Write migration docs to publish packages to v2

mtojek:
Fixes: https://github.com/elastic/package-storage/issues/2778

Preview: [link](https://github.com/mtojek/elastic-package/blob/write-migration-docs/docs/howto/use_package_storage_v2.md)

This PR adds a doc file describing the migrations from v1 to v2 storage. The goal is to prepare instructions in a way that customers can switch to the v2 model on their own.


docs/howto/use_package_storage_v2.md
```diff
@@ -0,0 +1,79 @@
...
+The Package Storage v2 is available publicly behind the endpoint: [package-storage.elastic.co](https://package-storage.elastic.co/)
+and exposes different package resources:
+* zipped packages (e.g. [barracuda-0.2.2.zip](https://package-storage.elastic.co/artifacts/packages/barracuda-0.2.2.zip))
+* package signatures (e.g. barracuda-0.2.2 [signature](https://package-storage.elastic.co/artifacts/packages/barracuda-0.2.2.zip.sig))
+* extracted static resources (e.g. cisco-0.11.5 [screenshot](https://package-storage.elastic.co/artifacts/static/cisco-0.11.5/img/kibana-cisco-asa.png))
+
+The Package Storage v2 has flat structure. It does not introduce any logic to divide packages into stages or a control panel to promote them.
+We recommend to use proper versioning instead and follow these rules:
+* a package with `version < 1.0.0` is experimental
+* a package with version containing a prerelease tag (beta1, SNAPSHOT, next) is experimental
```
jsoriano, line 17: Use of "Experimental" term is being replaced by "Technical preview".
```suggestion
* a package with `version < 1.0.0` is a technical preview.
* a package with `version >= 1.0.0` can contain prerelease tags (beta1, SNAPSHOT, next) on its version to indicate its prerelease state.
```

docs/howto/use_package_storage_v2.md
```diff
@@ -0,0 +1,79 @@
...
+
+## What is the goal of storage migration from v1 to v2?
+
+We identified a few issues in v1 design, we couldn't easily overcome or patch:
+1. Automatically release new Docker images of the Package Storage without missing packages due to a race condition
+   between CI jobs.
+2. Controle the Docker image size, which is constantly growing (as of today, >1GB). Packages with size >1GB must be served through the Package Registry too.
```
jsoriano, line 24:
```suggestion
2. Control the Docker image size, which is constantly growing (as of today, >1GB). Packages with size >1GB must be served through the Package Registry too.
```

docs/howto/use_package_storage_v2.md
```diff
@@ -0,0 +1,79 @@
...
+3. Deprecate promotion between stages. It caused a lot of frustration for customers and most of them didn't follow the recommended promotion path
```
jsoriano, line 25: Nit. I like "package developers" more than "customers" :slightly_smiling_face: 
```suggestion
3. Deprecate promotion between stages. It caused a lot of frustration for package developers and most of them didn't follow the recommended promotion path
```

docs/howto/use_package_storage_v2.md
```diff
@@ -0,0 +1,79 @@
...
+#### Requirements
+
+1. CI job signing credentials (`sign-artifacts-with-gpg-job`) - use them to call the signing pipeline on
+   the `internal-ci` Jenkins instance. The pipeline will sign artifacts uploaded to the signing bucket and upload there their signatures.
+2. Signing bucket credentials (`internal-ci-gcs-plugin`) - use them to upload zipped packages to be signed
+   and download matching signatures.
+3. Package Storage GCP uploader credentials (`upload-package-to-package-storage`) - use them to upload a package candidate to the "queue" bucket.
+   The candidates will be picked by the publishing job and removed after processing.
+4. Package Storage uploader secret (`secret/gce/elastic-bekitzur/service-account/package-storage-uploader`) - use it to kick off
+   the publishing job to process the uploaded candidate.
```
jsoriano, line 57: Could we provide a minimal pipeline as example of what would be needed in a project to publish a package?

This could help to identify if we can provide a one/two-liner helper that could be part of a jenkins library (or https://github.com/elastic/apm-pipeline-library/).

For example the functions of the package-storage jenkinsfile linked below could be parameterized and moved to a shared library.

jsoriano (commented):
Thanks for adding these docs!

I wonder if we could make it easier to introduce this in repositories with packages, by providing helper functions in a shared jenkins library.

docs/howto/use_package_storage_v2.md
v1v: Good question, there are three important benefits while using a shared library:
1) Easy to test it with UTs.
2) Easy to generalise therefore others could benefit from.
3) Simplify the pipeline logic.

On the other hand, it might require backward compatibility if a new change is added, since you might not know what other consumers use it.

mtojek: Fortunately, it calls two APIs, signing and publishing, so I don't expect it to be super complex. At least for now.

mtojek: PR to define a minimal publishing pipeline: https://github.com/elastic/elastic-package/pull/729

mtojek: PR to implement the minimal publishing pipeline: https://github.com/elastic/elastic-package/pull/731

mtojek:
I adjusted the .md file to refer to the minimal pipeline. I will create a follow-up issue to add a `packageStoragePublish` function to the [apm-pipeline-library](https://github.com/elastic/apm-pipeline-library).

jsoriano (approved):
Nice, thanks!

jlind23:
Congratulations @jsoriano @mtojek on finishing the first Milestone 👍🏼

PR #729, opened 2022-03-15
Define minimal publishing pipeline

mtojek:
Related: https://github.com/elastic/elastic-package/pull/728

This PR adds a definition for a minimal publishing pipeline. I will implement the groovy files in follow-ups.

### DIFF
--- a/.ci/jobs/package-storage-publish.yml
+++ b/.ci/jobs/package-storage-publish.yml
@@ -1,0 +1,41 @@
+---
+- job:
+    name: Ingest-manager/elastic-package/package-storage-publish
+    display-name: Publish a package to Package Storage v2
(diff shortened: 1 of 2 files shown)


jsoriano:
Is this pipeline going to be installed? I was thinking on a reference pipeline for documentation purpouses :slightly_smiling_face:

mtojek:
Yeah, but this way it can be tested, so I thing it's still worth connecting it :)

jsoriano:
> Yeah, but this way it can be tested, so I think it's still worth connecting it :)

Ok, but will this trigger the publication of any package? Or this job won't be ever triggered?

jsoriano (approved):

PR #731, opened 2022-03-16
Implement minimal pipeline to publish a package

mtojek:
Related: https://github.com/elastic/elastic-package/pull/728

This PR implements the minimal pipeline required to publish a package to the Package Storage v2. I extracted the publishing part from the main Jenkinsfile and put it in a separate mini-pipeline. We can discuss if/what we should we push to the apm-pipeline-library.

### DIFF
--- a/.ci/package-storage-publish.groovy
+++ b/.ci/package-storage-publish.groovy
@@ -53,7 +53,7 @@ pipeline {
         cleanup()
         useElasticPackage()
         dir("${BASE_DIR}/test/packages/package_storage_candidate") {
-          sh(label: 'Build package',script: "elastic-package build")
+          sh(label: 'Build package',script: "../../../elastic-package build")
         }
         stash(allowEmpty: true, name: 'build-package', useDefaultExcludes: false)
       }
--- a/.ci/package-storage-publish.groovy
+++ b/.ci/package-storage-publish.groovy
@@ -85,19 +85,28 @@ def packageStoragePublish(builtPackagesPath) {
 }
 
 def signUnpublishedArtifactsWithElastic(builtPackagesPath) {
-  findFiles()?.findAll{ it.name.endsWith('.zip') }?.collect{ it.name }?.sort()?.each {
-    def packageZip = it
-    if (isAlreadyPublished(packageZip)) {
-      return
-    }
+  def unpublished = false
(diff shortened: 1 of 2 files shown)


.ci/package-storage-publish.groovy
```diff
@@ -31,7 +45,30 @@ pipeline {
...
+          }
+        }
+        stash(allowEmpty: true, name: 'build-package', includes: "${BASE_DIR}/build/integrations/*.zip", useDefaultExcludes: false)
+      }
+    }
+    stage('Sign and publish package') {
+      steps {
+        cleanup(source: 'build-package')
+        dir("${BASE_DIR}") {
+          packageStoragePublish('build/integrations')
```
jsoriano, line 70: I'd vote for moving this to a shared library, so signing and publishing in any repository is basically a one line :)
We would need to see how to pass all the parameters in case environment is not recommended.

.ci/package-storage-publish.groovy
```diff
@@ -31,7 +45,30 @@ pipeline {
...
+        stash(allowEmpty: true, name: 'source', useDefaultExcludes: false)
+      }
+    }
+    stage('Build package') {
+      steps {
+        cleanup()
+        withGoEnv() {
+          dir("${BASE_DIR}") {
+            sh(label: 'Install elastic-package',script: "make install")
+            // sh(label: 'Install elastic-package', script: 'go build github.com/elastic/elastic-package')
```
jsoriano, line 57: Is this here for reference?

jsoriano (approved):
Nice, quite minimal pipeline at the end. Thanks for adding it!

.ci/package-storage-publish.groovy
mtojek: Yes, as for anyone outside of the elastic-package most likely it would be the other command.

.ci/package-storage-publish.groovy
mtojek: I guess that it would be a big input map with all properties :) I'm ok with moving it there, not sure if I should do this now or later. WDYT?

I guess it's a matter of priorities: indexing & package registry implementation vs. support in apm-pipeline-library

jsoriano: Yes, do it later, no need to do it now.

PR #721, opened 2022-03-07
test j2xunit

efd6:
TEST DO NOT MERGE

PR #740, opened 2022-03-18
Use gotestsum as test runner

jsoriano:
@efd6 found that current approach of parsing the output of `go test` can be problematic in some cases, as when running parallel tests.

`gotestsum` wraps `go test --json`, that is much less error prone that parsing plain `go test` output. It supports different output formats, including junit. This is also the approach [used in Beats](https://github.com/elastic/beats/blob/237937085a5a7337ba06f1268cfc55cd4b869e31/dev-tools/mage/gotest.go#L199).

Thanks @efd6 for raising the issue, and for trying `gotestsum` in https://github.com/elastic/elastic-package/pull/721.


mtojek (approved):
LGTM

PR #742, opened 2022-03-18
Optimize fields validation

jsoriano:
While trying some changes in pipeline tests I found that testing some packages was taking much more time than I would have expected.

I have enabled profiling and I have found that most of the time, `elastic-package` is compiling regular expressions.

Refactoring the code to don't use regular expressions drastically improves the execution times. For example in the case of the `panw` module it goes in my laptop from more than 3 minutes to less than 10 seconds.


internal/fields/validate.go
```diff
@@ -302,6 +302,21 @@ func FindElementDefinition(searchedKey string, fieldDefinitions []FieldDefinitio
 	return findElementDefinitionForRoot("", searchedKey, fieldDefinitions)
 }
 
+var compileCache map[string]*regexp.Regexp
+
+func init() {
+	compileCache = make(map[string]*regexp.Regexp)
+}
+
+func mustCompile(str string) *regexp.Regexp {
```
mtojek, line 311: nit: in elastic-package, we used to place such hacks or extensions in corresponding packages with -ext suffix, so `regexpext`. You can consider applying it here too.

mtojek, line 311: nit: `mustCompileCached`?

mtojek (approved):
I admit that I didn't expect it to happen. Good catch!

internal/fields/validate.go
```diff
@@ -302,6 +302,21 @@ func FindElementDefinition(searchedKey string, fieldDefinitions []FieldDefinitio
 	return findElementDefinitionForRoot("", searchedKey, fieldDefinitions)
 }
 
+var compileCache map[string]*regexp.Regexp
+
+func init() {
+	compileCache = make(map[string]*regexp.Regexp)
+}
```
efd6, line 309:
```suggestion
var compileCache = make(map[string]*regexp.Regexp)
```

jsoriano:
I have refactored the code to don't need regular expressions, I see similar improvements in time than using a regular expressions cache.

jsoriano:
I found profiling that `strings.Split` was also expensive, I have further refactored to do char by char comparison, this reduced the testing time of the `panw` module to less than 6 seconds (~3 mins original code, ~10 seconds with regexp cache or with first iteration without regexps, ~6 seconds with second iteration).

Shared profile with current code:
[cpu.profile.gz](https://github.com/elastic/elastic-package/files/8317475/cpu.profile.gz)

internal/fields/validate.go
```diff
@@ -303,34 +303,36 @@ func FindElementDefinition(searchedKey string, fieldDefinitions []FieldDefinitio
...
-	}
-	for i, part := range keyParts {
-		if part == "*" {
+	var i int
+	var j int
+	for i = 0; i < len(key); i++ {
+		if j >= len(searchedKey) {
+			return false
+		}
+		if key[i] == searchedKey[j] {
```
efd6, line 312: You could potentially save some bounds checks and syntax by making the outer loop `for i, k := range []byte(key) {` and using `k` in place of `key[i]` (note that the `[]byte` conversion does not allocate in this context see https://godbolt.org/z/Mhv7Kv9fq).

internal/fields/validate.go
```diff
@@ -303,34 +303,36 @@ func FindElementDefinition(searchedKey string, fieldDefinitions []FieldDefinitio
...
 			continue
 		}
-		if part == searchedParts[i] {
+		if key[i] == '*' {
+			for ; j < len(searchedKey); j++ {
+				if searchedKey[j] == '.' {
+					j--
+					break
+				}
+			}
```
efd6, line 322: Maybe:
```
if idx := strings.Index(searchedKey[j:], "."); idx != -1 {
    j += idx -1
}
```

internal/fields/validate.go
```diff
@@ -302,22 +302,49 @@ func FindElementDefinition(searchedKey string, fieldDefinitions []FieldDefinitio
 	return findElementDefinitionForRoot("", searchedKey, fieldDefinitions)
 }
 
+// compareKeys checks if `searchedKey` matches with the given `key`. `key` can contain
+// wildcards (`*`), that match any sequence of characters in `searchedKey` different to dots.
 func compareKeys(key string, def FieldDefinition, searchedKey string) bool {
-	k := strings.ReplaceAll(key, ".", "\\.")
-	k = strings.ReplaceAll(k, "*", "[^.]+")
+	// Loop over every byte in `key` to find if there is a matching byte in `searchedKey`.
```
mtojek, line 308: Do you think that we need some unit tests as it looks like a custom logic?

mtojek (approved):
Nice optimization, although the code readability might be impaired a bit. Do you think that you can get rid of "regexp" dependency here at all?

Before merging this PR, I suggest releasing a new minor version, so we can easily roll back if this PR becomes problematic.

internal/fields/validate.go
jsoriano: I thought that we already had good coverage for this, but adding a unit test has helped me found some wrong cases with wildcards. Good idea, thanks :+1:

jsoriano:
> Nice optimization, although the code readability might be impaired a bit.

Well, this is relative, others may also find counterintuitive to build regular expressions on the fly :slightly_smiling_face: In any case the code is quite localized inside a single function with well-defined expectations. I have added comments in case it helps.

> Do you think that you can get rid of "regexp" dependency here at all?

The other case where we use `regexp` is to validate patterns defined in the field definition, we need `regexp` compilation there. We could consider caching these regular expressions if they are a problem, but I think they are used only once most of the times, I haven't found these compilations while profiling.
We may consider optimizing this in a future PR if we find this is problematic.

> Before merging this PR, I suggest releasing a new minor version, so we can easily roll back if this PR becomes problematic.

:+1:

mtojek (approved):

PR #743, opened 2022-03-18
Compare test results using json diff

jsoriano:
Use json diff to compare results of pipeline tests. This avoids failures caused by different format of equal numbers, as was the case in https://github.com/elastic/integrations/pull/2845.

After this change two json documents are equal if their values are, independently of their format.

The biggest UX change is that this changes the diff to something like this:
```
panw/panos test-panw-panos-globalprotect-sample.log:
{
    "expected": [
        ...skipped 2 array elements...,
        {
            ...skipped 7 object properties...,
            "panw": {
                "panos": {
                    ...skipped 5 object properties...,
                    "sequence_number": 6920071768563516000 => 6920071768563516860,
                    ...skipped 6 object properties...
                }
            },
            ...skipped 3 object properties...
        },
        ...skipped 1 array element...,
        {
            ...skipped 7 object properties...,
            "panw": {
                "panos": {
                    ...skipped 5 object properties...,
                    "sequence_number": 6920071768563516000 => 6920071768563516847,
                    ...skipped 6 object properties...
                }
            },
            ...skipped 3 object properties...
        },
        ...
```

The library used has more [options](https://pkg.go.dev/github.com/nsf/jsondiff#Options) to customize the diff format.

We could also decide to show the previous diff if this one detects differences.


mtojek (approved):

mtojek (approved):
Maybe it's worth adding those problematic tests here?

jsoriano:
> Maybe it's worth adding those problematic tests here?

Do you mean to add the problematic packages as test cases here?

mtojek:
Yes, the ones that failed in the Integrations repository.

jsoriano:
> Yes, the ones that failed in the Integrations repository.

@mtojek I have added a simple sample package that reproduces the issues found in https://github.com/elastic/integrations/pull/2845. This will allow us to easily add test cases with numbers in documents.

andrewkroh (approved):
👍 The more concise output makes it easier to spot the differences.

kaiyan-sheng (approved):
+1 on this change to make it easier to compare! Thanks!

P1llus (approved):
Looks awesome! :)

internal/testrunner/runners/pipeline/test_result.go
```diff
@@ -70,87 +65,62 @@ func compareResults(testCasePath string, config *testConfig, result *testResult)
...
-	return b
-}
+func diffJson(want, got []byte) (string, error) {
+	opts := jsondiff.DefaultConsoleOptions()
+
+	// Remove colored output.
+	opts.Added = jsondiff.Tag{}
+	opts.Removed = jsondiff.Tag{}
+	opts.Changed = jsondiff.Tag{}
+	opts.Skipped = jsondiff.Tag{}
```
efd6, line 112: This loses information about the change when it's rendered to the page. Can we have some indication in the output about the kind of diff?

jsoriano: Umm, to rendering on what page?
These tags are used to add formatting, like colors, or html tags if rendered in a website, but without them it should be enough to see what is wrong. I am setting them to the zero values to avoid having colors in CI or other non-ttys, that are not going to render them properly.
Could you give an example of the problematic output and what would you expect?

efd6: Some sigil indicating the direction of the change would be fine. I've tried to find the PR I was looking at today that made this an issue, but can't find it. Looking at the diff, I was not able to tell what was happening. An indication of whether the diffs were additions/deletions/changes would have solved that.

jsoriano: Umm, I think that changes are clear, they are indicated as in the example in the description:
```
                    "sequence_number": 6920071768563516000 => 6920071768563516860,
```
But it is true that current diff doesn't distinguish between additions and deletions, and that may be confusing.

On the other hand, from the shared stacktrace it looks clear that there is some kind of problem with `host.hostname`.

I have opened a PR to add +/- tags before additions and deletions, let me know if you prefer this format: https://github.com/elastic/elastic-package/pull/770

efd6: Thank you for doing this. While I was looking into representing the problem better by code I found an issue with the underlying package; it does not give a minimal diff. This can be seen here: https://go.dev/play/p/lTcn7sGwdnv. The upshot is that there is no attempt to map corresponding elements within a list, which is unfortunate since all the outputs that it needs to give information on are lists.

jsoriano: Umm, yeah, not very nice in this case. Let's see if we find more problems with this package, we can try others if this one doesn't work well at the end.

efd6: Thanks

PR #748, opened 2022-03-22
Handle string ports in docker compose v2 config yaml

jsoriano:
Using docker-compose v2.3.x (go implementation), `elastic-package shellinit` fails with:
```
Error: shellinit failed: could not get Docker Compose configuration for service: could not unmarshal YAML map node: yaml: unmarshal errors:
  line 4: cannot unmarshal !!str `9200` into int
```

This seems to happen because some ports are represented as strings in the `docker-compose config` output.

Support both versions by allowing ports to be integers or numbers.


internal/compose/compose.go
```diff
@@ -68,17 +82,17 @@ func (p *portMapping) UnmarshalYAML(node *yaml.Node) error {
 
 		var s struct {
 			HostIP    string `yaml:"host_ip"`
-			Target    int
-			Published int
+			Target    intOrStringYaml
```
mtojek, line 85: If we start observing such problems, then I guess we'll end up soon with code switches depending on v1/v2.  Let's drop a comment here, why do we need to handle it this way.

mtojek (commented):
Is this the moment, when we should use Compose V2 in general for CI testing?

jsoriano:
> Is this the moment, when we should use Compose V2 in general for CI testing?

I think we are close, the main blocker I have found so far with v2 has been not being able to use `elastic-package shellinit`, what would be solved by this PR.
Apart of that, installation of packages during tests fail on my laptop, as if docker compose is not correctly recreating the package registry after packages are built. I would have to investigate this a bit more.

Something we might also investigate is to use the [Go API](https://github.com/docker/compose/blob/be187bae649521f5776033284a511da0fbaf8058/pkg/api) :slightly_smiling_face:, so we don't depend on the installed binary.

jsoriano:
Actually this may be a bug in docker compose. This is the [documented format](https://docs.docker.com/compose/cli-command-compatibility/):
```
    - mode: ingress
      target: 80
      published: 80
      protocol: tcp
```
This is what I am finding:
```
```
    - mode: ingress
      host_ip: 127.0.0.1
      target: 9200
      published: "9200"
      protocol: tcp
```
```

jsoriano:
Open issue in docker compose project: https://github.com/docker/compose/issues/9306

mtojek (approved):

PR #752, opened 2022-03-23
Import aditional indexing settings on external fields

### LINKED ISSUE #678: External ECS field imports not containing their multi-fields
### Summary

The expected multi-fields are not imported when [external ECS fields](https://github.com/elastic/elastic-package/blob/main/docs/howto/dependency_management.md#external-fields) are declared. I suspect this is the cause of some integrations not including the anticipated multi-fields in their index templates.

I first experienced on 7.16.3 using the Windows 1.5.0 integration. The `logs-windows.sysmon_operational` managed index template is missing `process.command_line.text`.

### Example

Declaring an external field with `process.command_line`:

```yml
- name: process.command_line
  external: ecs
```

produces an index mapping of:

```json
    "process": {
      "properties": {
            "command_line": {
              "type": "wildcard"
            }
          }
        }
```

In ECS, [`process.command_line`](https://www.elastic.co/guide/en/ecs/current/ecs-process.html#field-process-command-line) contains the `.text` multi-field.

The expected field mapping:

```json
    "process": {
      "properties": {
            "command_line": {
              "type": "wildcard",
              "fields": {
                "text": {
                  "type": "match_only_text"
                }
              }
            }
        }
```

jsoriano:
Fields imported from ECS may contain some settings that may affect how they are indexed:
* `multi_fields` allow to index the same field in different ways for different purposes.
* `index: false` disables indexing of the field.
* `doc_values: false` disables storing some internal data used when sorting and by some aggregations, this can help to save disk space with the cost of losing some functionality for those fields.

Fixes #678.


jsoriano:
I am going to move this back to draft, we may need to confirm how/if this is supported by fleet (ping @joshdover).

I have installed the modified test package, that includes `event.original` and `process.command_line` fields and the installed mappings are these ones:
```
    "event": {
      "properties": {
        "original": {
          "ignore_above": 1024,
          "index": false,
          "type": "keyword"
        }
```
```
    "process": {
       ...
        "command_line": {
          "ignore_above": 1024,
          "type": "wildcard",
          "fields": {}
        }
```
- `index` is correctly set :heavy_check_mark:
- `doc_fields` is not set (but maybe because it is redundant when `index: false` is also set?).
- `multi_fields` are not set, but interestingly `process.command_line` has an empty `fields`, should we be using a different format for multi fields?

FTR, imported definitions of these fields are these ones:
```
- description: |-
    Full command line that started the process, including the absolute path to the executable, and all ar
guments.
    Some arguments may be filtered to protect sensitive information.
  multi_fields:
    - name: text
      type: match_only_text
  name: process.command_line
  type: wildcard
```
```
- description: |-
    Raw text message of entire event. Used to demonstrate log integrity or where the full log message (be
fore splitting it up in multiple parts) may be required, e.g. for reindex.
    This field is not indexed and doc_values are disabled. It cannot be searched, but it can be retrieved
 from `_source`. If users wish to override this and index this field, please see `Field data types` in th
e `Elasticsearch Reference`.
  doc_values: false
  index: false
  name: event.original
  type: keyword
```

joshdover:
@juliaElastic I think you were recently in this mapping code for adding the dimensions settings. Would you be able to take a look and answer the questions above?

juliaElastic:
@jsoriano 
Regarding `doc_values`, it looks like fleet is not setting `doc_values` in mappings if `doc_values` is false, we could update it to work like `index` field:
https://github.com/elastic/kibana/blob/5903c417403b55418f74c3b9bfbcba07b12d402e/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L302-L307

For `multi_fields`, it looks like the type `match_only_text` is not covered by fleet, this can be updated as well:
https://github.com/elastic/kibana/blob/5903c417403b55418f74c3b9bfbcba07b12d402e/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L234-L255

jsoriano:
Thanks @juliaElastic, I think this explains what we are seeing.

> Regarding `doc_values`, it looks like fleet is not setting `doc_values` in mappings if `doc_values` is false, we could update it to work like `index` field:
> https://github.com/elastic/kibana/blob/5903c417403b55418f74c3b9bfbcba07b12d402e/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L302-L307

Ah yes, makes sense, I think it'd be good to have the same logic for `doc_values` in any case.

> For `multi_fields`, it looks like the type `match_only_text` is not covered by fleet, this can be updated as well:
> https://github.com/elastic/kibana/blob/5903c417403b55418f74c3b9bfbcba07b12d402e/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L234-L255

Ah ok, we recently had to add support for this one also in the package-spec (https://github.com/elastic/package-spec/pull/284). It'd be good to have support in fleet too. I will test this with a field of other type.

Could you take care of adding the missing pieces to Fleet?

jsoriano:
> is that okay to have `"ignore_above": 1024` as well in the resulting mappings?

I think so, I guess this is a default configured by fleet?

juliaElastic:
> > is that okay to have `"ignore_above": 1024` as well in the resulting mappings?
> 
> I think so, I guess this is a default configured by fleet?

yes

jsoriano:
Issues mentioned related to kibana handling of these settings have been solved in https://github.com/elastic/kibana/pull/128391 (thanks Julia!), so I am reopening this for review.

`doc_values` and `multi_fields` won't work for users using older versions of Kibana, but they don't seem to produce any issue, they are just not installed as they aren't now.

internal/fields/model.go
```diff
@@ -64,4 +75,40 @@ func (orig *FieldDefinition) Update(fd FieldDefinition) {
...
+				updatedFields = append(updatedFields, newField)
+			}
+		}
+		orig.MultiFields = updatedFields
+	}
+}
+
+// MultiFieldDefinition describes a multi field with its properties.
+type MultiFieldDefinition struct {
+	Name string `yaml:"name"`
```
mtojek, line 103: Multi-fields don't have other properties, right?

internal/fields/model.go
```diff
@@ -42,6 +45,14 @@ func (orig *FieldDefinition) Update(fd FieldDefinition) {
 	if fd.External != "" {
 		orig.External = fd.External
 	}
+	if fd.Index != nil {
+		v := *fd.Index
```
mtojek, line 49: Do we need to create a copy of `fd.Index`? The `fd FieldDefinition` is passed by value to `Update`.

internal/fields/model.go
```diff
@@ -64,4 +75,40 @@ func (orig *FieldDefinition) Update(fd FieldDefinition) {
 		}
 		orig.Fields = updatedFields
 	}
+
+	if len(fd.MultiFields) > 0 {
```
mtojek, line 79: Maybe extract the entire clause to a separate function?

internal/fields/model.go
jsoriano: Good point, I will change this.

internal/fields/model.go
mtojek: I'm wondering who can evaluate this. @ruflin Do you have any idea?

internal/fields/model.go
jsoriano: Done, we can consider using generics there when we update to 1.18 :) The updater for fields and for multi fields is basically the same.

internal/fields/model.go
jsoriano: Let's analyze what these models are used for:
* To import fields from ECS: Only `type` and `name` is defined there for multi fields. @ebeahan do you know if there are plans to add more properties to multi fields in ECS?
* To generate documentation: Should we document somehow these multi fields? Maybe as part of their parent fields? In any case we wouldn't need more properties beyond its name.
* To validate documented fields: Probably we don't need more properties, but maybe we need to add validation for multi fields if they start appearing in test documents.

jsoriano: > * To validate documented fields: Probably we don't need more properties, but maybe we need to add validation for multi fields if they start appearing in test documents.

Fixed in 633f703e6e16a943203ff059df189ec955aee454.

internal/fields/dependency_manager.go
```diff
@@ -221,5 +229,21 @@ func transformImportedField(fd FieldDefinition) common.MapStr {
...
+			i := transformImportedMultiField(f)
+			t = append(t, i)
+		}
+		m.Put("multi_fields", t)
+	}
 	return m
 }
+
+func transformImportedMultiField(fd MultiFieldDefinition) common.MapStr {
+	return common.MapStr{
```
mtojek, line 245: nit: double-checking, it isn't possible that `multi_fields` have children, right :D ?

internal/fields/dependency_manager.go
```diff
@@ -221,5 +229,21 @@ func transformImportedField(fd FieldDefinition) common.MapStr {
 		}
 		m.Put("fields", t)
 	}
+
+	if len(fd.MultiFields) > 0 {
+		var t []common.MapStr
+		for _, f := range fd.MultiFields {
+			i := transformImportedMultiField(f)
```
mtojek, line 236: nit: an alternative implementation - you can introduce a new interface `IFieldDefinition` with `GetName`, `GetDescription`, `GetType`, ..., so that you can unify the transformation - single function `transformImportedField`.

mtojek (commented):
I left a few nit-picks, but in general, I'm wondering if, from a Go code perspective,`multi_fields` shouldn't be treated the same way as standard fields. Maybe the implementation logic will be simpler then, as it's just another `fields` field to visit in terms of validation. Once we open the gate for generics, we could improve it.

Have you considered this option?

internal/fields/model.go
ruflin: @mtojek What you mean by evaluate here? You mean how much we should support?

mtojek: Yeah, I mean which prorities are the high priority, which ones can be postponed for later, and which ones can be "forgotten".

ruflin: I would say the ones that exist already in integrations today we need to support, everything else we should do based on request.

jsoriano:
> I left a few nit-picks, but in general, I'm wondering if, from a Go code perspective,`multi_fields` shouldn't be treated the same way as standard fields. Maybe the implementation logic will be simpler then, as it's just another `fields` field to visit in terms of validation. Once we open the gate for generics, we could improve it.
> 
> Have you considered this option?

I considered this, but I found that there are several fields that are not common, e.g. a multi field doesn't have a description, or other sub-fields. But maybe we can assume these differences, I can give it another try.

internal/fields/dependency_manager.go
jsoriano: I don't think so :)

internal/fields/model.go
jsoriano: > * To generate documentation: Should we document somehow these multi fields? Maybe as part of their parent fields? In any case we wouldn't need more properties beyond its name.

Added multi-fields to documentation in https://github.com/elastic/elastic-package/pull/752/commits/fbb0809f1b91310c73bde8c658ea901d895d98d3, let me know what you think.

This will require to regenerate READMEs in packages using multi-fields.

Issue #615, opened 2021-12-08
Field validator should check for ECS fields marked as arrays


andrewkroh:
It would be helpful if elastic-package could trigger errors on ECS event fields that are not arrays. For example, [host.ip](https://www.elastic.co/guide/en/ecs/current/ecs-host.html#field-host-ip) is specified to be an array so events that contain the field should be an array of ips.

<img width="745" alt="Screen Shot 2021-12-08 at 8 49 04 AM" src="https://user-images.githubusercontent.com/4565752/145219412-387a58ea-a250-494e-9d15-de5d35623809.png">

The generated YAML specification indicates the array type. https://github.com/elastic/ecs/blob/9b463c9116eb594a3be10db01c3236fd6b44b996/generated/ecs/ecs_flat.yml#L4536-L4543

```yaml
host.ip:
  dashed_name: host-ip
  description: Host ip addresses.
  flat_name: host.ip
  level: core
  name: ip
  normalize:
  - array
  short: Host ip addresses.
  type: ip
```

andrewkroh:
Another validation that would be nice to apply is for the `allowed_values` listed in the schema for fields like `event.{kind,type,category}`.

https://github.com/elastic/ecs/blob/4cbe6f7f9cb78160153659802eb66dd135fb4677/schemas/event.yml#L75-L76

andrewkroh:
`pattern` has now been added to the ECS schema. See https://github.com/elastic/ecs/issues/1741 https://github.com/elastic/ecs/pull/1834.

So the `elastic-package` tool could now check data against the `pattern` defined in ECS for a given field.

jsoriano:
Hey @andrewkroh,

I am implementing the check for arrays in https://github.com/elastic/elastic-package/pull/765, and I have found that pipeline tests of several packages in the integrations repository (35 out of 140) would fail if we enable this. So I wonder if we should do it.

What would be the advantage of checking that fields are formatted as arrays? In principle for Elasticsearch it is the same.

andrewkroh:
> What would be the advantage of checking that fields are formatted as arrays? In principle for Elasticsearch it is the same.

You're right, for Elasticsearch indexing it doesn't make a difference whether the field is a scalar or an array. ECS added an array indicator to the schema primarily for things other than Elasticsearch that might be producing or consuming the data.

[...]

PR #765, opened 2022-03-29
Validate expected fields normalization

jsoriano:
ECS fields can include normalization rules, indicated by the `normalize` parameter.

By now only the `array` rule exist, that indicates that values should be included in the
field as an array, even if there is a single element.

Validate this on test results for packages with `format_version` >= `2.0.0`.

Things to check in this review:
* This new validation is only enabled for packages using package spec v2. This is the first time we do something like that in `elastic-package`.
* Field validators can be configured now with the spec version of the package, this is done for all test runners that validate fields.
* Apache test package has been upgraded to version 2.0.0 of the spec. This illustrates the kind of changes that package developers will need to do when updating to this major. In this case it was:
  * Modify the pipelines to change `event.category` and `event.type` to arrays as for this PR.
  * Fix the license fields as for https://github.com/elastic/package-spec/pull/355.
* ECS has been upgraded in the Apache test package to a version that has normalization rules. 

Part of https://github.com/elastic/elastic-package/issues/615 and https://github.com/elastic/package-spec/issues/399.


jsoriano:
`normalize` is not included in the file we import, we need to have these fields included in this file, or start using one of the generic ecs files (https://github.com/elastic/elastic-package/pull/766).

jsoriano:
Blocked till we have some way of disabling checks per package.

mrodm (approved):
:+1:

Issue #750, opened 2022-03-23
[lint] fail when referencing ECS fields outside of allowed field reuse locations


andrewkroh:
I found a number of Fleet integrations referencing ECS fields that should not exist. For example, the `as` (autonomous system) field set should not be used at the top-level. The only valid "reuse" locations are specified in the schema. **I think it should be treated as an error to reference an ECS field outside of it's prescribed field reuse locations.**

From the [ECS docs](https://www.elastic.co/guide/en/ecs/current/ecs-as.html#_field_reuse):

<img width="718" alt="Screen Shot 2022-03-22 at 22 06 18" src="https://user-images.githubusercontent.com/4565752/159608827-ec3b857b-5255-41dc-8882-b361bc8d0eae.png">

The generated `ecs_flat.yml` already takes field reuse locations into account so this could be used when resolving the external ECS definitions to simplify logic in elastic-package.

Linked below are the fields I found that don't exist in ECS (note that this is checked against ECS 8.2 and some of these integrations are building against earlier ECS versions, so there are some false positives for fields like `process.ppid`).

https://gist.github.com/andrewkroh/35000805a649a4137e59251bc712b0e0#file-invalid_ecs_fields-md

jsoriano:
I think this check can have some problems:
* A package could be legally using a field that is not in ECS, this field is later added to ECS as a reusable field. Then the package must continue using an old version of ECS, or introduce a breaking change, to avoid failing this check.
* ECS fields are quite generic, I can imagine for example that a service may have its own `user` or `file` data (e.g. `mysql.user`, `mysql.database.innodb.file.name`...). Or the proposal would be to fail only on ECS fields reused at the root level or other ECS fields?

andrewkroh:
I tested it, and thankfully no it is not.

```
- name: mysql
  type: group
  release: beta
  fields:
    - name: user
      external: ecs
```

```
error calling fields: collecting fields files failed: visiting fields failed: recursive visiting fields failed (namePrefix: ): can't import field: field definition not found in schema (name: mysql.user)
```

So that reduces the problem to restricting top-level field reuse to those that are defined in the schema. 👍

PR #766, opened 2022-03-29
Import ecs fields from the ecs_nested.yml source file

jsoriano:
Import ECS fields from `ecs/ecs_nested.yml` instead of from `beats/fields.ecs.yml`.

Beats fields don't include all the information provided by ECS, for example they don't include the `normalize` rules, required for https://github.com/elastic/elastic-package/issues/615. I guess this can be expected, as there can be features of ECS that are not used or needed in Beats, but could be used by other tools.

`ecs_nested.yml` takes into account where nested objects can be reused. This fixes https://github.com/elastic/elastic-package/issues/750.


test/packages/parallel/gcp/data_stream/compute/fields/ecs.yml
```diff
@@ -1,5 +1,3 @@
-- external: ecs
-  name: cloud
```
jsoriano, line 2: Not sure about how we want to support this, I see two options:
* As it is in current `main` branch, importing the definition of the group, that is used only for documentation. We would need to use `ecs_nested.yml` instead of `ecs_flat.yml` for this in this PR.
  This may be error prone, as it may look like if someone can use the `cloud` field.
* Importing the whole object, making easier to import groups of ECS fields.
* Don't import anything, and require explicit imports, as is done by now in this PR.

jsoriano: > * As it is in current `main` branch, importing the definition of the group, that is used only for documentation. We would need to use `ecs_nested.yml` instead of `ecs_flat.yml` for this in this PR.

I am going to do this, otherwise this is going to be breaking for many packages.

mtojek (approved):
LGTM, but to be honest we should get rid of this script as soon as possible. I don't know what is the percentage of migration completeness.

jsoriano:
> LGTM, but to be honest we should get rid of this script as soon as possible. I don't know what is the percentage of migration completeness.

@mtojek take into account that this is for the fields imported for the `external: ecs` property, not for the information imported from beats.

The mention to beats is because before this change we are using the ecs fields document generated for beats (like [this one](https://raw.githubusercontent.com/elastic/ecs/8.2/generated/beats/fields.ecs.yml)), and after this change we would be using one of the generic ones (like [this one](https://raw.githubusercontent.com/elastic/ecs/8.2/generated/ecs/ecs_nested.yml)). This way we can import anything from the ECS definitions, and not only what is relevant to Beats.

I guess we don't want to get rid of `external: ecs` :slightly_smiling_face:

internal/fields/model.go
```diff
@@ -82,3 +89,67 @@ func updateFields(origFields, fields []FieldDefinition) []FieldDefinition {
...
+		err := value.Decode(&fields)
+		if err != nil {
+			return err
+		}
+		*fds = fields
+		return nil
+	case yaml.MappingNode:
+		// Fields are defined as a map, this happens in ecs fields files.
+		var fields []FieldDefinition
+		for i := 0; i < len(value.Content); i += 2 {
```
mtojek, line 112: Is it safe to assume it's always a pair? Is there a chance that it will become a triple :)? a comment, or second value...

mtojek (approved):

internal/fields/model.go
jsoriano: In principle yes, the content of a mapping node should be pairs of key/values. Comments are included in the significative nodes themselves.

But let me add an additional constraint here to avoid reading the `i+1` value when it doesn't exist.

mtojek (approved):

PR #767, opened 2022-03-30
Enable the Cloud Security Posture Kibana plugin

eyalkraft:
Enable the Cloud Security Posture Kibana plugin by default when using elastic-package.

- Relates https://github.com/elastic/integrations/pull/2930


mtojek (commented):
Could you please fill in the issue description and link to the original issue?

Please take a look at the build artifacts. It looks like Kibana doesn't support it:

```
Attaching to elastic-package-stack_kibana_1
[36mkibana_1                     |[0m [2022-03-30T14:42:56.139+00:00][INFO ][plugins-service] Plugin "metricsEntities" is disabled.
[36mkibana_1                     |[0m [2022-03-30T14:42:56.207+00:00][FATAL][root] Error: Unknown configuration key(s): "xpack.cloudSecurityPosture.enabled". Check for spelling errors and ensure that expected plugins are installed.
[36mkibana_1                     |[0m     at ensureValidConfiguration (/usr/share/kibana/src/core/server/config/ensure_valid_configuration.js:35:11)
[36mkibana_1                     |[0m     at Server.preboot (/usr/share/kibana/src/core/server/server.js:164:5)
[36mkibana_1                     |[0m     at Root.preboot (/usr/share/kibana/src/core/server/root/index.js:48:14)
[36mkibana_1                     |[0m     at bootstrap (/usr/share/kibana/src/core/server/bootstrap.js:99:9)
[36mkibana_1                     |[0m     at Command.<anonymous> (/usr/share/kibana/src/cli/serve/serve.js:216:5)
[36mkibana_1                     |[0m 
[36mkibana_1                     |[0m  FATAL  Error: Unknown configuration key(s): "xpack.cloudSecurityPosture.enabled". Check for spelling errors and ensure that expected plugins are installed.
[36mkibana_1                     |[0m 
```

Is it generally available in 8.x?

mtojek:
> Kibana 8.2-SNAPSHOT should contain the plugin I believe

I can confirm that 8.2 stack booted up without issues. 

That complicates things as we need to support another [Kibana config variant](https://github.com/elastic/elastic-package/blob/8d9732d0c4364d28ba1c34f2ad7fe91dfc7d2785/internal/stack/variants.go#L16). I guess we can help with this one. Sadly Kibana can't ignore unknown configuration properties.

Please tell me how urgent is this change as we have other stuff on the board?

eyalkraft:
Is this a blocker for merging the integration?
If so, it is urgent for us. We can open the PR to support the additional variant if you are short in time - so you will only need to review.

If this is not a blocker for merging the integration, We can work with elastic-package just fine configuring the `kibana.yml` in our own local profiles.
This flag is only required for installing our integration.

mtojek:
Then it's a blocker as correct installation of the integration is required as part of the validation process. If you have spare time to work on this, then please assign us as reviewers.

AFAIR here are the required steps:

1. Prepare an additional config file based on [kibana_config_8x.yml](https://github.com/elastic/elastic-package/blob/main/internal/profile/_static/kibana_config_8x.yml), for example `kibana_config_80.yml`.
2. Register the file in [here](https://github.com/elastic/elastic-package/blob/770c7bc97918edfecdfe3e55fb8bdce99bfa0099/internal/profile/static.go).
3. Add new entries to [managedProfileFiles](https://github.com/elastic/elastic-package/blob/770c7bc97918edfecdfe3e55fb8bdce99bfa0099/internal/profile/profile.go#L38). 
This map keeps tracking of all configs and it's used to create files in `~/.elastic-package/profiles`. Probably you will have to add virtual records for any non-existing 8.0 configs too. We don't want to duplicate config files for Elasticsearch and Elastic Agent in the source code. You can find all config variants [here](https://github.com/elastic/elastic-package/blob/770c7bc97918edfecdfe3e55fb8bdce99bfa0099/internal/profile/_static/docker-compose-stack.yml) by searching for `STACK_VERSION_VARIANT`.

As we can see, the CI failed already for many steps but succeeded for 8x, as of now it's 8.2.0-SNAPSHOT, which means that it can boot up correctly with the plugin. We just need to strip the entry from the other Kibana config.

mtojek (commented):
Left a small comment around `selectStackVersion`. Otherwise, it's ready for shipping. Well done!

internal/stack/variants.go
```diff
@@ -10,15 +10,19 @@ import (
...
 	return fmt.Sprintf("STACK_VERSION_VARIANT=%s", selectStackVersion(version))
 }
 
 func selectStackVersion(version string) string {
 	if strings.HasPrefix(version, "8.") {
+		if len(version) > 2 && (int(version[2])-'0') < 2 {
+			return "80"
+		}
 		return "8x"
 	}
```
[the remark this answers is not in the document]
eyalkraft: Nice! 
I wasn't too comfortable with this as well.

internal/stack/variants.go
```diff
@@ -6,20 +6,34 @@ package stack
 
 import (
 	"fmt"
-	"strings"
+
+	"github.com/Masterminds/semver/v3"
```
mtojek, line 10: just a small comment, I see you imported the "v3". Is there anything missing in the `github.com/Masterminds/semver` as we are using the base release? If not, would you mind switching to that version?

internal/stack/variants.go
```diff
@@ -6,20 +6,34 @@ package stack
...
+// * 8x, supports different configuration options in Kibana, covers stack versions 8.2.0+
 func stackVariantAsEnv(version string) string {
 	return fmt.Sprintf("STACK_VERSION_VARIANT=%s", selectStackVersion(version))
 }
 
 func selectStackVersion(version string) string {
-	if strings.HasPrefix(version, "8.") {
-		return "8x"
+	if v, err := semver.NewVersion(version); err == nil {
+		if checkVersion(v, "8.0-0 - 8.1-0") {
```
mtojek, line 24: 👍 

much clearer now!

internal/stack/variants.go
```diff
@@ -6,20 +6,34 @@ package stack
...
+// * 80, covers stack versions 8.0.0 to 8.1.x
+// * 8x, supports different configuration options in Kibana, covers stack versions 8.2.0+
 func stackVariantAsEnv(version string) string {
 	return fmt.Sprintf("STACK_VERSION_VARIANT=%s", selectStackVersion(version))
 }
 
 func selectStackVersion(version string) string {
-	if strings.HasPrefix(version, "8.") {
-		return "8x"
+	if v, err := semver.NewVersion(version); err == nil {
```
mtojek, line 23: Actually, you can convert these if-conditions into a `configurationVariantMap` (key: constraint, value: config variant).

internal/stack/variants.go
eyalkraft: Done.
I had some trouble finding the correct v1 syntax for the range constraint as it differs from the v3 a bit.

mtojek (approved):

PR #769, opened 2022-03-30
Add new changelog subcommand

jsoriano:
Add a new subcommand with utilities for the changelog.

It includes an `add` command that can be used to add changelog entries,
this can be useful in automation.

For example the following command creates a new patch version including a change with a description, link and type:
```
elastic-package changelog add --next patch --description "Add documentation for multi-fields" --type enhancement --link https://github.com/elastic/integrations/pull/2916
```

If no `--next` or `--version` flags are passed, change is included in current version.

### DIFF
--- a/cmd/changelog.go
+++ b/cmd/changelog.go
@@ -1,0 +1,211 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/Masterminds/semver"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	"github.com/elastic/elastic-package/internal/cobraext"
+	"github.com/elastic/elastic-package/internal/packages"
+	"github.com/elastic/elastic-package/internal/packages/changelog"
+)
+
+const changelogLongDescription = `Use this command to work with the changelog of the package.`
+
+const changelogAddLongDescription = `Use this command to add an entry to the changelog file.
+
+The entry is added on top of the last entry in the current version. Or optionally in the next
+major, minor or patch version, or as a new given version.`
+
+func setupChangelogCommand() *cobraext.Command {
+	addChangelogCmd := &cobra.Command{
+		Use:   "add",
+		Short: "Add an entry to the changelog file",
+		Long:  changelogAddLongDescription,
+		RunE:  changelogAddCmd,
+	}
+	addChangelogCmd.Flags().String(cobraext.ChangelogAddNextFlagName, "", cobraext.ChangelogAddNextFlagDescription)
+	addChangelogCmd.Flags().String(cobraext.ChangelogAddVersionFlagName, "", cobraext.ChangelogAddVersionFlagDescription)
+
+	addChangelogCmd.Flags().String(cobraext.ChangelogAddDescriptionFlagName, "", cobraext.ChangelogAddDescriptionFlagDescription)
+	addChangelogCmd.MarkFlagRequired(cobraext.ChangelogAddDescriptionFlagName)
+	addChangelogCmd.Flags().String(cobraext.ChangelogAddTypeFlagName, "", cobraext.ChangelogAddTypeFlagDescription)
+	addChangelogCmd.MarkFlagRequired(cobraext.ChangelogAddTypeFlagName)
+	addChangelogCmd.Flags().String(cobraext.ChangelogAddLinkFlagName, "", cobraext.ChangelogAddLinkFlagDescription)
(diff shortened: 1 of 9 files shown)


cmd/changelog.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/Masterminds/semver"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/elastic-package/internal/cobraext"
+	"github.com/elastic/elastic-package/internal/packages"
+	"github.com/elastic/elastic-package/internal/packages/changelog"
+)
+
+const changelogLongDescription = `Use this command to work with the changelog of the package.`
```
mtojek, line 20: nit:  You can say a few more words here about what to do with this command as it will land in the README.

cmd/changelog.go
```diff
@@ -0,0 +1,168 @@
...
+
+const changelogAddLongDescription = `Use this command to add an entry to the changelog file.
+
+The entry is added on top of the last entry in the current version. Or optionally in the next
```
mtojek, line 24: nit (comma): minor, or patch version

cmd/changelog.go
```diff
@@ -0,0 +1,168 @@
...
+func setupChangelogCommand() *cobraext.Command {
+	addChangelogCmd := &cobra.Command{
+		Use:   "add",
+		Short: "Add an entry to the changelog file",
+		Long:  changelogAddLongDescription,
+		RunE:  changelogAddCmd,
+	}
+	addChangelogCmd.Flags().String(cobraext.ChangelogAddNextFlagName, "", cobraext.ChangelogAddNextFlagDescription)
+	addChangelogCmd.Flags().String(cobraext.ChangelogAddVersionFlagName, "", cobraext.ChangelogAddVersionFlagDescription)
+
```
mtojek, line 36: nit: empty line

internal/packages/changelog/yaml.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/Masterminds/semver"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+// PatchYAML looks for the proper place to add the new revision in the changelog,
+// trying to conserve original format and comments.
+func PatchYAML(d []byte, patch Revision) ([]byte, error) {
+	var nodes []yaml.Node
```
mtojek, line 18: I understand why you did it this way (generic YAML parsing), but maybe it's fair to depend on schema and unmarshal the changelog to the proper structure? I don't mind if we lose some comments in the meantime.

internal/packages/changelog/yaml.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode manifest")
+	}
+
+	// Manifest is a document, with a single element, that should be a map.
+	if len(node.Content) == 0 || node.Content[0].Kind != yaml.MappingNode {
+		return nil, errors.Wrap(err, "unexpected manifest content")
+	}
+
+	setYamlMapValue(node.Content[0], "version", version)
```
mtojek, line 112: Same concern as for the changelog. In this case, we already have the logic to read the package manifest.

internal/packages/changelog/yaml.go
```diff
@@ -0,0 +1,160 @@
...
+
+	var buf bytes.Buffer
+	encoder := yaml.NewEncoder(&buf)
+	encoder.SetIndent(2)
```
mtojek, line 116: FYI we have the [formatter](https://github.com/elastic/elastic-package/blob/main/internal/formatter/yaml_formatter.go) in the source code, so it would be better not to introduce another `SetIndent`  call :)

internal/packages/changelog/yaml.go
jsoriano: I would prefer to change the files the minimum possible and maintain the current format, at the end these files are still maintained by humans.

I gave a try to yaml patch libraries for this, and they work well for simple changes as setting the version in the manifest, or appending elements, but I didn't manage to add the changelog entry on top. I can give this approach another try so we don't have to maintain this parsing code, but we can still keep the format and comments.

mtojek: I guess that it's a matter of personal preference, YAML patching vs strict model. I don't have any strong points for switching to mapping except for simpler implementation and less node-level processing.

BTW as an alternative approach, you can consider [mapstr](https://github.com/elastic/elastic-package/blob/main/internal/common/mapstr.go). It's up to you to decide which option seems less error-prone :)

mtojek (approved):
I left a comment about an alternative approach, but please don't consider it as blocker for this PR.

mtojek:
FYI @marc-gr as your team may benefit from this issue :)

internal/packages/changelog/yaml.go
jsoriano: Yeah, this code is a bit too much for what it does, but I didn't find another reliable way to do it with the requisite (maybe self-imposed :) ) of maintaining the format as much as possible.

In any case I have tried to maintain the interfaces of the two exposed methods clean, and tests cover them, we can change their implementation if we get another idea.

PR #771, opened 2022-03-31
Check allowed values for fields

jsoriano:
Fields can include a list of allowed values. Check that the values of
these fields are between the allowed ones.

For arrays, validation is now done for all fields, as we not only need
to validate their type but also their values.

Part of https://github.com/elastic/elastic-package/issues/615.


internal/fields/validate.go
```diff
@@ -354,60 +354,74 @@ func compareKeys(key string, def FieldDefinition, searchedKey string) bool {
...
 
-	var valid bool
-	switch definition.Type {
-	case "constant_keyword":
-		var valStr string
-		valStr, valid = val.(string)
-		if !valid {
-			break
-		}
+			if err := ensureConstantKeywordValueMatches(key, valStr, definition.Value); err != nil {
```
mtojek, line 367: It looks like the switch evolved since I remember :) Do you think we could transform the field type-driven logic into properties-driven? I mean, for every field we could check keyword, pattern, allowed values, and simply skip testing if not applicable. Otherwise, this switch will grow up indefinitely...

jsoriano: Umm, we still need custom logic per type:
* Only for `constant_keyword` we check the value is the defined one.
* Only for `ip` we check that is one of the allowed ips.
* Only for `numbers` we check that they are actually numbers.

To skip testing when not applicable we would still need some conditional logic per type.

I think the switch fits well for this. One thing I can do is to group back `keyword`, `text` and `date`. Though I am not sure why `date` was checked as if were `text`.

mtojek: > I think the switch fits well for this. One thing I can do is to group back keyword, text and date. Though I am not sure why date was checked as if were text.

I don't remember now to be honest. Maybe @adriansr remembers this?

jsoriano: Handling of `date` changed more in https://github.com/elastic/elastic-package/pull/781, so I guess that this is going to continue being separated.

internal/fields/validate.go
```diff
@@ -490,3 +497,12 @@ func ensureConstantKeywordValueMatches(key, value, constantKeywordValue string)
 	}
 	return nil
 }
+
+// ensureAllowedValues validates that the document's field value
+// is one of the allowed values.
+func ensureAllowedValues(key, value string, allowedValues AllowedValues) error {
```
mtojek, line 503: nit: in theory, you don't need to pass `key` here as it's just used to render the error message :)

mtojek (approved):

jsoriano:
Blocked till https://github.com/elastic/integrations/issues/3016 is solved.

internal/fields/validate.go
jsoriano: Umm, but error message is rendered here, `key` is used in this function for that.

I would be fine with refactoring this error generation, but in a different PR, making the other ensure method consistent with this.

jsoriano:
I have moved the changes to check values in arrays to https://github.com/elastic/elastic-package/pull/835, so we can start benefiting of this in other cases as with IP validation (https://github.com/elastic/integrations/issues/3408).

jsoriano:
Opening this for final review as blockers in https://github.com/elastic/integrations/issues/3016 are close to be completed.

@mtojek could you please take another look?

mtojek (approved):
Ship it!

PR #777, opened 2022-04-01
Fix importing of ECS tracing fields

jsoriano:
Some groups (`base` and `tracing` afaik) contains fields that are not intended to be nested, but belong to the top level.

`base` was already handled this way, add `tracing` to the exceptions.


mtojek (approved):

mtojek:
Error is unrelated, I will merge this PR.

PR #778, opened 2022-04-04
Fix: do not comment on notifyBuildResult

mtojek:
Related: https://github.com/elastic/elastic-package/commit/8a8e3d214988370320e6051c7bde29a9e1de2439#r70180022

This PR modifies the minimal publishing pipeline so that it doesn't update the relevant Github comment. Currently, there is a race condition between two pipelines and the same comment is randomly updated.


v1v (approved):
Thanks 👍

PR #783, opened 2022-04-06
Document build tags for cloud resources

endorama:
Following https://github.com/elastic/elastic-package/pull/755/ is possible to use a set of environment variables to tag cloud resources created for system tests through Terraform.

The available vars were not documented, this PR aims to fix that.

### DIFF
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -169,6 +169,16 @@ Terraform is often used to interact with Cloud Providers. This require Cloud Pro
 
 Injecting credentials can be achieved with functions from the [`apm-pipeline-library`](https://github.com/elastic/apm-pipeline-library/tree/main/vars) Jenkins library. For example look for `withAzureCredentials`, `withAWSEnv` or `withGCPEnv`.
 
+#### Tagging/labelling created Cloud Provider resources
+
+Leveraging Terraform to create cloud resources is useful but risks creating leftovers resources that are difficult to remove.
+
+There are 4 environment variables that should be leveraged to overcome this issue; these variables are already injected to be used by Terraform (through `TF_VAR_`):
+- `TEST_RUN_ID`: an unique ID for the test run, allows to distinguish each run
+- `REPO_NAME`: the repository name the CI run is linked to
+- `CHANGE_ID`: the PR number the CI run is linked to
+- `BUILD_NUMBER`: incremental number providing the current CI run number
+
 ### Kubernetes service deployer
 
 The Kubernetes service deployer requires the `_dev/deploy/k8s` directory to be present. It can include additional `*.yaml` files to deploy
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -171,10 +171,10 @@ Injecting credentials can be achieved with functions from the [`apm-pipeline-lib
 
 #### Tagging/labelling created Cloud Provider resources
 
-Leveraging Terraform to create cloud resources is useful but risks creating leftovers resources that are difficult to remove.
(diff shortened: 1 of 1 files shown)


docs/howto/system_testing.md
```diff
@@ -169,6 +169,16 @@ Terraform is often used to interact with Cloud Providers. This require Cloud Pro
 
 Injecting credentials can be achieved with functions from the [`apm-pipeline-library`](https://github.com/elastic/apm-pipeline-library/tree/main/vars) Jenkins library. For example look for `withAzureCredentials`, `withAWSEnv` or `withGCPEnv`.
 
+#### Tagging/labelling created Cloud Provider resources
+
+Leveraging Terraform to create cloud resources is useful but risks creating leftovers resources that are difficult to remove.
```
jsoriano, line 174:
```suggestion
Leveraging Terraform to create cloud resources is useful but risks creating leftover resources that are difficult to remove.
```

docs/howto/system_testing.md
```diff
@@ -169,6 +169,16 @@ Terraform is often used to interact with Cloud Providers. This require Cloud Pro
...
+
+There are 4 environment variables that should be leveraged to overcome this issue; these variables are already injected to be used by Terraform (through `TF_VAR_`):
+- `TEST_RUN_ID`: an unique ID for the test run, allows to distinguish each run
```
jsoriano, line 177:
```suggestion
- `TEST_RUN_ID`: a unique identifier for the test run, allows to distinguish each run
```

jsoriano, line 177: Umm, this variable is referenced as `TF_VAR_TEST_RUN_ID` in https://github.com/elastic/elastic-package/pull/755/files#diff-1d4bd2c1ad56472b4dc4c97566e6ce9dc00bda82eb59cf14d1ab8d5c930a1fe0R8

Is this correct?

jsoriano (commented):
:+1:

docs/howto/system_testing.md
mtojek: From the Terraform perspective, these vars are defined without the `TF_VAR_` prefix: [vars.tf](https://github.com/elastic/elastic-package/blob/main/test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/vars.tf), but if you want to map extra ones in the [env.yml](https://github.com/elastic/elastic-package/blob/main/test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/env.yml), then you have to add the prefix.

jsoriano: Yep, I see, but in the [docker compose file for the terraform deployer](https://github.com/elastic/elastic-package/blob/d3345012cf5ce4c565abdc4917c5258dee4fec71/internal/install/_static/terraform_deployer.yml#L8) it seems that we are reading this variable as `TF_VAR_TEST_RUN_ID` (`${TF_VAR_TEST_RUN_ID:-detached}`).
For example for `REPO_NAME` we are reading the variable without this prefix `${REPO_NAME:-unknown}`.

So I wonder if it should be `${TEST_RUN_ID:-detached}` in the docker compose file, and if it shouldn't, I wonder if `TEST_RUN_ID` works or it should be `TF_VAR_TEST_RUN_ID` here in the docs.
```suggestion
- `TF_VAR_TEST_RUN_ID`: an unique ID for the test run, allows to distinguish each run
```

mtojek: AFAIR the source code, TF_VAR_ variables aren't included in the agent policy. It would be a good exercise to check if REPO_NAME is included :) 

Nevertheless, I would focus here on documenting the existing state, so answering your question: `it should be TF_VAR_TEST_RUN_ID here in the docs`.

jsoriano (approved):
thanks!

PR #786, opened 2022-04-11
Add config option to deploy custom elastic-agents as test services

marc-gr:
For some integrations eg: auditbeat or winlogbeat dependant ones. Would be useful to define custom agents. This adds the ability to define them as a system test service and enroll them during the tests instead of the elastic-package stack ones.

Example test service definition:

```yaml
version: '2.3'
services:
  auditd-agent:
    hostname: auditd-agent
    image: "docker.elastic.co/beats/elastic-agent-complete:8.2.0-SNAPSHOT"
    pid: host
    cap_add:
      - AUDIT_CONTROL
      - AUDIT_READ
    user: root
    healthcheck:
      test: "elastic-agent status"
      retries: 180
      interval: 1s
    environment:
      FLEET_ENROLL: "1"
      FLEET_INSECURE: "1"
      FLEET_URL: "http://fleet-server:8220"
```

Example service test config:

```yaml
service: auditd-agent
custom_agent: true
data_stream:
  vars:
    audit_rules:
      - "-a always,exit -F arch=b64 -S execve,execveat -k exec"
    preserve_original_event: true
```


internal/testrunner/runners/system/servicedeployer/factory.go
```diff
@@ -51,6 +51,11 @@ func Factory(options FactoryOptions) (ServiceDeployer, error) {
 			}
 			return NewDockerComposeServiceDeployer([]string{dockerComposeYMLPath}, sv)
 		}
+	case "agent":
+		dockerComposeYMLPath := filepath.Join(serviceDeployerPath, "docker-compose.yml")
+		if _, err := os.Stat(dockerComposeYMLPath); err == nil {
```
jsoriano, line 56: Umm, if err is not nil here, this is going to fail with `unsupported service deployer (name: agent)`, right? This can be a bit counter intuitive. I would say that if this directory exist, it should contain this file.

jsoriano (commented):
This approach looks cleaner to me than the previous one.

Could you add an example test package to `test/packages` that uses this approach?

internal/testrunner/runners/system/servicedeployer/factory.go
```diff
@@ -51,6 +51,17 @@ func Factory(options FactoryOptions) (ServiceDeployer, error) {
 			}
 			return NewDockerComposeServiceDeployer([]string{dockerComposeYMLPath}, sv)
 		}
+	case "agent":
+		dockerComposeYMLPath := filepath.Join(serviceDeployerPath, "docker-compose.yml")
+		if _, err := os.Stat(dockerComposeYMLPath); err != nil {
+			return nil, errors.Wrap(err, "can't find expected file docker-compose.yml")
```
jsoriano, line 57: Nit. Include the path where the file was expected to be found.

test/packages/with-custom-agent/apache/_dev/build/docs/README.md
```diff
@@ -0,0 +1,17 @@
+# Apache Integration
```
jsoriano, line 1: Nit. The apache integration doesn't actually need this deployer. It'd be nice to have a test package that illustrates a real use case where the custom agent would be needed.

Makefile
```diff
@@ -63,9 +63,9 @@ test-stack-command-7x:
 test-stack-command-8x:
 	./scripts/test-stack-command.sh 8.2.0-SNAPSHOT
 
-test-stack-command: test-stack-command-default test-stack-command-7x test-stack-command-800 test-stack-command-8x
+test-stack-command: test-stack-command-default test-stack-command-7x test-stack-command-8x
```
jsoriano, line 66: Unrelated change?

docs/howto/system_testing.md
```diff
@@ -106,6 +107,45 @@ volumes:
...
+      - AUDIT_READ
+    user: root
+    healthcheck:
+      test: "elastic-agent status"
+      retries: 180
+      interval: 1s
+    environment:
+      FLEET_ENROLL: "1"
+      FLEET_INSECURE: "1"
+      FLEET_URL: "http://fleet-server:8220"
```
jsoriano, line 136: Umm, this makes me think that we might need some way to pass variables to these configs. These connection settings depend on the stack and how it is started. For example in #789 I am changing these options, and this would break scenarios with these environment variables.

An option can be to create a different configuration file, expected for example in `_dev/deploy/agent/config.yml`, that is used to "patch" a base agent configuration. So for example in this case, it could be something like this:
```
image: "docker.elastic.co/beats/elastic-agent-complete:8.2.0"
pid: host
cap_add:
  - AUDIT_CONTROL
  - AUDIT_READ
user: root
```
This way every package using this deployer only needs to configure the minimal relevant set of settings. And we can control the settings needed for enrollment, or other settings that may be dependent of `elastic-package stack`.

Another middle-ground option could be to use `env_file` here to include an environment file that would be generated by the deployer, and that includes this kind of environment variables needed by agents to connect with the stack.

jsoriano (commented):
This is looking good, I added some comments, the only blocking one would be to decide how we want to configure the custom agent. Allowing any kind of docker-compose configuration may complicate maintainability of `elastic-package`, see my comment about this.

docs/howto/system_testing.md
mtojek: It looks like you're referring to this [issue](https://github.com/elastic/elastic-package/issues/787#issuecomment-1095148116) (Extend "profiles" with local patches).

Do you think that we need a proposal to sketch the final look and then iterate on that? Maybe we should focus on this specific issue. It might be tricky if we want "patches" to be backward compatible with older stacks.

jsoriano: I would consider this deployer a separate thing to possible general patches for profiles or the stack subcommand. I think that something like this "agent" deployer has value on itself. Even if later we consider more advanced configurations for profiles or the stack subcommand.

The problem I see with general local patches for `elastic-package stack` is that they may become a source of many reproducibility problems. You may patch the stack to work on one package, and later you start working on a different package and something unexpectedly doesn't work. Or it may be difficult to know or remember that a certain package needs a patched stack.
This could cause a new set of problems similar to the ones with the packages contained in the registry depending on where `elastic-package stack up` was executed (https://github.com/elastic/elastic-package/pull/599).
And as you mention it may be difficult to support patches with multiple versions of the stack in a general way.

I think it is fine to look for a way to start/patch specialized agents on test time as is being done in this PR. These agents are disposable and developers have more awareness on when they are being started. Packages could also select the version of the agent to use, limiting the problems of supporting multiple versions. Variants could help to test with multiple versions or configurations if there are differences.

And if some day we also have stack-level patches, I think it would be ok if the patches are different to the ones used for this "agent" deployer, at the end they are different things.

mtojek: > I think it is fine to look for a way to start/patch specialized agents on test time as is being done in this PR. 

What about unpatching? Did you plan for this too or is it like the Kubernetes agent, once installed it stays there.

jsoriano: > > I think it is fine to look for a way to start/patch specialized agents on test time as is being done in this PR.
> 
> What about unpatching? Did you plan for this too or is it like the Kubernetes agent, once installed it stays there.

No need to unpatch. Current implementation starts this patched agent as a docker compose service, and destroys it on tear down. So it doesn't stay. I think this is a good approach.

jsoriano: (If I understand it correctly, please @marc-gr correct me if I am wrong :slightly_smiling_face: )

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -4,33 +4,245 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "marshal custom-agent config")
+	}
+
+	dir, err := os.MkdirTemp("", "elastic-package")
+	if err != nil {
+		return "", errors.Wrap(err, "create tmp dir")
+	}
+
+	tf, err := os.CreateTemp(dir, "custom-agent")
```
jsoriano, line 237: `dir` is never removed? Maybe you can create the config directly on tmp.
```suggestion
	tf, err := os.CreateTemp(os.TempDir(), "custom-agent")
```

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -4,33 +4,245 @@
...
+//go:embed custom-agent-base-config.yml
+var customAgentYml string
+
 // CustomAgentDeployer knows how to deploy a custom elastic-agent defined via
 // a Docker Compose file.
 type CustomAgentDeployer struct {
-	cd *DockerComposeServiceDeployer
+	cfg      string
+	ymlPaths []string
+}
```
jsoriano, line 30: Umm, in the previous version of the code this `CustomAgentDeployer` was a wrapper over `DockerComposeServiceDeployer`.
Are there more differences apart of the additional config?

jsoriano (commented):
I like this approach :)

Added a couple of comments about implementation details.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
marc-gr: Makes sense, will just create in on tmp. 👍

internal/testrunner/runners/system/servicedeployer/custom_agent.go
marc-gr: Also is needed to send appconfig env vars, but I think it makes sense to try to use a wrapper approach if possible. Will work on that.

docs/howto/system_testing.md
marc-gr: It is correct 👍

Makefile
mtojek: +1

internal/testrunner/runners/system/servicedeployer/custom-agent-base-config.yml
```diff
@@ -0,0 +1,15 @@
...
+    image: "${ELASTIC_AGENT_IMAGE_REF}"
+    healthcheck:
+      test: "elastic-agent status"
+      retries: 180
+      interval: 1s
+    hostname: custom-agent
+    environment:
+      - FLEET_ENROLL=1
+      - FLEET_INSECURE=1
+      - FLEET_URL=http://fleet-server:8220
```
mtojek, line 13: Based on this [file](https://github.com/elastic/elastic-package/blob/main/internal/profile/_static/elastic-agent_8x.env), isn't `FLEET_TOKEN_POLICY_NAME` not required?

internal/testrunner/runners/system/servicedeployer/custom-agent-base-config.yml
```diff
@@ -0,0 +1,15 @@
+version: "2.3"
+services:
+  custom-agent:
+    image: "${ELASTIC_AGENT_IMAGE_REF}"
+    healthcheck:
+      test: "elastic-agent status"
+      retries: 180
+      interval: 1s
+    hostname: custom-agent
```
mtojek, line 9: nit: docker-custom-agent

In case that somebody uses `custom-agent` for enrolling a tru custom agent.

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -78,16 +86,25 @@ func (d *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
...
+			)...,
+		),
 		ExtraArgs: []string{"--build", "-d"},
 	}
 	err = p.Up(opts)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not boot up service using Docker Compose")
 	}
 
+	// Connect service network with stack network (for the purpose of metrics collection)
```
mtojek, line 102: Is it safe to move `docker.ConnectToNetwork` before checking if the service is healthy? Do you think that it won't break if the service doesn't boot up in time?

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	appConfig, err := install.Configuration()
+	if err != nil {
+		return nil, errors.Wrap(err, "can't read application configuration")
+	}
+
+	cd, err := newDockerComposeServiceDeployer(
+		append(ymlPaths, path),
+		appConfig.StackImageRefs(install.DefaultStackVersion).AsEnv(),
```
mtojek, line 49: Does it mean that it won't be possible to use other stack versions?

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// SetUp sets up the service and returns any relevant information.
+func (d *CustomAgentDeployer) SetUp(inCtxt ServiceContext) (DeployedService, error) {
+	ds, err := d.cd.SetUp(inCtxt)
+	if err != nil {
+		return nil, err
+	}
+
+	dds, ok := ds.(*dockerComposeDeployedService)
```
mtojek, line 69: Hm... I used to go without casting as [here](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/servicedeployer/kubernetes.go#L103). Is it mandatory really here?

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,131 @@
...
+
+type deployedCustomAgent struct {
+	cfg string
+	*dockerComposeDeployedService
+}
+
+var _ DeployedService = &deployedCustomAgent{}
+
+// NewCustomAgentDeployer returns a new instance of a deployedCustomAgent.
+func NewCustomAgentDeployer(cfgPath string, ymlPaths []string, sv ServiceVariant) (*CustomAgentDeployer, error) {
```
mtojek, line 36: What's the difference between `cfgPath` and `ymlPaths`? Are both used?

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -33,14 +34,20 @@ type dockerComposeDeployedService struct {
...
 // NewDockerComposeServiceDeployer returns a new instance of a DockerComposeServiceDeployer.
 func NewDockerComposeServiceDeployer(ymlPaths []string, sv ServiceVariant) (*DockerComposeServiceDeployer, error) {
+	return newDockerComposeServiceDeployer(ymlPaths, nil, sv)
+}
+
+func newDockerComposeServiceDeployer(ymlPaths, appEnv []string, sv ServiceVariant) (*DockerComposeServiceDeployer, error) {
 	return &DockerComposeServiceDeployer{
 		ymlPaths: ymlPaths,
 		sv:       sv,
+		appEnv:   appEnv,
```
mtojek, line 50: Is there an option not to pass `appEnv` through many places?

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,131 @@
...
+		return "", errors.Wrap(err, "open custom-agent config")
+	}
+
+	cv := map[string]interface{}{}
+	if err := yaml.Unmarshal(cb, &cv); err != nil {
+		return "", errors.Wrap(err, "unmarshal custom-agent config")
+	}
+
+	for k, v := range cv {
+		bc.Services["custom-agent"][k] = v
```
mtojek, line 112: Could you please clarify a bit what is the purpose of the processing here?

mtojek (commented):
Good job! I left a few comments, but maybe it's better to sync over zoom to explain the technical decisions. Up to you, Marc :)

internal/testrunner/runners/system/servicedeployer/custom-agent-base-config.yml
marc-gr: Would only be needed if we want it to be enrolled to the current stack policy, but we are just interested in it during the test, so the policy will change right after for the test one, running the config useless.

mtojek: Ah, you're right! Could you please drop a comment there in case somebody else will come here.

internal/testrunner/runners/system/servicedeployer/compose.go
marc-gr: From my understanding it makes no difference for the other cases, and in this one is a requirement (healthcheck only passes once enrolled, and needs to be in the same network to be enrolled). Do you think of any specific scenarios where this might be troublesome?

internal/testrunner/runners/system/servicedeployer/custom_agent.go
marc-gr: Other stacks could be used by adding the `image` option in the custom-agent.yml file or using service variants.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
marc-gr: To have access to the internal context to override it we need the specific type, since the interface does not expose it.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
marc-gr: cfgPath is the custom agent options and is always mandatory. If you define other containers then ymlpaths is also used, but they are optional. `cfgPath` is used to create a temporary ymlPath for the custom agent container though.

internal/testrunner/runners/system/servicedeployer/compose.go
marc-gr: I couldn't find any non-intrusive one, I thought is better to pass this aroung that modifying the composer code too much.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
marc-gr: we take all the custom options from `custom-agent.yml` and use them to override the base config.

internal/testrunner/runners/system/servicedeployer/compose.go
mtojek: Yeah, but the `compose.go` code is generic, so it may affect also other services like kind or docker based. 

You could verify the behavior by simulating a broken Docker service, for instance: a container that immediately fails setup.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
mtojek: I'm not sure if it doesn't loosen the dependency between Docker service deployer and Custom Agent deployer. Take a look at the [terraform](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/servicedeployer/terraform.go#L32) deployer. It looks like you need to define the Compose project. WDYT?

internal/testrunner/runners/system/servicedeployer/custom_agent.go
mtojek: Is the base config one from `~/.elastic-agent`?

internal/testrunner/runners/system/servicedeployer/compose.go
mtojek: Yeah yeah, that's why I proposed to loosen the dependency between Compose service deployer and this one :)

internal/testrunner/runners/system/servicedeployer/custom_agent.go
mtojek: Out of curiosity: does it mean that it's possible to boot up Elastic stack 8.1 and the agent 8.3? I'm wondering if we need a helper to select the proper stack. See [kind deployer](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/servicedeployer/kubernetes.go#L151)

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,198 @@
...
+		bc.Services[dockerCustomAgentName][k] = v
+	}
+	bc.Services[dockerCustomAgentName]["hostname"] = dockerCustomAgentName
+
+	b, err := yaml.Marshal(bc)
+	if err != nil {
+		return "", errors.Wrap(err, "marshal custom-agent config")
+	}
+
+	tf, err := os.CreateTemp("", dockerCustomAgentName)
```
mtojek, line 187: nit: There is `~/.elastic-package/tmp` directory and probably also a function to reach out to it.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,198 @@
...
+	defer func() {
+		if err := os.Remove(s.cfg); err != nil {
+			logger.Errorf("cleaning up tmp file (path: %s): %v", s.cfg, err)
+		}
+	}()
+	return s.dockerComposeDeployedService.TearDown()
+}
+
+func createCustomAgentYaml(cfgPath string) (string, error) {
+	bc := struct {
```
mtojek, line 159: nit: could you please replace shortcut-named vars with something meaningful? I admit that I got lost, cv, cb, bc, .. :)

docs/howto/system_testing.md
```diff
@@ -106,6 +107,55 @@ volumes:
...
+### Agent service deployer
+
+When using the Agent service deployer, the `elastic-agent` provided by the stack
+will not be used. An agent will be deployed as a Docker compose service named `docker-custom-agent`
+which base configuration is provided [here](../../internal/testrunner/runners/system/servicedeployer/custom-agent-base-config.yml).
+This configuration will be merged with the one provided in the `custom-agent.yml` file.
+This is useful if you need different capabilities than the provided by the
+`elastic-agent` used by the `elastic-package stack` command. 
+
+`custom-agent.yml`
```
mtojek, line 119: As the `custom-agent.yml` will be part of a package, we need to cover it with package-spec. You will need to open one more PR.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,198 @@
...
+	s := serviceComposeConfig.Services[serviceName]
+	outCtxt.Ports = make([]int, len(s.Ports))
+	for idx, port := range s.Ports {
+		outCtxt.Ports[idx] = port.InternalPort
+	}
+
+	// Shortcut to first port for convenience
+	if len(outCtxt.Ports) > 0 {
+		outCtxt.Port = outCtxt.Ports[0]
+	}
```
mtojek, line 137: nit: I don't remember if this condition is still required.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -0,0 +1,198 @@
...
+		return "", errors.Wrap(err, "marshal custom-agent config")
+	}
+
+	tf, err := os.CreateTemp("", dockerCustomAgentName)
+	if err != nil {
+		return "", errors.Wrap(err, "create tmp file")
+	}
+	defer tf.Close()
+
+	if _, err := tf.Write(b); err != nil {
```
mtojek, line 193: You don't need to use the temp dir if you "install" `custom-agent.yml` to the `~/.elastic-package/deployer/custom-agent` similarly to `~/.elastic-package/deployer/terraform/terraform-deployer.yml`. This would be a very similar mechanism for extending Compose definition.

internal/testrunner/runners/system/servicedeployer/custom_agent.go
marc-gr: if we keep it then the port can be referenced from the test config and that can be useful in some cases, not a hard requirement though so we can remove it if there is any concern

mtojek (approved):

PR #793, opened 2022-04-20
Add Check to lint to validate changelog PR link

r00tu53r:
On occasion the PR link in the `changelog` is not updated with the correct number and is left with `xxxx` or with no number. This PR adds a linter check to validate the latest entry's change PR links.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -5,6 +5,10 @@
 package cmd
 
 import (
+	"net/url"
+	"path"
+	"strconv"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
@@ -13,6 +17,7 @@ import (
 	"github.com/elastic/elastic-package/internal/cobraext"
 	"github.com/elastic/elastic-package/internal/docs"
 	"github.com/elastic/elastic-package/internal/packages"
+	"github.com/elastic/elastic-package/internal/packages/changelog"
 )
 
 const lintLongDescription = `Use this command to validate the contents of a package using the package specification (see: https://github.com/elastic/package-spec).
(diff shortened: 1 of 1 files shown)


mtojek (commented):
The idea of an extra check is, of course, handy, but maybe we can move it to the package-spec? WDYT?

jsoriano:
+1 to add this check to the spec, maybe it is enough with adding `format: uri` here: https://github.com/elastic/package-spec/blob/a3d66cc9dc3fe05ff28b561e69d1438d714caa12/versions/1/changelog.spec.yml#L35

This will be a breaking change, we would have to check if this breaks integrations with wrong urls.

Also take into account that this doesn't solve the problem of wrong or no numbers in urls.

r00tu53r:
The idea of this PR was to check if a PR number ID is specified. There have been merges of packages to `main` without the PR number. Checking URL format is just a part of it. This check should break the CI and hence prevent the merge.

jsoriano:
> The idea of this PR was to check if a PR number ID is specified. There have been merges of packages to `main` without the PR number. Checking URL format is just a part of it. This check should break the CI and hence prevent the merge.

We can add both things, the URL format check in the package-spec if we see that this doesn't break too many things, and another check for the URL in the changelog.

I wouldn't couple a check in elastic-package to github url formats. This won't work for developers using other issue trackers, or linking to other websites explaining the issue. It also still doesn't prevent developers temporarily using copied links, or arbitrary numbers instead of `https://github.com/.../xxx`.

One thing that could be done, to allow having this check, in a more reliable way and without coupling it to the issue tracker, could be to have a check that explicitly validates that the URL of the pull request is used in the changelog. The URL can be known by Jenkins or other CI tools, it could be passed to `elastic-package` using some flag or environment variable.

Imagine for example:
```
elastic-package lint --changelog-link https://github.com/elastic/elastic-package/pull/793
```
Or
```
export ELASTIC_PACKAGE_CHANGELOG_LINK=https://github.com/elastic/elastic-package/pull/793
elastic-package lint
```
The advantage of the environment variable vs the flag is that the flag might need to be added to all the subcommands that do package validation. Though if we add this I would prefer to add this as a flag instead of a hidden environment variable.

@mtojek wdyt?

mtojek:
I'd rather be for making baby steps, so far we haven't seen any link referring to the other tracker as Github, so let's add conditional logic to package-spec which checks if the link belongs to Github and then it parses its format.

r00tu53r:
Thanks @jsoriano @mtojek very helpful discussion. I'm looking at package-spec. Will update shortly.

r00tu53r:
Opened a package-spec PR https://github.com/elastic/package-spec/pull/324. Closing this one.

PR #800, opened 2022-04-26
Fix fleet managed spec

DaveSys911:
What does this pr do?

Updates the cluster role permissions required for cloudbeat's k8s API fetcher.

image


oren-zohar (changes requested):
tested locally and it didn't have kube-api findings 
image

DaveSys911:
> tested locally and it didn't have kube-api findings image

These are the correct permissions for what we require, I couldn't test with elastic-package, since my local build of it isn't usable for some reason. Tested successfully with a standalone agent.

oren-zohar:
@DaveSys911 I agree that in theory these are the permissions we need for the `kube-api` data, but if we want to edit the elastic-agent-managed manifest, we need to test it with a managed deployment.

PR #819, opened 2022-05-12
Permissive validation of arrays of objects

jsoriano:
Some integrations are storing arrays of objects in group types. This is
valid, even if not recommended. Be permissive on these cases by now.

Changes in #818 breaks validation of fields in some integrations that
store arrays of objects (https://beats-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Fintegrations%2FPR-3348/detail/PR-3348/1/tests/).

### DIFF
--- a/internal/fields/validate_test.go
+++ b/internal/fields/validate_test.go
@@ -237,6 +237,34 @@ func Test_parseElementValue(t *testing.T) {
 			},
 			fail: true,
 		},
+		// arrays of objects can be stored in groups, even if not recommended
+		{
+			key: "host",
+			value: []interface{}{
+				map[string]interface{}{
+					"id":       "somehost-id",
+					"hostname": "somehost",
+				},
+				map[string]interface{}{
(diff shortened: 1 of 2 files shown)


internal/fields/validate_test.go
```diff
@@ -237,6 +237,34 @@ func Test_parseElementValue(t *testing.T) {
...
+				map[string]interface{}{
+					"id":       "somehost-id",
+					"hostname": "somehost",
+				},
+				map[string]interface{}{
+					"id":       "otherhost-id",
+					"hostname": "otherhost",
+				},
+			},
+			definition: FieldDefinition{
```
mtojek, line 253: I'm wondering if there isn't any parser like this in the ECS project? I have a feeling that we're reimplementing it.

mtojek (approved):

mtojek:
I will merge this PR to unclog this issue https://github.com/elastic/integrations/pull/3366

PR #835, opened 2022-05-30
Check values in arrays

jsoriano:
Add validation for all values in arrays.

This is needed when some validation of the values is desired, apart of type validation. This is the case with IPs (https://github.com/elastic/integrations/issues/3408) or event types (https://github.com/elastic/elastic-package/issues/615#issuecomment-1021884406).

This change was part of https://github.com/elastic/elastic-package/pull/771, but this is being delayed due to https://github.com/elastic/integrations/issues/3016.

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -352,85 +352,83 @@ func compareKeys(key string, def FieldDefinition, searchedKey string) bool {
 }
 
 func (v *Validator) parseElementValue(key string, definition FieldDefinition, val interface{}) error {
-	val, ok := ensureSingleElementValue(val)
-	if !ok {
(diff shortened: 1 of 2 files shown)


jsoriano:
> A more generic non-blocking comment - it's really easy to get lost in the logic due to multiple switch-cases and "if" clauses. Maybe you should drop a few comments to show field examples including edge cases.

I have added comments, and refactored the code to explicitly return type errors instead of having the `if valid` condition at the end. I have also moved the code to a function for single values, what reduces the nesting level of the main code.

@mtojek let me know what you think.

mtojek (approved):
Thank you! That's exactly what I meant.

PR #840, opened 2022-06-03
Enable developer extensions

mtojek:
Meta-issue: https://github.com/elastic/package-spec/issues/319

Changes:
* render the table with fields in docs
* import fields from ECS

Related:
https://github.com/elastic/package-spec/pull/350


internal/fields/validate.go
```diff
@@ -89,16 +90,25 @@ func CreateValidatorForDataStream(dataStreamRootPath string, opts ...ValidatorOp
...
 	if v.disabledDependencyManagement {
 		return v, nil
 	}
 
-	packageRoot := filepath.Dir(filepath.Dir(dataStreamRootPath))
+	packageRoot, found, err := packages.FindPackageRoot()
+	if err != nil {
+		return nil, errors.Wrap(err, "can't find package root")
+	}
+	if !found {
```
jsoriano, line 107: Could this hide problems with malformed packages?

On what cases is it ok to not find the package root? Could you add a comment here about this?

internal/builder/external_fields.go
```diff
@@ -37,11 +37,18 @@ func resolveExternalFields(packageRoot, destinationDir string) error {
...
 	}
 
-	fieldsFile, err := filepath.Glob(filepath.Join(destinationDir, "data_stream", "*", "fields", "*.yml"))
+	dataStreamFieldsFiles, err := filepath.Glob(filepath.Join(destinationDir, "data_stream", "*", "fields", "*.yml"))
 	if err != nil {
 		return err
 	}
-	for _, file := range fieldsFile {
+
+	packageFieldsFiles, err := filepath.Glob(filepath.Join(destinationDir, "fields", "*.yml"))
```
jsoriano, line 45: Maybe we should refactor `resolveExternalFields` to accept only the paths where fields are, this way the method can be reused for any set of directories containing fields files.

Then the method to build integration packages would call `resolveExternalFields` with the data stream directories, and the method to build input packages would call it with its fields directory.

I see we are reusing the same `BuildPackage` method for both types of packages, maybe we should have different build methods for each kind of package.

This can be left for a future refactor, maybe when we have a third kind of package with fields :)

jsoriano (commented):
Looks good.

Added some comments about possible future refactors that may be done to support multiple package types.

My only concern is about cases where we may be accepting that we cannot find the package root.

internal/fields/validate.go
mtojek: I added a comment, let me know what do you think.

internal/fields/validate.go
```diff
@@ -89,16 +90,25 @@ func CreateValidatorForDataStream(dataStreamRootPath string, opts ...ValidatorOp
 
 	v.allowedCIDRs = initializeAllowedCIDRsList()
 
-	v.Schema, err = loadFieldsForDataStream(dataStreamRootPath)
+	fieldsDir := filepath.Join(fieldsParentDir, "fields")
+	v.Schema, err = loadFieldsFromDir(fieldsDir)
 	if err != nil {
-		return nil, errors.Wrapf(err, "can't load fields for data stream (path: %s)", dataStreamRootPath)
+		return nil, errors.Wrapf(err, "can't load fields for data stream (path: %s)", fieldsParentDir)
```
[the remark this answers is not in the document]
mtojek: Fixed

internal/builder/external_fields.go
mtojek: It sounds like a good idea, I can create an issue for this as I don't want to swamp in that refactoring while introducing input packages.

mtojek: https://github.com/elastic/elastic-package/issues/843

internal/fields/validate.go
jsoriano: Ok, we can revisit this if found to be problematic at some point. Would it make sense to add a debug message here?

mtojek: Sure!

PR #841, opened 2022-06-06
Allow cancellation of clean-up wait

efd6:
This allows waits to hear signals and uses this capacity in deferred clean-up waits.

Please take a look.


internal/testrunner/runners/system/runner.go
```diff
@@ -95,7 +95,7 @@ func (r *runner) TearDown() error {
 func (r *runner) tearDownTest() error {
 	if r.options.DeferCleanup > 0 {
 		logger.Debugf("waiting for %s before tearing down...", r.options.DeferCleanup)
-		time.Sleep(r.options.DeferCleanup)
+		signal.Sleep(r.options.DeferCleanup)
```
mtojek, line 98: Ok, so I see that this is enabled only for `options.DeferCleanup`? If so, then I guess it's safe to merge.

mtojek (approved):

PR #844, opened 2022-06-07
Update elastic-agent-managed.yaml.tmpl clusterrole

MichaelKatsoulis:
Add permissions in `elastic-agent` ClusterRole for monitoring 
1.  serviceaccounts
2.  persistentvolumes
3.  persistentvolumeclaims
4.  daemonsets

Syncing the manifest with https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-managed-kubernetes.yaml


ChrsMark:
> > Wondering how we could ensure to getting this synced with the upstream in order to avoid having inconsistencies that might lead to a failing CI.
> 
> On what cases is the current situation leading to a failing CI? Changes are introduced in a controlled way now both in elastic-package and integrations.

Whenever we change something in Beats (or Agent) core that requires changes in the manifests too. In those cases we adjust the manifests at the respective repositories too. However, the CI that tests the k8s package in the Integrations repository will start failing if the manifests used for testing are not updated too. This could be manageable so far with few contributors to these areas but as soon as we have more people involved it can become frustrating and counter-productive.
Maybe just having docs can help here but relying on people's tribal knowledge would be quite error prune to my mind.

@gizas any ideas here?

jsoriano:
> Whenever we change something in Beats (or Agent) core that requires changes in the manifests too.

Take into account that if we make changes in Beats or Agent, that also require changes in the manifests to avoid failures, they will also require changes in user's manifests, what would be breaking changes. We should avoid this kind of changes.

> However, the CI that tests the k8s package in the Integrations repository will start failing if the manifests used for testing are not updated too.

Do they start failing without changes in the package? Or do they faile when changes are applied to the package?
If they start failing after changes in the package I think this is fine, a change to support some things is needed in the manifest included in elastic-package.
If they start failing without changes in the package, this needs to be investigated.

We should aim to have reproducible builds, to reduce surprises, this implies to introduce changes in controlled ways everywhere.

An option to give more control to package developers can be to have some way to configure the agent with some configuration in the package. For example in https://github.com/elastic/elastic-package/pull/786 we introduced a way to deploy custom agents (using docker compose), and it includes [a mechanism to override settings](https://github.com/elastic/elastic-package/blob/810353f59deb7c88ff8a241663b5671d9d4f1f40/docs/howto/system_testing.md#agent-service-deployer). We might introduce a similar way to override settings in the agent started by the kubernetes deployer. Would something like this be preferable for your team?

ChrsMark:
> > Whenever we change something in Beats (or Agent) core that requires changes in the manifests too.
> 
> Take into account that if we make changes in Beats or Agent, that also require changes in the manifests to avoid failures, they will also require changes in user's manifests, what would be breaking changes. We should avoid this kind of changes.


> > However, the CI that tests the k8s package in the Integrations repository will start failing if the manifests used for testing are not updated too.
> 
> Do they start failing without changes in the package? Or do they faile when changes are applied to the package? If they start failing after changes in the package I think this is fine, a change to support some things is needed in the manifest included in elastic-package. If they start failing without changes in the package, this needs to be investigated.
> 
> We should aim to have reproducible builds, to reduce surprises, this implies to introduce changes in controlled ways everywhere.
> 
> An option to give more control to package developers can be to have some way to configure the agent with some configuration in the package. For example in #786 we introduced a way to deploy custom agents (using docker compose), and it includes [a mechanism to override settings](https://github.com/elastic/elastic-package/blob/810353f59deb7c88ff8a241663b5671d9d4f1f40/docs/howto/system_testing.md#agent-service-deployer). We might introduce a similar way to override settings in the agent started by the kubernetes deployer. Would something like this be preferable for your team?


See as an example https://github.com/elastic/elastic-package/pull/702. In general we don't consider changes at [manifests](https://github.com/elastic/beats/tree/2f398145c29b9a025364bda21140b29c81d139f8/deploy/kubernetes) as breaking changes and change those files at will. For example we had completely refactored those to introduce k8s leaderelection. 

Having changes that would fail critically without changes in the manifests is sth serious but is a matter of implementation. Ideally a new feature that would require a change in permissions should not fail critically. But this should be taken into account during the feature implementation by answering the question "what will happen if users do not grand these permissions? Will Metricbeat crash completely?" I don't have a fresh example to illustrate this at the moment but I think it's a bit diverging from my main point here which is to provide a smooth way to maintain our stuff with less human prone/interaction approach.

Speaking of the specific manifests I would just find it ok to use the "official" manifests in every testing we do either it is manual or CI based. This is why I wonder why to maintain manifests in several places.

gizas:
> > > Wondering how we could ensure to getting this synced with the upstream in order to avoid having inconsistencies that might lead to a failing CI.
> > 
> > 
> > On what cases is the current situation leading to a failing CI? Changes are introduced in a controlled way now both in elastic-package and integrations.
> 
> Whenever we change something in Beats (or Agent) core that requires changes in the manifests too. In those cases we adjust the manifests at the respective repositories too. However, the CI that tests the k8s package in the Integrations repository will start failing if the manifests used for testing are not updated too. This could be manageable so far with few contributors to these areas but as soon as we have more people involved it can become frustrating and counter-productive. Maybe just having docs can help here but relying on people's tribal knowledge would be quite error prune to my mind.
> 
> @gizas any ideas here?

I am thinking of a simple file diff initially might work for us as manual step to check differences only with main brunch.

Issue #654, opened 2022-01-19
Enable SSL mode in the Elastic stack


mtojek:

jsoriano:
WIP in https://github.com/elastic/elastic-package/pull/789

PR #847, opened 2022-06-14
Add support for custom CAs for TLS validation in clients

jsoriano:
Initially implemented as part of https://github.com/elastic/elastic-package/pull/789, moved to a separate PR for clarity.

Add support for custom CAs for TLS in clients for Elasticsearch and Kibana.
Include helpers to create certificate pools that include a given CA certificate file.

Related to https://github.com/elastic/elastic-package/issues/654.


internal/certs/pool.go
```diff
@@ -0,0 +1,58 @@
...
+// PoolWithCACertificate returns a new pool that includes the CA certificate
+// in the given path.
+func PoolWithCACertificate(path string) (*x509.CertPool, error) {
+	pool := x509.NewCertPool()
+	err := addCACertificateToPool(pool, path)
+	if err != nil {
+		return nil, err
+	}
+	return pool, nil
+}
```
mrodm, line 23: **question** Currently, it is just being used the function related to the system certs pool. Should this one be kept too? It's true that it could allow us to test just a subset of CAs in case it is needed easily.

mrodm (approved):

internal/stack/shellinit.go
```diff
@@ -26,6 +26,7 @@ var (
 	ElasticsearchUsernameEnv = elasticPackageEnvPrefix + "ELASTICSEARCH_USERNAME"
 	ElasticsearchPasswordEnv = elasticPackageEnvPrefix + "ELASTICSEARCH_PASSWORD"
 	KibanaHostEnv            = elasticPackageEnvPrefix + "KIBANA_HOST"
+	CACertificateEnv         = elasticPackageEnvPrefix + "CACERT"
```
mtojek, line 29: nit: CA_CERT?

internal/kibana/client.go
```diff
@@ -34,32 +37,49 @@ type ClientOption func(*Client)
...
+
+	if c.host == "" {
+		return nil, stack.UndefinedEnvError(stack.KibanaHostEnv)
+	}
+
 	return c, nil
 }
 
+// Address option sets the host to use to connect to Kibana.
+func Address(address string) ClientOption {
```
mtojek, line 63: I suppose that this function will be used in the original PR, right?

internal/certs/testdata/ca-key.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN EC PRIVATE KEY-----
```
mtojek, line 1: Maybe drop a CONTRIBUTING.md with instructions to generate those :)

internal/certs/pool.go
```diff
@@ -0,0 +1,58 @@
...
+
+func addCACertificateToPool(pool *x509.CertPool, path string) error {
+	d, err := ioutil.ReadFile(path)
+	if err != nil {
+		return err
```
mtojek, line 42: fmt.Errorf?

mtojek (approved):
only few nit-picks, nothing blocking

internal/certs/pool.go
jsoriano: In https://github.com/elastic/elastic-package/pull/789 I will also use it to [check](https://github.com/elastic/elastic-package/pull/789/files#diff-5f4340f6bec3f455c2fdb1aed7ab1ae1fd23cc179d8520cdbf275461794c504eR133) that the CA in the profile verifies the service certificates.

I can remove it by now, it may happen that we don't use it at the end, depending on the other PR :slightly_smiling_face:

internal/kibana/client.go
jsoriano: Added for completitude, this is only used on tests by now https://github.com/elastic/elastic-package/pull/847/files/bff27e46db7e789f768740b4a4b18cc7c403793b#diff-9c4e0882a886054c20aa4c09e6c0e667826fd54b97f98e5ec36066eb0798feb8R30

internal/certs/testdata/ca-key.pem
jsoriano: Added as comment in the tests.

mtojek (approved):
LGTM

PR #862, opened 2022-06-22
Add new subcommand to dump Agent Policies

### LINKED ISSUE #671: Dump policy using elastic-agent
Related: https://github.com/elastic/elastic-package/pull/666

This problem [strikes back](https://beats-ci.elastic.co/job/Ingest-manager/job/elastic-package/job/PR-666/1/artifact/build/elastic-stack-dump/stack/7.16.3-SNAPSHOT/logs/elastic-agent.log):

```
[36melastic-agent_1              |[0m Policy selected for enrollment:  2016d7cc-135e-5583-9758-3ba01f5a06e5
[36melastic-agent_1              |[0m 2022-01-26T17:27:04.290Z	WARN	[tls]	tlscommon/tls_config.go:98	SSL/TLS verifications disabled.
[36melastic-agent_1              |[0m 2022-01-26T17:27:05.283Z	INFO	cmd/enroll_cmd.go:454	Starting enrollment to URL: http://fleet-server:8220/
[36melastic-agent_1              |[0m Successfully enrolled the Elastic Agent.
[36melastic-agent_1              |[0m 2022-01-26T17:27:06.077Z	INFO	cmd/enroll_cmd.go:252	Elastic Agent might not be running; unable to trigger restart
[36melastic-agent_1              |[0m Error: could not read configuration file /usr/share/elastic-agent/state/elastic-agent.yml: yaml: line 47: could not find expected ':'
[36melastic-agent_1              |[0m For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/7.16/fleet-troubleshooting.html
```

We can either use `elastic-agent diagnostics` or simply dump the mentioned file. In both cases, there is a risk that the policy will contain sensitive data that even our developers shouldn't see. Consider redacting or trimming.

mrodm:
Fixes #671 

Add a new subcommand under `dump` named `agent-policies` that is in charge of getting and writing the required agent policies.
Example of the usage output for this new subcommand:

```
 $ ./elastic-package dump agent-policies -h
Use this command to dump agent policies created by Fleet as part of a package installation.

Use this command as a exploratory tool to dump agent policies as they are created by Fleet when installing a package. Dumped agent policies are stored in files as they are returned by APIs of the stack, without any processing.

Usage:
  elastic-package dump agent-policies [flags]

Flags:
      --agent-policy string   name of the agent policy
  -h, --help                  help for agent-policies

Global Flags:
  -o, --output string    path to directory where exported assets will be stored (default "package-dump")
  -P, --package string   name of the package
  -v, --verbose          verbose mode
```

Examples of usages of this new subcommand:
- Dump a specific agent policy:
  ```
  elastic-package dump agent-policies --agent-policy <agent_policy_id>
  ```
- Dump all agent policies set in Fleet:
  ```
  elastic-package dump agent-policies
  ```
- Dump the agent policies where a package:
  ```
  elastic-package dump agent-policies --package <package_name>
  ```

Currently, these agent policies are obtained directly from Fleet API using this endpoint: `<fleet_api>/api/fleet/agent_policies`

These agent policies are being stored as they are returned by the APIs without any processing. If there any secrets that must be redacted, probably it would be better to wait until this https://github.com/elastic/package-spec/pull/339


internal/kibana/policies.go
```diff
@@ -72,6 +72,50 @@ func (c *Client) GetPolicy(policyID string) (*Policy, error) {
...
+
+	return resp.Item, nil
+}
+
+// ListRawPolicy fetches all the Policies in the Ingest Manager.
+func (c *Client) ListRawPolicy() ([]json.RawMessage, error) {
+	statusCode, respBody, err := c.get(fmt.Sprintf("%s/agent_policies?full=true", FleetAPI))
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get policy")
+	}
```
mrodm, line 102: This query returns a paginated response. Need to check about how to retrieve all the agent policies.

cmd/dump.go
```diff
@@ -19,30 +20,43 @@ const dumpInstalledObjectsLongDescription = `Use this command to dump objects in
 
 Use this command as a exploratory tool to dump objects as they are installed by Fleet when installing a package. Dumped objects are stored in files as they are returned by APIs of the stack, without any processing.`
 
+const dumpAgentPoliciesLongDescription = `Use this command to dump agent policies created by Fleet as part of a package installation.
+
+Use this command as a exploratory tool to dump agent policies as they are created by Fleet when installing a package. Dumped agent policies are stored in files as they are returned by APIs of the stack, without any processing.`
```
jsoriano, line 25: Mention somewhere that if no flag is provided, all policies are dumped.

cmd/dump.go
```diff
@@ -76,3 +90,56 @@ func dumpInstalledObjectsCmd(cmd *cobra.Command, args []string) error {
...
+			return errors.Wrap(err, "dump failed")
+		}
+		cmd.Printf("Dumped agent policy %s to %s\n", agentPolicy, outputPath)
+	case packageName != "":
+		dumper := dump.NewAgentPoliciesDumper(kibanaClient)
+		count, err := dumper.DumpAgentPoliciesFileteredByPackage(cmd.Context(), packageName, outputPath)
+		if err != nil {
+			return errors.Wrap(err, "dump failed")
+		}
+		cmd.Printf("Dumped %d agent policies filtering by package name %s to %s\n", count, packageName, outputPath)
```
jsoriano, line 135: Nit. Log a different message if no policies are dumped. It prints now:
```
Dumped 0 agent policies filtering by package name apache to package-dump
```

internal/dump/agentpolicies.go
```diff
@@ -0,0 +1,187 @@
...
+const AgentPoliciesDumpDir = "agent_policies"
+
+type AgentPolicyDumper struct {
+	name   string
+	client *kibana.Client
+
+	policy *AgentPolicy
+}
+
+type AgentPoliciesDumper struct {
```
jsoriano, line 26: Nit. Could this dumper be generalized to dump also a single policy? If a name is given, it dumps only this one, if no name is given, it dumps all. So we wouldn't need two dumper implementations for the same thing.

jsoriano (commented):
This is looking good, I also tried it and works well. Added some suggestions.

It'd be also good to add tests, also to be able to see what kind of documents are dumped. You can follow the same strategy used in the `dump installed-objects` command, where an elasticsearch client is mocked with pre-recorded responses, see https://github.com/elastic/elastic-package/blob/dec6298119687e42c3352a6fec9d953280c05c1a/internal/dump/installedobjects_test.go#L81.

internal/dump/agentpolicies.go
mrodm: Done
Now there is just one struct `AgentPoliciesDumper` to manage both scenarios

internal/kibana/policies.go
mrodm: Added support to loop over all the pages returned by the API.

cmd/dump.go
```diff
@@ -76,3 +95,64 @@ func dumpInstalledObjectsCmd(cmd *cobra.Command, args []string) error {
...
+	if tlsSkipVerify {
+		clientOptions = append(clientOptions, kibana.TLSSkipVerify())
+	}
+	kibanaClient, err := kibana.NewClient(clientOptions...)
+	if err != nil {
+		return errors.Wrap(err, "failed to initialize Kibana client")
+	}
+
+	switch {
+	case agentPolicy != "":
```
jsoriano, line 127: Nit. Should we add a case to log an error if both agent policy and package name are set?

cmd/dump.go
```diff
@@ -76,3 +95,64 @@ func dumpInstalledObjectsCmd(cmd *cobra.Command, args []string) error {
...
+		dumper := dump.NewAgentPoliciesDumper(kibanaClient, &agentPolicy)
+		err = dumper.DumpAgentPolicy(cmd.Context(), outputPath)
```
jsoriano, line 129: Nit. This API is using different approaches to dump per agent policy or per package name. For agent policy the parameter is passed on the constructor, and for package name in the dump method.

I would suggest to use a consistent approach for both ways of duping policies.

For example it could be like this:
```
dumper := dump.NewAgentPoliciesDumper(kibanaClient)
err = dumper.DumpAgentPolicy(cmd.Context(), outputPath, dump.AgentPolicyOptions{Name: agentPolicy})
...
err = dumper.DumpAgentPolicy(cmd.Context(), outputPath, dump.AgentPolicyOptions{PackageName: packageName})
```
Or with explicit methods:
```
dumper := dump.NewAgentPoliciesDumper(kibanaClient)
err = dumper.DumpByName(cmd.Context(), outputPath, agentPolicy)
...
err = dumper.DumpByPackage(cmd.Context(), outputPath, packageName)
...
err = dumper.DumpAll(cmd.Context(), outputPath)
```
Or initialize the dumper for the given configuration:
```
dumper := dump.NewAgentPoliciesDumper(kibanaClient, agentPolicy, packageName)
err = dumper.Dump(cmd.Context())
```

internal/dump/agentpolicies.go
```diff
@@ -0,0 +1,182 @@
...
+	"path/filepath"
+
+	"github.com/elastic/elastic-package/internal/common"
+	"github.com/elastic/elastic-package/internal/kibana"
+)
+
+const AgentPoliciesDumpDir = "agent_policies"
+
+type AgentPoliciesDumper struct {
+	name   *string
```
jsoriano, line 20: When possible, it'd be better to use the zero value to indicate that this is not set, de-referencing pointers is always a risk of panics.
```suggestion
	name   string
```

And then you can check if this is set by checking if `name != ""`.

cmd/dump.go
mrodm: True, as it is now it does not make sense to have both set at the same time.
I'll add a new case for that.
Thanks!

internal/dump/agentpolicies.go
mrodm: Ok, I'll change to be a regular string.

cmd/dump.go
mrodm: You're right, there was a mix of different approaches in these calls. Updated using explicit methods.

mrodm:
> This is looking good, I also tried it and works well. Added some suggestions.
> 
> It'd be also good to add tests, also to be able to see what kind of documents are dumped. You can follow the same strategy used in the `dump installed-objects` command, where an elasticsearch client is mocked with pre-recorded responses, see
> 
> https://github.com/elastic/elastic-package/blob/dec6298119687e42c3352a6fec9d953280c05c1a/internal/dump/installedobjects_test.go#L81
> 
> .

Thanks for all your comments.
Regarding testing, if there is no inconvenience, I'll add those tests in a following branch

internal/dump/agentpolicies.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/elastic/elastic-package/internal/common"
+	"github.com/elastic/elastic-package/internal/kibana"
+)
+
+const AgentPoliciesDumpDir = "agent_policies"
+
+type AgentPoliciesDumper struct {
+	client *kibana.Client
+
+	policies []AgentPolicy
```
jsoriano, line 22: Maybe we don't need to cache the policies here, this would simplify the logic. In the installed objects dumper we cache the objects to avoid querying objects multiple times when ones depends on others.
```suggestion
```

internal/kibana/policies.go
```diff
@@ -72,6 +72,62 @@ func (c *Client) GetPolicy(policyID string) (*Policy, error) {
...
+			return nil, fmt.Errorf("could not get policies; API status code = %d; response body = %s", statusCode, respBody)
+		}
+
+		if err := json.Unmarshal(respBody, &resp); err != nil {
+			return nil, errors.Wrap(err, "could not convert policies (response) to JSON")
+		}
+
+		itemsRetrieved += len(resp.Items)
+		currentPage += 1
+		items = append(items, resp.Items...)
```
jsoriano, line 125: Nit. To avoid having to store all policies in memory, we could pass a visitor method here, so policies can be directly processed.
The signature of the function would be something like this:
```
func (c *Client) ForEachRawPolicy(visitFn func(json.RawMessage) error) error {
```
And then here:
```
for _, item := range resp.Items {
    if err := visitFn(item); err != nil {
        return err
    }
}
```

But no need to do it now, we can make this change later if this is a problem at some point.

internal/kibana/policies.go
```diff
@@ -72,6 +72,62 @@ func (c *Client) GetPolicy(policyID string) (*Policy, error) {
...
+	currentPage := 1
+	var items []json.RawMessage
+	var resp struct {
+		Items   []json.RawMessage `json:"items"`
+		Total   int               `json:"total"`
+		Page    int               `json:"page"`
+		PerPage int               `json:"perPage"`
+	}
+
+	for finished := false; !finished; finished = itemsRetrieved == resp.Total {
```
jsoriano, line 109: Nit.Would this be equivalent?
```suggestion
	for itemsRetrieved < resp.Total {
```

jsoriano (commented):
Looks good, added some suggestions but nothing really blocking.

internal/kibana/policies.go
mrodm: At least as it is coded, it is not equivalent. 
The first iteration of the loop resp.Total has the default value (zero) since there was no query run against fleet yet. This struct gets the value in each iteration. With that suggestion, the loop never gets executed.

Using that for loop in that way, I tried to mimic a do-while loop.

internal/dump/agentpolicies.go
mrodm: Makes sense. At least now, the same agent policy is not being queried several times.

internal/dump/agentpolicies.go
```diff
@@ -0,0 +1,175 @@
...
+
+		for _, policy := range policies {
+			err = json.Unmarshal(policy, &policyPackages)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get Agent Policy ID: %w", err)
+			}
+			packageNames := getPackagesUsingAgentPolicy(policyPackages.PackagePolicies)
+			if !common.StringSliceContains(packageNames, packageName) {
+				continue
+			}
```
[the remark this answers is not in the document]
mrodm: Good catch! Much better reusing that function

internal/dump/agentpolicies.go
```diff
@@ -0,0 +1,175 @@
...
+func (d *AgentPoliciesDumper) DumpAll(ctx context.Context, dir string) (count int, err error) {
+	d.policies = nil
+	agentPolicies, err := d.getAllAgentPolicies(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get agent policy: %w", err)
+	}
+
+	dir = filepath.Join(dir, AgentPoliciesDumpDir)
+	for _, agentPolicy := range agentPolicies {
+		err := dumpInstalledObject(dir, agentPolicy)
```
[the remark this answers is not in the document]
mrodm: I've just renamed it to dumpJSONResource.
Do you think it is worthy to move it to different file in the same package dump? @jsoriano  

Something like internal/dump/utils.go ? I would move that function and formatJSON

jsoriano (approved):
:rocket:

internal/dump/agentpolicies.go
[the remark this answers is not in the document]
jsoriano: Yeah, this could be moved to some common file, maybe called `json.go`, as you prefer.

jsoriano:
Failure in CI is related to this change.

mtojek (commented):
Do you think that there is an option to cover the implementation with tests? I'm thinking about  some end-to-end tests like [bash scripts](https://github.com/elastic/elastic-package/tree/main/scripts).
You could boot up the stack, dump default policies, and take it down.

EDIT:

I guess that you can also modify the [test-stack-command](https://github.com/elastic/elastic-package/blob/main/scripts/test-stack-command.sh) and run the dump command before taking down the stack.

cmd/dump.go
```diff
@@ -76,3 +97,66 @@ func dumpInstalledObjectsCmd(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.AgentPolicyFlagName)
+	}
+
+	outputPath, err := cmd.Flags().GetString(cobraext.DumpOutputFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.DumpOutputFlagName)
+	}
+
+	tlsSkipVerify, _ := cmd.Flags().GetBool(cobraext.TLSSkipVerifyFlagName)
```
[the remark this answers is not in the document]
jsoriano: Yes, this flag should continue working after the changes in https://github.com/elastic/elastic-package/pull/847.

internal/kibana/policies.go
```diff
@@ -72,6 +72,62 @@ func (c *Client) GetPolicy(policyID string) (*Policy, error) {
 	return &resp.Item, nil
 }
 
+// GetRawPolicy fetches the given Policy with all the fields in the Ingest Manager.
```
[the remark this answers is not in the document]
mrodm: Updating comments here. I was not totally sure about whether or not they were referring to the same.
Thanks!

mrodm:
> Do you think that there is an option to cover the implementation with tests? I'm thinking about some end-to-end tests like [bash scripts](https://github.com/elastic/elastic-package/tree/main/scripts). You could boot up the stack, dump default policies, and take it down.
> 
> EDIT:
> 
> I guess that you can also modify the [test-stack-command](https://github.com/elastic/elastic-package/blob/main/scripts/test-stack-command.sh) and run the dump command before taking down the stack.


I was planing to add tests in a following branch if there is no inconvenience.
I will check those scripts and the `test-stack-command`, and I also wanted to give it a chance to what @jsoriano mentioned [in this comment](https://github.com/elastic/elastic-package/pull/862#pullrequestreview-1016671480)

internal/kibana/client.go
```diff
@@ -80,19 +80,19 @@ func CertificateAuthority(certificateAuthority string) ClientOption {
...
 
-func (c *Client) put(resourcePath string, body []byte) (int, []byte, error) {
+func (c *Client) Put(resourcePath string, body []byte) (int, []byte, error) {
 	return c.sendRequest(http.MethodPut, resourcePath, body)
 }
 
-func (c *Client) delete(resourcePath string) (int, []byte, error) {
+func (c *Client) Delete(resourcePath string) (int, []byte, error) {
 	return c.sendRequest(http.MethodDelete, resourcePath, nil)
 }
```
mrodm, line 97: In order to be used in `internal/kibana/test/http_test.go` I had to set get method as public. The others I've changed it to ensure consistency.

internal/dump/json.go
```diff
@@ -0,0 +1,44 @@
...
+type DumpableJSONResource interface {
+	Name() string
+	JSON() []byte
+}
+
+func dumpJSONResource(dir string, object DumpableJSONResource) error {
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return fmt.Errorf("failed to create dump directory: %w", err)
+	}
+	formatted, err := formatJSON(object.JSON())
```
mtojek, line 25: There is `formatter` package in elastic-package, maybe you can use it here.

mtojek (approved):

internal/kibana/test/httptest.go
```diff
@@ -0,0 +1,74 @@
...
+	}))
+}
+
+var pathReplacer = strings.NewReplacer("/", "-", "*", "_", "&", ".")
+
+// FIXME duplicated in internal/elasticsearch/test/http_test.go
+func pathForURL(url string) string {
+	clean := strings.Trim(url, "/")
+	if len(clean) == 0 {
+		return "root.jsn"
```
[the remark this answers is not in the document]
mrodm: Right, it must be json
Thanks!

jsoriano (approved):

internal/dump/json.go
mrodm: Oh, ok! Better reuse the formatter package.
I'll use it and update all the dumped files.
Thanks!

internal/kibana/agents.go
```diff
@@ -41,7 +41,7 @@ func (a *Agent) String() string {
 
 // ListAgents returns the list of agents enrolled with Fleet.
 func (c *Client) ListAgents() ([]Agent, error) {
-	statusCode, respBody, err := c.get(fmt.Sprintf("%s/agents", FleetAPI))
+	statusCode, respBody, err := c.Get(fmt.Sprintf("%s/agents", FleetAPI))
```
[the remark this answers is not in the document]
mrodm: I'll check this option. I prefer to keep those methods as private if possible.

internal/dump/testdata/elasticsearch-7-apache-dump-all/ingest_pipelines/logs-apache.access-1.3.4.json
```diff
@@ -1,202 +1,202 @@
...
+            "set": {
+                "field": "event.category",
+                "value": "web"
+            }
+        },
+        {
+            "set": {
+                "field": "event.outcome",
+                "value": "success",
+                "if": "ctx?.http?.response?.status_code != null \u0026\u0026 ctx.http.response.status_code \u003c 400"
```
jsoriano, line 77: Take into account that the JSONformatter is replacing invalid HTML characters with the UTF-8 encoding, this is done [by default](https://pkg.go.dev/encoding/json#Marshal) for HTML purpouses, but not really needed here. Previous formatter was just indenting the files for readability, without Marshalling and Unmarshalling them.

In any case I guess it doesn't mind here if we are comparing documents formatted in the same way.

jsoriano (approved):

internal/dump/testdata/elasticsearch-7-apache-dump-all/ingest_pipelines/logs-apache.access-1.3.4.json
mtojek: I would keep it as is for consistency reasons - the same formatter used everywhere, unless HTML entities are considered to be a bug in this case.

jsoriano: Well, thinking a second time about this, this is not only used for testing. We have to take into account that these are dumps from Elasticsearch/Kibana, we are modifying them here, what could be unexpected, a user dumping these objects with `elastic-package` is going to see something different to what is seen in Kibana or using the APIs.

Also, if we modify them, we should also check that they continue being valid and generate the same result if imported back in Elasticsearch/Kibana.

Thinking about this, and the purpouse of the dump command I would suggest to go back to the previous formatting.

mrodm: Ok, I see now the reason there was an explicit formatJSON method previously.
As you mentioned, it should be equivalent. However, as these dump commands are intended to be used as an exploratory tool for users, I think it would be better to just revert this change/commit and use the previous one, so these files can be inspected/debugged easier.

internal/dump/json.go
mrodm: According to this https://github.com/elastic/elastic-package/pull/862#discussion_r908489425 , I'll keep using this method (formatJSON) to write the files.

mtojek: Yes, please do :)

PR #872, opened 2022-06-27
internal/testrunner/runners/pipeline: replace jsondiff output with difflib

efd6:
This gives a correct output for arrays and provides meaningful context for finding issues in test-case source. (Note also that this is the library that is used in beats for diff output).

Please take a look.

### DIFF
--- a/internal/testrunner/runners/pipeline/test_result.go
+++ b/internal/testrunner/runners/pipeline/test_result.go
@@ -12,8 +12,9 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/nsf/jsondiff"
+	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
+	"github.com/pmezard/go-difflib/difflib"
 
 	"github.com/elastic/elastic-package/internal/common"
 	"github.com/elastic/elastic-package/internal/testrunner"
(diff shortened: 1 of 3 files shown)


internal/testrunner/runners/pipeline/test_result.go
```diff
@@ -103,24 +104,37 @@ func compareJsonNumbers(a, b json.Number) bool {
...
+	}
+	err = jsonUnmarshalUsingNumber(got, &gotVal)
+	if err != nil {
+		return "", fmt.Errorf("invalid got value: %w", err)
+	}
+	if cmp.Equal(gotVal, wantVal, cmp.Comparer(compareJsonNumbers)) {
+		return "", nil
+	}
 
-	result, diff := jsondiff.Compare(want, got, &opts)
```
jsoriano, line 118: We could keep this comparison method even if later we change the diff that we print, right?

internal/testrunner/runners/pipeline/test_result.go
```diff
@@ -103,24 +104,37 @@ func compareJsonNumbers(a, b json.Number) bool {
...
-	switch result {
-	case jsondiff.FirstArgIsInvalidJson, jsondiff.SecondArgIsInvalidJson, jsondiff.BothArgsAreInvalidJson:
-		return "", errors.New("invalid json")
+	got, err = marshalNormalizedJSON(gotVal)
+	if err != nil {
+		return "", err
 	}
-	return diff, nil
+	want, err = marshalNormalizedJSON(wantVal)
```
jsoriano, line 124: `marshalNormalizedJSON` marshals and unmarshals using `jsonUnmarshalUsingNumber`, something that has been already done here. We could do the final marshal directly here instead of repeating the operations.
Not so important though as this code is only used if the documents are different.

jsoriano (commented):
I am ok with the general approach to change the diff that is printed. Added some questions.

internal/testrunner/runners/pipeline/test_result.go
```diff
@@ -103,24 +104,37 @@ func compareJsonNumbers(a, b json.Number) bool {
...
-
-	// Remove colored output.
-	opts.Added = jsondiff.Tag{Begin: "+ "}
-	opts.Removed = jsondiff.Tag{Begin: "- "}
-	opts.Changed = jsondiff.Tag{}
-	opts.Skipped = jsondiff.Tag{}
-
-	// Configure diff.
-	opts.SkipMatches = true
-	opts.CompareNumbers = compareJsonNumbers
```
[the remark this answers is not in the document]
efd6: I was originally thinking that, but the logic behind using it made a lot of sense, to I used it as the `Comparer` for `cmp.Equal`.

internal/testrunner/runners/pipeline/test_result.go
efd6: Yes, that's right.

internal/testrunner/runners/pipeline/test_result.go
efd6: I was being conservative to ensure that the values were being treated the same way and to ensure that the reader was clear that this was happening.

PR #879, opened 2022-06-30
Include server auth key usage in service certificates

jsoriano:
Including this key usage makes Chrome on OSX to show the "Proceed
anyway" button.

Seen in https://stackoverflow.com/questions/58802767/no-proceed-anyway-option-on-neterr-cert-invalid-in-chrome-on-macos/64309893#64309893

Thanks @mtojek for the suggestion!


mrodm:
Checked in Chrome on OSX, adding that extension the button "Proceed to localhost" is shown 💯 

Warning_with_proceed

mrodm (approved):

PR #882, opened 2022-06-30
Include licenses in packages

jsoriano:
Include a new `LICENSE.txt` file when building packages that include the `source.license` field in their manifests.

Follow up of https://github.com/elastic/package-spec/pull/355.
Part of https://github.com/elastic/package-spec/issues/298.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -152,3 +152,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/elastic/package-spec => github.com/elastic/package-spec v1.12.2-0.20220629092545-9ee240e557a5
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -26,9 +26,6 @@ type BuildOptions struct {
 
 	CreateZip   bool
 	SignPackage bool
-
-	// buildDir is the base build directory to use. Used for testing now.
-	buildDir string
 }
 
 // BuildDirectory function locates the target build directory. If the directory doesn't exist, it will create it.
@@ -71,14 +68,15 @@ func findBuildDirectory() (string, bool, error) {
 }
 
 // BuildPackagesDirectory function locates the target build directory for the package.
-func BuildPackagesDirectory(buildDir, packageRoot string) (string, error) {
-	var err error
-	if buildDir == "" {
-		buildDir, err = buildPackagesRootDirectory()
-		if err != nil {
-			return "", errors.Wrap(err, "can't locate build packages root directory")
-		}
+func BuildPackagesDirectory(packageRoot string) (string, error) {
+	buildDir, err := buildPackagesRootDirectory()
+	if err != nil {
+		return "", errors.Wrap(err, "can't locate build packages root directory")
 	}
+	return buildPackagesDirectoryWithBuildDir(packageRoot, buildDir)
+}
+
+func buildPackagesDirectoryWithBuildDir(packageRoot, buildDir string) (string, error) {
 	m, err := packages.ReadPackageManifestFromPackageRoot(packageRoot)
 	if err != nil {
 		return "", errors.Wrapf(err, "reading package manifest failed (path: %s)", packageRoot)
@@ -148,7 +146,16 @@ func FindBuildPackagesDirectory() (string, bool, error) {
 
 // BuildPackage function builds the package.
 func BuildPackage(options BuildOptions) (string, error) {
-	destinationDir, err := BuildPackagesDirectory(options.buildDir, options.PackageRoot)
+	buildDir, err := buildPackagesRootDirectory()
+	if err != nil {
+		return "", err
+	}
+
+	return buildPackageWithBuildDir(options, buildDir)
+}
+
+func buildPackageWithBuildDir(options BuildOptions, buildDir string) (string, error) {
+	destinationDir, err := buildPackagesDirectoryWithBuildDir(options.PackageRoot, buildDir)
 	if err != nil {
 		return "", errors.Wrap(err, "can't locate build directory")
 	}
@@ -191,7 +198,7 @@ func BuildPackage(options BuildOptions) (string, error) {
 	}
 
 	if options.CreateZip {
-		return buildZippedPackage(options, destinationDir)
+		return buildZippedPackage(options, buildDir, destinationDir)
 	}
 
 	err = validator.ValidateFromPath(destinationDir)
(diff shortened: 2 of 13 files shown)


go.mod
```diff
@@ -152,3 +152,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/elastic/package-spec => github.com/elastic/package-spec v1.12.2-0.20220629092545-9ee240e557a5
```
jsoriano, line 156: To be removed before merging, once https://github.com/elastic/package-spec/pull/355 is included in the used package-spec.

internal/packages/packages.go
```diff
@@ -110,7 +110,10 @@ type PackageManifest struct {
 	Owner           Owner            `config:"owner" json:"owner" yaml:"owner"`
 	Description     string           `config:"description" json:"description" yaml:"description"`
 	License         string           `config:"license" json:"license" yaml:"license"`
-	Categories      []string         `config:"categories" json:"categories" yaml:"categories"`
+	Source          struct {
+		License string `config:"license" json:"license" yaml:"license"`
```
mtojek, line 114: Do you plan to deprecate the license field in this PR?

internal/builder/packages.go
```diff
@@ -150,6 +173,18 @@ func BuildPackage(options BuildOptions) (string, error) {
 		return "", errors.Wrap(err, "clearing package contents failed")
 	}
 
+	m, err := packages.ReadPackageManifestFromPackageRoot(options.PackageRoot)
+	if err != nil {
+		return "", errors.Wrapf(err, "reading package manifest failed (path: %s)", options.PackageRoot)
+	}
+	if license := m.Source.License; license != "" {
+		logger.Debugf("Write license text for %q", license)
+		err := licenses.WriteTextToFile(license, filepath.Join(destinationDir, "LICENSE.txt"))
```
mtojek, line 182: Will LICENSE.txt be defined in the package-spec?

mtojek, line 182: Is there an option to use your own license? Will it break the building logic here?

internal/builder/packages.go
```diff
@@ -138,9 +152,18 @@ func FindBuildPackagesDirectory() (string, bool, error) {
...
+	if options.buildDir == "" {
+		destinationDir, err = BuildPackagesDirectory(options.PackageRoot)
+		if err != nil {
+			return "", errors.Wrap(err, "can't locate build directory")
+		}
+	} else {
+		destinationDir, err = buildPackagesDirectory(options.buildDir, options.PackageRoot)
+		if err != nil {
+			return "", errors.Wrap(err, "can't locate build directory")
+		}
```
mtojek, line 166: Why exactly do you need this if-else block? It looks like it complicated the current logic.

internal/builder/packages.go
```diff
@@ -25,6 +26,9 @@ type BuildOptions struct {
 
 	CreateZip   bool
 	SignPackage bool
+
+	// buildDir is the base build directory to use. Used for testing now.
+	buildDir string
```
mtojek, line 31: Similar question: why is this field required?

mtojek (commented):
I think that you may want to take a look also at the `create` command and add an option to select licenses there.

internal/packages/packages.go
jsoriano: It is [deprecated in the package-spec](https://github.com/elastic/package-spec/pull/355/files#diff-654fe06534f569b9f336e21e2cac527bfb3d15a9d19cf6f3995915ed2445f208R35).
I think we are not doing anything with deprecations in `elastic-package`, do you have a proposal for something that we could do here?

internal/builder/packages.go
jsoriano: I hadn't considered it, but it could very well be there, I will open a PR.

jsoriano: Not at the moment, do we want to support it?

If we want, we could relax the check of unknown licenses, and include whatever the developer places in the `source.license` field and the `LICENSE.txt` file.

jsoriano: (it will break the current logic because it assumes that `source.license` if set, is set to one of the values enumerated in the spec)

internal/builder/packages.go
jsoriano: I added `options.buildDir` to have a predictable place where packages are built for testing.

I can move this logic to the new internal method. I didn't modify the behaviour or interface of `BuildPackagesDirectory` because it is a exposed method, though it is in an internal package.

internal/builder/packages.go
jsoriano: To have a predictable place where packages are built for testing, without affecting the developer working copy.

internal/builder/packages.go
jsoriano: Moved logic to the current `BuildPackagesDirectory` and changed the other use we have of this.

internal/builder/packages.go
jsoriano: https://github.com/elastic/package-spec/pull/367

internal/packages/packages.go
mtojek: I would drop a comment in the source in this place as both fields are modeled. Frankly speaking, it's another argument for switching to a common place with business/model structures.

internal/builder/packages.go
mtojek: It's probably a matter preference, but I would rather keep exposed fields only in `BuildOptions` and introduce another function for testing purposes. For example:

`BuildPackage(options BuildOptions)` calls `buildPackageWithBuildDir(options BuildOptions, buildDir string)`. You can use the second one for testing.  Same story with `BuildPackagesDirectory`.

WDYT?

internal/licenses/_static/Apache-2.0.txt
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
mtojek, line 190: Should we render this data or remove the entry?

jsoriano: This comment https://github.com/elastic/package-spec/pull/367#discussion_r911148495 applies here.

internal/packages/packages.go
jsoriano: > Frankly speaking, it's another argument for switching to a common place with business/model structures.

Yep, another +1 for something like https://github.com/elastic/package-spec/issues/322.

mtojek (commented):
[Recent changes](https://github.com/elastic/elastic-package/pull/882/commits/82da4f7f791db0b3390157ca4890629b2a15ea34) look good to me, but will hold my "approval" until we clarify [this point](https://github.com/elastic/package-spec/pull/367#discussion_r912747074).

Issue #584, opened 2021-11-15
Make sure that our tools support the Package Storage v2


mtojek:
The goal of this issue is to make that elastic-package fully supports packages distributed in the zipped form, not just building them, but also consuming ones integrated into existing Package Storage distributions. At the moment, we're using unzipped packages to build custom Package Registry distributions.

mtojek:
As I'm blocked with other issues (deployment, adjusting Integrations pipeline) I can take a look at this one.

PR #886, opened 2022-07-07
Use .zip packages by default in Elastic stack

mtojek:
Fixes: https://github.com/elastic/elastic-package/issues/584


internal/files/copy.go
```diff
@@ -50,9 +56,10 @@ func copy(sourcePath, destinationPath string, skippedDirs []string) error {
...
 
-func shouldDirectoryBeSkipped(name string, skippedDirs []string) bool {
+// shouldDirectoryBeSkipped function checks if absolute path or last element should be skipped.
+func shouldDirectoryBeSkipped(path string, skippedDirs []string) bool {
 	for _, d := range skippedDirs {
-		if name == d {
+		if path == d || filepath.Base(path) == d {
 			return true
 		}
 	}
```
mrodm, line 65: As a suggestion, `skippedDirs` parameter could be changed to be of type `map[string]bool` and just look whether or not the key is present. That would avoid to compare all the elements for every element in the directory when executing `CopyWithSkipped`.

It's also true that this would just have some benefits in the boot process (BootUp function) according to the current usages of this function, so not sure if it can be kept just an array for now. WDYT ?

Some data, with the tests I've being doing with metrics I saw that there are at least 1584 packages being indexed currently.

```go
func shouldDirectoryBeSkipped(path string, skippedDirs map[string]bool) bool {
    if _, found := skippedDirs[path]; found {
        return true
    }
    if _, found := skippedDirs[filepath.Base(path)]; found {
        return true
    }
    return false
}
```

mrodm (approved):
LGTM, just a suggestion regarding `shouldDirectoryBeSkipped` function

internal/files/copy.go
mtojek: Keep in mind that the real use case for this is building one package at a time and including it in the stack. Also, the skippedDirs list will contain max. 1-2 entries on average, so I would keep the array here.

Let me know what you think :)

mrodm: Ah! Ok, I thought that the `BootUp` function was related to walk through all the packages and versions. If that is intended for just one package with just a few entries, then let's just keep the array of directories 👍

PR #887, opened 2022-07-11
Bump up default Elastic stack version to 8.3.2

mtojek:
This PR bumps up the default Elastic stack version to v8.3.2. It looks like we missed the v8.3.0 release.

### DIFF
--- a/internal/install/stack_version.go
+++ b/internal/install/stack_version.go
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "8.2.0"
+	DefaultStackVersion = "8.3.2"
 )


internal/install/stack_version.go
```diff
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "8.2.0"
+	DefaultStackVersion = "8.3.2"
```
mrodm, line 9: Should the Makefile target ([test-stack-command-8x](https://github.com/elastic/elastic-package/blob/main/Makefile#L64)) be updated to use that snapshot version ?

mtojek: Actually, the `8.3.0-SNAPSHOT` will pick up always latest `8.3.x`. In this case, it would the next patch release after 8.3.2, so we're safe :)

mrodm: Ok! 👍

mrodm (approved):

PR #897, opened 2022-07-18
Add required subscription on package creation

jsoriano:
It adds by now the old `license` field and the new `elastic.subscription` condition.

### DIFF
--- a/cmd/create_package.go
+++ b/cmd/create_package.go
@@ -19,13 +19,14 @@ const createPackageLongDescription = `Use this command to create a new package.
 The command can bootstrap the first draft of a package using embedded package template and wizard.`
 
 type newPackageAnswers struct {
-	Name          string
-	Version       string
-	Title         string
-	Description   string
-	Categories    []string
-	KibanaVersion string `survey:"kibana_version"`
-	GithubOwner   string `survey:"github_owner"`
+	Name                string
+	Version             string
+	Title               string
+	Description         string
+	Categories          []string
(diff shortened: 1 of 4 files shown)


cmd/create_package.go
```diff
@@ -81,10 +82,19 @@ func createPackageCommandAction(cmd *cobra.Command, args []string) error {
...
-				Default: surveyext.DefaultConstraintValue(),
+				Default: surveyext.DefaultKibanaVersionConditionValue(),
 			},
 			Validate: survey.ComposeValidators(survey.Required, surveyext.ConstraintValidator),
 		},
+		{
+			Name: "elastic_subscription",
+			Prompt: &survey.Select{
+				Message: "Required Elastic subscription:",
+				Options: []string{"basic", "gold", "platinum", "enterprise"},
```
mrodm, line 93: nit: I would change the order to be shown like [here](https://www.elastic.co/subscriptions)

```suggestion
				Options: []string{"basic", "platinum", "enterprise", "gold"},
```

mrodm (approved):
LGTM

cmd/create_package.go
jsoriano: Interesting, I didn't know that gold is discontinued, I guess that this is the only reason why it appears later there. But technically platinum > gold, we have been using this order everywhere. See the spec for example: https://github.com/elastic/package-spec/blob/0456e4f000a5bf85130d4e5e9c6d329a627523ec/versions/1/integration/manifest.spec.yml#L54

PR #902, opened 2022-07-21
Build command: skip package validation

mtojek:
Related: https://github.com/elastic/endpoint-package/pull/272

The goal of this PR is to expose an option to disable validation of built packages if we have acknowledged all spec issues. It blocks endpoint-package to publish artifacts to Package Storage v2.

PR #903, opened 2022-07-21
Fix: read skip-validation flag

mtojek:
This PR fixes the bug in the [previous PR](https://github.com/elastic/elastic-package/pull/902). The `--skip-validation` flag wasn't read from CLI.


mtojek:
self-approved to unblock the related issue.

PR #904, opened 2022-07-21
Pass policy_template name to Kibana for system tests

### LINKED ISSUE #899: System tests fail in packages using policy_templates
While running the system tests for AWS we saw abnormal behavior. Before adding securityhub_findings and securityhub_insights data streams in the AWS package system tests are getting passed(only ec2_metrics and redshift data streams have system tests).

When the system tests are added for securityhub data stream it gives us an error in the previously passing tests. Below mentioned are the two scenarios that we observed in the AWS package.

1st Scenario

When the policy template of securityhub is kept first in the manifest file, the system test fails. It gives the error `"Input template not found, unable to find input type aws/metrics"` (ec2_metrics data-stream). In ec2_metrics <service deployer> is `terraform`.
And in this scenario, if we run a system test for securityhub_findings and securityhub_insights data streams only ( elastic-package test system -d securityhub_findings,securityhub_insights ) it is passing.

image

2nd Scenario

When the policy template of securityhub is kept at the last level in the manifest file, the system test fails. It gives the error `"input template not found, unable to find input type httpjson"` (securityhub data-streams). In securityhub <service deployer> is `docker`.

image

andrewkroh:
When calling `POST /api/fleet/package_policies` for packages that
containing multiple policy_templates, the name of the policy_template
associated to the input under test should be specified.

If the policy_template is not specified then Kibana will default to
the first policy_template which may not contain the input or may
not contain the expected package level variables for the test.

This affected testing packages like AWS.

The error that this fixes is:

    Error: error running package system tests: could not complete test run:
    could not add data stream config to policy: could not add package to policy;
    API status code = 500; response body = {"statusCode":500,
    "error":"Internal Server Error","message":"Input template not found, unable
    to find input type logfile"}

Fixes #899


jsoriano (approved):

jsoriano:
@andrewkroh sorry for the late review, can you please take a look to the merge conflicts? Thanks!

andrewkroh:
Conflicts addressed. ✅

jsoriano (approved):

Issue #855, opened 2022-06-21
Performance testing for ingest pipelines


adriansr:
This is a proposed first iteration of performance benchmarks for Elastic packages, part of a broader effort to provide tooling to fully understand the performance or storage impacts of changes to ingest pipelines and other assets.

The objective of this iteration is to provide visibility into a package's ingest-pipelines performance for a developer working on a package. To achieve this, it makes use of existing facilities and data already available to `elastic-package` in order to reduce development time and cross-team dependencies:

- Uses existing [Elasticsearch's Nodes Stats API ](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html), already used by `elastic-package` for pipeline coverage.
- Uses existing testing logs/documents from ingest pipeline tests.
- Reports it's output to stdout in a format similar to tests.
- Works as an additional stage for pipeline tests (i.e. `elastic-package test pipeline --bench`), in opposition to a completely new action in elastic-agent (`elastic-package benchmark pipeline`). This reduces code complexity.

### First stage

- [x] Add performance measurement to package's pipeline tests (`--bench`). #958
- [x] Text output: Raw (eps) performance, top N ingest processors by time spent. #958
- ~~[ ] IP address randomization to properly exercise geoip processors.~~ Unfeasible with pipeline tests, as only a limited number of IP addresses is allowed.

### Further development

- [x] xUnit output for CI reports. Needs investigation as benchmark representation in xUnit is not clearly specified and may need a bit of trial and error. #958 (will need improvements).
- ~[ ] #1040~

marc-gr:
> * [ ]  Benchmarking against previous package version. Need to figure out what's the best way to implement this, possibly just comparing the output of two benchmark runs, similar to Go's benchcmp / [benchstat.](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat)

This requires to add the git commit as part of the registry metadata (https://github.com/elastic/package-spec/issues/446)

marc-gr:
Moved last point to https://github.com/elastic/elastic-package/issues/1040 as it is blocked and requires of specific work and is out of the first stage

PR #906, opened 2022-07-21
Add benchmarks to pipeline tests

adriansr:
This PR adds initial benchmarking support to pipeline tests.

### Configuration

3 new flags are added to the `test` subcommand:

* `--bench`: Enables benchmarking.
* `--bench-count`: Number of documents to use for benchmarking (default 1000).
* `--bench-duration`: Adjust the number of docs so that the benchmark runs for this duration (count above is ignored).

### Output (text format)

```
╭─────────────────────────────────╮
│ parameters                      │
├──────────────────┬──────────────┤
│ package          │        azure │
│ data_stream      │ activitylogs │
│ source doc count │            5 │
│ doc count        │         1367 │
╰──────────────────┴──────────────╯
╭──────────────────────╮
│ ingest performance   │
├─────────────┬────────┤
│ ingest time │  4.87s │
│ eps         │ 280.64 │
╰─────────────┴────────╯
╭─────────────────────────────────────────────╮
│ processors by total time                    │
├─────────────────────────────────────┬───────┤
│ script @ default.yml:192            │ 4.09% │
│ script @ default.yml:220            │ 4.04% │
│ script @ default.yml:199            │ 3.61% │
│ script @ default.yml:206            │ 3.59% │
│ json @ default.yml:20               │ 3.20% │
│ grok @ azure-shared-pipeline.yml:72 │ 2.16% │
│ script @ default.yml:231            │ 1.99% │
│ script @ default.yml:92             │ 1.79% │
│ script @ default.yml:68             │ 1.75% │
│ grok @ azure-shared-pipeline.yml:52 │ 1.66% │
╰─────────────────────────────────────┴───────╯
╭─────────────────────────────────────────────────╮
│ processors by average time per doc              │
├─────────────────────────────────────┬───────────┤
│ script @ default.yml:192            │ 145.574µs │
│ script @ default.yml:220            │ 144.111µs │
│ script @ default.yml:199            │ 128.749µs │
│ script @ default.yml:206            │ 128.017µs │
│ json @ default.yml:20               │ 114.118µs │
│ grok @ azure-shared-pipeline.yml:72 │  95.978µs │
│ grok @ azure-shared-pipeline.yml:52 │   74.04µs │
│ grok @ azure-shared-pipeline.yml:62 │  71.297µs │
│ script @ default.yml:231            │  70.958µs │
│ grok @ azure-shared-pipeline.yml:31 │  68.555µs │
╰─────────────────────────────────────┴───────────╯
```

### Output (xUnit)

Uses the xUnit format as [defined in the Jenkins Benchmark plugin.](https://github.com/Autodesk/jenkinsci-benchmark-plugin/blob/master/doc/EXAMPLE_SCHEMA_XML_DEFAULT.md)

```
<?xml version="1.0" encoding="UTF-8"?>
<group name="pipeline benchmark for azure/activitylogs">
  <parameter name="package">
    <value>azure</value>
  </parameter>
  <parameter name="data_stream">
    <value>activitylogs</value>
  </parameter>
  <parameter name="source doc count">
    <value>5</value>
  </parameter>
  <parameter name="doc count">
    <value>1000</value>
  </parameter>
  <test name="ingest performance">
    <result name="ingest time">
      <description>time elapsed in ingest processors</description>
      <unit>s</unit>
      <value>0.267</value>
    </result>
    <result name="eps">
      <description>ingested events per second</description>
      <value>3745.318352059925</value>
    </result>
  </test>
</group>
```

For the plugin to work it's necessary to generate an output file for each datastream, so the report generation had to be modified a bit.

The detailed processor reports are not included in the xUnit format as they add a too much data for the Benchmark plugin to handle.

### Related issues

 - https://github.com/elastic/elastic-package/issues/855


leehinman:
I was wondering how this would do against a known performance issue so I ran this against version 1.3.2, of the crowdstrike package.  The `--bench` option shows a script processor that was taking up an outsized portion of the processing time.  One oddity, the output says line 54, but in the default.yml file the script is at line 83 in the file on disk.  

## 1.3.2 output
```
╭────────────────────────────────╮
│ parameters                     │
├──────────────────┬─────────────┤
│ package          │ crowdstrike │
│ data_stream      │         fdr │
│ source doc count │         125 │
│ doc count        │       10000 │
╰──────────────────┴─────────────╯
╭───────────────────────╮
│ ingest performance    │
├─────────────┬─────────┤
│ ingest time │   6.32s │
│ eps         │ 1582.03 │
╰─────────────┴─────────╯
╭─────────────────────────────────────────╮
│ processors by total time                │
├────────────────────────────────┬────────┤
│ script @ default.yml:54        │ 32.15% │
│ script @ default.yml:1273      │  3.72% │
│ json @ default.yml:9           │  3.61% │
│ append @ default.yml:396       │  3.27% │
│ geoip @ default.yml:947        │  2.82% │
│ append @ default.yml:1139      │  2.58% │
│ community_id @ default.yml:876 │  2.56% │
│ script @ default.yml:522       │  2.18% │
│ append @ default.yml:401       │  1.41% │
│ date @ default.yml:36          │  1.27% │
╰────────────────────────────────┴────────╯
╭─────────────────────────────────────────────────╮
│ processors by average time per doc              │
├─────────────────────────────────────┬───────────┤
│ script @ default.yml:613            │  231.25µs │
│ script @ default.yml:54             │   203.2µs │
│ script @ default.yml:522            │  143.75µs │
│ registered_domain @ default.yml:999 │  106.25µs │
│ uri_parts @ default.yml:936         │ 104.166µs │
│ date @ default.yml:1194             │      75µs │
│ date @ default.yml:1188             │      50µs │
│ script @ default.yml:1009           │   43.75µs │
│ set @ default.yml:749               │  35.416µs │
│ append @ default.yml:879            │  26.851µs │
╰─────────────────────────────────────┴───────────╯
```

I also ran against 1.4.1 to see how it looked after the fix, and we see that the performance issue with the script processor is fixed. :-)

## 1.4.1
```
╭────────────────────────────────╮
│ parameters                     │
├──────────────────┬─────────────┤
│ package          │ crowdstrike │
│ data_stream      │         fdr │
│ source doc count │         126 │
│ doc count        │       10000 │
╰──────────────────┴─────────────╯
╭───────────────────────╮
│ ingest performance    │
├─────────────┬─────────┤
│ ingest time │   4.15s │
│ eps         │ 2411.96 │
╰─────────────┴─────────╯
╭─────────────────────────────────────────╮
│ processors by total time                │
├─────────────────────────────────┬───────┤
│ script @ default.yml:2258       │ 5.91% │
│ json @ default.yml:9            │ 5.28% │
│ append @ default.yml:1337       │ 3.88% │
│ append @ default.yml:2123       │ 3.84% │
│ geoip @ default.yml:1891        │ 3.35% │
│ script @ default.yml:2051       │ 2.92% │
│ date @ default.yml:14           │ 2.89% │
│ script @ default.yml:1463       │ 2.46% │
│ append @ default.yml:1342       │ 1.69% │
│ community_id @ default.yml:1820 │ 1.57% │
╰─────────────────────────────────┴───────╯
│ processors by average time per doc    │
├───────────────────────────┬───────────┤
│ date @ default.yml:2172   │ 202.531µs │
│ date @ default.yml:2178   │ 151.898µs │
│ set @ default.yml:1876    │ 130.801µs │
│ script @ default.yml:1555 │ 125.786µs │
│ script @ default.yml:1463 │ 106.918µs │
│ date @ default.yml:44     │  75.949µs │
│ append @ default.yml:1359 │  50.632µs │
│ date @ default.yml:34     │  50.632µs │
│ script @ default.yml:2051 │  43.635µs │
│ script @ default.yml:1953 │  37.735µs │
╰───────────────────────────┴───────────╯
```

internal/testrunner/runners/pipeline/benchmark.go
```diff
@@ -0,0 +1,388 @@
...
+	dataStreamPath, found, err := packages.FindDataStreamRootForPath(options.TestFolder.Path)
+	if err != nil {
+		return ingestResult{}, errors.Wrap(err, "locating data_stream root failed")
+	}
+	if !found {
+		return ingestResult{}, errors.New("data stream root not found")
+	}
+
+	testCase := testCase{
+		events: docs,
```
leehinman, line 306: This might be better as a feature request.

I'm wondering if we could introduce `batch_size` here, and divide the number of docs into those batches and send each batch with `simulatePipelineProcessing`.  That would get around the limitation we have right now wher the number of docs must fit in one request and would make the processing more like what the users sees in production.  Might even be possible to concurrently call `simulatePipelineProcessing` to again be more like what the users experiences in production.

cmd/testrunner.go
```diff
@@ -68,6 +69,9 @@ func setupTestCommand() *cobraext.Command {
 	cmd.PersistentFlags().StringP(cobraext.ReportFormatFlagName, "", string(formats.ReportFormatHuman), cobraext.ReportFormatFlagDescription)
 	cmd.PersistentFlags().StringP(cobraext.ReportOutputFlagName, "", string(outputs.ReportOutputSTDOUT), cobraext.ReportOutputFlagDescription)
 	cmd.PersistentFlags().BoolP(cobraext.TestCoverageFlagName, "", false, cobraext.TestCoverageFlagDescription)
+	cmd.PersistentFlags().BoolP(cobraext.TestBenchFlagName, "", false, cobraext.TestBenchFlagDescription)
+	cmd.PersistentFlags().IntP(cobraext.TestBenchCountFlagName, "", 1000, cobraext.TestBenchCountFlagDescription)
+	cmd.PersistentFlags().DurationP(cobraext.TestPerfDurationFlagName, "", time.Duration(0), cobraext.TestPerfDurationFlagDescription)
```
mtojek, line 74: What does `time.Duration(0)` mean? Is it an infinite benchmark? Maybe it's fair to replace it with a finite value, e.g. 1 minute?

cmd/testrunner.go
```diff
@@ -122,6 +126,21 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
...
+		if err != nil {
+			return cobraext.FlagParsingError(err, cobraext.TestBenchFlagName)
+		}
+
+		testBenchCount, err := cmd.Flags().GetInt(cobraext.TestBenchCountFlagName)
+		if err != nil {
+			return cobraext.FlagParsingError(err, cobraext.TestBenchCountFlagName)
+		}
+
+		testBenchDur, err := cmd.Flags().GetDuration(cobraext.TestPerfDurationFlagName)
```
mtojek, line 139: nit: testBenchDuration

internal/testrunner/runners/pipeline/benchmark.go
```diff
@@ -0,0 +1,388 @@
...
+	// Minimum acceptable length for a benchmark result.
+	minDurationSeconds = 0.001 // 1ms
+
+	// How many top processors to return.
+	numTopProcs = 10
+)
+
+func BenchmarkPipeline(options testrunner.TestOptions) (*testrunner.BenchmarkResult, error) {
+	// Load all test documents
+	docs, err := loadAllTestDocs(options.TestFolder.Path)
```
mtojek, line 42: I guess that you need to prepare a spec PR first.

mtojek (commented):
Let's start with CLI design!

Based on your implementation, @adriansr, I have a feeling that there are many conditions to determine whether this is a pipeline test or a pipeline benchmark. Same with formatting reports - you need to pass/return another structure to process the correct test type.

What do you think about creating a new command action: `elastic-package benchmark`, which executes `elastic-package benchmark pipeline`. We will leave `benchmark` open for other subjects to measure and DON'T couple it only with `pipeline tests`.

@ecosystem Happy to hear your comments on this.

BTW we should start with the spec PR first, as it introduces new _dev artifacts to the package.

marc-gr:
Changes made:

- Add a dedicated `benchmark` command
- Use a dedicated `_dev/benchmark` folder to define config and samples instead of flags
- Moved pipeline management code to `elasticsearch/ingest`

internal/testrunner/runners/pipeline/benchmark.go
marc-gr: will wait to open that one until we agree on the options and structure here

cmd/benchrunner.go
```diff
@@ -0,0 +1,177 @@
...
+
+func benchTypeCommandActionFactory(runner benchrunner.BenchRunner) cobraext.CommandAction {
+	benchType := runner.Type()
+	return func(cmd *cobra.Command, args []string) error {
+		cmd.Printf("Run %s benchmarks for the package\n", benchType)
+
+		failOnMissing, err := cmd.Flags().GetBool(cobraext.FailOnMissingFlagName)
+		if err != nil {
+			return cobraext.FlagParsingError(err, cobraext.FailOnMissingFlagName)
+		}
```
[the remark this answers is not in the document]
marc-gr: I guess it depends. I was wondering if would be practical to add a config option to use the pipeline tests events as samples, to avoid repetition and basically to be able to leverage benchmarks for any package with pipeline tests without changes. If added, would fallback to lookup pipeline test samples, and I guess than in a scenario like this failing could be useful? WDYT?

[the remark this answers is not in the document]
mtojek: > If added, would fallback to lookup pipeline test samples, and I guess than in a scenario like this failing could be useful

Yes, in this case, it makes sense.

Based on your experience, how many cases would be covered by borrowing pipeline test events? Is it the majority or just w few samples? If that feature doesn't look to be popular, I would rather drop it.

cmd/benchrunner.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/elastic/elastic-package/internal/cobraext"
+	"github.com/elastic/elastic-package/internal/common"
+	"github.com/elastic/elastic-package/internal/elasticsearch"
+	"github.com/elastic/elastic-package/internal/packages"
+	"github.com/elastic/elastic-package/internal/signal"
+)
+
+const benchLongDescription = `Use this command to run benchmarks on a package. Currently, the following types of benchmarks are available:
+
+#### Pipeline Benchmarks
```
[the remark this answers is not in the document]
marc-gr:
```
Use this command to run benchmarks on a package. Currently, the following types of benchmarks are available:

#### Pipeline Benchmarks
These benchmarks allow you to benchmark any Ingest Node Pipelines defined by your packages.

For details on how to configure pipeline benchmarks for a package, review the [HOWTO guide](https://github.com/elastic/elastic-package/blob/main/docs/howto/pipeline_benchmarking.md).

Context: package

Usage:
  elastic-package benchmark [flags]
  elastic-package benchmark [command]

Available Commands:
  pipeline    Run pipeline benchmarks

Flags:
  -m, --fail-on-missing        fail if tests are missing
  -h, --help                   help for benchmark
      --report-format string   format of test report (default "human")
      --report-output string   output location for test report (default "stdout")

Global Flags:
  -v, --verbose   verbose mode

Use "elastic-package benchmark [command] --help" for more information about a command.
```

Similar fashion as the test help :)

cmd/benchrunner.go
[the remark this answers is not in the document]
marc-gr: I think would be nice to aim to have this in some capacity as part of the CI, so I think all packages would end up using it in one way or another. cc @leehinman

cmd/benchrunner.go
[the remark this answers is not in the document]
mtojek: I was thinking about the header `#### Pipeline Benchmarks`, but if it's aligned with `test`, you can leave it :)

mtojek:
BTW Please ignore the failing `elastic-package-package-storage-publish/pr-merge`. We have a different issue to solve it.

marc-gr:
Spec PR in https://github.com/elastic/elastic-package/pull/906

internal/elasticsearch/ingest/datastream.go
```diff
@@ -20,37 +19,20 @@ import (
...
-type simulatePipelineResponse struct {
-	Docs []pipelineIngestedDocument `json:"docs"`
-}
-
-type pipelineIngestedDocument struct {
-	Doc pipelineDocument `json:"doc"`
-}
-
-func installIngestPipelines(api *elasticsearch.API, dataStreamPath string) (string, []ingest.Pipeline, error) {
+func InstallDataStreamPipelines(api *elasticsearch.API, dataStreamPath string) (string, []Pipeline, error) {
```
mtojek, line 27: nit: is this renaming intentional? AFAIR there are packages with ingest pipelines detached from data streams, but most likely we don't support them here.

scripts/test-check-packages.sh
```diff
@@ -38,7 +38,7 @@ OLDPWD=$PWD
 for d in test/packages/${PACKAGE_TEST_TYPE:-other}/${PACKAGE_UNDER_TEST:-*}/; do
   (
     cd $d
-    elastic-package check -v
+    elastic-package build -v
```
mtojek, line 41: Actually, the `check` commands runs also `format` and `validate`. This test script is to verify the `check` call. Is there an option to revert this change?

test/packages/benchmarks/pipeline_benchmark/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "999.999.999"
+  changes:
+    - description: initial release
+      type: enhancement # can be one of: enhancement, bugfix, breaking-change
+      link: https://github.com/elastic/elastic-package/pull/209
```
mtojek, line 6: nit: let's link to this PR :)

scripts/test-check-packages.sh
marc-gr: this slipped through 🙏

internal/elasticsearch/ingest/datastream.go
marc-gr: Both signatures received datastream as a parameter, since I had to expose it, it seemed clearer to me to rename it.

internal/benchrunner/runners/pipeline/benchmark.go
```diff
@@ -0,0 +1,292 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/benchrunner"
+	"github.com/elastic/elastic-package/internal/elasticsearch/ingest"
+)
+
+const (
+	// How many top processors to return.
+	numTopProcs = 10
```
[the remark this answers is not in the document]
marc-gr: If we expose it through the cli, it will use the value for all the datastreams in the package, in some cases based on the size of the events used to test, this might cause some simulations to fail with a 413. This means some packages might require different num_doc values for each datastream, and this is why I added it to the benchmark config instead. Plus if the intention is to use it in CI, it is much less flexible to use because the same reason. WDYT?

jsoriano:
@marc-gr you may need to rebase the PR or recreate it, CI seems to be looking for Adrian as member of Elastic.

marc-gr:
Reopened https://github.com/elastic/elastic-package/pull/958

PR #910, opened 2022-07-26
document --version for stack up

djptek:
Mention `--version` in the documentation for `elastic-package stack up`

Adding this because it is not clear how to set a specific version when spinning up the stack.

I put this in the description to avoid confusion with e.g.

`elastic-package help version` which shows the version of `elastic-package`

I didn't add as a flag, since this is only relevant to `stack up` - that would really require breaking `stack` down into its components, however, that remains an option. Let me know


djptek:
note:

`elastic-package stack help` generates a help message at a higher level than `--version`
`elastic-package stack up help` starts the stack, no help message
`elastic-package stack up --help` generates a help message with detail on `--version`

mtojek (approved):

PR #913, opened 2022-07-27
feat: use 1.17 'go run' syntax in make tasks

kruskall:
Go 1.17 adds a go run syntax to run a specific version of
a binary.

We can replace the tools.go file with the new syntax and
avoid polluting the dependency graph.

See https://tip.golang.org/doc/go1.17#go%20run


mtojek:
As this repository is controlled by dependabot, I will close this PR.

PR #920, opened 2022-08-02
Include repository license in package when available

jsoriano:
If the package doesn't include a `LICENSE.txt` file, look for one in the repository, and include it if found.

Implements https://github.com/elastic/package-spec/issues/298#issuecomment-1202153924.

### DIFF
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -159,6 +159,7 @@ func BuildPackage(options BuildOptions) (string, error) {
 		return "", errors.Wrap(err, "copying package contents failed")
 	}
 
+	logger.Debug("Copy license file if needed")
 	err = copyLicenseTextFile(filepath.Join(destinationDir, licenseTextFileName))
 	if err != nil {
 		return "", errors.Wrap(err, "copying license text file")
@@ -243,20 +244,20 @@ func signZippedPackage(options BuildOptions, zippedPackagePath string) error {
 func copyLicenseTextFile(licensePath string) error {
 	_, err := os.Stat(licensePath)
 	if err == nil {
-		// License is already there, nothing to do.
+		logger.Debug("License file in the package will be used")
 		return nil
 	}
 
 	sourceLicensePath, err := findRepositoryLicense()
 	if errors.Is(err, os.ErrNotExist) {
-		logger.Debug("No license text file is included in package.")
+		logger.Debug("No license text file is included in package")
 		return nil
 	}
 	if err != nil {
 		return errors.Wrap(err, "failure while looking for license in repository")
 	}
 
-	logger.Debugf("License text found in %q will be included in package", sourceLicensePath)
+	logger.Infof("License text found in %q will be included in package", sourceLicensePath)
 	err = sh.Copy(licensePath, sourceLicensePath)
 	if err != nil {
 		return errors.Wrap(err, "can't copy license from repository")
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/magefile/mage/sh"
 	"github.com/pkg/errors"
 
 	"github.com/elastic/package-spec/code/go/pkg/validator"
@@ -19,6 +20,7 @@ import (
 )
 
 const builtPackagesFolder = "packages"
+const licenseTextFileName = "LICENSE.txt"
 
 type BuildOptions struct {
 	PackageRoot string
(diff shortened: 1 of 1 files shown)


internal/builder/packages.go
```diff
@@ -233,7 +240,53 @@ func signZippedPackage(options BuildOptions, zippedPackagePath string) error {
...
+	sourceLicensePath, err := findRepositoryLicense()
+	if errors.Is(err, os.ErrNotExist) {
+		logger.Debug("No license text file is included in package.")
+		return nil
+	}
+	if err != nil {
+		return errors.Wrap(err, "failure while looking for license in repository")
+	}
+
+	logger.Debugf("License text found in %q will be included in package", sourceLicensePath)
```
mtojek, line 259: maybe info?

internal/builder/packages.go
```diff
@@ -233,7 +240,53 @@ func signZippedPackage(options BuildOptions, zippedPackagePath string) error {
...
+	}
+
+	return nil
+}
+
 func createBuildDirectory(dirs ...string) (string, error) {
+	dir, err := findRepositoryRootDirectory()
+	if errors.Is(err, os.ErrNotExist) {
+		return "", errors.New("package can be only built inside of a Git repository (.git folder is used as reference point)")
+	}
```
mtojek, line 272: Does it change the implementation logic or is it just refactoring?

internal/builder/packages.go
```diff
@@ -157,6 +159,11 @@ func BuildPackage(options BuildOptions) (string, error) {
 		return "", errors.Wrap(err, "copying package contents failed")
 	}
```
mtojek, line 161: Please leave `logger.Debugf` here for consistency with other steps (clear target, encode dashboards, resolve fields).

internal/builder/packages.go
```diff
@@ -233,7 +240,53 @@ func signZippedPackage(options BuildOptions, zippedPackagePath string) error {
 	return nil
 }
 
+func copyLicenseTextFile(licensePath string) error {
+	_, err := os.Stat(licensePath)
+	if err == nil {
+		// License is already there, nothing to do.
```
mtojek, line 246: Debug message?

internal/builder/packages.go
```diff
@@ -233,7 +240,53 @@ func signZippedPackage(options BuildOptions, zippedPackagePath string) error {
 	return nil
 }
 
+func copyLicenseTextFile(licensePath string) error {
```
mtojek, line 243: I guess that you don't check if the license content has changed?

mtojek (commented):
I know that some packages are using NOTICE.txt to carry license and Kibana shows that file. I don't know what's the difference between NOTICE.txt and LICENSE.txt, but maybe this is something we need to sort out too.

internal/builder/packages.go
jsoriano: Here it does different things depending on the situation, so it will log two messages for the same thing, do we still want the log message here?

internal/builder/packages.go
jsoriano: 1. Ok.

internal/builder/packages.go
jsoriano: Just refactoring, this is basically a copy of what there was [in the loop](https://github.com/elastic/elastic-package/pull/920/files#diff-a018efea59016591a1ef36042025a57b31c426b7f3c0c45f83b6e8af81580585L247) looking for `.git`.

internal/builder/packages.go
jsoriano: Ok, message added, now it looks like this:

When copied from the repo:
```
2022/08/02 15:59:57 DEBUG Copy license file if needed
2022/08/02 15:59:57  INFO License text found in "/home/jaime/gocode/src/github.com/elastic/integrations/LICENSE.txt" will be included in package
```

When already available in the package:
```
2022/08/02 16:01:32 DEBUG Copy license file if needed
2022/08/02 16:01:32 DEBUG License file in the package will be used
```

internal/builder/packages.go
jsoriano: Added.

internal/builder/packages.go
mtojek: For example, the `check` command fails if README.md is outdated due to changed fields.

jsoriano:
> I know that some packages are using NOTICE.txt to carry license and Kibana shows that file. I don't know what's the difference between NOTICE.txt and LICENSE.txt, but maybe this is something we need to sort out too.

There was a discussion about this here: https://github.com/elastic/package-spec/issues/298#issuecomment-1173531846

internal/builder/packages.go
mtojek: Ok 👍

mtojek (approved):

PR #921, opened 2022-08-09
Add support for any shell for stack shellinit

endorama:
`elastic-package stack shellinit` output shell code needed to setup the testing environment and connect to the stack instance started via `elastic-package stack up`.

The current implementation emits POSIX shell code, compatible with `sh`, `zsh` and `bash`. The shellcode breaks on other shells, like `fish` - that I'm using. There are workarounds (like the awesome `babelfish`) but I think a built in solution would be preferred.

This PR aims to address this, adding a simple template mechanism that allows to support any shell through a custom shell code template, adds supports for `fish` shell and a `--shell` flag on the `shellinit` command to select the desired shell output.

This change is not breaking, as by default it outputs POSIX compatible shell code.


jsoriano (changes requested):
Nice! Maybe this is also a way to support other shells like powershell in the future :slightly_smiling_face:

Added some comments, the most important ones would be to decide if we want separated types for the different posix shells, and to define the new flag and their description in `cobraext` as is done in other modules. The rest of comments are suggestions about code organization.

andrewkroh (commented):
Thinking about UX and I'm wondering it it would be possible attempt determine the shell automatically such that it automatically does the correct thing for the user's environment (e.g. if we detect fish then output in the appropriate `fish` format).

I think the most reliable and cross-platform means (assuming we need this to work on Windows to output PowerShell format) of detecting the shell would be to get the executable name of the parent process. Like

```go
        // Using github.com/elastic/go-sysinfo.
	parent, err := sysinfo.Process(os.Getppid())
	if err != nil {
		log.Fatal(err)
	}

	parentInfo, err := parent.Info()
	if err != nil {
		log.Fatal(err)
	}

	shell := filepath.Base(parentInfo.Exe)
```

Reading the `SHELL` env var could be a simple option, but if a user has launched a subshell explicitly then it will be wrong (e.g. opening `fish` from `bash`).

mtojek (dismissed):
Could you please explain what's wrong with the original implementation? I might be missing something. Fish seems to be supported too.

```bash
$ elastic-package completion
Generate the autocompletion script for elastic-package for the specified shell.
See each sub-command's help for details on how to use the generated script.

Usage:
  elastic-package completion [command]

Available Commands:
  bash        Generate the autocompletion script for bash
  fish        Generate the autocompletion script for fish
  powershell  Generate the autocompletion script for powershell
  zsh         Generate the autocompletion script for zsh
```

```bash
$ elastic-package completion fish
```

mtojek:
I misread your intention, thanks @jsoriano for pointing it. I thought it's about `completion` not `shellinit`. In this case, I would research the solution proposed by Andrew. You could use `SHELL` env to detect the shell.

internal/stack/shellinit/shell/types.go
```diff
@@ -0,0 +1,17 @@
+package shell
+
+// AvailableShellTypes returns all available shell types in human format
+func AvailableShellTypes() []string {
+	return []string{"bash", "fish", "sh", "zsh"}
```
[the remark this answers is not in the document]
endorama: This change makes a lot of sense to me!

internal/stack/shellinit/shell/code.go
```diff
@@ -0,0 +1,19 @@
+package shell
```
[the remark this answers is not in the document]
endorama: I think this can go in 2 direction: 1 is removing the package, the other is making this more generic (so the template is not a string allows creating the init code through a builder pattern for example).

I'm moving everything in a single file to keep it simple (given the POSIX enhancement you proposed) as we don't need anything fancy at the moment.

endorama:
I updated the code with Jaime suggestion:
- remove the `shell` package and put everything in `stack`
- updated cobra flag to `cobraext`
- simplified templates to `posix` and `fish`

Instead of exposing `posix` to users, I kept `bash`, `sh`, `zsh` as I think is more ergonomic (as a user I don't have to worry about if my shell is compatible, just which shell I'm using) and I want to dig into automating this.

I would strongly advise against using `SHELL`, as [from the docs](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08), it 
> shall represent a pathname of the user's preferred command language interpreter

This can create issues when running the command in subshell or automating it and depends on a parameter that may not be configurable (think running this in the CI), leading to "fail first, configure later" scenario that I think are 

I like Andrew's approach, just want to double check if you are ok adding `elastic/go-sysinfo` as a dependency, as is not currently used. If you're ok, I would change default flag value to empty and use that as a signal to trigger the autodetection. I would still allow overriding this with an explicit value.

cmd/stack.go
```diff
@@ -198,19 +198,25 @@ func setupStackCommand() *cobraext.Command {
 				return cobraext.FlagParsingError(err, cobraext.ProfileFlagName)
 			}
 
+			s, err := cmd.Flags().GetString(cobraext.ShellInitShellFlagName)
```
jsoriano, line 201: Please use a more meaningful name here.
```suggestion
			shellName, err := cmd.Flags().GetString(cobraext.ShellInitShellFlagName)
```

internal/stack/shellinit.go
```diff
@@ -23,24 +24,57 @@ var (
...
 }
+
+const (
+	// shell init code for POSIX compliant shells.
+	// IEEE POSIX Shell and Tools portion of the IEEE POSIX specification (IEEE Standard 1003.1)
+	posixTemplate = `export %s=%s
+	export %s=%s
+	export %s=%s
+	export %s=%s
+	export %s=%s
```
jsoriano, line 51: This is adding spaces before each `export` right? Can you check the output of `elastic-package stack shellinit`?

internal/stack/shellinit.go
```diff
@@ -23,24 +24,57 @@ var (
...
+	case "bash":
+		return posixTemplate
+	case "fish":
+		return fishTemplate
+	case "sh":
+		return posixTemplate
+	case "zsh":
+		return posixTemplate
+	default:
+		panic("shell type is unknown, should be one of " + strings.Join(availableShellTypes, ", "))
```
jsoriano, line 78: `s` comes directly from user input, we shouldn't panic in this case.

jsoriano (changes requested):

jsoriano:
> I like Andrew's approach, just want to double check if you are ok adding `elastic/go-sysinfo` as a dependency, as is not currently used. If you're ok, I would change default flag value to empty and use that as a signal to trigger the autodetection. I would still allow overriding this with an explicit value.

I think it would be fine to add this dependency if it helps to have a better UX.

internal/stack/shellinit.go
endorama: fixed in [`7a98775` (#921)](https://github.com/elastic/elastic-package/pull/921/commits/7a9877558c9dc80c47fdfd3bb15fc1a481f070d3)

cmd/stack.go
endorama: fixed in [`18364c6` (#921)](https://github.com/elastic/elastic-package/pull/921/commits/18364c69f1cb0aa809e48a32b48d538767b939c5)

endorama:
Added shell detection using the script provided by Andrew in [`57cf67f` (#921)](https://github.com/elastic/elastic-package/pull/921/commits/57cf67f295ce894e02e9cf1ea678093ecb6ebf0b)

jsoriano:
CI failures seem related, `go-sysinfo` should be added to go mod.

cmd/stack.go
```diff
@@ -198,19 +201,34 @@ func setupStackCommand() *cobraext.Command {
...
+			if err != nil {
+				return cobraext.FlagParsingError(err, cobraext.ShellInitShellFlagName)
+			}
+
+			if shellName == "detect" {
+				shellName, err = detectShell()
+				if err != nil {
+					return fmt.Errorf("cannot detect parent shell from current process: %w", err)
+				}
+				fmt.Fprintf(cmd.OutOrStderr(), "detected shell: %s\n", shellName)
```
jsoriano, line 214: Leftover? It is expected that `elastic-package stack shellinit` output can be evaluated.

cmd/stack.go
```diff
@@ -332,3 +350,20 @@ func printStatus(cmd *cobra.Command, servicesStatus []stack.ServiceStatus) {
...
+	if err != nil {
+		return "", fmt.Errorf("cannot retrieve information for process %d: %w", ppid, err)
+	}
+
+	parentInfo, err := parent.Info()
+	if err != nil {
+		return "", fmt.Errorf("cannot retrieve information for parent of process %d: %w", ppid, err)
+	}
+
+	shell := filepath.Base(parentInfo.Exe)
```
jsoriano, line 367: In bash in Windows this detects `bash.exe`, and fails to find the shell type.

FTR, in Windows terminal it detects `cmd.exe`, and in PowerShell, `powershell.exe`. No need to add support for these now though.

internal/stack/shellinit_internal_test.go
```diff
@@ -0,0 +1,42 @@
...
+		want string
+	}{
+		{"bash code template", args{"bash"}, posixTemplate},
+		{"fish code template", args{"fish"}, fishTemplate},
+		{"sh code template", args{"sh"}, posixTemplate},
+		{"zsh code template", args{"zsh"}, posixTemplate},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := initTemplate(tt.args.s); got != tt.want {
```
jsoriano, line 27: CI complains here.
```suggestion
			if got, err := initTemplate(tt.args.s); got != tt.want {
```

jsoriano (commented):
:+1:

cmd/stack.go
endorama: That string is sent to `stderr`, so should not affect evaluation:

```
$ ./elastic-package stack shellinit | source
detected shell: fish
$ printenv | ag ELASTIC_PACKAGE_
ELASTIC_PACKAGE_ELASTICSEARCH_HOST=https://127.0.0.1:9200
ELASTIC_PACKAGE_KIBANA_HOST=https://127.0.0.1:5601
ELASTIC_PACKAGE_ELASTICSEARCH_PASSWORD=changeme
ELASTIC_PACKAGE_ELASTICSEARCH_USERNAME=elastic
ELASTIC_PACKAGE_CA_CERT=/home/endorama/.elastic-package/profiles/default/certs/ca-cert.pem
```

Is this acceptable?

cmd/stack.go
endorama: I added a trim suffix to remove any `.exe` suffix.

cmd/stack.go
```diff
@@ -332,3 +350,22 @@ func printStatus(cmd *cobra.Command, servicesStatus []stack.ServiceStatus) {
...
```
jsoriano, line 367: Found another corner case.

When running `elastic-package` with `go run`, it detects `go` as the shell:
```
detected shell: go
Error: shellinit failed: cannot get shell init template: shell type is unknown, should be one of bash, fish, sh, zsh
exit status 1
```

This happens when running `elastic-package` with `go run` in repositories where this package is in `go.mod`. This is done in CI in some cases, and for example I use it when trying to reproduce failures in the integrations repository with the same version of elastic-package that is used in CI.

This could be considered a regression, could we do something about this? Maybe it is possible to detect if `elastic-package` is being run with `go run`, or we could get the parent of the parent when it is `go`.

cmd/stack.go
jsoriano: Ah ok, then I guess that this is fine, we print other messages and warnings to stderr.

As nitpicking:
```suggestion
				fmt.Fprintf(cmd.OutOrStderr(), "Detected shell: %s\n", shellName)
```

cmd/stack.go
jsoriano: Or maybe another option to avoid regressions, we could directly return `bash` and log something to stderr instead of failing when this method cannot detect the shell.

endorama: This use case is supported by not using auto detection (passing `--shell=bash`). I think that is preferable than deciding to fallback to BASH because that can lead to errors that are very hard to troubleshoot and less explanatory.

We may want to make this explicit using a different error message.

jsoriano: Well, let's separate my two comments:
* Not breaking the not so rare `go run` case. Possible workaround for this could be: if found shell is `go`, try to get the parent of the parent instead.
* General solution to don't break unknown cases that work now. I propose to use `bash` in case of error only to keep the current behavior, where only bash is supported. I agree that it could be more discussed as I don't have more use cases beyond the mentioned ones, and it would be better to have explicit behavior.

endorama: > Not breaking the not so rare go run case. Possible workaround for this could be: if found shell is go, try to get the parent of the parent instead.

I added this in https://github.com/elastic/elastic-package/pull/921/commits/12ddc3d18b9c71a8fd7b1fa72717a5ee4e5cecd4 (with tests)

> General solution to don't break unknown cases that work now. I propose to use bash in case of error only to keep the current behavior, where only bash is supported.

I didn't consider this POV, but it make complete sense, as the behaviour would not change from current implementation. I will change the default case.

cmd/stack.go
```diff
@@ -352,21 +353,42 @@ func printStatus(cmd *cobra.Command, servicesStatus []stack.ServiceStatus) {
...
+
+func detectShell() (string, error) {
+	ppid := os.Getppid()
+	parentInfo, err := getParentInfo(ppid)
+	if err != nil {
+		return "", err
+	}
+
+	shell := getShellName(parentInfo.Exe)
+	if shell == "go" {
```
jsoriano, line 385: Nit. Please add a comment here about why we have this special case.

jsoriano:
> The CI fails on `check-static`, complaining that `go.mod` and `go.sum` have changes, but `make check-static` passes on my local machine. Any idea how to solve?

CI uses go 1.19, is it possible that you are using an older version?

cmd/stack.go
```diff
@@ -198,19 +203,34 @@ func setupStackCommand() *cobraext.Command {
...
+			if err != nil {
+				return cobraext.FlagParsingError(err, cobraext.ShellInitShellFlagName)
+			}
+
+			if shellName == cobraext.ShellInitShellDetect {
+				shellName, err = detectShell()
+				if err != nil {
+					return fmt.Errorf("cannot detect parent shell from current process: %w", err)
+				}
+				fmt.Fprintf(cmd.OutOrStderr(), "detected shell: %s\n", shellName)
```
jsoriano, line 216: Nit.
```suggestion
				fmt.Fprintf(cmd.OutOrStderr(), "Detected shell: %s\n", shellName)
```

jsoriano (approved):
Apart of some last nitpicking and the issues with CI and go.mod it LGTM, thanks!

cmd/stack.go
endorama: What do you think of?
```suggestion
	// NOTE: handle the case of this CLI being run with "go run"
	// See https://github.com/elastic/elastic-package/pull/921
	if shell == "go" {
```

endorama:
> CI uses go 1.19, is it possible that you are using an older version?

I'm using go 1.19.1 locally, which is the same as reported in `.go-version`

cmd/stack.go
jsoriano: Looks good, thanks.

jsoriano:
> > CI uses go 1.19, is it possible that you are using an older version?
> 
> I'm using go 1.19.1 locally, which is the same as reported in `.go-version`

If I run `make check-static` locally with the code of this branch, with go 1.19.2, I see the same errors as in CI. If I run `go mod tidy`, I see that `go.mod` and `go.sum` get changes. :thinking: 

Shall I push the changes I see to this branch?

endorama:
@jsoriano yes please, I cannot reproduce even with go 1.19.2 :(

jsoriano (approved):

endorama:
Thank you!

PR #928, opened 2022-08-18
Bump up default stack version go 8.3.3

jsoriano:


mrodm (approved):

jsoriano:
Investigating the issue in CI, it seems that parameters are not being properly sent between jobs.

Issue #880, opened 2022-06-29
[docs] Add new docs template language to `elastic-package create package`


colleenmcginnis:
In elastic/integrations#3433, we're working on the first iteration of more detailed documentation guidelines for integrations (including template language). In that PR @jsoriano suggested:

>`elastic-package create package` can be used to create an initial scaffolding for new packages, as a follow up we could apply these recommendations in the templates there.

We should do this after elastic/integrations#3433 is merged.

cc @jamiehynds @elastic/obs-docs

colleenmcginnis:
@jsoriano is this something I could help with or is it something your team needs to do?

jsoriano:
@colleenmcginnis it would be great if you can open a PR with the proposed changes, we would be happy to review :slightly_smiling_face:

The code to generate packages is around here: https://github.com/elastic/elastic-package/tree/3e52239e64ffa9c6f912592204d119cd88451a2a/internal/packages/archetype

I see that this is mostly go code, we could maybe migrate it first to use templates if this makes it easier to make changes there.

Let us know what approach do you prefer.

colleenmcginnis:
Thanks @jsoriano! I'll give it a try next week.

PR #936, opened 2022-08-22
Add package readme howto guide

### LINKED ISSUE #349: HOWTO: Package README files
We should create a HOWTO document for package developers who want to write `README.md` files for their packages. It should cover things like the `_dev/build/docs/README.md` template, the generated `README.md` file, the various placeholders (e.g. `event` and `fields`) available in the README template and what they mean + where they pull their replacement contents from, etc.

mrodm:
This PR adds a how to guide for developers to have documentation to write the required README.md files.

A link to this documentation has been added in the `elastic-package build` command description.

Closes #349


mtojek (approved):
LGTM!

bmorelli25:
@colleenmcginnis added detailed documentation guidelines for integration readme files in https://github.com/elastic/integrations/pull/3433. Does it make more sense to move that content to this new file? Where can these docs be seen by the greatest number of people?

mrodm:
@bmorelli25 thanks for pointing out that doc in integrations. Looking at both PRs, I would see that they are both complementary to each other.

According to what I see, on the one hand, the PR in integrations repository by @colleenmcginnis (elastic/integrations#3433) are guidelines to follow in that repository to create new readme files, and I would say it could be kept in that repository.

On the other hand, this PR is specifically about what syntax and features are allowed by `elastic-package build` command to be used when writing those readme templates.

Would it make sense to keep both as they are and add a link from https://github.com/elastic/integrations/blob/main/docs/documentation_guidelines.md to this file https://github.com/elastic/elastic-package/blob/main/docs/howto/add_package_readme.md ? Thereby, developers would be able to check what they can use in those readme templates from the guidelines. WDYT ?

Feel free to chime in @ecosystem

bmorelli25:
Gotcha. This PR is about `elastic-package build`. And I'm thinking about the documentation template more in the context of `elastic-package create package`.

I think the best solution is to integrate the doc template into `elastic-package create package`—which Colleen will be attempting as a part of https://github.com/elastic/elastic-package/issues/880.

Meanwhile, as you suggested above, cross linking between the [`add_package_readme.md`](https://github.com/elastic/elastic-package/blob/main/docs/howto/add_package_readme.md) and [`documentation_guidelines.md`](https://github.com/elastic/integrations/blob/main/docs/documentation_guidelines.md) also probably makes sense.

PR #938, opened 2022-08-23
Avoid hardcoded links in READMEs

mrodm:
Closes https://github.com/elastic/elastic-package/issues/948

This PR adds a new function ("url") available into Readme templates to render links based on the `event` and `fields` functions. All links will be defined in a common file that will be the source of truth.

This intended to be used in [integrations repository](https://github.com/elastic/integrations/) where a specific ~csv file `links_table.csv`~ YAML file `links_table.yml` containing all the available links is defined.
This file should be placed at the root of the integrations repository and it should follow this pattern:
```
links:
  link_key: http://url1.test
  foo_key: http://url2.test
``` 

Being there at the root of the repository, this file (all the links) are available and shared by all the packages defined in the repository in building time (`elastic-package build`).

Example of usages in the README using the above file as basis:
- README template:
  ```
  Interesting links:
  - {{ url "link_key" }}
  - {{ url "foo_key" "howto guide" }}
  ```
- README generated:
  ```
  Interesting links:
  - http://url1.test
  - [howto guide](http://url2.test)
  ```


mtojek (commented):
I'm not sure if I fully understand what's the purposes of this PR. Do you want to add a new file to a package? Is it necessary to be a .csv file?

If so, I would start with a package-spec issue to decide about the form of the file first.

mrodm:
> I'm not sure if I fully understand what's the purposes of this PR. Do you want to add a new file to a package? Is it necessary to be a .csv file?
> 
> If so, I would start with a package-spec issue to decide about the form of the file first.

I see that the description of the PR lacked some information regarding this proposal as a first step to solve https://github.com/elastic/integrations/issues/2652. I'll give some more information here (and I'll update the description accordingly afterwards). 

With this PR, my proposal is adding a file at the root of "integrations" repository that contains all the links to elastic documentation pages. Being there, all these links will be available and shared by all the packages in building time (`elastic-package build`). Similarly, as License file.

As that file would be out of the package scope, at least with this design, it should not be added to the package-spec.

mtojek:
Yeah, but keep in mind that Elastic packages shouldn't have a strong coupling between with elastic/integrations repository. They can be stored in any place.

mrodm:
> I'm wondering if we can decouple `links_table.csv` from elastic-package and integrations, and make it more generic. I'm thinking about a dedicated ENV which contains a path to the `links_table.csv` (assuming that we want that format instead of YAML). Whenever elastic-package finds the `{{ url ... }}`, it will look up the ENV to read the value from there.

That way is way more flexible. Avoiding hard-coding the filename and the path where it should be located will help to avoid issues. Maybe later it needs to be changed to some other folder that file or just being renamed, and that should be independent of elastic-package.

In case ENV is not defined, it could be kept looking in the root of the repository.

I will look into it to update this PR accordingly. Thanks for the suggestion!

About csv vs YAML, as first step I just wanted to follow the same structure it was shown as example in the main issue ([example](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_doc_ids/table.csv)). But definetly, this can be changed later to YAML format. Probably easier to read.

Just some other possible format using YAML:
```yaml
links:
  link_key: http://url1.test
  foo_key: http://url2.test
```

internal/common/env.go
```diff
@@ -0,0 +1,9 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
mtojek, line 1: In most cases, directories like `common`, `utils`, `tools` are considered to be an antipattern. Do you think that you can place it in a more specific package?

internal/common/env.go
```diff
@@ -0,0 +1,9 @@
...
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package common
+
+const (
+	ElasticPackageEnvPrefix = "ELASTIC_PACKAGE_"
```
mtojek, line 8: This way somebody can overwrite the `ElasticPackageEnvPrefix`. I would wrap with a dedicated function, for example: `withElasticPackageEnv(...)`.

internal/common/fs.go
```diff
@@ -0,0 +1,35 @@
...
+package common
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+func FindRepositoryRootDirectory() (string, error) {
```
mtojek, line 14: What was wrong with the `builder` directory?

mtojek (commented):
First review round, I haven't looked at the entire scope.

internal/common/fs.go
mrodm: I wanted to use this function in `docs` package to be able to get the default location of the yaml file with links when generating the readme.
As I thought this function was more related to files than the builder process (looking for the root of the working copy), I was thinking to move this function to another package.

As you mentioned, `common` is not the best place, what about moving it to `files` packages (e.g. internal/files/scm.go)? if it is not clear, I would revert this change and keep it under `builder` package.

internal/common/env.go
mrodm: Totally agreed, I'll check to move it to another package (or create a new one)

internal/common/fs.go
mtojek: The `internal/files` directory seems to be more precise location. Alternatively, you can consider `internal/git/repository`. It doesn't look like we will support other SCMs.

internal/common/env.go
mrodm: Updated :+1: 
As a note, I had to change some constants to variables as a consequence of being this a function.

internal/docs/readme.go
```diff
@@ -160,7 +170,7 @@ func findReadmeTemplatePath(fileName, packageRoot string) (string, bool, error)
 	return templatePath, true, nil
 }
 
-func renderReadme(fileName, packageRoot, templatePath string) ([]byte, error) {
+func renderReadme(fileName, packageRoot, templatePath string, linksMap linkMap) ([]byte, error) {
```
mtojek, line 173: Do you think we can stop passing the `linksMap` through all these functions and just define it in the deepest function?

docs/howto/add_package_readme.md
```diff
@@ -81,3 +82,53 @@ List of placeholders that can be used in the Markdown templates:
...
+This file by default is located at the root of the repository with the name `links_table.csv`.
+It can be overwritten by setting the environment variable `ELASTIC_PACKAGE_LINKS_FILE_PATH` with
+the path to another file.
+
+The format of this file must be:
+```
+<key_1>,<url_1>
+<key_2>,<url_2>
+<key_3>,<url_3>
+```
```
mtojek, line 134: The only thing I'm not convinced about is the format of this file. I prefer YAML to the comma-separated form, but I'm fine if this is Elastic standard.

mtojek (commented):
Could you please add another test package which uses the links feature?

-- 

There is one caveat, URL rendering will be static, performed during the build time. It means that it won't be possible to magically replace URL destinations based on Kibana version.
Also, we will need to republish packages as new versions to expose updated URLs. It would be good to describe it in the README somewhere.

docs/howto/add_package_readme.md
mrodm: This was a leftover in the docs, I've updated this file with the latest changes that I introduced to being this file a YAML. The format of the YAML would be:
```yaml
links:
  key_1: url_1
  key_2: url_2
```

internal/docs/readme.go
mrodm: I wanted to try to do that, but doing so the links file would need to be read again for each markdown file. That's why I read that file at the beginning of the process to be used for all of them, since those links are not going to change.

Currently, there are not so many readme files per package, and maybe it would be preferable to move for now this read to the deepest function. WDYT?

internal/docs/links_map.go
```diff
@@ -0,0 +1,118 @@
...
+
+	err = yaml.Unmarshal(contents, &links)
+	if err != nil {
+		return linkMap{}, err
+	}
+
+	return links, nil
+}
+
+func (l linkMap) RenderUrl(key string) (string, error) {
```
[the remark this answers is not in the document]
mrodm: I'll check how to improve this :+1: 
I wanted to differentiate between rendering just the url or the url with a caption

mrodm:
> Could you please add another test package which uses the links feature?
> 

I've added new tests for `generateReadme` function instead of adding a new package under `test/packages` to try to test all the usages of the functions available in readme (fields, event and now url). These tests along with the ones to test `linksDefinitionsFilePath` would probably cover this.

Would it be better to add a new package? Would that test something else ?

mtojek (commented):
> Would it be better to add a new package? Would that test something else?

If you ask me, more tests are always better. The benefit of adding a package is that we can catch if the `README.md` changed. With unit tests, you won't catch it if the calling function suddenly disappears.

If you don't want to create a test package, you can modify an existing one.

internal/docs/readme.go
mrodm: Updated. I've just kept linksMap as a parameter in renderReadme function, so testing could be done easier.

jsoriano (approved):
:+1:

PR #957, opened 2022-09-06
EPR: enable proxy mode

mtojek:
Issue: https://github.com/elastic/package-registry/issues/770

This PR enables the proxy mode feature in the elastic-package stack.

### DIFF
--- a/internal/profile/_static/Dockerfile.package-registry
+++ b/internal/profile/_static/Dockerfile.package-registry
@@ -1,5 +1,10 @@
 FROM __BASE_IMAGE__
-
 ARG PROFILE
+
+# Disable package validation (already done).
+ENV EPR_DISABLE_PACKAGE_VALIDATION=true
+
 COPY profiles/${PROFILE}/stack/package-registry.config.yml /package-registry/config.yml
 COPY stack/development/ /packages/development
+
+ENTRYPOINT ["./package-registry", "--feature-proxy-mode", "--log-level", "debug"]
--- a/internal/profile/_static/Dockerfile.package-registry
+++ b/internal/profile/_static/Dockerfile.package-registry
@@ -4,7 +4,7 @@ ARG PROFILE
 # Disable package validation (already done).
 ENV EPR_DISABLE_PACKAGE_VALIDATION=true
 
+ENV EPR_FEATURE_PROXY_MODE=true
+
 COPY profiles/${PROFILE}/stack/package-registry.config.yml /package-registry/config.yml
 COPY stack/development/ /packages/development
-
-ENTRYPOINT ["./package-registry", "-feature-proxy-mode"]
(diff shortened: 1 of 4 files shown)


internal/profile/_static/Dockerfile.package-registry
```diff
@@ -1,5 +1,10 @@
...
-
 ARG PROFILE
+
+# Disable package validation (already done).
+ENV EPR_DISABLE_PACKAGE_VALIDATION=true
+
 COPY profiles/${PROFILE}/stack/package-registry.config.yml /package-registry/config.yml
 COPY stack/development/ /packages/development
+
+ENTRYPOINT ["./package-registry", "--feature-proxy-mode", "--log-level", "debug"]
```
mrodm, line 10: Just one dash ? 

```suggestion
ENTRYPOINT ["./package-registry", "-feature-proxy-mode", "-log-level", "debug"]
```

mtojek: Fixed

internal/profile/_static/Dockerfile.package-registry
```diff
@@ -1,5 +1,10 @@
...
-
 ARG PROFILE
+
+# Disable package validation (already done).
+ENV EPR_DISABLE_PACKAGE_VALIDATION=true
+
 COPY profiles/${PROFILE}/stack/package-registry.config.yml /package-registry/config.yml
 COPY stack/development/ /packages/development
+
+ENTRYPOINT ["./package-registry", "-feature-proxy-mode"]
```
jsoriano, line 10: It should be also possible to use environment variables to configure this: `EPR_FEATURE_PROXY_MODE=true`. This may be better than setting the entrypoint, as we are already using environment variables to disable validation.

internal/profile/_static/package_registry.yml
```diff
@@ -1,5 +1,2 @@
 package_paths:
-  - /packages/development
-  - /packages/production
-  - /packages/staging
-  - /packages/snapshot
```
jsoriano, line 5: :fire: :fire: :fire:

internal/profile/static.go
```diff
@@ -124,7 +123,8 @@ func newPackageRegistryConfig(_ string, profilePath string) (*simpleFile, error)
 }
 
 // PackageRegistryBaseImage is the base Docker image of the Elastic Package Registry.
-const PackageRegistryBaseImage = "docker.elastic.co/package-registry/distribution:snapshot"
+// commit SHA of enabled proxy mode (technical preview feature)
+const PackageRegistryBaseImage = "docker.elastic.co/package-registry/package-registry:14455fcb7e415d6b4aed48e00d2289718f056162"
```
jsoriano, line 127: Do you want to wait to have a released version with this?

jsoriano (commented):
:+1:

internal/profile/static.go
mtojek: Yes, as I'm afraid that we may need to add more fixes, and going with the standard release will slow it down. I will replace it with a proper release after some time.

internal/profile/_static/Dockerfile.package-registry
mtojek: Fixed

jsoriano (approved):

mrodm (approved):

Issue #821, opened 2022-05-16
[system testing] Add assertion for expected number of events


andrewkroh:
In some system tests we have a fixed number of events that are reproducible for each run. In order to verify the input configuration is working as expected I would like to be able to assert that the system test produced a specific number of events.

For example, in an integration that uses the httpjson input to request events from an API I want to validate that input requests the initial batch, makes requests to paginate, and then makes a follow up request after some interval. Given that we mock the API responses we know the expected number of events that we should have after all of those calls complete. So if the test had an assertion then we could validate this behavior.

An example config file for such a test:

```
input: httpjson
service: cloudflare
vars:
  api_url: http://{{Hostname}}:{{Port}}
  ssl.verification_mode: none
data_stream:
  vars:
    preserve_original_event: true
    auth_email: user@example.com
    auth_key: xxxxxxxxxx
    account: aaabbbccc

# PROPOSED
system_test:
  expected_event_count: 10
```

ebeahan:
Added in #960

PR #960, opened 2022-09-07
[system testing] Implement assertion for expected number of events

djptek:
## What does this PR do?

This PR will enhance `system testing` through the implementation of an assertion for expected number of events. 

## Checklist

- [x] unpack events embedded in a message/data/events/etc. field within a hit
- [x] add test case to `httpjson` integration package
- [x] update readme
- [x] ensure test does not fail when test parameters are absent
- ~[ ] handle persistence between calls~ _NOT YET REQUIRED_
- ~[ ] sum events across calls~ _IMPLICIT_

## Author's Checklist

- [x] run system tests vs elasticsearch + 4 randomly selected integrations to check for regressions
    - [x] 1password
    - ~[ ] cisco_aironet~
    - [x] elasticsearch
    - [x] hadoop
    - ~[ ] juniper_junos~
    - ~[ ] kafka_log~
    - [x] nginx
    - [x] traefik
    - ~[ ] zookeeper~
- [x] add test under `other` 

Pull this branch and install elastic-package with `make install`

Clone `elastic/integrations`

Configure the httpjson (or related) integration as follows

For an http request, consider this example from the `httpjson/generic` data stream's `test-expected-hit-count-config.yml`, shown below.

```
input: httpjson
service: httpjson
data_stream:
  vars:
    data_stream.dataset: httpjson.generic
    username: test
    password: test
    request_url: http://{{Hostname}}:{{Port}}/testexpectedhits/api
    response_split: |- 
      target: body.hits
      type: array
      keep_parent: false     
assert:
  hit_count: 4
```

The `data_stream.vars.request_url` corresponds to a test-stub path in the `_dev/deploy/docker/files/config.yml` file.

```
  - path: /testexpectedhits/api
    methods: ["GET"]
    request_headers:
      Authorization:
        - "Basic dGVzdDp0ZXN0"
    responses:
      - status_code: 200
        headers:
          Content-Type:
            - "application/json; charset=utf-8"
        body: |-
          {"total":3,"hits":[{"message": "success"},{"message": "success"},{"message": "success"}]}
```

As `assert.hit_count` is defined and `> 0` the system test will only pass the assertion when the number of elements in the top-level array of the JSON body is equal to this value. In this case the assertion requires 4 hits when the actual count will be 3, so the assertion should fail.

Run the system tests using `elastic-package test system -v`

The test results should include the following line:

```
│ httpjson │ generic     │ system    │ expected-hit-count │ FAIL: observed hit count 3 did not match expected hit count 4 │ 21.123980175s │
```

Next, amend `assert.hit_count` so that this matches the number of elements (3) in the response to the `/testexpectedevents/api` path. The test should now pass as follows

```
│ httpjson │ generic     │ system    │ expected-hit-count │ PASS   │ 20.979543465s │
```

## Related issues

- Relates https://github.com/elastic/elastic-package/issues/821

~## Screenshots~


internal/testrunner/runners/system/runner.go
```diff
@@ -478,6 +479,17 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+	// Check Event Count within docs, if 0 then it has not been specified
+	if config.SystemTest.ExpectedEventCount != 0 {
+		if result.EventCount, err = getEventCount(docs); err != nil {
+			return result.WithError(err)
+		}
+		if result.EventCount != config.SystemTest.ExpectedEventCount {
+			result.FailureMsg = fmt.Sprintf("observed event count %d did not match expected event count %d", result.EventCount, config.SystemTest.ExpectedEventCount)
+			return result.WithError(fmt.Errorf("%s", result.FailureMsg))
+		}
+	}
```
jsoriano, line 491: Other validations are executed before generating sample events, could you also make this validation before?

internal/testrunner/runners/system/runner.go
```diff
@@ -747,6 +759,43 @@ func validateFields(docs []common.MapStr, fieldsValidator *fields.Validator, dat
...
+		if reflect.ValueOf(v).Kind() == reflect.Slice {
+			e := events[k].([]interface{})
+			logger.Debugf("found %d events under \"%s\"", len(e), k)
+			return len(e)
+		}
+	}
+	return 0
+}
+
+func getEventCount(docs []common.MapStr) (lenEvents int, err error) {
```
jsoriano, line 773: It doesn't look obvious how events are counted here, it could be just `len(docs)`, but this is also looking for events in the `message`. Could you add a comment about this here?

jsoriano (commented):
Thanks for adding this. System tests are a bit unpredictable by nature, so counting events may be a bit flaky, though I guess we are fine in cases where we control the source events generated.

Could you add a test package that leverages this feature?

internal/testrunner/runners/system/test_config.go
```diff
@@ -38,6 +38,11 @@ type testConfig struct {
 		Vars map[string]packages.VarValue `config:"vars"`
 	} `config:"data_stream"`
 
+	SystemTest struct {
+		// Length of first array found within the message field inside a Hit
+		ExpectedEventCount int `config:"expected_event_count"`
+	} `config:"system_test"`
```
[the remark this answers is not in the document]
andrewkroh: SGTM.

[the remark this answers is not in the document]
djptek: Thanks, I've applied a global change  `SystemTest.ExpectedEventCount` to `Assert.EventsCount`in config, logic and docs

internal/testrunner/runners/system/runner.go
```diff
@@ -747,6 +759,43 @@ func validateFields(docs []common.MapStr, fieldsValidator *fields.Validator, dat
...
+	for _, doc := range docs {
+		var message interface{}
+		var events interface{}
+
+		if message, err = doc.GetValue("message"); err != nil {
+			return 0, testrunner.ErrTestCaseFailed{
+				Reason:  fmt.Sprintf("could not find message in doc %v", doc),
+				Details: err.Error(),
```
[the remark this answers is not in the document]
djptek: From what I've seen, the `httpjson` input packages the response received from the target api within a `message` object. If the response contains multiple events, these are packaged as an array under an inner object whose name is not mandatory, for example `"events": [{},{},{}]`, `"data": [{},{},{}]` etc.

The `func getEventCount()` is only called when the assertion regarding the number of those events was previously configured in the system test, so it checks for the existence of `message` before calling `getDocEventCount()` to look for the arbitrarily named array (`"events": []` or `"data": []` etc.) and add its length to the number of events. 

See also: https://github.com/elastic/beats/blob/main/x-pack/filebeat/input/httpjson/input.go#L272

docs/howto/system_testing.md
```diff
@@ -395,6 +395,43 @@ you can use the `input` option to select the stream to test. The first stream
...
+    request_url: http://{{Hostname}}:{{Port}}/testexpectedevents/api
+system_test:
+  expected_event_count: 3
+```
+
+The `data_stream.vars.request_url` corresponds to a path in the `_dev/deploy/docker/files/config.yml` file
+
+The `system_test.expected_event_count` field corresponds to the expected number of events in the corresponding path in the `_dev/deploy/docker/files/config.yml` file
+
+For example
```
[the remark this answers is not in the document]
djptek: added :+1:

internal/testrunner/runners/system/runner.go
```diff
@@ -747,6 +759,43 @@ func validateFields(docs []common.MapStr, fieldsValidator *fields.Validator, dat
 	return nil
 }
 
+func getDocEventCount(events map[string]interface{}) int {
+	for k, v := range events {
+		if reflect.ValueOf(v).Kind() == reflect.Slice {
+			e := events[k].([]interface{})
```
andrewkroh, line 765: I don't think this function will be needed based on other comments. I still wanted to mention that reflection does not look required. A [type assertion](https://go.dev/ref/spec#Type_assertions) will work.

`if e, ok := v.([]interface{}); ok {`

internal/testrunner/runners/system/runner.go
```diff
@@ -478,6 +479,17 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 		}
 	}
```
andrewkroh, line 481: When this new assertion is used the condition inside of `waitUntilTrue` should be changed. It should continue searching until it reaches at least the expected count.

internal/testrunner/runners/system/runner.go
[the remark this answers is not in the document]
djptek: Done

djptek:
@jsoriano 

>> Could you add a test package that leverages this feature?

I added to `httpjson` see: https://github.com/elastic/integrations/pull/4152

internal/testrunner/runners/system/runner.go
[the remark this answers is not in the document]
djptek: >> The hit count (length of the .hits.hits[] in the Elasaticsearch response)

I looked at that first & found at this point in the testing, `response.split:` has not been applied, so API calls to a source other than Elasticsearch will have an array of events packaged inside a single `"message"`. I checked for `httpjson` and there were no system test cases where the response was NDJSON, it's always JSON with an arbitrarily named top level element inside `"message"`. I also tried adding multiple elements into the response & that didn't work atall (expected).

docs/howto/system_testing.md
```diff
@@ -395,6 +395,43 @@ you can use the `input` option to select the stream to test. The first stream
...
+input: httpjson
+service: httpjson
+data_stream:
+  vars:
+    data_stream.dataset: httpjson.generic
+    username: test
+    password: test
+    request_url: http://{{Hostname}}:{{Port}}/testexpectedevents/api
+assert:
+  events_count: 3
```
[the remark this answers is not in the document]
djptek: applied

jsoriano:
> @jsoriano
> 
> > > Could you add a test package that leverages this feature?
> 
> I added to `httpjson` see: [elastic/integrations#4152](https://github.com/elastic/integrations/pull/4152)

Sorry, I meant a test package here in `elastic-package`, you can add it to `test/packages`, or modify one of the ones there.

internal/testrunner/runners/system/runner.go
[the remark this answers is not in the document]
djptek: @jsoriano any message may contain one or more events. This logic is counting events within messages, so I think `event_count` remains accurate

@andrewkroh when I was looking at the `hits.hits` vs `hits.total` my thought was a future issue to add `hit_count` as an alternative configuration parameter, in parallel to `event_count`

That would allow assertions to cover both scenarios as appropriate for each integration and ensure the distinction between hits & events is explicit

Prior to this PR, the existing logic admits the possibility of multiple `docs` so the assertion is designed to handle this as/when such test-cases  are added in a possible future.

I also added `TestResult.EventCount` to track the sum across a request as a whole as it's theoretically possible for a test to have multiple hits, each containing a message where the message contains multiple events. Running both assertions for that edge case could require `event_count` to match the sum total of events across messages within all hits and the number of hits to match the expected total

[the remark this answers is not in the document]
andrewkroh: I'm not following something here.

Elastic Agent generates documents (aka events). Those documents are written to Elasticsearch. For a given system test I know the expected number of documents that should be in Elasticsearch. After the test completes I want to assert that there are exactly the expected amount of documents in Elasticsearch.

The contents of those documents are beyond the scope of this assertion. So there's no need to examine the `message`.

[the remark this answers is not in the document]
djptek: @andrewkroh 

`Documents : Events` is a `1:n` relationship for at least one integration (`httpjson`)

 >> After the test completes I want to assert that there are exactly the expected amount of documents in Elasticsearch

This is easier to test (it's what I wrote first) - however, it was counting documents, not events. I can definitely do that.

I´ll add the the `hit_count` logic in parallel as an additional assertion.

Context: I've been working with the `httpjson` integration as my point of reference. In that integration, it appears `message` is split to constituent events downstream of the system test. So, `hits.total == len(hits.hits[]) == 1` every time.

Did you have a specific integration in mind when you conceived the assertion? I know others _do_ return `hits.total > 1` and I could use that as a test case to ensure both cases are covered & then look into the edge case where both assertions might w be defined. If not, I can just pick one.

internal/testrunner/runners/system/runner.go
djptek: @andrewkroh 

The assertion validates how many docs were harvested, asking "Did we get X docs?"

Continuing to search until a determined number of documents have been acquired feels to me more like setting a target, requesting "Go get X docs?"

These could resolve to the same value in many cases, which could make the assertion itself redundant.

I can see 2 ways to do this, either add an additional parameter, so:

```
assert:
  hit_count: 3
target:
  doc_count: 3
```

or alternatively rename `assert` to `target`, so 


```
target:
  doc_count: 3
```

or maybe 

```
target:
  hit_count: 3
```

What do you think?

djptek:
@jsoriano 

> Thanks for adding this. System tests are a bit unpredictable by nature, so counting events may be a bit flaky, though I guess we are fine in cases where we control the source events generated.
> 
> Could you add a test package that leverages this feature?

I've added a (passing) test package `test/packages/other/hit_count_assertion/`

To provoke a fail, I changed

`test/packages/other/hit_count_assertion/data_stream/test/_dev/test/system/test-hits-config.yml:  hit_count: 100`

to

`test/packages/other/hit_count_assertion/data_stream/test/_dev/test/system/test-hits-config.yml:  hit_count: 101`

Which yields the following output in CI test `make install test-check-packages-other check-git-clean`

```
[2022-12-23T08:51:32.907Z] 2022/12/23 08:51:32 DEBUG found 100 hits in logs-hit_count_assertion.test-ep data stream
[2022-12-23T08:51:33.007Z] 2022/12/23 08:51:32 DEBUG assert hit count expected 101, observed 100
...
[2022-12-23T08:51:48.363Z] Error: one or more test cases failed
```

Restoring the value of 100 allows the test to pass.

internal/testrunner/runners/system/runner.go
djptek: Moving back to review pending feedback from @andrewkroh on this remaining issue

djptek: Hi @andrewkroh would you prefer the additional parameter suggested above, or to rename? Thanks

djptek: sumarising offline conversation fine to keep

```
assert:
  hit_count: 3
```

jsoriano (approved):

internal/testrunner/runners/system/runner.go
```diff
@@ -478,6 +478,11 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
 
+	// Check Hit Count within docs, if 0 then it has not been specified
+	if assertion_pass, message := assertHitCount(config.Assert.HitCount, docs); !assertion_pass {
```
andrewkroh, line 482: Go naming conventions are `mixedCaps`. https://go.dev/doc/effective_go#mixed-caps

```suggestion
	if assertionPass, message := assertHitCount(config.Assert.HitCount, docs); !assertionPass {
```

internal/testrunner/runners/system/runner.go
```diff
@@ -747,6 +752,17 @@ func validateFields(docs []common.MapStr, fieldsValidator *fields.Validator, dat
 	return nil
 }
 
+func assertHitCount(expected int, docs []common.MapStr) (pass bool, message string) {
+	observed := len(docs)
```
andrewkroh, line 756: The variable should be declared as close to where it is used as possible to minimize scope. So this can be moved within the `if` block.

internal/testrunner/runners/system/runner.go
```diff
@@ -478,6 +478,11 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 		}
 	}
 
+	// Check Hit Count within docs, if 0 then it has not been specified
```
andrewkroh, line 481: Up on line 446 there is code to wait for documents to be received.

```go
docs, err = r.getDocs(dataStream)
return len(docs) > 0, err
```

If I am using `assert.hit_count` then my expectation is that the test will wait for at least the expected number of hits to be received. Imagine a test where the documents are read in slowly (say it takes 30 seconds to pull down the expected documents and they are posted in a few separate _bulks). Currently this will stop waiting as soon as any documents are received. Then when the count is checked here it will be wrong because the it didn't wait long enough. I think the check on L446 needs to be amended to wait.

internal/testrunner/runners/system/runner.go
djptek: applied

internal/testrunner/runners/system/runner.go
djptek: Double checking logic here: I am accustomed to assert in the sense of "flag an error if not condition satisfied" but what you actually want is "keep doing action until condition satisfied" correct?

andrewkroh: It should do both. If it the test runner doesn't wait until the condition can be satisfied then it will be brittle.

djptek: I'm really not getting this. Assertions would normally run after a test, not during. If an assertion evaluates to not true, that test has ended and is flagged as a fail. However, I think what you want is to keep running the "assertion" as part of the mechanics of test setup until enough docs/hits have arrived & only to flag a fail if, after trying "n" times, the "assertion" fails. If the loop repeats until the assertion is true then the assertion will never fail. @andrewkroh this is why I am suggesting to separate the loop exit condition from the assertion, even if they share the same target quantity - does this make sense?

andrewkroh: When employing this assertion, I think the wait condition for the test running needs to be `len(docs) >= config.Assert.HitCount`.

djptek: Hi Andrew, I'm not asking how to compare a greater than or equal value, my question is whether it's OK to continue a test after an assert has failed, or alternatively do the `>=` comparison separately to the assert. I appreciate they will have the same target value in some use cases, it's more a point of the nature of an assertion

andrewkroh: I'm not asking to have the call to `assertHitCount` moved. It's good where it's at.

I want this closure to wait for at least the expected number of hits to be received when the `assert.hit_count` feature is in use.
https://github.com/elastic/elastic-package/blob/74bc7061097d3cab12506b78674e1bcd5ed39d50/internal/testrunner/runners/system/runner.go#L446-L453

internal/testrunner/runners/system/runner.go
```diff
@@ -443,6 +443,11 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 
 		var err error
 		docs, err = r.getDocs(dataStream)
+
+		if config.Assert.HitCount > 0 {
```
ebeahan, line 447: @andrewkroh Since I already had context, I added this logic per your guidance. If this case needs more careful handling, I'll have someone else refactor.

andrewkroh (approved):

test/packages/other/hit_count_assertion/manifest.yml
```diff
@@ -0,0 +1,24 @@
+format_version: 1.0.0
+name: hit_count_assertion
+title: Hit Count Assertion test
+# version is set to something very large to so this test package can
+# be installed in the package registry regardless of the version of
+# the actual apache package in the registry at any given time.
+version: 999.999.999
+description: Test for Hit Count Assertion
+categories: ["network"]
+release: experimental
```
jsoriano, line 10: Tests are failing in CI because the package is not available for the version of the stack used. Experimental packages are not returned by the package registry by default, try removing this line.
```suggestion
```

test/packages/other/hit_count_assertion/manifest.yml
```diff
@@ -0,0 +1,24 @@
+format_version: 1.0.0
```
jsoriano, line 1: Nit. Use a more recent version of the spec.
```suggestion
format_version: 2.5.0
```

jsoriano (changes requested):

ebeahan:
@jsoriano thanks for taking another look and for the feedback. I think the CI issues are resolved.

ebeahan:
@jsoriano I believe I need a final ✅ if everything looks ok from your side.

jsoriano (approved):
Sorry for the delay!

Issue #837, opened 2022-05-31
[field validation] Check event.type values are aligned with event.category


andrewkroh:
The `event.type` values are constrained based on the `event.category` values. Each category value has an expected list of `event.type` values that can be used. ECS lists these values `expected_event_types` in `ecs_flat.yml` [^1].

As a package developer I would like elastic-package to validate the event.category and event.type combinations used events to ensure they are allowed by ECS.

Related

- https://github.com/elastic/elastic-package/issues/615
- https://github.com/elastic/integrations/pull/3424#discussion_r881667201

[^1]: https://github.com/elastic/ecs/blob/b2df3bc395e125b9380a25e009eddef1b1890476/generated/ecs/ecs_flat.yml#L2645-L2648

jsoriano:
Putting this also on hold till we have something like https://github.com/elastic/package-spec/issues/313.

PR #961, opened 2022-09-07
Validate that event.type is aligned with event.category

jsoriano:
Fields like `event.category` in ECS define a list of allowed values for `event.type`, check that these values are aligned.

This check is only enabled for packages using at least format version 2.0.0.

Fixes https://github.com/elastic/elastic-package/issues/837.
Part of https://github.com/elastic/package-spec/issues/399.

### DIFF
--- a/test/packages/parallel/apache/data_stream/access/_dev/test/pipeline/test-access-basic.log-expected.json
+++ b/test/packages/parallel/apache/data_stream/access/_dev/test/pipeline/test-access-basic.log-expected.json
@@ -13,10 +13,13 @@
                     "web"
                 ],
                 "created": "2020-04-28T11:07:58.223Z",
-                "ingested": "2022-09-06T07:11:46.614845014Z",
+                "ingested": "2022-09-08T10:05:05.863598691Z",
                 "kind": "event",
                 "original": "::1 - - [26/Dec/2016:16:16:29 +0200] \"GET /favicon.ico HTTP/1.1\" 404 209",
-                "outcome": "failure"
+                "outcome": "failure",
+                "type": [
+                    "access"
+                ]
             },
             "http": {
                 "request": {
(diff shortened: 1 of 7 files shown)


internal/fields/validate.go
```diff
@@ -475,6 +476,11 @@ func (v *Validator) parseSingleElementValue(key string, definition FieldDefiniti
 		if err := ensureAllowedValues(key, valStr, definition); err != nil {
 			return err
 		}
+		if !v.specVersion.LessThan(semver.MustParse("2.0.0")) {
```
mtojek, line 479: nit: you can initialize `semver.MustParse("2.0.0")` somewhere at the top

internal/fields/validate_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/elastic/elastic-package/internal/common"
```
mtojek, line 12: nit: sorting

mtojek (commented):
I admit that it wasn't easy to go through the logic :) Maybe it's the first symptom to refactor the `fields` code.

jsoriano:
> I admit that it wasn't easy to go through the logic :) Maybe it's the first symptom to refactor the `fields` code.

Yeah, this is the first time we need to validate the values of two fields at the same time, this is why I ended up passing the whole document through all the calls chain so when validating one field we can check the value of the other.

Two possible alternatives would be:
* Internally instantiate a document validator for each document, this validator could keep state for each document validation, or the document itself.
* Given that by now we only need this for the `event.type` and `event.category` fields, validate them explicitly in `ValidateDocumentMap`, but this would give special consideration to these fields, what can be tricky in the future if ECS decides to use `expected_event_types` in other fields apart of `event.category`.

mtojek (approved):
Frankly speaking, both propositions are rather hackish. I think that I'd rather keep your current proposal (PR).

jsoriano:
> Frankly speaking, both propositions are rather hackish. I think that I'd rather keep your current proposal (PR).

Yep, I also prefer the current approach even if not so nice to have to pass the document around.

mtojek (approved):

PR #968, opened 2022-09-08
Validate the proper value of event.dataset

jsoriano:
If the `event.dataset` and `data_stream.dataset` fields are used, they must have as values the name of the package and the name of the data stream, separated by a dot.

Fixes https://github.com/elastic/package-spec/issues/217.
Part of https://github.com/elastic/package-spec/issues/399.

### DIFF
--- a/docs/howto/update_major_package_spec.md
+++ b/docs/howto/update_major_package_spec.md
@@ -62,14 +62,15 @@ For example if a document contains `event.category: web`, the value of
 
 ### field "event.dataset" should have value ..., it has ...
 
-The field `event.dataset` should contain the name of the package and the name of
-the data stream that generates it, separated by a dot. For example for documents
-of the "access" data stream of the Apache module, it should be `apache.access`.
+The fields `event.dataset` and `data_stream.dataset` should contain the name of
+the package and the name of the data stream that generates it, separated by a
+dot. For example for documents of the "access" data stream of the Apache module,
+it should be `apache.access`.
 
-To fix this, review what is generating the value of this field. Depending on
-your case, you may apply one of the following solutions.
+If these fields are not being correctly populated, look for the source of the
+value.
 
-Fix the value in the field definition using a `constant_keyword`:
+If it is a constant keyword, review the configured value.
 ```
 - name: event.dataset
   type: constant_keyword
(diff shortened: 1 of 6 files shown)


docs/howto/update_major_package_spec.md
```diff
@@ -59,3 +59,30 @@ some of the expected values.
...
+the data stream that generates it, separated by a dot. For example for documents
+of the "access" data stream of the Apache module, it should be `apache.access`.
+
+To fix this, review what is generating the value of this field. Some ways to fix
+it are the following.
+
+Fix the value in the field definition using a `constant_keyword`:
+```
+- name: event.dataset
+  type: constant_keyword
```
jsoriano, line 75: This is not possible now. We don't allow to override the type of an imported value. But I think it would make sense to support the change from keyword to constant_keyword. I will open a PR to support it.

jsoriano: Added in https://github.com/elastic/elastic-package/pull/969.

mrodm (approved):
LGTM

docs/howto/update_major_package_spec.md
```diff
@@ -59,3 +59,32 @@ some of the expected values.
 
 For example if a document contains `event.category: web`, the value of
 `event.type` must be `access`, `error` or `info` according to ECS 8.4.
+
+### field "event.dataset" should have value ..., it has ...
+
+The fields `event.dataset` and `data_stream.dataset` should contain the name of
+the package and the name of the data stream that generates it, separated by a
+dot. For example for documents of the "access" data stream of the Apache module,
+it should be `apache.access`.
```
mrodm, line 68: In this section, I think it should also be added `data_stream.dataset`. Same error could happen in that field.

```suggestion
### field "event.dataset"/"data_stream.dataset" should have value ..., it has ...

Both fields `event.dataset` and `data_stream.dataset` should contain the name of
the package and the name of the data stream that generates it, separated by a
dot. For example for documents of the "access" data stream of the Apache module,
it should be `apache.access`.
```

mrodm: Ignore this comment, it was before you pushing the latest changeset.

mrodm (approved):

PR #969, opened 2022-09-08
Allow to override ECS fields type from keyword to constant_keyword

jsoriano:
Support the case of defining the value of a constant_keyword for fields that are defined as keyword in ECS.

Current workaround is to define the whole field on each data stream instead of importing it, so the `constant_keyword` type can be set along with a value. Something like this:
```
- name: event.dataset
  type: constant_keyword
  description: Event dataset
  value: apache.access
```

With this change, it will be allowed to reuse the definition from ECS:
```
- name: event.dataset
  external: ecs
  type: constant_keyword
  value: apache.access
```

This is helpful to have consistent descriptions and other settings in more complex fields.


mtojek (approved):
Thanks for leaving a comment there :) You can get lost with all these rules.

PR #977, opened 2022-09-13
Add docs about validation of ILM policy

jsoriano:
Add troubleshooting docs for validation added in https://github.com/elastic/package-spec/pull/413.

### DIFF
--- a/docs/howto/update_major_package_spec.md
+++ b/docs/howto/update_major_package_spec.md
@@ -124,3 +124,27 @@ You can instead use the object type directly as type:
 - name: ciphersuites
   type: keyword
 ```
+
+### Invalid field ilm_policy
+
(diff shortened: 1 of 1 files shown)


mrodm (approved):

docs/howto/update_major_package_spec.md
```diff
@@ -124,3 +124,27 @@ You can instead use the object type directly as type:
...
+```
+
+And an ILM policy definition exists in:
+```
+./data_streams/{data stream name}/elasticsearch/ilm/{ilm file name without extension}.json
+```
+
+It could be the case that you are trying to reference to an ILM policy defined
+in some other place out of the package. This is not supported. If you have a use
+case for this, please open an issue in https://github.com/elastic/package-spec
```
mtojek, line 149: nit: maybe a link?

mtojek (approved):

PR #978, opened 2022-09-15
Add docs about how to migrate invalid properties in fields definitions

jsoriano:
Troubleshooting docs for the changes in https://github.com/elastic/package-spec/pull/420.

### DIFF
--- a/docs/howto/update_major_package_spec.md
+++ b/docs/howto/update_major_package_spec.md
@@ -22,6 +22,17 @@ conditions:
   elastic.subscription: basic
 ```
 
+### field ...: Additional property ... is not allowed
+
+Validation of properties on fields definitions is stricter now. If you see any
+error about properties not allowed in fields definitions, this probably means
+one of these two things:
+
+There is a typo, please check that this is the property you wanted to use.
+
+The property is not supported, then you should probably remove it. If you think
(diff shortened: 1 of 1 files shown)


docs/howto/update_major_package_spec.md
```diff
@@ -22,6 +22,20 @@ conditions:
...
+### field ...: Additional property ... is not allowed
+
+Validation of properties on fields definitions is stricter now. If you see any
+error about properties not allowed in fields definitions, this probably means
+one of these two things:
+
+There is a typo, please check that this is the property you wanted to use.
+
+The property is not supported, then you should probably remove it. If you think
+that the property should be supported, please [open an issue](https://github.com/elastic/package-spec/issues/new?assignees=&labels=discuss&template=Change_Proposal.md&title=%5BChange+Proposal%5D+)
```
mrodm, line 34: Just a suggestion, adding the two common issues in a list 

```suggestion
Validation of properties on fields definitions is stricter now. If you see any
error about properties not allowed in fields definitions, this probably means
one of these two things:

- There is a typo, please check that this is the property you wanted to use.
- The property is not supported, then you should probably remove it. If you think
that the property should be supported, please [open an issue](https://github.com/elastic/package-spec/issues/new?assignees=&labels=discuss&template=Change_Proposal.md&title=%5BChange+Proposal%5D+)
```

mrodm (approved):

PR #979, opened 2022-09-16
[system tests] Use datastream dataset if it exists

klacabane:
### Summary
The system tests didn't account for the [overwriting dataset](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/manifest.spec.yml#L109) defined in a data stream's manifest.

For example the [logstash.node](https://github.com/elastic/integrations/blob/main/packages/logstash/data_stream/node/manifest.yml#L4) data stream that overwrites the dataset fails with the following:

`Error: error running package system tests: could not complete test run: could not add data stream config to policy: could not add package to policy; API status code = 500; response body = {"statusCode":500,"error":"Internal Server Error","message":"Stream template not found, unable to find dataset logstash.node"}`

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -533,7 +533,7 @@ func createPackageDatastream(
 			Enabled: true,
 			DataStream: kibana.DataStream{
 				Type:    ds.Type,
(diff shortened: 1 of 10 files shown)


jsoriano (approved):
Thanks!

PR #986, opened 2022-09-21
Show error if assets are being exported using a Elastic stack SNAPSHOT version

mrodm:
Closes https://github.com/elastic/package-spec/issues/318

This PR shows an error when assets are exported from an Elastic stack SNAPSHOT version. Using those assets generated by a snapshot version may lead to use features that are not released at the end, or maybe they are in a different way.

In case developers need to use an SNAPSHOT version, a new parameter has been added `--allow-snapshots` to allow this export process.

Output using SNAPSHOT Version:
- Running `elastic-package export dashboards`:
  ```
   $ elastic-package export dashboards 
  Export Kibana dashboards
  Error: exporting dashboards from a SNAPSHOT version of Kibana (8.4.0-SNAPSHOT) is discouraged. It could lead to invalid dashboards (for example if they use features that are reverted or modified before the final release). --allow-snapshot flag can be used to ignore this error
  ```
- Running `elastic-package export dashboards --allow-snapshot`:
  ```
   $ elastic-package export dashboards --allow-snapshot
  Export Kibana dashboards
  Warning: exporting dashboards from a SNAPSHOT version of Kibana (8.4.0-SNAPSHOT) is discouraged. It could lead to invalid dashboards (for example if they use features that are reverted or modified before the final release)
  ? Which dashboards would you like to export?  [Use arrows to move, space to select, <right> 
  to all, <left> to none, type to filter]
  ...
  ```

Tested with 8.0.0, 8.4.0-SNAPSHOT and 7.17.6 and 7.17.0-SNAPSHOT versions.

In this PR, the [`/api/status` endpoint](https://www.elastic.co/guide/en/kibana/current/access.html#status) is being tested to retrieve the version information. This includes a dedicated field about whether or not this is a snapshot build (`build_snapshot` field).
- Running command `elastic-package stack up -v -d --version 8.0.0`:
```
 $ curl -u elastic:changeme -H 'Content-Type: application/json' -H 'kbn-xsrf: true' -k -s https://localhost:5601/api/status? | jq -r .version
{
  "number": "8.0.0",
  "build_hash": "57ca5e139a33dd2eed927ce98d8231a1f217cd15",
  "build_number": 49192,
  "build_snapshot": false
}
```
- Running command `elastic-package stack up -v -d --version 8.4.0-SNAPSHOT`:
```
 $ curl -u elastic:changeme -H 'Content-Type: application/json' -H 'kbn-xsrf: true' -k -s https://localhost:5601/api/status | jq -r .version
{
  "number": "8.4.0",
  "build_hash": "7ffc375be69ab56267d3fc71bf7f27da6dfacdfb",
  "build_number": 55408,
  "build_snapshot": true
}
```

In case of build_snapshot true , it must be added the suffix "-SNAPSHOT" to the version.number field.


cmd/export.go
```diff
@@ -59,11 +60,29 @@ func exportDashboardsCmd(cmd *cobra.Command, args []string) error {
...
 		return errors.Wrap(err, "can't create Kibana client")
 	}
 
+	kibanaVersion, isSnapshot, err := kibanaClient.Version()
+	if err != nil {
+		return errors.Wrap(err, "can't get Kibana status information")
+	}
+
+	if isSnapshot {
+		message := fmt.Sprintf("Exporting dashboards from a Elastic stack SNAPSHOT version (%s) is discouraged. It could lead to invalid dashboards (e.g. using features that could be not released or they could changed)", kibanaVersion)
```
jsoriano, line 79: Nit. Some rephrasing here.
```suggestion
		message := fmt.Sprintf("exporting dashboards from a SNAPSHOT version of Kibana (%s) is discouraged. It could lead to invalid dashboards (for example if they use features that are reverted or modified before the final release)", kibanaVersion)
```

cmd/export.go
```diff
@@ -59,11 +60,29 @@ func exportDashboardsCmd(cmd *cobra.Command, args []string) error {
...
+		if !allowSnapshot {
+			return errors.Errorf("using Elastic stack SNAPSHOT version: %s. %s", kibanaVersion, message)
```
jsoriano, line 81: Nit. The version is already in the message. Maybe we can add here a mention about the flag.
```suggestion
			return errors.Errorf("%s. --%s flag can be used to ignore this error", message, cobraext.AllowSnapshotFlagName)
```

jsoriano (commented):
:+1: some nitpicking in the error messages.

internal/kibana/status.go
```diff
@@ -0,0 +1,47 @@
...
+
+type statusType struct {
+	Version struct {
+		Number        string `json:"number"`
+		BuildSnapshot bool   `json:"build_snapshot"`
+	} `json:"version"`
+}
+
+// Version method returns the version of Kibana (Elastic stack)
+func (c *Client) Version() (string, bool, error) {
```
mrodm, line 25: Version method from `internal/kibana/injected_metadata.go` deleted in favor of this one.

This new method uses the `/api/status` endpoint, and it has a dedicated field to indicate whether or not the version is a snapshot one ([build_snapshot](https://github.com/elastic/kibana/blob/main/packages/core/status/core-status-server-internal/src/routes/status.ts#L123)).

internal/kibana/status.go
```diff
@@ -0,0 +1,47 @@
...
+	err = json.Unmarshal(respBody, &status)
+	if err != nil {
+		return "", false, errors.Wrapf(err, "unmarshalling response failed (body: \n%s)", respBody)
+	}
+
+	stackVersion := status.Version.Number
+	if status.Version.BuildSnapshot {
+		stackVersion = fmt.Sprintf("%s%s", stackVersion, SNAPSHOT_SUFFIX)
+	}
+
```
mrodm, line 45: This API returns the version without `-SNAPSHOT`. https://github.com/elastic/kibana/blob/main/packages/core/status/core-status-server-internal/src/routes/status.ts#L120-L123

That's why it is needed to be added here.

Example of output (Elastic stack 7.17.0-SNAPSHOT):
```
 $ curl -u elastic:changeme -H 'Content-Type: application/json' -H 'kbn-xsrf: true' -k -s https://localhost:5601/api/status | jq -r .version
{
  "number": "7.17.0",
  "build_hash": "0ab325e4b36168a264907fbc84f6913519671acf",
  "build_number": 46537,
  "build_snapshot": true
}
```

internal/kibana/status.go
mrodm: Sure! Added that new object.

PR #993, opened 2022-09-28
render diff for readme checks

efd6:
This renders the difference between the stored and rendered readme files. Like so:
```
[2022-09-27T07:35:28.126Z] README.md is outdated. Rebuild the package with 'elastic-package build'
[2022-09-27T07:35:28.126Z]   strings.Join({
[2022-09-27T07:35:28.126Z]   	... // 49896 identical bytes
[2022-09-27T07:35:28.126Z]   	"| host.os.name | Operating system name, without the version. | k",
[2022-09-27T07:35:28.126Z]   	"eyword |\n| host.os.name.text | Multi-field of `host.os.name`. | ",
[2022-09-27T07:35:28.126Z] - 	"match_only_",
[2022-09-27T07:35:28.126Z]   	"text |\n| host.os.platform | Operating system platform (such cent",
[2022-09-27T07:35:28.126Z]   	"os, ubuntu, windows). | keyword |\n| host.os.version | Operating ",
[2022-09-27T07:35:28.126Z]   	... // 10306 identical bytes
[2022-09-27T07:35:28.126Z]   }, "")
[2022-09-27T07:35:28.126Z] Error: checking package failed: checking readme files are up-to-date failed: files do not match
```
This is important due to differential behaviour between CI and local `elastic-package build`.

### DIFF
--- a/cmd/lint.go
+++ b/cmd/lint.go
@@ -47,7 +47,7 @@ func lintCommandAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		for _, f := range readmeFiles {
 			if !f.UpToDate {
-				cmd.Printf("%s is outdated. Rebuild the package with 'elastic-package build'\n", f.FileName)
+				cmd.Printf("%s is outdated. Rebuild the package with 'elastic-package build'\n%s", f.FileName, f.Diff)
 			}
 			if f.Error != nil {
 				cmd.Printf("check if %s is up-to-date failed: %s\n", f.FileName, f.Error)
--- a/internal/docs/readme.go
+++ b/internal/docs/readme.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"text/template"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 
 	"github.com/elastic/elastic-package/internal/builder"
@@ -22,6 +23,7 @@ import (
 type ReadmeFile struct {
 	FileName string
 	UpToDate bool
+	Diff     string
 	Error    error
 }
 
@@ -40,11 +42,12 @@ func AreReadmesUpToDate() ([]ReadmeFile, error) {
 	var readmeFiles []ReadmeFile
 	for _, filePath := range files {
 		fileName := filepath.Base(filePath)
-		ok, err := isReadmeUpToDate(fileName, packageRoot)
+		ok, diff, err := isReadmeUpToDate(fileName, packageRoot)
 		if !ok || err != nil {
 			readmeFile := ReadmeFile{
 				FileName: fileName,
 				UpToDate: ok,
+				Diff:     diff,
 				Error:    err,
 			}
 			readmeFiles = append(readmeFiles, readmeFile)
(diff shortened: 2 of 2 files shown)


jsoriano:
> This is important due to differential behaviour between CI and local `elastic-package build`.

To what different behavior do you refer?

When this error happens, locally you can run `elastic-package build`, and see with `git diff` what are the changes.

> [2022-09-27T07:35:28.126Z] README.md is outdated. Rebuild the package with 'elastic-package build'
> [2022-09-27T07:35:28.126Z]   strings.Join({
> [2022-09-27T07:35:28.126Z]   	... // 49896 identical bytes
> [2022-09-27T07:35:28.126Z]   	"| host.os.name | Operating system name, without the version. | k",
> [2022-09-27T07:35:28.126Z]   	"eyword |\n| host.os.name.text | Multi-field of `host.os.name`. | ",
> [2022-09-27T07:35:28.126Z] - 	"match_only_",

Is it expected for this diff to show the `strings.Join(...`... text? This looks like non evaluated Go code. Would it be possible to show a unified diff?

efd6:
At the moment the builds on CI are consistently rendering readme files differently to how they are rendered locally.

The diff is as it's intended to be, I would not like to see huge diffs here, just enough to be able to get an idea of what is being misrendered.

efd6:
If you would prefer, we can do what we have in the pipeline result diffing
```
	var buf bytes.Buffer
	err = difflib.WriteUnifiedDiff(&buf, difflib.UnifiedDiff{
		A:        difflib.SplitLines(string(existing)),
		B:        difflib.SplitLines(string(rendered)),
		FromFile: "want",
		ToFile:   "got",
		Context:  3,
	})
	return false, buf.String(), err
```

jsoriano:
Yes, I think I would prefer a more standard diff, could you give it a try and show an example?

efd6:
Sure. I'm thinking a context of 2 would be enough for this situation. WDYT?

jsoriano:
Agree with a small diff :+1:

efd6:
I tried 2 and I think it's still too much. This is 1:
```
Format the package
Done
Lint the package
buildmanifest.Dependencies{ECS:buildmanifest.ECSDependency{Reference:"git@v8.4.0-rc1"}}
README.md is outdated. Rebuild the package with 'elastic-package build'
--- want
+++ got
@@ -308,3 +308,2 @@
 | rsa.internal.cid | This is the unique identifier used to identify a NetWitness Concentrator. This key should never be used to parse Meta data from a session (Logs/Packets) Directly, this is a Reserved key in NetWitness | keyword |
-| rsa.internal.data | Deprecated key defined only in table map. | keyword |
 | rsa.internal.dead | Deprecated key defined only in table map. | long |
@@ -342,4 +341,2 @@
 | rsa.internal.msg | This key is used to capture the raw message that comes into the Log Decoder | keyword |
-| rsa.internal.msg_id | This is the Message ID1 value that identifies the exact log parser definition which parses a particular log session. This key should never be used to parse Meta data from a session (Logs/Packets) Directly, this is a Reserved key in NetWitness | keyword |
-| rsa.internal.msg_vid | This is the Message ID2 value that identifies the exact log parser definition which parses a particular log session. This key should never be used to parse Meta data from a session (Logs/Packets) Directly, this is a Reserved key in NetWitness | keyword |
 | rsa.internal.node_name | Deprecated key defined only in table map. | keyword |
@@ -347,3 +344,2 @@
 | rsa.internal.obj_id | Deprecated key defined only in table map. | keyword |
-| rsa.internal.obj_server | Deprecated key defined only in table map. | keyword |
 | rsa.internal.obj_val | Deprecated key defined only in table map. | keyword |
Error: checking package failed: checking readme files are up-to-date failed: files do not match
```

jsoriano (approved):

mrodm (approved):

Issue #881, opened 2022-06-30
Profiles without cert files cannot be listed


mrodm:
After being updated elastic-package to enable SSL https://github.com/elastic/elastic-package/pull/789 , old profiles that do not have CA or certificate files cannot be listed.
This causes that `elastic-package profiles list` command do not work.

Output example:

```bash
 $ elastic-package profiles list
Error: error listing all profiles: error loading profile default_dec6298_1656436608: error reading in profile default_dec6298_1656436608: error reading in profile: reading filed failed (path: /Users/mariorodriguez/.elastic-package/profiles/default_dec6298_1656436608/certs/fleet-server/key.pem): open /Users/mariorodriguez/.elastic-package/profiles/default_dec6298_1656436608/certs/fleet-server/key.pem: no such file or directory
```

Issue #589, opened 2021-11-17
Old profiles no longer work with 0.29.0 (and profiles list breaks)


endorama:
I recently updated to `0.29.0`, and when running `profiles list` I encounter this error:

```
$ elastic-package profiles list
Error: error listing all profiles: error loading profile 7.12.0: error reading in profile 7.12.0: error reading in profile: reading filed failed (path: /.../.elastic-package/profiles/7.12.0/stack/elasticsearch.config.default.yml): open /.../.elastic-package/profiles/7.12.0/stack/elasticsearch.config.default.yml: no such file or directory
```

I think this is a side effect of all the changes happening around profiles and configurations to support 8.0, but this is quite a pain. I have some old profiles with custom configuration for testing some specific use cases that require manual updating. While not expecting backward compatibility, this kind of issue create effort downstream and make relying on this tool more risky.

To me this highlights that profiles are quite fragile today. I have some improvements idea that gathers around simplifying the current profile structure, also given that after the introduction of `--version` flag for `stack` commands is possible to combine profiles and versions (making this troublesome to manage and debug).

Are there any proposals to improve the current profiles to make them more future proof? (I've seen https://github.com/elastic/elastic-package/issues/557 but seems more code centric than user centric).

mtojek:
Thanks for raising the issue.

I'm afraid we have to rewrite this piece of code as there is some strong coupling between embedded files, source code, and logic responsible for creating profiles.

There will be always an issue with integrity, but what we can do in this case is try to use a default file.

Anyway, bugfix PRs are welcome!

jsoriano:
Profiles list fixed in https://github.com/elastic/elastic-package/pull/1006, the approach implemented there should be more future proof, till we implement something like https://github.com/elastic/elastic-package/issues/557.

PR #1006, opened 2022-10-17
Ignore missing files while loading profiles metadata

jsoriano:
While loading profiles, ignore missing files. These errors can be caused by old or corrupted profiles, and shouldn't prevent elastic-package to read the rest of the profile or its metadata.

Fixes https://github.com/elastic/elastic-package/issues/881.
Solves the part of https://github.com/elastic/elastic-package/issues/589 about broken profiles list.


mrodm (approved):
:+1:

PR #1015, opened 2022-10-25
Update package-registry to use EPR production as proxy

mrodm:
This PR updates the Dockerfile that creates the Elastic Package Registry image to configure the Elastic Package Registry production DNS (https://epr.elastic.co) to be used as proxy. Currently, the default value used is the snapshot one (https://epr-snapshot.elastic.co)

### DIFF
--- a/internal/profile/_static/Dockerfile.package-registry
+++ b/internal/profile/_static/Dockerfile.package-registry
@@ -5,6 +5,7 @@ ARG PROFILE
 ENV EPR_DISABLE_PACKAGE_VALIDATION=true
 
 ENV EPR_FEATURE_PROXY_MODE=true
+ENV EPR_PROXY_TO=https://epr.elastic.co
 
 COPY profiles/${PROFILE}/stack/package-registry.config.yml /package-registry/config.yml
 COPY stack/development/ /packages/development


internal/profile/_static/Dockerfile.package-registry
```diff
@@ -5,6 +5,7 @@ ARG PROFILE
 ENV EPR_DISABLE_PACKAGE_VALIDATION=true
 
 ENV EPR_FEATURE_PROXY_MODE=true
+ENV EPR_PROXY_TO=https://epr.elastic.co
```
jsoriano, line 8: Is it needed if we are going to change the default in package-registry itself?

jsoriano (approved):

internal/profile/_static/Dockerfile.package-registry
mrodm: I just add it in case we want to push this change quick into elastic-package.

This is not totally mandatory, if we can wait until the package-registry release and the required update here in elastic-package (update the docker image used for package-registry - https://github.com/elastic/elastic-package/blob/main/internal/profile/static.go#L126).

I'll keep the PR opened in case we see it is worthy.

PR #1020, opened 2022-10-27
Remove packages from bucket dry-run execution

mrodm:
As publishing-job-remote is executed in dry-run mode, zip package and its signature files that were uploaded to the publishing queue are not being deleted from there when this job is run.
Therefore, it has been added the needed commands needed to ensure that there are no left files in the publishing queue (bucket).


mrodm:
Closing PR, the account used to upload files has no permission to remove objects.

PR #1024, opened 2022-11-02
Fix to correct policy assignment to unhealthy agents when using custom agents

### LINKED ISSUE #1023: elastic-package fails if there is more than one scenario in with-custom-agent
Seen in this PR https://github.com/elastic/elastic-package/pull/1009
Relates https://github.com/elastic/elastic-package/issues/997

After adding a new package into the test suite of `with-custom-agent`, it looks like that after running the first run of the system testing (with auditbeat package), there are some leftovers or conditions that make the second run with the oracle pckage to fail.

agithomas:
Fixes #1023

Elastic-package is not checking the status of the agent (if unhealthy or not) when applying the policy when using custom-agent service deployer.


mrodm (approved):
Thanks @agithomas for fixing this bug!

PR #1032, opened 2022-11-08
Allow to configure repository license using environment variable

jsoriano:
Allow to use `ELASTIC_PACKAGE_REPOSITORY_LICENSE` to select the default repository license to use in a repository with multiple licenses. Previously the file `LICENSE.txt` at the root of the repository was used.

Required for https://github.com/elastic/integrations/pull/4593.


mrodm (approved):
:+1:

PR #1042, opened 2022-11-16
Fix fleet-server healthcheck for 8.6

jsoriano:
Fleet server healthcheck has changed its capitalization, before 8.6, it was:
```
{"name":"fleet-server","status":"HEALTHY"}
```
In 8.6 it is:
```
{"name":"fleet-server","status":"healthy"}
```


mrodm (approved):
Good catch! :+1:

mrodm (approved):

PR #1047, opened 2022-11-23
Use consistent dataset in input tests configuration

jsoriano:
It seems that when the agent policy template includes a dataset ([as is the case of the jolokia input](https://github.com/elastic/integrations/blob/99dcdbe8ca003435d800c1abde5e4493c1fe4122/packages/jolokia_input/agent/input/input.yml.hbs#L11-L12)), this is used as part of the data stream name. If names are not consistent there, `elastic-package` cannot find documents when running system tests.


mrodm (approved):
:+1:

PR #1048, opened 2022-11-23
Find overrides of variables with names in their dots

jsoriano:
Variables in policy templates can have dots in their names. This names are defined as plain strings in yaml files. For example like this:
```
      - name: jmx.mappings
        type: yaml
        title: JMX Mappings
        multi: false
        required: true
        show_user: true
```
These variables can be overridden in system tests configuration files, for example like this:
```
vars:
  hosts:
    - {{Hostname}}:{{Port}}
  jmx.mappings: |
    - mbean: 'java.lang:type=Runtime'
      attributes:
        - attr: Uptime
          field: uptime
```
These configuration files are parsed with go-ucfg, configured to use `.` as path separator, so variable names containing dots are expanded. Also, when parsed, they are parsed into an opaque `VarVariable` structure, so it is not possible to access their content, so overrides cannot be defined for variables with dots in their names.
This happens for example in https://github.com/elastic/integrations/pull/4706.

This issue could be easily solved if `ucfg` would support `preserve` as defined in https://github.com/elastic/go-ucfg/issues/124.

This change introduces a hacky workaround to give access to child elements, so they can be also used for overrides.

Alternatives considered:
* Completely disable variable expansion when parsing this configuration. This could produce breaking changes, and could be misleading, as developers are used to this kind of configurations where handling of objects structure is transparent for them.
* Implement `preserve` as defined in https://github.com/elastic/go-ucfg/issues/124. Probably a rabbit hole.
* Parse the configuration as `common.MapStr` instead of `map[string]VarValue`, and generate the `VarValue`s later. This could be cleaner, but require some refactors. I may give it a try though.

### DIFF
--- a/internal/packages/packages.go
+++ b/internal/packages/packages.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/elastic/elastic-package/internal/common"
 	"github.com/elastic/go-ucfg"
 	"github.com/elastic/go-ucfg/yaml"
 )
@@ -34,6 +35,7 @@ const (
 // VarValue represents a variable value as defined in a package or data stream
 // manifest file.
 type VarValue struct {
+	dict   map[string]interface{}
 	scalar interface{}
 	list   []interface{}
 }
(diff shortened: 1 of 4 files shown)


jsoriano:
> * Parse the configuration as `common.MapStr` instead of `map[string]VarValue`, and generate the `VarValue`s later. This could be cleaner, but require some refactors. I may give it a try though.

Ok, this ended up being much more clear: https://github.com/elastic/elastic-package/pull/1049.

PR #1049, opened 2022-11-23
Fix override of variables with dots in their names in system tests

jsoriano:
Use `common.MapStr` to be able to get variables independently of them being expanded because of having dots or not.

Variables in policy templates can have dots in their names. This names are defined as plain strings in yaml files. For example like this:
```
      - name: jmx.mappings
        type: yaml
        title: JMX Mappings
        multi: false
        required: true
        show_user: true
```
These variables can be overridden in system tests configuration files, for example like this:
```
vars:
  hosts:
    - {{Hostname}}:{{Port}}
  jmx.mappings: |
    - mbean: 'java.lang:type=Runtime'
      attributes:
        - attr: Uptime
          field: uptime
```
These configuration files are parsed with go-ucfg, configured to use `.` as path separator, so variable names containing dots are expanded. Also, when parsed, they are parsed into an opaque `VarVariable` structure, so it is not possible to access their content, so overrides cannot be defined for variables with dots in their names.
This happens for example in https://github.com/elastic/integrations/pull/4706.

This issue could be easily solved if `ucfg` would support `preserve` as defined in https://github.com/elastic/go-ucfg/issues/124.

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -639,14 +639,15 @@ func createInputPackageDatastream(
 	return r
 }
 
-func setKibanaVariables(definitions []packages.Variable, values map[string]packages.VarValue) kibana.Vars {
+func setKibanaVariables(definitions []packages.Variable, values common.MapStr) kibana.Vars {
 	vars := kibana.Vars{}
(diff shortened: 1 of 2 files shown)


mrodm (approved):
LGTM

PR #1052, opened 2022-11-25
Check cluster health before trying to run tests

jsoriano:
Trying to run tests with an unhealthy cluster may lead to obscure failures. Explicitly check the status of the cluster before trying to run tests.

### DIFF
--- a/internal/elasticsearch/client.go
+++ b/internal/elasticsearch/client.go
@@ -5,8 +5,11 @@
 package elasticsearch
 
 import (
+	"context"
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 
@@ -27,6 +30,9 @@ type IngestSimulateRequest = esapi.IngestSimulateRequest
 
 // IngestGetPipelineRequest configures the Ingest Get Pipeline API request.
 type IngestGetPipelineRequest = esapi.IngestGetPipelineRequest
+
+// ClusterStateRequest configures the Cluster State API request.
+type ClusterStateRequest = esapi.ClusterStateRequest
 
 // clientOptions are used to configure a client.
 type clientOptions struct {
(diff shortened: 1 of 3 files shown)


internal/elasticsearch/client.go
```diff
@@ -108,5 +120,114 @@ func Client(customOptions ...ClientOption) (*elasticsearch.Client, error) {
...
+
+	if status := clusterHealth.Status; status != "green" && status != "yellow" {
+		if status != "red" {
+			return errors.Errorf("cluster in unhealthy state: %q", status)
+		}
+		cause, err := client.redHealthCause(ctx)
+		if err != nil {
+			return errors.Wrapf(err, "cluster in unhealthy state, failed to identify cause")
+		}
+		return errors.Errorf("cluster in unhealthy state: %s", cause)
```
mrodm, line 155: I was thinking if it would be helpful to diagnose further to include here something like:
> More info running inside the elasticsearch container: curl -k "https://localhost:9200/_internal/_health?pretty"

No need to run this command inside the elasticsearch contianer.

mrodm (approved):

internal/elasticsearch/client.go
jsoriano: Umm, this `cause` is obtained from `_internal/_health`. From all the information there I am only collecting the causes of impacts with highest severity. Should we collect something else?

mrodm: Just re-thinking about it, probably there is no need to add that message.

Let's consider that there are two "red" issues in the Elasticsearch cluster with different severity levels. In the first run, the cause of the issue with the highest severity would be shown to the user. Once solved, in the second run/attempt would be shown the other issue with the lowest severity if it is still active.

Taking into account that, I would say this could be a good way of handling these errors/issues. Displaying the issue with the highest severity. :+1: So, just dismiss this.

Moreover, in the case of using `elastic-package` in CI that message would not be helpful at all.

jsoriano: Well, I could ignore the severity and just return all the causes. But for the example of the disk, it would also append the cause for shard unavailability, that would be redundant. We may consider a different approach if we find other cases to handle.

Issue #952, opened 2022-08-29
CLI command to attach a package policy to the Agent


klacabane:
### Summary

The `elastic-package install` command currently installs a package assets. It would be useful to also have a command to attach an integration policy of the installed package to the elastic-agent started by the stack, after prompting for the inputs configuration. This will avoid UI configuration and enable automation of the package installation.

It could be a flag provided to `install` or its own dedicated command

klacabane:
The [system tests](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md#test-case-definition) already do that by enrolling a new agent with policies defined in test configs under `_dev/test/system/`. Would it make sense to reuse that logic to have an agent continuously running with these configs ?

jsoriano:
As discussed over slack, this feature makes sense, it could be something like `elastic-package install --attach <agent id>`. https://github.com/elastic/kibana/issues/70582 would help to use this feature with any package on any cluster.

klacabane:
After a first look at the workflow and code, the following two steps are required:

* Create and configure the integration. We need a way to build the integration inputs and the ability to provide variables like hosts, credentials... In system tests we define a static yml configuration ([example](https://github.com/elastic/integrations/blob/main/packages/elasticsearch/data_stream/cluster_stats/_dev/test/system/test-default-config.yml)) that allows configuration for a single data stream.
This should not be much different than what is already done, we'll only have to extend the capability to cover multiple streams: similar to the system tests we can build a policy by parsing the package and data stream's manifest, but instead of reading a single stream we'll have to parse every ds defined in `data_streams` directory. We'll then use user-defined configuration file (under `<package>/_dev/policy` ?) that will provide both package and ds variables that need to be substituted. The configuration file's format would like like this:
```yaml
inputs:
  - type: elasticsearch/metrics
    vars:
      hosts:
        - "http://{{Hostname}}:5066"
    streams:
      - dataset: elasticsearch.stack_monitoring.cluster_stats
```

* Attach the integration to an agent policy
Ability to provide an agent id `--attach <agent-id>` or fallback to the first available agent if none provided. [This code already exists](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/runner.go#L401-L405) in the system test runner's logic and we should be able to reuse it

klacabane:
Logging observations made while digging further:
- variables can be defined at package, input and data stream levels
[...]

PR #1065, opened 2022-12-12
Enroll cmd

klacabane:
### Summary

Add a command that spawns a agent with preconfigured policy. The mechanism extends the system tests approach to have a configuration that accepts definition of multiple data streams. Agents are created in unique container under the `elastic-package-service` project and can be run in parallel.

**Usage `elastic-package enroll --config default_config.yml`**

The goal was a working POC for spacetime week so I didn't test this extensively, only a couple scenarios in the stack packages. I also went for quick wins in implementations by tweaking existing mechanisms like the custom agent to allow our new use case.

The initial use case is to allow automated installation of a package (see https://github.com/elastic/elastic-package/issues/952) but the mechanism could be reused for further testing scenarios, like spinning a specific version of an agent (maybe allow local binaries as well?) or replace the current system tests approach. For the latter the thinking is that the Enroll mechanism could allow parallel test execution for packages that have large sets of streams + variants (see https://github.com/elastic/integrations/pull/4442#pullrequestreview-1146776038) and speed things up, but there is maybe a simpler way for these gains.

### Process
- Agent policy is created from the configuration[1] provided to the command
- Agent is then spawned with the previously policy attached so it starts to collect metrics as soon as it is setup
- todo: cleanup agent, policy and streams when done

### Testing
- create a `default_config.yml` under `<integrations-repo>/packages/elasticsearch/_dev/policy` with the following configuration

_[1] example configuration_
```
policy_templates:
  - name: elasticsearch
    inputs:
      - type: elasticsearch/metrics
        vars:
          hosts:
            - https://elasticsearch:9200
          username: elastic
          password: changeme
        streams:
          - dataset: elasticsearch.stack_monitoring.cluster_stats
          - dataset: elasticsearch.stack_monitoring.index_recovery
          - dataset: elasticsearch.stack_monitoring.node_stats
```
- start a stack `elastic-package stack up -v -d --version 8.7.0-SNAPSHOT`
- within the `elasticsearch` package scope, enroll an agent with the configuration `elastic-package enroll --config default_config.yml`
- an agent is spawned under `elastic-package-service` project
- navigate to Fleet UI and verify a corresponding integration was created. It should be named `elastic-package_elasticsearch_<timestamp>`


internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -26,16 +28,51 @@ const dockerCustomAgentName = "docker-custom-agent"
...
 	}, nil
 }
 
+func NewAgentDeployer() *CustomAgentDeployer {
+	return &CustomAgentDeployer{
+		serviceName: fmt.Sprintf("elastic-package-agent_%d", time.Now().UnixNano()),
+	}
+}
+
+func (d *CustomAgentDeployer) buildYml() string {
```
klacabane, line 49: The docker-compose file is now built on demand to allow dynamic service names

internal/testrunner/runners/system/servicedeployer/custom_agent.go
```diff
@@ -110,12 +157,6 @@ func (d *CustomAgentDeployer) SetUp(inCtxt ServiceContext) (DeployedService, err
 		return nil, errors.Wrap(err, "could not boot up service using Docker Compose")
 	}
 
-	// Connect service network with stack network (for the purpose of metrics collection)
-	err = docker.ConnectToNetwork(p.ContainerName(serviceName), stack.Network())
-	if err != nil {
-		return nil, errors.Wrapf(err, "can't attach service container to the stack network")
-	}
```
klacabane, line 117: I moved this to the docker-compose file as I've seen [this call fail](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/cmd/container.go#L494) in agent containers. I assumed it is caused by the period where the agent is started but not attached yet, and it is, but the retry mechanism eventually succeeds so this is not necessary

sharbuz:
Hi everyone,

Issue #1030, opened 2022-11-08
shellinit shell detection fails sometimes with "unimplemented"


jsoriano:
In some machines, `eval $(elastic-package stack shellinit)` fails with:
```
Error: cannot detect parent shell from current process: cannot retrieve information for process 60983: unimplemented
```

Seen with elastic-package for Mac amd64, downloaded from https://github.com/elastic/elastic-package/releases/tag/v0.67.0.

It could be that [go-sysinfo](https://github.com/elastic/go-sysinfo) doesn't support getting process information in this OS or architecture.

jsoriano:
This may be the issue: https://github.com/elastic/go-sysinfo/issues/127

PR #1068, opened 2022-12-13
Use bash as shell in shellinit when detection is not implemented

jsoriano:
Elastic Package relies on sysinfo to get information about the running shell. This relies on features that are not implemented for all platforms. This is a frequent issue with some builds for Mac users. Assume bash when detection relies on unimplemented features.

Fixes https://github.com/elastic/elastic-package/issues/1030.

### DIFF
--- a/internal/stack/shellinit.go
+++ b/internal/stack/shellinit.go
@@ -15,6 +15,7 @@ import (
 	"github.com/elastic/go-sysinfo/types"
 
 	"github.com/elastic/elastic-package/internal/environment"
+	"github.com/elastic/elastic-package/internal/logger"
 	"github.com/elastic/elastic-package/internal/profile"
 )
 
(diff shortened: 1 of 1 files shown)


internal/stack/shellinit.go
```diff
@@ -131,6 +132,14 @@ func getShellName(exe string) string {
 func detectShell() (string, error) {
 	ppid := os.Getppid()
 	parentInfo, err := getParentInfo(ppid)
+	if errors.Is(err, types.ErrNotImplemented) {
+		// Sysinfo doesn't implement some features in some platforms.
+		// This mainly affects osx when building without CGO.
+		// Assume bash in that case.
+		// See https://github.com/elastic/elastic-package/issues/1030.
+		logger.Debugf("Failed to determine parent process info while detecting shell, will assume bash")
```
mrodm, line 140: nit. No variable used in the message.
```suggestion
		logger.Debug("Failed to determine parent process info while detecting shell, will assume bash")
```

mrodm (approved):

mrodm (approved):

Issue #1018, opened 2022-10-26
Temporarily import hard-coded list of common fields


jsoriano:
_This issue is a temporary workaround for https://github.com/elastic/package-spec/issues/441 and https://github.com/elastic/elastic-package/issues/1017, **discard it if the other issues are prioritized first**._

For packages importing fields from ECS, import a hard-coded list of fields that are commonly used and can be a source of conflicts if they don't have any mapping.

**Update**: instead of adding specific fields, include dynamic mappings as described in https://github.com/elastic/elastic-package/issues/1018#issuecomment-1340844735. We have to try how to add the dynamic mappings:
* If we add these dynamic mappings using `fields.yml`, this would only be supported by versions of kibana implementing https://github.com/elastic/kibana/pull/137772 (>= 8.4.0).
* If we add the dynamic mappings in the manifest, we will cover more versions of the stack, but we have to be careful of merging correctly with other existing configurations, and it may be difficult to check for documented fields in tests.
    - [x] Add support for `elasticsearch.index_template` key in packages and a new flag to opt in https://github.com/elastic/package-spec/pull/455
    - [ ] Import templates in elastic-package https://github.com/elastic/elastic-package/pull/1073
    - [ ] Add imported templates as definitions to be used in tests https://github.com/elastic/elastic-package/pull/1090

Fields imported this way should be imported in a new well-known fields file that can be discarded later by Fleet if https://github.com/elastic/kibana/issues/144048 is implemented. For example they could be included in a file called `fields/imported_elastic_package.yml`.

Fixes https://github.com/elastic/integrations/issues/4236.

P1llus:
Don't think I forgot anything here, but feel free to double check @andrewkroh @jsoriano .
We also have a few processors adding non-ecs fields, a bit unsure what we want to do with those, but added them to the list.

**Agent:**
```
- external: ecs
  name: agent.name
- external: ecs
  name: agent.type
- external: ecs
  name: agent.id
- external: ecs
  name: agent.ephemeral_id
- external: ecs
  name: agent.version
```

**Host:**
```
- external: ecs
  name: host.architecture
- external: ecs
  name: host.domain
- external: ecs
  name: host.hostname
- external: ecs
  name: host.id
- external: ecs
  name: host.ip
- external: ecs
  name: host.mac
- external: ecs
  name: host.name
- external: ecs
  name: host.os.family
- external: ecs
  name: host.os.kernel
- external: ecs
  name: host.os.name
- external: ecs
[...]

PR #1069, opened 2022-12-14
Add IndexTemplate field under elasticsearch field

mrodm:
According to the package-spec [data_stream manifest definition](https://github.com/elastic/package-spec/blob/40622f8/spec/integration/data_stream/manifest.spec.yml#L196-L230), the structure under `elasticsearch` field should be:

```yaml
elasticsearch:
  index_template:
    mappings:
      ...
    ingest_pipeline:
      name:
```

Currently, DataStreamManifest definition does not include `index_template` field https://github.com/elastic/elastic-package/blob/019ab37/internal/packages/packages.go#L141-L145

This PR adds the `index_template` key to be compliant with the package-spec. Currently, there is no package defining/using this key in the integrations repository.

This addition is also required if at some point `mappings` need to be taken into account too.

Relates https://github.com/elastic/elastic-package/issues/1018

### DIFF
--- a/internal/packages/packages.go
+++ b/internal/packages/packages.go
@@ -139,9 +139,11 @@ type DataStreamManifest struct {
 	Hidden        bool   `config:"hidden" json:"hidden" yaml:"hidden"`
 	Release       string `config:"release" json:"release" yaml:"release"`
 	Elasticsearch *struct {
-		IngestPipeline *struct {
-			Name string `config:"name" json:"name" yaml:"name"`
-		} `config:"ingest_pipeline" json:"ingest_pipeline" yaml:"ingest_pipeline"`
+		IndexTemplate *struct {
+			IngestPipeline *struct {
+				Name string `config:"name" json:"name" yaml:"name"`
+			} `config:"ingest_pipeline" json:"ingest_pipeline" yaml:"ingest_pipeline"`
+		} `config:"index_template" json:"index_template" yaml:"index_template"`
 	} `config:"elasticsearch" json:"elasticsearch" yaml:"elasticsearch"`
 	Streams []struct {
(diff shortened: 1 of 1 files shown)


jsoriano:
> This PR adds the `index_template` key to be compliant with the package-spec. Currently, there is no package defining/using this key in the integrations repository.

This is used by the APM package, that is not in the integrations repository, see here for example: https://github.com/elastic/apm-server/blob/a705bd788cba5f78d82844dab8e3485745844010/apmpackage/apm/data_stream/app_metrics/manifest.yml#L7

mrodm:
> > This PR adds the `index_template` key to be compliant with the package-spec. Currently, there is no package defining/using this key in the integrations repository.
> 
> This is used by the APM package, that is not in the integrations repository, see here for example: https://github.com/elastic/apm-server/blob/a705bd788cba5f78d82844dab8e3485745844010/apmpackage/apm/data_stream/app_metrics/manifest.yml#L7

Right, `apm` package and also the `endpoint` package use this key (https://github.com/elastic/endpoint-package/search?q=index_template&type=code). In both repositories, they are using the structure of the manifest in data_streams as defined in package-spec, so that's correct.

`elastic-package` uses this information to retrieve the pipeline name (`elasticsearch.index_template.ingest_pipeline.name`) in this function:
https://github.com/elastic/elastic-package/blob/main/internal/packages/packages.go#L257
I tried to look for usages of `ingest_pipeline` in packages, but I didn't find any. I would say this change is needed so elastic-package gets that field value according to what package-spec defines. If not, I think it will always use the default value.

@jsoriano do you think this change could cause any issue?

jsoriano (approved):

PR #1072, opened 2022-12-15
Fix expected dataset check in test runners

jsoriano:
Data stream manifests can override the dataset, this is configured in the policy, but then when checking generated documents, it fails with:
```
[0] field "data_stream.dataset" should have value "beat.stats", it has "beats.stack_monitoring.stats"
[1] field "event.dataset" should have value "beat.stats", it has "beats.stack_monitoring.stats"
```
Seen in https://github.com/elastic/integrations/pull/4708.

Also fix issue that was silently skipping static tests for input packages, because they were looking for the sample event in the wrong path.

### DIFF
--- a/test/packages/other/with_dataset/manifest.yml
+++ b/test/packages/other/with_dataset/manifest.yml
@@ -1,15 +1,14 @@
-format_version: 1.0.0
+format_version: 2.2.0
 name: with_dataset
 title: With dataset test
-version: 0.0.1
+version: 0.0.2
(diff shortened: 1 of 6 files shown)


test/packages/other/with_dataset/manifest.yml
```diff
@@ -1,15 +1,14 @@
-format_version: 1.0.0
+format_version: 2.2.0
```
jsoriano, line 1: Dataset checks are only executed for since 2.0.0.

PR #1073, opened 2022-12-15
Add common dynamic mappings and properties from ECS automatically

mrodm:
Relates https://github.com/elastic/elastic-package/issues/1018

This PR adds support to include some dynamic templates and properties from ECS that are commonly used and can be a source of conflicts if they don't have any mapping.

These mappings are added if packages enable the flag `import_common_dynamic_mappings` in `_dev/build/build.yml`. This flag is introduced in https://github.com/elastic/package-spec/pull/455 , and it would be available in a future version 2.3.0 of the spec.

**Option 1:**
All the dynamic templates names are prefixed with `_embedded_ecs`, so they can be distinguished easily by Fleet in case it is needed. For instance:

```yaml
elasticsearch:
  index_template:
    mappings:
      dynamic_templates:
        - _embedded_ecs.ecs_timestamp:
            mapping:
              ignore_malformed: false
              type: date
            path_match: '@timestamp'
        - _embedded_ecs.data_stream_to_constant:
            mapping:
              type: constant_keyword
            path_match: data_stream.*
``` 

**Option 2 (Discarded for now)**: 
All the mappings added by this flag are added into the `_meta` field. Thereby, they are marked in case it is needed to know which ones were automatically added. Properties and dynamic templates are added to `ecs_properties_added` and `ecs_dynamic_templates_added` keys, respectively.
Example:

```json
    "mappings": {
      "_meta": {
        "ecs_properties_added": [
          "@timestamp"
        ],
        "package": {
          "name": "elastic_package_registry"
        },
        "managed_by": "fleet",
        "managed": true,
        "ecs_dynamic_templates_added": [
          "data_stream_to_constant",
          "resolved_ip_to_ip",
          "forwarded_ip_to_ip",
          "ip_to_ip",
          "port_to_long",
          "thread_id_to_long",
          "status_code_to_long",
          "line_to_long",
          "priority_to_long",
          "code_to_long",
          "code_to_long",
          "bytes_to_long",
          "packets_to_long",
          "public_key_exponent_to_long",
          "severity_to_long",
          "duration_to_long",
          "pid_to_long",
          "uptime_to_long",
          "sequence_to_long",
          "entropy_to_long",
          "size_to_long",
          "entrypoint_to_long",
          "ttl_to_long",
          "major_to_long",
          "minor_to_long",
          "as_number_to_long",
          "pgid_to_long",
          "exit_code_to_long",
          "chi_to_long",
          "args_count_to_long",
          "virtual_address_to_long",
          "io_text_to_wildcard",
          "strings_to_wildcard",
          "path_to_wildcard",
          "message_id_to_wildcard",
          "command_line_to_multifield",
          "error_stack_trace_to_multifield",
          "http_content_to_multifield",
          "url_full_to_multifield",
          "url_original_to_multifield",
          "user_agent_original_to_multifield",
          "error_message_to_match_only",
          "message_match_only_text",
          "agent_name_to_keyword",
          "service_name_to_keyword",
          "sections_name_to_keyword",
          "resource_name_to_keyword",
          "observer_name_to_keyword",
          "question_name_to_keyword",
          "group_name_to_keyword",
          "geo_name_to_keyword",
          "host_name_to_keyword",
          "severity_name_to_keyword",
          "title_to_multifield",
          "executable_to_multifield",
          "file_path_to_multifield",
          "file_target_path_to_multifield",
          "name_to_multifield",
          "full_name_to_multifield",
          "os_full_to_multifield",
          "working_directory_to_multifield",
          "timestamp_to_date",
          "delivery_timestamp_to_date",
          "not_after_to_date",
          "not_before_to_date",
          "accessed_to_date",
          "origination_timestamp_to_date",
          "created_to_date",
          "installed_to_date",
          "creation_date_to_date",
          "ctime_to_date",
          "mtime_to_date",
          "ingested_to_date",
          "start_to_date",
          "end_to_date",
          "score_base_to_float",
          "score_temporal_to_float",
          "score_to_float",
          "score_norm_to_float",
          "usage_to_float",
          "location_to_geo_point",
          "same_as_process_to_boolean",
          "established_to_boolean",
          "resumed_to_boolean",
          "max_bytes_per_process_exceeded_to_boolean",
          "interactive_to_boolean",
          "exists_to_boolean",
          "trusted_to_boolean",
          "valid_to_boolean",
          "go_stripped_to_boolean",
          "coldstart_to_boolean",
          "exports_to_flattened",
          "structured_data_to_flattened",
          "imports_to_flattened",
          "attachments_to_nested",
          "segments_to_nested",
          "elf_sections_to_nested",
          "pe_sections_to_nested",
          "macho_sections_to_nested",
          "trigger_to_nested"
        ]
      },   
```


internal/builder/dynamic_mappings.go
```diff
@@ -0,0 +1,256 @@
...
+	bm, ok, err := buildmanifest.ReadBuildManifest(packageRoot)
+	if err != nil {
+		return errors.Wrap(err, "can't read build manifest")
+	}
+	if !ok {
+		logger.Debugf("Build manifest hasn't been defined for the package")
+		return nil
+	}
+	if !bm.ImportCommonDynamicMappings() {
+		logger.Debugf("Package doesn't have to import common dynamic mappings")
```
jsoriano, line 42: I would log on the opposite, when the package is importing mappings :slightly_smiling_face:

internal/builder/dynamic_mappings.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil, err
+	}
+
+	contents, err = formatResult(&doc)
+	if err != nil {
+		logger.Errorf("Error formatting %s", err)
+		return nil, err
+	}
+
+	logger.Debugf("New Contents manifest:\n%s", contents)
```
jsoriano, line 116: Nit. This may be too verbose, this content can be also seen by looking at the built package, right?

jsoriano (commented):
Looks good, but added a couple of comments that may help to simplify this feature.

Ideally we will also need to take these mappings into account when validating field definitions on tests. Tests will start failing if developers remove their ECS definitions after enabling this import, although the final template will be correct.
But let's think about this in a followup. Current workaround would be to keep definitions required by tests.

internal/builder/dynamic_mappings.go
mrodm: mmm this was just intended for debugging purposes. Just a leftover, I'll remove it.

internal/builder/_static/ecs_mappings.json
```diff
@@ -0,0 +1,886 @@
+{
+  "mappings": {
+    "properties": {
+      "@timestamp": {
+        "type": "date",
+        "ignore_malformed": false
+      }
```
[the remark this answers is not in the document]
mrodm: I used the same json that was built here https://gist.github.com/P1llus/e0de7b3a7824a41a29660e253c6cce6b

I think it could be changed to be a dynamic template as you mentioned. @P1llus do you think in any inconvenient if it is changed ?

I would say there is no problem and it would simplify the code to just manage dynamic templates.

internal/builder/dynamic_mappings.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = addEcsMappings(&doc, ecsMappings)
+	if err != nil {
+		return nil, err
+	}
+
+	err = addEcsMappingsListMeta(&doc, ecsMappings)
```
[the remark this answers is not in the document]
mrodm: It looks better this approach than adding new fields into `_meta`. And it will ensure to just take into account the right templates/mappings (thanks to the prefix) to take the desired actions (e.g. drop them).

I will dig into this approach, so the current lists built and added into `_meta` could be removed.

go.mod
```diff
@@ -160,3 +160,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/elastic/package-spec/v2 => github.com/mrodm/package-spec/v2 v2.0.0-20221220145202-70d4563fbf33
```
mrodm, line 164: to be removed before merging

internal/builder/dynamic_mappings.go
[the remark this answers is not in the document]
mrodm: Pushed the change to add a prefix to all the dynamic templates. In case of properties, I think this cannot be done. However, if properties can be left apart and just use dynamic templates, there should not be needed to keep any list in the `_meta` field.

internal/builder/_static/ecs_mappings.json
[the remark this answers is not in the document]
mrodm: For now I've moved that property to a dynamic template as suggested by @jsoriano.

If there is any inconvenient because of that, it could be re-introduced.

Currently, it has been followed the approach to add a prefix (`_embedded_ecs.*`) to all the names of the dynamic templates in the static file, so they can be identified easily. If properties are needed too, it should be added another way (using `_meta` field and lists as it was in a previous change?).

internal/builder/dynamic_mappings.go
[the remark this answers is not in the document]
jsoriano: > In case of properties, I think this cannot be done.

Yeah, I don't think so. But yes, let's try to add only dynamic mappings with this feature.

internal/builder/packages.go
```diff
@@ -189,6 +195,7 @@ func BuildPackage(options BuildOptions) (string, error) {
 		return destinationDir, nil
 	}
 
+	logger.Infof("Validating package %s", destinationDir)
```
jsoriano, line 198: Nit. Can you add an equivalent message for the zip validation? Should this be at the debug level?

internal/builder/dynamic_mappings.go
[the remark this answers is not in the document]
jsoriano: ccing @hop-dev about this change over what we talked.

jsoriano (approved):

internal/builder/packages.go
mrodm: Added and changed to Debug :+1:

jsoriano (approved):

internal/builder/_static/ecs_mappings.json
```diff
@@ -0,0 +1,889 @@
+{
```
hop-dev, line 1: Should we record which version of ECS this is? And is it useful to know which version of ECS was applied to a given package? I am not sure how often ECS is updated.

internal/builder/dynamic_mappings.go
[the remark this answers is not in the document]
hop-dev: I like the new approach 👍

hop-dev (approved):
I added one question but looks good to me, I like the new style of prefixing the name

internal/builder/_static/ecs_mappings.json
mrodm: Currently, this set of dynamic templates are not related to a specific version of ECS. I would say it would not be needed to keep track of that.

jsoriano (approved):

mrodm:
I've checked building and installing a test package in different versions of Kibana. In kibana 8.6.0-SNASPSHOT there was no issue. But in versions < 8.6.0 , the dot in the dynamic template names is causing this error:

> Error installing elastic_package_registry 0.0.7: mapper_parsing_exception: [mapper_parsing_exception] Reason: Failed to parse mapping: Illegal dynamic template parameter: [ecs_timestamp]

I've updated how the name of each dynamic template is generated to replace the dot "." by a dash "-" @jsoriano @hop-dev 

Thereby, the dynamic template generated would be like:

```yaml
elasticsearch:
  index_template:
    mappings:
      dynamic_templates:
        - _embedded_ecs-ecs_timestamp:
            mapping:
              ignore_malformed: false
              type: date
            path_match: '@timestamp'
        - _embedded_ecs-data_stream_to_constant:
            mapping:
              type: constant_keyword
            path_match: data_stream.*
        - _embedded_ecs-resolved_ip_to_ip:
            mapping:
              type: ip
            match: resolved_ip
```

Tested with the following Kibana versions 7.17.8-SNAPSHOT, 8.0.0-SNAPSHOT, 8.5.0-SNAPSHOT and 8.6.0-SNAPSHOT

PR #1076, opened 2022-12-17
Enable APM tracing by default for stack up

joshdover:
Turns on tracing by default for all components

image

TODO:
- [ ] Figure out how to enable tracing for managed Agent
	- Setting `agent.monitoring.tracing` and `agent.monitoring.apm` settings in `elastic-agent.yml` didn't work
- [ ] Cleanup Kibana configs, prefer env vars over yaml if possible
- [ ] Log correlation for Fleet Server and Agent logs (which are already collected)
- [ ] Setup certs & https for APM for RUM support

Things that could be improved in future PRs:
- Agent and Fleet Server should allow enabling tracing via env vars, rather than depending on yaml or policy configuration
- Trace context propagation from Agent to Fleet Server
- Log collection and correlation for Kibana, Elasticsearch, and Package Registry

### DIFF
--- a/internal/profile/_static/docker-compose-stack.yml
+++ b/internal/profile/_static/docker-compose-stack.yml
@@ -37,6 +37,7 @@ services:
     environment:
       # Is there a better way to add certificates to Kibana/Fleet?
       - "NODE_EXTRA_CA_CERTS=/usr/share/kibana/config/certs/ca-cert.pem"
+      - "ELASTIC_APM_ENVIRONMENT=dev"
     volumes:
       - "./kibana.config.${STACK_VERSION_VARIANT}.yml:/usr/share/kibana/config/kibana.yml"
       - "../certs/kibana:/usr/share/kibana/config/certs"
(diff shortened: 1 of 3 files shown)


ruflin:
++ on having apm enabled in elastic-package by default. I wonder if it would be cleaned to run a separate elastic-agent with the apm-integration.

joshdover:
Agree on using a separate agent and making it optional via the existing `--services` parameter. I think this separate agent should also run the Kibana and Elasticsearch integrations with shared volumes to capture logs from ES and Kibana. Any opposition to that?

jsoriano:
> via the existing `--services` parameter.

We don't have any other optional service now, but this could be an option, yes.

> I think this separate agent should also run the Kibana and Elasticsearch integrations with shared volumes to capture logs from ES and Kibana. Any opposition to that?

Sounds good, if this is also optional, and not used to test the Kibana and Elasticsearch integrations.

florianl:
`elastic/elastic-package` comes in handy for full system integration and for local development. Having an option to easily enable APM would really come handy.

graphaelli:
superseded by #1598

PR #1077, opened 2022-12-19
Bulk update of several dependencies

jsoriano:
Bulk update dependencies that haven't been automatically updated.


mrodm (approved):
:+1: 

Maybe golang.org/x/oauth2 package has not been updated due to this https://github.com/dependabot/dependabot-core/issues/3017 ?

jsoriano:
> Maybe golang.org/x/oauth2 package has not been updated due to this [dependabot/dependabot-core#3017](https://github.com/dependabot/dependabot-core/issues/3017) ?

Looks like that, yes.

Issue #1079, opened 2022-12-19
Panic on shellinit detection


jsoriano:
Running shell detection on mac fails in v0.70.1. Could be related to https://github.com/elastic/elastic-package/pull/1068.

```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/elastic/go-sysinfo/providers/darwin.kern_procargs(0xc00022f720?, 0xc0000b8d20)
	/var/lib/jenkins/workspace/_manager_elastic-package_v0.70.1/pkg/mod/github.com/elastic/go-sysinfo@v1.9.0/providers/darwin/process_darwin.go:207 +0x58f
github.com/elastic/go-sysinfo/providers/darwin.(*process).Info(0xc0000b8d20)
	/var/lib/jenkins/workspace/_manager_elastic-package_v0.70.1/pkg/mod/github.com/elastic/go-sysinfo@v1.9.0/providers/darwin/process_darwin.go:104 +0x12d
github.com/elastic/elastic-package/internal/stack.getParentInfo(0xc000243b00?)
	/var/lib/jenkins/workspace/_manager_elastic-package_v0.70.1/src/github.com/elastic/elastic-package/internal/stack/shellinit.go:165 +0x15c
github.com/elastic/elastic-package/internal/stack.detectShell()
	/var/lib/jenkins/workspace/_manager_elastic-package_v0.70.1/src/github.com/elastic/elastic-package/internal/stack/shellinit.go:134 +0x58
github.com/elastic/elastic-package/internal/stack.init.0()
	/var/lib/jenkins/workspace/_manager_elastic-package_v0.70.1/src/github.com/elastic/elastic-package/internal/stack/shellinit.go:35 +0x17
```

Thanks @ajosh0504 for reporting!

PR #1080, opened 2022-12-19
Use gopsutil instead of go-sysinfo for shell detection

jsoriano:
go-sysinfo has been causing issues with shell detection, see https://github.com/elastic/elastic-package/issues/1079 or https://github.com/elastic/elastic-package/issues/1030.

Use gopsutil instead, that seems to have broader platform support.

In all cases, fallback to bash if shell detection fails.


internal/stack/shellinit.go
```diff
@@ -11,8 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/elastic/go-sysinfo"
-	"github.com/elastic/go-sysinfo/types"
+	"github.com/shirou/gopsutil/process"
```
mrodm, line 14: To remove the `+incompatible` version in go.mod of this package, this import could be changed to:

```suggestion
	"github.com/shirou/gopsutil/v3/process"
```
I think it comes from because there is no go.mod in [v3.21.11 version](https://github.com/shirou/gopsutil/tree/v3.21.11)

Changin this, that suffix would be removed in go.mod (and there are also added a couple of indirect deps):
```diff
        github.com/olekukonko/tablewriter v0.0.5
        github.com/pkg/errors v0.9.1
        github.com/pmezard/go-difflib v1.0.0
-       github.com/shirou/gopsutil v3.21.11+incompatible
+       github.com/shirou/gopsutil/v3 v3.22.11
        github.com/spf13/cobra v1.6.1
        github.com/stretchr/testify v1.8.1
        golang.org/x/oauth2 v0.3.0
```
Just did a quick test for bash and it looks like it worked too:
```bash
 $ make build
go build -ldflags "-X github.com/elastic/elastic-package/internal/version.CommitHash=`git describe --always --long --dirty` -X github.com/elastic/elastic-package/internal/version.BuildTime=`date +%s` -X github.com/elastic/elastic-package/internal/version.Tag=`(git describe --exact-match --tags 2>/dev/null || echo '') | tr -d '\n'`" -o elastic-package

 $ ./elastic-package stack shellinit
Detected shell: bash
export ELASTIC_PACKAGE_ELASTICSEARCH_HOST=https://127.0.0.1:9200
export ELASTIC_PACKAGE_ELASTICSEARCH_USERNAME=elastic
export ELASTIC_PACKAGE_ELASTICSEARCH_PASSWORD=changeme
export ELASTIC_PACKAGE_KIBANA_HOST=https://127.0.0.1:5601
export ELASTIC_PACKAGE_CA_CERT=/home/mariorodriguez/.elastic-package/profiles/default/certs/ca-cert.pem
```

mrodm (commented):
LGTM!
Agreed to use `bash` as the default shell :+1: 

Just a comment to update the import to remove the incompatible suffix in the package.

internal/stack/shellinit.go
jsoriano: Ah, good one, thanks!

mrodm (approved):

PR #1082, opened 2022-12-21
Add CA fingerprint to agents configuration in 8.x versions

jsoriano:
Try again something like https://github.com/elastic/elastic-package/pull/789/commits/bb84977ac4e3710587a26cbf3eb752483b5b65b9 for 8.x stacks.

Adding this setting was evaluated and discarded in https://github.com/elastic/elastic-package/pull/789 because it didn't help configuring SSL in agents started by `elastic-package stack up`, as it was not supported in all versions of the stack supported by `elastic-package`. More about this can be read in https://github.com/elastic/elastic-package/pull/789.

The option implemented then and being used now is to install the certificate in the Elastic Agent containers, in `/etc/ssl`. This works well for Agents managed by `elastic-package`, but is not enough for external Elastic Agents, where installing the CA in the system may be more cumbersome.

Try to add the fingerprint back, as it would be helpful in some testing scenarios using external Elastic Agents.

Fix https://github.com/elastic/elastic-package/issues/1083.

### DIFF
--- a/internal/certs/certs.go
+++ b/internal/certs/certs.go
@@ -10,8 +10,10 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"encoding/pem"
 	"fmt"
 	"io"
@@ -19,6 +21,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/elastic/elastic-package/internal/common"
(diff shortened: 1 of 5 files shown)


jsoriano:
Ah no, this setting cannot be set here, it would need to be set in the policies.
Not sure what we can do then from the elastic-package side for policies and agents configured manually.

jsoriano:
> Ah no, this setting cannot be set here, it would need to be set in the policies.

As @agithomas mentioned, this setting goes not in the policies, but in the output configuration. PR updated.

agithomas:
Yes, it works. Tested with 8.5.1 stack and agent version

image

agithomas (commented):
Works well. 

Recommend to  add the documentation in the elastic-package about how to add the agent 

example:
`sudo ./elastic-agent install --certificate-authorities=$ELASTIC_PACKAGE_CA_CERT --fleet-server-es-ca=$ELASTIC_PACKAGE_CA_CERT --url=https://fleet-server:8220 --enrollment-token=`

mrodm (approved):

PR #1090, opened 2023-01-10
Add ECS mappings during test runner executions

mrodm:
Closes https://github.com/elastic/elastic-package/issues/1018

Follow-up of #1073 

This PR adds all the [ECS field definitions](https://github.com/elastic/ecs/tree/main/generated/ecs) available in test executions to check whether or not a field in the sample event is defined. None of these field definitions are added in the built package.

The ECS version used for that regard is the version set at `_dev/build/build.yml` in the package itself.

ECS definitions are used instead of trying to convert the [`ecs_mappings.json`](https://github.com/elastic/elastic-package/pull/1073/files#diff-97a49a89f8f463695e0e915b859ffd0e4ee86719365cdee666dc968059781ded) to fields in elastic-package mainly because:
- those dynamic templates in `ecs_mappings.json`  use `match`, `patch_unmatch`, that do not have a relation 1 to 1 with the field definitions as managed in elastic-package.
- ECS schema explicitly defines all the field definitions.

These mappings are going to be used just when that flag is enabled.

A new test package has been added with the `import_mappings` flag is enabled, and for that it requires package-spec 2.3.0 to be updated in elastic-package.

This PR is based on the branch of https://github.com/elastic/elastic-package/pull/1073. To review just the commits from this PR, it should be checked from commit 26c8cfa (inclusive).

Requisites:
- [x] package-spec 2.3.0 published and updated in elastic-package (required by the test package)


internal/builder/packages.go
```diff
@@ -180,7 +180,7 @@ func BuildPackage(options BuildOptions) (string, error) {
 		return "", errors.Wrap(err, "resolving external fields failed")
 	}
 
-	logger.Debug("Add dynamic mappings if needed")
+	logger.Debug("Add dynamic mappings if needed (technical preview)")
```
mrodm, line 183: Would it be interesting to add this ? Maybe changing to INFO instead ?

internal/fields/validate.go
```diff
@@ -287,6 +309,21 @@ func (v *Validator) validateScalarElement(key string, val interface{}, doc commo
 	if definition == nil && skipValidationForField(key) {
 		return nil // generic field, let's skip validation for now
 	}
+
+	if definition == nil && !v.disabledImportAllECSSchema {
+		// import all fields from external schema
+		ecsSchema, err := v.FieldDependencyManager.ImportAllFields(defaultExternal)
+		if err != nil {
+			return err
+		}
```
jsoriano, line 318: This will import ECS for each field, right?
Would it be possible to combine the `v.Schema` and the result of `v.FieldDependencyManager.ImportAllFields(defaultExternal)` in `CreateValidatorForDirectory(...)`? So we would only import ECS once and this logic here wouldn't be needed.

jsoriano (commented):
Nice to see it working, added a possible improvement.

internal/builder/packages.go
jsoriano: Sounds good to add this. If changed to info it should probably appear only if the feature is used.

internal/fields/validate.go
mrodm: I was worried in case all these ECS field definitions were added also for instance in the exported field documentation.

I've been checking again this, and when it is used to [get the exported fields](https://github.com/elastic/elastic-package/blob/main/internal/docs/exported_fields.go#L28):

```golang
	validator, err := fields.CreateValidatorForDirectory(fieldsParentDir)
	if err != nil {
		return "", errors.Wrapf(err, "can't create fields validator instance (path: %s)", fieldsParentDir)
	}
```

no spec version is used as a ValidatorOption parameter, so the default value (0.0.0) is taken. That makes sure that all these fields are not used at all in this stage since it must be at least 2.3.0.

I'll do the changes to move this logic directly in `CreateValidatorForDirectory()`

Thanks!

internal/builder/packages.go
mrodm: Ok! I'll move this message as an INFO to be shown just when the feature is enabled/used.

internal/docs/exported_fields.go
```diff
@@ -25,7 +25,8 @@ type fieldsTableRecord struct {
 var escaper = strings.NewReplacer("*", "\\*", "{", "\\{", "}", "\\}", "<", "\\<", ">", "\\>")
 
 func renderExportedFields(fieldsParentDir string) (string, error) {
-	validator, err := fields.CreateValidatorForDirectory(fieldsParentDir)
+	validator, err := fields.CreateValidatorForDirectory(fieldsParentDir,
+		fields.WithDisabledImportAllECSSChema())
```
jsoriano, line 29: We use `CreateValidatorForDirectory` in more places, shouldn't the import be disabled there too? Unless mappings are included in the package.

Could you add a test package that misses an ECS field definition, doesn't import it and fails validation?

jsoriano (commented):
I am wondering now if we are correctly deciding when to import the schema for tests. It should be only imported for packages that include the ECS dynamic mappings.

internal/fields/validate.go
```diff
@@ -48,6 +53,8 @@ type Validator struct {
 
 	enabledAllowedIPCheck bool
 	allowedCIDRs          []*net.IPNet
+
+	disabledImportAllECSSchema bool
```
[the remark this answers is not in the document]
mrodm: That's true, I'll rename it and do the required changes.

internal/fields/validate.go
```diff
@@ -178,6 +178,17 @@ func CreateValidatorForDirectory(fieldsParentDir string, opts ...ValidatorOption
...
+	if v.disabledImportAllECSSchema {
+		return v, nil
+	}
+
+	ecsSchema, err := v.FieldDependencyManager.ImportAllFields(defaultExternal)
+	if err != nil {
+		return nil, err
+	}
+	logger.Debug("Adding to the schema the ECS field definitions")
+	v.Schema = append(v.Schema, ecsSchema...)
```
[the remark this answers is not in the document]
mrodm: I had some issues when creating this type of tests. In tests, it is not found the package root and therefore the dependency management is disabled because of that.

I could add another datastream in the test package where there is a field defined in both in ECS and the package fields.

internal/docs/exported_fields.go
mrodm: > Could you add a test package that misses an ECS field definition, doesn't import it and fails validation?

The test packages that we have currently are expected to be run/validated successfully. Because of that I've added a test `TestValidate_WithDisabledImportAllECSSchema` that checks that. Would that be ok ?

jsoriano: > The test packages that we have currently are expected to be run/validated successfully.

Ah, you are right. We should have a way to add negative tests too, let's leave this for a future change. 

> Because of that I've added a test `TestValidate_WithDisabledImportAllECSSchema` that checks that. Would that be ok ?

Yes, this will make it by now.

jsoriano (approved):

PR #1099, opened 2023-01-18
Move package-spec check version for dynamic mappings import

mrodm:
This PR moves down the check performed to import dynamic mappings related to the package-spec version.
This check should just be applied in case it is defined `import_mappings` in the package, but they are not using at least `format_version: 2.3.0`

Before the following debug message was showed when building a package:
```
2023/01/18 10:34:42 DEBUG Required spec version >= 2.3.0 to import ECS mappings
```
but the package had no enable the import mappings feature.


jsoriano (approved):
:+1:

PR #1108, opened 2023-01-23
[Fleet] Enable experimental toggles in Fleet UI from Kibana config

kpollich:
Enable the experimental data stream toggles (synthetic source, TSDB, doc-values-only) in Fleet's policy editor by toggling a feature flag in the Kibana config used for launching the local Kibana instance.

Ref https://github.com/elastic/kibana/pull/148418
Ref https://github.com/elastic/kibana/issues/148317

Not sure if this should also be turned on in the other Kibana config files:
- `kibana_config_8x.yml`
- `kibana_config_80.yml`

If they should be edited too, let me know and I'll happily make the changes.

### DIFF
--- a/internal/profile/_static/kibana_config_8x.yml
+++ b/internal/profile/_static/kibana_config_8x.yml
@@ -5,7 +5,7 @@ server.ssl.certificate: "/usr/share/kibana/config/certs/cert.pem"
 server.ssl.key: "/usr/share/kibana/config/certs/key.pem"
 server.ssl.certificateAuthorities: ["/usr/share/kibana/config/certs/ca-cert.pem"]
 
-elasticsearch.hosts: [ "https://elasticsearch:9200" ]
+elasticsearch.hosts: ["https://elasticsearch:9200"]
 elasticsearch.ssl.certificateAuthorities: "/usr/share/kibana/config/certs/ca-cert.pem"
 elasticsearch.serviceAccountToken: "AAEAAWVsYXN0aWMva2liYW5hL2VsYXN0aWMtcGFja2FnZS1raWJhbmEtdG9rZW46b2x4b051SWNRa0tYMHdXazdLWmFBdw"
 
@@ -15,6 +15,7 @@ xpack.fleet.registryUrl: "https://package-registry:8080"
 xpack.fleet.agents.enabled: true
 xpack.fleet.agents.elasticsearch.hosts: ["https://elasticsearch:9200"]
 xpack.fleet.agents.fleet_server.hosts: ["https://fleet-server:8220"]
+xpack.fleet.enableExperimental: ["experimentalDataStreamSettings"] # Enable experimental toggles in Fleet UI
 
 xpack.encryptedSavedObjects.encryptionKey: "12345678901234567890123456789012"
 
@@ -55,7 +56,7 @@ xpack.fleet.outputs:
   - id: fleet-default-output
     name: default
     type: elasticsearch
-    hosts: [ https://elasticsearch:9200 ]
+    hosts: [https://elasticsearch:9200]
     ca_trusted_fingerprint: "${ELASTIC_PACKAGE_CA_TRUSTED_FINGERPRINT}"
     is_default: true
     is_default_monitoring: true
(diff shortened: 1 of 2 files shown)


internal/profile/_static/kibana_config_default.yml
```diff
@@ -17,5 +17,6 @@ xpack.fleet.registryUrl: "https://package-registry:8080"
 xpack.fleet.agents.enabled: true
 xpack.fleet.agents.elasticsearch.host: "https://elasticsearch:9200"
 xpack.fleet.agents.fleet_server.hosts: ["https://fleet-server:8220"]
+xpack.fleet.enableExperimental: ["experimentalDataStreamSettings"] # Enable experimental toggles in Fleet UI
```
jsoriano, line 20: This should be probably added to `kibana_config_8x.yml` only. `kibana_config_80.yml` is used for 8.0 version, and `kibana_config_default.yml` is used for older versions.

jsoriano, line 20: Does it enable the features or only the toggles?

kpollich: Just the toggles are enabled - the features themselves still require user input or package spec opt-in by the integration.

internal/profile/_static/kibana_config_8x.yml
```diff
@@ -55,7 +56,7 @@ xpack.fleet.outputs:
   - id: fleet-default-output
     name: default
     type: elasticsearch
-    hosts: [ https://elasticsearch:9200 ]
+    hosts: [https://elasticsearch:9200]
```
ruflin, line 59:
```suggestion
    hosts: ["https://elasticsearch:9200"]
```

Mainly to be consistent.

ruflin (commented):
LGTM

jsoriano:
Failures in CI seem related to this change, should this setting be applied only for versions >= 8.7? (We don't have an easy way to do it).
```
[36mkibana_1                     |[0m [2023-01-23T16:12:56.735+00:00][INFO ][node] Kibana process configured with roles: [background_tasks, ui]
[36mkibana_1                     |[0m [2023-01-23T16:13:10.753+00:00][FATAL][root] Error: [config validation of [xpack.fleet].enableExperimental]: [experimentalDataStreamSettings] is not allowed. Allowed values are: createPackagePolicyMultiPageLayout, packageVerification, showDevtoolsRequest, diagnosticFileUploadEnabled
```

kpollich:
> Failures in CI seem related to this change, should this setting be applied only for versions >= 8.7? (We don't have an easy way to do it).

Yes, this option is only available in 8.7+. Due to the nature of Kibana's config validator, the setting will result in an invalid config for older versions.

ruflin:
@jsoriano @kpollich For our current testing, is there a manual way to overwrite it by setting some environment variables or similar?

jsoriano:
> For our current testing, is there a manual way to overwrite it by setting some environment variables or similar?

I don't think so.

Can these flags be changed on runtime through the API?

joshdover:
Kibana supports setting configs via env vars in the Docker images: https://www.elastic.co/guide/en/kibana/current/docker.html#environment-variable-config

Probably something like `XPACK_FLEET_ENABLEEXPERIMENTAL=true` would work.

I think this isn't automatic though, so there might be some Docker entrypoint script that needs to be updated for the image to support this. Kibana operations team would be the one to ask.

ruflin:
As a manual workaround what I did for now is add `xpack.fleet.enableExperimental: ["experimentalDataStreamSettings"]` as an entry into `.elastic-package/profiles/default/stack/kibana.config.8x.yml` and run `elastic-package stack up --version=8.7.0-SNAPSHOT -v` and it works. This is a manual way to make sure we don't get blocked for testing.

ruflin:
> Can these flags be changed on runtime through the API?

@jsoriano Unfortunately no.

ruflin:
Can we get this moving? Most of these flags were introduced so we can test it with packages. Not having it available in elastic-package is problematic as this is where most of us tests packages.

jsoriano:
> Can we get this moving? Most of these flags were introduced so we can test it with packages. Not having it available in elastic-package is problematic as this is where most of us tests packages.

Let me take a quick look to adding a new config file for Kibana 8.7.

jsoriano:
https://github.com/elastic/elastic-package/pull/1128

jsoriano:
Change merged as part of https://github.com/elastic/elastic-package/pull/1128, closing this one.

PR #1112, opened 2023-01-25
all: replace github.com/pkg/errors with errors

florianl:
Signed-off-by: Florian Lehner <florian.lehner@elastic.co>

This PR replaces the no longer maintained dependency `github.com/pkg/errors` with its standard library equivalent. 

The standard library package `errors` was introduced with Go 1.13.

### DIFF
--- a/cmd/benchmark.go
+++ b/cmd/benchmark.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"github.com/elastic/elastic-package/internal/benchrunner"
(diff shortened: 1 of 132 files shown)


jsoriano (approved):
Thanks!

cmd/benchmark.go
```diff
@@ -106,10 +105,10 @@ func benchTypeCommandActionFactory(runner benchrunner.BenchRunner) cobraext.Comm
 
 		packageRootPath, found, err := packages.FindPackageRoot()
 		if !found {
-			return errors.New("package root not found")
+			return fmt.Errorf("package root not found")
 		}
 		if err != nil {
-			return errors.Wrap(err, "locating package root failed")
+			return fmt.Errorf("locating package root failed: %s", err)
```
efd6, line 111: This is not right. If a caller uses `errors.Is` or `errors.As` on this error it will always fail since `err` is no longer wrapped in the returned value. The verb should be %w to replace `errors.Wrap(errr, ...`.

I have a set of gofmt rewrites that will do the changes mechanically if that would be helpful.

efd6 (changes requested):
The changes here from `errors.Wrap(err, ...` to `fmt.Errorf("...%s...", err)` change the semantics of the error values.

cmd/benchmark.go
jsoriano: Good catch :+1:

florianl: > I have a set of gofmt rewrites that will do the changes mechanically if that would be helpful.

happy to see the contribution :+1:

efd6: The details are in the commit message of elastic/beats@4ab90f9b3339c6f8a5d198ad469305dcc7bbad53.

PR #1118, opened 2023-01-26
Update default stack version to 8.6.1

jsoriano:
Add also some options to try to reduce flakiness in `elastic-package stack up`:
* retry limits have been replaced with start periods.
* fleet-server and elastic-agent healthcheck retries have been increased.
* if docker-compose up fails and elastic-agent container is the one failing, retry once, this may happen if it tries to enroll while fleet-server is being restarted or reconfigured.


internal/profile/_static/docker-compose-stack.yml
```diff
@@ -124,7 +125,8 @@ services:
         condition: service_healthy
     healthcheck:
       test: "elastic-agent status"
-      retries: 180
+      start_period: 60s
+      retries: 360
```
mrodm, line 129: Even using these settings, there are some cases where elastic_agent_is_ready container still has unhealthy status :disappointed: 
`ERROR: for elastic-agent_is_ready  Container "066fd24a8fc6" is unhealthy.`

Not sure if at least those will help us to make it a bit better 

I've not been able to reproduce it locally to check what it is the output of the healthchecks using `docker inspect`

mrodm (approved):

mrodm (approved):

internal/stack/boot.go
```diff
@@ -60,12 +60,41 @@ func BootUp(options Options) error {
...
+		// At least starting on 8.6.0, fleet-server may be reconfigured or
+		// restarted after being healthy. If elastic-agent tries to enroll at
+		// this moment, it fails inmediately, stopping and making `docker-compose up`
+		// to fail too.
+		// As a workaround, try to give another chance to docker-compose if only
+		// elastic-agent failed.
+		if onlyElasticAgentFailed() {
+			fmt.Println("Elastic Agent failed to start, trying again.")
+			err = dockerComposeUp(options)
+		}
```
mrodm, line 72: :+1:

mrodm (approved):
Great! I see that the latest builds run in Jenkins have been successful :muscle: 

Thanks!

Issue #1110, opened 2023-01-25
Add a "generate" command for producing integration data


aspacca:
In the scope of the discussion at https://github.com/elastic/elastic-package/issues/984, we'd like to add a `generate` command to `elastic-package`

The command to run will be something like:
```
elastic-package data-generate --package=bar --dataset=foo
```

Implementation details:
- The `generate` command will consume the [elastic-integration-corpus-generator-tool package](https://github.com/elastic/elastic-integration-corpus-generator-tool)
-  For the first iteration the required artifacts (field yaml definitions, corpus-generator-tool configs and templates) will be fetched from the [elastic-integration-corpus-generator-tool repo assets folder](https://github.com/elastic/elastic-integration-corpus-generator-tool/tree/main/assets/templates). Future iterations will define the proper entries in [`package-spec`](https://github.com/elastic/package-spec) and the artifacts (or their equivalent) will be available directly from the package definitions
- For the first iteration the output will be a file in the current directly. Future iterations will add support for an `--output` parameter

aspacca:
Closed by #1122

PR #1122, opened 2023-01-30
Generate comand

aspacca:
Addresses #1110


internal/integration-corpus-generator/assets.go
```diff
@@ -0,0 +1,119 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package integration_corpus_generator
```
aspacca, line 5: this package may be better named as `genlib`: current implementation fetches the artifacts from the integration corpus generator repo, but once we will update the package-spec to carry those information directly in the package content, it will be more loosing coupled to the integration corpus generator, using just the `Config` and `Fields` parsers and the `Generator`

internal/integration-corpus-generator/assets.go
```diff
@@ -0,0 +1,119 @@
...
+
+func tmpGenlibDir() (string, error) {
+	buildDir, err := builder.BuildDirectory()
+	if err != nil {
+		return "", errors.Wrap(err, "locating build directory failed")
+	}
+	return filepath.Join(buildDir, "genlib"), nil
+}
+
+func writeFieldsYamlFile(fieldsYamlContent []byte, packageName, dataStreamName string) (string, error) {
```
aspacca, line 97: this is currently needed because we don't have anything else than `fields.LoadFieldsWithTemplate` (requiring a path to a local file) to loads `Fields` from custom content. I will refactor/add a method that accepts `[]byte` instead in the dependency package, so that we can skip this

go.mod
```diff
@@ -9,6 +9,8 @@ require (
 	github.com/aymerick/raymond v2.0.2+incompatible
 	github.com/boumenot/gocover-cobertura v1.2.0
 	github.com/cespare/xxhash/v2 v2.2.0
+	github.com/dustin/go-humanize v1.0.0
+	github.com/elastic/elastic-integration-corpus-generator-tool v0.3.1-0.20230126015529-f42e909f4536
```
aspacca, line 13: a few changes are needed in github.com/elastic/elastic-integration-corpus-generator-tool before we can tag a proper version to use a required package here

internal/integration-corpus-generator/client.go
```diff
@@ -0,0 +1,75 @@
...
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/pkg/errors"
+
+	"github.com/elastic/elastic-package/internal/logger"
+)
+
+const integrationCorpusGeneratorAssetsBaseURL = "https://raw.githubusercontent.com/elastic/elastic-integration-corpus-generator-tool/main/assets/templates/"
```
aspacca, line 18: this is a temporary solution in place before we will have the relevant assets content from the integration corpus generator in package-spec.

until then I'm unsure how and if we should manage the version/tag of the assets: any extra complexity related to this will be finally removed once we can rely on package-spec, so I'd rather assume `main` branch and do not handle version/tag of the assets for the moment

cmd/generate.go
```diff
@@ -0,0 +1,28 @@
...
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/elastic/elastic-package/internal/cobraext"
+)
+
+const generateLongDescription = `Use this command to generate benchmarks data for a package. Currently, only data for what we have related assets on https://github.com/elastic/elastic-integration-corpus-generator-tool are supported.`
```
aspacca, line 13: maybe it is worth mentioning that this command has to be considered in beta: I'm not sure about the standard for beta/ga commands in elastic-package

@jsoriano any hint?

README.md
```diff
@@ -170,6 +170,12 @@ Use this command to format the package files.
 
 The formatter supports JSON and YAML format, and skips "ingest_pipeline" directories as it's hard to correctly format Handlebars template files. Formatted files are being overwritten.
 
+### `elastic-package generate`
```
jsoriano, line 173: "generate" may be a too generic command name, consider using a more specifc name or subcommand.

If the data is intended to be used in benchmars, maybe it could live under the benchmark command, something like `elastic-package benchmark generate-data`.

cmd/generate.go
jsoriano: As discussed, it is ok to add beta (or technical preview) commands. Please, add a mention to this here in the description, and consider logging an info message when this command is used.

Also, take into account that `elastic-package` is still 0.x, so we still reserve the right to make breaking changes to commands, although we try to avoid it.

internal/integration-corpus-generator/assets.go
jsoriano: I'd propose to avoid underscores in the name of this package. `corpusgenerator` can be a good name.

internal/integration-corpus-generator/client.go
jsoriano: Please add a `TODO`/`FIXME` comment here about this.

jsoriano (commented):
Thanks for integrating this tool into elastic-package! Added some overall comments.

README.md
ruflin: There might be both commands. The `generate` command is focused on generating the sample data sets. I expect there is a benchmark command that can have a flag to generate or use some pregenerated data.

jsoriano: > The `generate` command is focused on generating the sample data sets.

But `elastic-package generate` at the root level may sound as if it generates packages. If we want it independent of benchmarking, maybe we can keep it as `elastic-package generate-data` or `elastic-package dataset generate` if there can be more actions on datasets.

cmd/generate_data_stream.go
```diff
@@ -0,0 +1,92 @@
...
+		buf.WriteByte('\n')
+
+		if _, err = f.Write(buf.Bytes()); err != nil {
+			return err
+		}
+
+		currentSize += uint64(buf.Len())
+	}
+
+	return generator.Close()
```
mrodm, line 65: nit. What about adding these lines to its own function ? It could be something like `generateData` or `generateDatastreamData`

If so, would it make sense to move it to `integration-corpus-generator` module ?

internal/integration-corpus-generator/client.go
mrodm: Could it be set to a specific tag or reference instead of `main`? That would help to identify which templates are used in each elastic-package version.

cmd/generate_data_stream.go
```diff
@@ -0,0 +1,92 @@
...
+
+	cfg, err := genLibClient.GetConf(packageName, dataStreamName)
+	if err != nil {
+		return nil, err
+	}
+	flds, err := genLibClient.GetFields(packageName, dataStreamName)
+
+	if err != nil {
+		return nil, err
+	}
```
mrodm, line 80: nit. I would use the full word
```suggestion
	fields, err := genLibClient.GetFields(packageName, dataStreamName)
	if err != nil {
		return nil, err
	}
```

internal/integration-corpus-generator/client.go
aspacca: > Could it be set to a specific tag or reference instead of `main`? That would help to identify which templates are used in each elastic-package version.

I completely understand and agree with the value of identify which template are used in each elastic-package version, but the only way to reference a specific tag or reference instead of main is either:
1. pass the tag as argument of the command, so that's resolved at runtime
2. keep the tag in the elastic package codebase: this will require that whenever we add new templates on the corpus generator repo we need to file a tag, updated in the elastic package code and release a new elastic package version

I pinned main exactly to avoid 2 and at the same time being able to "consume" the artifacts as soon as they are merged

since fetching the artifacts from the corpus generator repo it's just a temporary solution I would also not introduce 1. since it will be later removed: when the information contained by the artifacts will be part of package-specs we'll have to define the integrations version, rather than the tag in the corpus generator repo.

what do you think about going for 1., using `package-version` or similar, so that we can later use the same flag to just execute a new logic?

it will be a little misleading while we are in the intermediate phase, but future proof

mrodm: @aspacca Thanks for sharing these options and information!

As you mentioned the idea is to have this as part of the package-spec repo, and if we add that new parameter to specify the tag, then it should be removed afterwards (once it is in the package-spec)... I think it can be kept as using `main` as it does not affect currentcly CI, but as @jsoriano suggested adding that TODO or FIXME comment.

aspacca:
how can I login to buildkite to check what's the error?

mrodm:
> how can I login to buildkite to check what's the error?

Didn't know that this pipeline was under the login. 
@aspacca This is the error (running make check-static)
```
cmd/benchmark.go:252:5: no value of type uint64 is less than 0 (SA4003)
```

aspacca:
> Didn't know that this pipeline was under the login. @aspacca This is the error (running make check-static)
> 
> ```
> cmd/benchmark.go:252:5: no value of type uint64 is less than 0 (SA4003)
> ```

I fixed that and other things and `make check-static` returns no error locally
buildkite action is still red

mrodm:
> I fixed that and other things and `make check-static` returns no error locally
> buildkite action is still red

This is the error:
```
+ '[' benchmarks == benchmarks ']'
--
+ rm -rf /opt/buildkite-agent/builds/bk-agent-elastic-gcp-1678259899748376435/elastic/elastic-package/build/benchmark-results
+ elastic-package benchmark -v --report-format xUnit --report-output file --fail-on-missing
Error: unknown flag: --report-format
+ cleanup
+ r=1
+ elastic-package stack dump -v --output build/elastic-stack-dump/check-benchmarks
```

cmd/benchmark.go
```diff
@@ -198,3 +211,62 @@ func benchTypeCommandActionFactory(runner benchrunner.BenchRunner) cobraext.Comm
...
+	}
+
+	generator, err := corpusgenerator.GetGenerator(packageName, dataStreamName, commit, totSizeInBytes)
+	if err != nil {
+		return errors.Wrap(err, "can't generate benchmarks data corpus for data stream")
+	}
+
+	err = corpusgenerator.RunGenerator(generator)
+	if err != nil {
+		return errors.Wrap(err, "can't generate benchmarks data corpus for data stream")
```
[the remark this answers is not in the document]
aspacca: we have to steps: 
1. init the generator (`corpusgenerator.NewGenerator`)
2. run the generator (`corpusgenerator.RunGenerator`) 

in the end, if any of the above fails, we cannot generate the data. so I'd say the message can be the same, without exposing details about which one of the two steps failed: this information I would expect to come from the original error we are wrapping.

but I'm find having two different messages if you think it's better

aspacca:
> This is the error:
> 
> ```
> + '[' benchmarks == benchmarks ']'
> --
> + rm -rf /opt/buildkite-agent/builds/bk-agent-elastic-gcp-1678259899748376435/elastic/elastic-package/build/benchmark-results
> + elastic-package benchmark -v --report-format xUnit --report-output file --fail-on-missing
> Error: unknown flag: --report-format
> ```

yes, I've found that thanks. finally got access to buildkite :)

cmd/benchmark.go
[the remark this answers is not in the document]
mrodm: > I'd say the message can be the same, without exposing details about which one of the two steps failed: this information I would expect to come from the original error we are wrapping.

Ok! the original error would add the needed information :+1:

mrodm (approved):

ruflin:
@aspacca It would be great to have a bit more details in the PR description that can get someone just started. A copy / paste command with existing datasets. What is in the linked issue is not accurate anymore.

Issue #1102, opened 2023-01-19
[K8s testing] Support use of Kustomize


constanca-m:
## What does this issue try to achieve?
Updating kube state metrics version on testing at most one time.

## What is the problem?
Right now, the current kube state metrics version being used for Kubernetes integration testing is `2.3.0`. 
However, this version is no longer aligned with the Kubernetes version being supported, which range between
`1.23.x`-`1.26.x`. This implies that, at least, there should be support for `2.4.2` version - and this version should keep being incremented at the same time Kubernetes versions are released and supported by the integration. 

Example of a problem the current being used version is causing (also documented [here](https://github.com/elastic/integrations/blob/main/packages/kubernetes/docs/kube-state-metrics.md#state_cronjob)):
```
DEBUG found 0 hits in metrics-kubernetes.state_cronjob-ep data stream: index_not_found_exception: no such index [metrics-kubernetes.state_cronjob-ep] Status=404
```
This causes our tests to fail and makes it impossible to merge PRs on our Kubernetes integration.

The problem is that changing this version, doesn't apply changing only one tag. We need to:
- Update tag on five different files (the ones related to kube state metrics), and, on top of that:
- We need to update other things, such [as permissions](https://github.com/elastic/integrations/pull/5012#pullrequestreview-1254056620).

This can be seen on [this PR](https://github.com/elastic/integrations/pull/5061).

This is, of course, not ideal and this process should not require such high maintenance. Since, Kustomize is already being used ([see from this PR](https://github.com/elastic/elastic-agent/pull/2104)), the same approach should be followed.

## Steps this would require

We would need to:

1. [In elastic package repository](https://github.com/elastic/elastic-package) Change flag `-f` on kubectl to `-k` in [here](https://github.com/elastic/elastic-package/blob/main/internal/kubectl/kubectl.go) to allow (and enforce) the use of Kustomize;
2. [In integrations repository](https://github.com/elastic/integrations) Update [files](https://github.com/elastic/integrations/tree/main/packages/kubernetes/_dev/deploy/k8s) in a way that a new `kustomization.yaml` file is deployed, that will simultaneously apply kube-state-metrics from the original repository. Same approach as shown in this merged [PR](https://github.com/elastic/elastic-agent/pull/2104);
3. Document this (maybe in this [file](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md#kubernetes-service-deployer)).

[...]

PR #1125, opened 2023-01-31
[K8s testing] Support use of Kustomize

constanca-m:
# What

Support use of Kustomize to remove having to maintain KSM version according to the latest supported Kubernetes version. See more details in [this issue](https://github.com/elastic/elastic-package/issues/1102).

**Note**: This change only affects the testing of Kubernetes integration. More information on this [here](https://github.com/elastic/elastic-package/issues/1102).

# Walk through

There are four steps to run the tests (check the code [here](https://github.com/elastic/elastic-package/blob/ed8b57f1bc760f05f7ea1cd6b4f83f9fdfde19ed/internal/testrunner/runners/system/servicedeployer/kubernetes.go#L81-L111)):
1. Verify context.
2. Connect to the elastic stack network.
3. Install Elastic Agent.
4. Install custom definitions. These are present in the folder `kubernetes/_dev/deploy/k8s`.

Of these steps, manifests are being deployed through 3 and 4. However, step 3 does not use files from our custom definitions, and therefore, is not important for the specific testing files of Kubernetes integration.

To not install custom definitions, we need the presence of an `.empty` file. Taking [`apiserver` datastream](https://github.com/elastic/integrations/tree/e001c8650917a1144500540573a16b95fd2b6a92/packages/kubernetes/data_stream/apiserver/_dev/deploy/k8s) as an example, since we have the `.empty` file, nothing from `kubernetes/_dev/deploy/k8s` will be deployed.
On the other hand, [`state_cronjob` datastream](https://github.com/elastic/integrations/tree/e001c8650917a1144500540573a16b95fd2b6a92/packages/kubernetes/data_stream/state_cronjob) does not have that file, so custom definitions will be installed.

The main function responsible for applying the **custom** manifests is this [one](https://github.com/elastic/elastic-package/blob/ed8b57f1bc760f05f7ea1cd6b4f83f9fdfde19ed/internal/kubectl/kubectl.go#L30-L51). **This function is only used to deploy the custom definitions**, not the elastic agent. So changing it won't affect anything related to it. Our specific goal to Kubernetes integration is to deploy the newest KSM version in an automated way, so we need to use a `kustomization.yaml` file to ensure that. This will stop requiring maintenance in regard to future KSM version. For that, we need to change our `-f` flag to `-k` flag.

To determine the correct path for the custom files we use this [function](https://github.com/elastic/elastic-package/blob/ed8b57f1bc760f05f7ea1cd6b4f83f9fdfde19ed/internal/testrunner/runners/system/servicedeployer/kubernetes.go#L135), which is, once again, **only used to installed custom definitions**, so step 3 will not be affected. It will be refactored to:
1. Check if `_dev/deploy/k8s` directory given as argument exists. (Same as before)
2. Check if `kustomization.yaml` file exists.
   - If it does not exist, then the `.empty` file **needs** to exist. Otherwise, testing will fail. This condition, although not really necessary, remains present to keep the same logic as before and not cause confusion. 
   - Otherwise, deploy it.
```go
func containsKustomization(definitionsPath string) (bool, error) {
	if _, err := os.Stat(definitionsPath); err != nil {
		return false, errors.Wrapf(err, "can't read definitions directory (path: %s)", definitionsPath)
	}
	// check if kustomization.yaml exists
	if _, err := os.Stat(filepath.Join(definitionsPath, "kustomization.yaml")); err != nil {
		// if it does not exist, then the .empty file needs to be present
		if _, err := os.Stat(filepath.Join(definitionsPath, ".empty")); err != nil {
			return false, errors.Errorf("kustomization.yaml file is missing (path: %s). Add one or create an .empty file"+
				" if no custom definitions are required.", definitionsPath)
		}
		return false, nil
	}
	return true, nil
}
```

# Outcome

Without changing anything from Kubernetes integration package, running the tests with these changes would result in an error once the first `state_*` datastream testing starts:

```
2023/01/31 16:39:24 DEBUG install custom Kubernetes definitions (directory: /home/c/go/src/github.com/elastic/integrations/packages/kubernetes/_dev/deploy/k8s)
Error: error running package system tests: could not complete test run: could not setup service: can't install custom definitions in the Kubernetes cluster: can't find Kubernetes definitions in given directory (path: /home/c/go/src/github.com/elastic/integrations/packages/kubernetes/_dev/deploy/k8s): kustomization.yaml file is missing (path: /home/c/go/src/github.com/elastic/integrations/packages/kubernetes/_dev/deploy/k8s)
```

If we update the custom definitions folder to have the right `kustomization.yaml` file - i.e. one that deploys kube state metrics and all the needed resources - the logs for each `state_*` datastream would be like this:

```
2023/01/31 16:58:01 DEBUG Running system tests for data stream
2023/01/31 16:58:01 DEBUG running test with configuration 'default'
2023/01/31 16:58:01 DEBUG setting up service...
2023/01/31 16:58:01 DEBUG ensure that kind context is selected
2023/01/31 16:58:01 DEBUG output command: /usr/local/bin/kubectl config current-context
2023/01/31 16:58:02 DEBUG find "kind-control-plane" container
2023/01/31 16:58:02 DEBUG output command: /usr/local/bin/docker ps --filter name=kind-control-plane --format {{.ID}}
2023/01/31 16:58:02 DEBUG check network connectivity between service container kind-control-plane (ID: 7765398977cd) and the stack network elastic-package-stack_default
2023/01/31 16:58:02 DEBUG output command: /usr/local/bin/docker network inspect elastic-package-stack_default
2023/01/31 16:58:02 DEBUG container kind-control-plane is already attached to the elastic-package-stack_default network
2023/01/31 16:58:02 DEBUG install Elastic Agent in the Kubernetes cluster
2023/01/31 16:58:02 DEBUG GET https://127.0.0.1:5601/api/status
2023/01/31 16:58:02 DEBUG Prepare YAML definition for Elastic Agent running in stack v8.6.0
2023/01/31 16:58:02 DEBUG Apply Kubernetes stdin
2023/01/31 16:58:02 DEBUG run command: /usr/local/bin/kubectl apply -f - -o yaml
2023/01/31 16:58:03 DEBUG Handle "apply" command output
2023/01/31 16:58:03 DEBUG Extract resources from command output
2023/01/31 16:58:03 DEBUG Wait for ready resources
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-package-certs (kind: Secret, namespace: kube-system)
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent (kind: DaemonSet, namespace: kube-system)
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent (kind: ClusterRoleBinding, namespace: )
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent (kind: RoleBinding, namespace: kube-system)
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent-kubeadm-config (kind: RoleBinding, namespace: kube-system)
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent (kind: ClusterRole, namespace: )
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent (kind: Role, namespace: kube-system)
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent-kubeadm-config (kind: Role, namespace: kube-system)
2023/01/31 16:58:03 DEBUG Sync resource info: elastic-agent (kind: ServiceAccount, namespace: kube-system)
2023/01/31 16:58:03 DEBUG beginning wait for 9 resources with timeout of 10m0s
2023/01/31 16:58:03 DEBUG install custom Kubernetes definitions (directory: /home/c/go/src/github.com/elastic/integrations/packages/kubernetes/_dev/deploy/k8s)
2023/01/31 16:58:03 DEBUG Apply Kubernetes custom definitions
2023/01/31 16:58:03 DEBUG run command: /usr/local/bin/kubectl apply -k /home/c/go/src/github.com/elastic/integrations/packages/kubernetes/_dev/deploy/k8s -o yaml
2023/01/31 16:58:05 DEBUG Handle "apply" command output
2023/01/31 16:58:05 DEBUG Extract resources from command output
2023/01/31 16:58:05 DEBUG Wait for ready resources
2023/01/31 16:58:05 DEBUG Sync resource info: pods-high (kind: ResourceQuota, namespace: default)
2023/01/31 16:58:05 DEBUG Sync resource info: kube-state-metrics (kind: ServiceAccount, namespace: kube-system)
2023/01/31 16:58:06 DEBUG Sync resource info: kube-state-metrics (kind: ClusterRole, namespace: )
2023/01/31 16:58:06 DEBUG Sync resource info: kube-state-metrics (kind: ClusterRoleBinding, namespace: )
2023/01/31 16:58:06 DEBUG Sync resource info: example-redis-config (kind: ConfigMap, namespace: default)
2023/01/31 16:58:06 DEBUG Sync resource info: kube-state-metrics (kind: Service, namespace: kube-system)
2023/01/31 16:58:06 DEBUG Sync resource info: task-pv-volume (kind: PersistentVolume, namespace: )
2023/01/31 16:58:06 DEBUG Sync resource info: task-pv-claim (kind: PersistentVolumeClaim, namespace: default)
2023/01/31 16:58:06 DEBUG Sync resource info: kube-state-metrics (kind: Deployment, namespace: kube-system)
2023/01/31 16:58:06 DEBUG Sync resource info: web (kind: StatefulSet, namespace: default)
2023/01/31 16:58:06 DEBUG Sync resource info: hello (kind: CronJob, namespace: default)
2023/01/31 16:58:06 DEBUG Sync resource info: fluentd-elasticsearch (kind: DaemonSet, namespace: kube-system)
2023/01/31 16:58:06 DEBUG Sync resource info: hello (kind: Job, namespace: default)
2023/01/31 16:58:06 DEBUG beginning wait for 13 resources with timeout of 10m0s
2023/01/31 16:58:06 DEBUG Deployment is not ready: kube-system/kube-state-metrics. 0 out of 1 expected pods are ready
2023/01/31 16:58:08 DEBUG Deployment is not ready: kube-system/kube-state-metrics. 0 out of 1 expected pods are ready
2023/01/31 16:58:10 DEBUG Deployment is not ready: kube-system/kube-state-metrics. 0 out of 1 expected pods are ready
2023/01/31 16:58:12 DEBUG Deployment is not ready: kube-system/kube-state-metrics. 0 out of 1 expected pods are ready
2023/01/31 16:58:14 DEBUG Deployment is not ready: kube-system/kube-state-metrics. 0 out of 1 expected pods are ready
2023/01/31 16:58:16 DEBUG StatefulSet is ready: default/web. 1 out of 1 expected pods are ready
2023/01/31 16:58:16 DEBUG creating test policy...
2023/01/31 16:58:16 DEBUG POST https://127.0.0.1:5601/api/fleet/agent_policies
2023/01/31 16:58:20 DEBUG adding package data stream to test policy...
2023/01/31 16:58:20 DEBUG POST https://127.0.0.1:5601/api/fleet/package_policies
2023/01/31 16:58:23 DEBUG deleting old data in data stream...
2023/01/31 16:58:23 DEBUG found 0 hits in metrics-kubernetes.state_storageclass-ep data stream
2023/01/31 16:58:23 DEBUG GET https://127.0.0.1:5601/api/fleet/agents
2023/01/31 16:58:23 DEBUG filter agents using criteria: NamePrefix=kind-control-plane
2023/01/31 16:58:23 DEBUG found 1 enrolled agent(s)
2023/01/31 16:58:23 DEBUG GET https://127.0.0.1:5601/api/fleet/agent_policies/134afa00-a180-11ed-be41-0b0968616500
2023/01/31 16:58:23 DEBUG assigning package data stream to agent...
2023/01/31 16:58:23 DEBUG PUT https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560/reassign
2023/01/31 16:58:25 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:25 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"134afa00-a180-11ed-be41-0b0968616500","local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:25 DEBUG Wait until the policy (ID: 134afa00-a180-11ed-be41-0b0968616500, revision: 2) is assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:27 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:27 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"134afa00-a180-11ed-be41-0b0968616500","local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:27 DEBUG Wait until the policy (ID: 134afa00-a180-11ed-be41-0b0968616500, revision: 2) is assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:29 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:29 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"134afa00-a180-11ed-be41-0b0968616500","local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:29 DEBUG Wait until the policy (ID: 134afa00-a180-11ed-be41-0b0968616500, revision: 2) is assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:31 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:31 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"134afa00-a180-11ed-be41-0b0968616500","policy_revision":2,"local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:31 DEBUG Policy revision assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:31 DEBUG checking for expected data in data stream...
2023/01/31 16:58:31 DEBUG found 0 hits in metrics-kubernetes.state_storageclass-ep data stream
2023/01/31 16:58:32 DEBUG found 0 hits in metrics-kubernetes.state_storageclass-ep data stream
2023/01/31 16:58:33 DEBUG found 0 hits in metrics-kubernetes.state_storageclass-ep data stream
2023/01/31 16:58:34 DEBUG found 1 hits in metrics-kubernetes.state_storageclass-ep data stream
2023/01/31 16:58:34 DEBUG reassigning original policy back to agent...
2023/01/31 16:58:34 DEBUG PUT https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560/reassign
2023/01/31 16:58:36 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:36 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"elastic-agent-managed-ep","local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:36 DEBUG Wait until the policy (ID: elastic-agent-managed-ep, revision: 3) is assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:38 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:38 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"elastic-agent-managed-ep","local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:38 DEBUG Wait until the policy (ID: elastic-agent-managed-ep, revision: 3) is assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:40 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:40 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"elastic-agent-managed-ep","local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:40 DEBUG Wait until the policy (ID: elastic-agent-managed-ep, revision: 3) is assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:42 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:42 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"elastic-agent-managed-ep","local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:42 DEBUG Wait until the policy (ID: elastic-agent-managed-ep, revision: 3) is assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:44 DEBUG GET https://127.0.0.1:5601/api/fleet/agents/25fd07a7-522f-4399-af4d-c441b2867560
2023/01/31 16:58:44 DEBUG Agent data: {"id":"25fd07a7-522f-4399-af4d-c441b2867560","policy_id":"elastic-agent-managed-ep","policy_revision":3,"local_metadata":{"host":{"name":"kind-control-plane"}}}
2023/01/31 16:58:44 DEBUG Policy revision assigned to the agent (ID: 25fd07a7-522f-4399-af4d-c441b2867560)...
2023/01/31 16:58:44 DEBUG deleting test policy...
2023/01/31 16:58:44 DEBUG POST https://127.0.0.1:5601/api/fleet/agent_policies/delete
2023/01/31 16:58:47 DEBUG tearing down service...
2023/01/31 16:58:47 DEBUG uninstall custom Kubernetes definitions (directory: /home/c/go/src/github.com/elastic/integrations/packages/kubernetes/_dev/deploy/k8s)
2023/01/31 16:58:47 DEBUG run command: /usr/local/bin/kubectl delete -k /home/c/go/src/github.com/elastic/integrations/packages/kubernetes/_dev/deploy/k8s
2023/01/31 16:58:49 DEBUG deleting data in data stream...
```

# Related issues
- Closes https://github.com/elastic/elastic-package/issues/1102


internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -132,15 +132,17 @@ func (ksd KubernetesServiceDeployer) installCustomDefinitions() error {
...
+	if _, err := os.Stat(definitionsDir); err != nil {
+		return "", errors.Wrapf(err, "can't read definitions directory (path: %s)", definitionsDir)
 	}
-
-	var definitionPaths []string
-	definitionPaths = append(definitionPaths, files...)
-	return definitionPaths, nil
+	if _, err := os.Stat(filepath.Join(definitionsDir, "kustomization.yaml")); err != nil {
+		if _, err := os.Stat(filepath.Join(definitionsDir, ".empty")); err != nil {
+			return "", errors.Errorf("kustomization.yaml file is missing (path: %s)", definitionsDir)
```
gsantoro, line 141: maybe we should also say the alternative way to fix this. Either add a .empty file or add a kustomization.yaml.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -132,15 +132,17 @@ func (ksd KubernetesServiceDeployer) installCustomDefinitions() error {
 
 var _ ServiceDeployer = new(KubernetesServiceDeployer)
 
-func findKubernetesDefinitions(definitionsDir string) ([]string, error) {
-	files, err := filepath.Glob(filepath.Join(definitionsDir, "*.yaml"))
-	if err != nil {
-		return nil, errors.Wrapf(err, "can't read definitions directory (path: %s)", definitionsDir)
+func findKubernetesDefinitions(definitionsDir string) (string, error) {
```
gsantoro, line 135: what about `checkKubernetesDefinition(definitionsDir string) (error)` since it just validates that directory either contains an .empty file or a kustomization.yaml.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -132,15 +132,17 @@ func (ksd KubernetesServiceDeployer) installCustomDefinitions() error {
...
+	if _, err := os.Stat(definitionsDir); err != nil {
+		return "", errors.Wrapf(err, "can't read definitions directory (path: %s)", definitionsDir)
```
gsantoro, line 137: `definitions directory` is not consistent with `definitionPath` any longer. I think we should either use the plural in every case (eg. definitionsPath) or come up with a better name

docs/howto/system_testing.md
```diff
@@ -302,12 +302,11 @@ There are some specific environment variables that should be leveraged to overco
 
 ### Kubernetes service deployer
 
-The Kubernetes service deployer requires the `_dev/deploy/k8s` directory to be present. It can include additional `*.yaml` files to deploy
-custom applications in the Kubernetes cluster (e.g. Nginx deployment). If no resource definitions (`*.yaml` files ) are needed,
-the `_dev/deploy/k8s` directory must contain an `.empty` file (to preserve the `k8s` directory under version control).
+The Kubernetes service deployer requires the `_dev/deploy/k8s` directory to be present. 
+It needs to include either a `kustomization.yaml` file or an `.empty` file. The first is needed to ensure the deployment of each resource.
```
gsantoro, line 306: `It needs to include either a `kustomization.yaml` file or an `.empty` file. The first is needed to ensure the deployment of each resource.... **The second in case there are no resources to deploy.** `

I would maybe removed the following phrase since it is better expressed by the suggested statement above and it is redundant with what said before about the folder being needed anyway.

gsantoro (changes requested):

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -25,31 +25,31 @@ import (
...
-	definitionsDir string
+	definitionsPath string
 }
 
 func (s kubernetesDeployedService) TearDown() error {
-	logger.Debugf("uninstall custom Kubernetes definitions (directory: %s)", s.definitionsDir)
+	logger.Debugf("uninstall custom Kubernetes definitions (directory: %s)", s.definitionsPath)
 
-	definitionPath, err := findKubernetesDefinitions(s.definitionsDir)
+	definitionsPath, err := checkKubernetesDefinition(s.definitionsPath)
```
gsantoro, line 40:
```suggestion
	err := checkKubernetesDefinition(s.definitionsPath)
```

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -25,31 +25,31 @@ import (
...
 	if err != nil {
-		return errors.Wrapf(err, "can't find Kubernetes definitions in given directory (path: %s)", s.definitionsDir)
+		return errors.Wrapf(err, "can't find Kubernetes definitions in given directory (path: %s)", s.definitionsPath)
 	}
 
-	if len(definitionPath) == 0 {
-		logger.Debugf("no custom definitions found (directory: %s). Nothing will be uninstalled.", s.definitionsDir)
+	if len(definitionsPath) == 0 {
```
gsantoro, line 45:
```suggestion
	if len(s.definitionsPath) == 0 {
```

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -25,31 +25,31 @@ import (
...
+		logger.Debugf("no custom definitions found (directory: %s). Nothing will be uninstalled.", s.definitionsPath)
 		return nil
 	}
 
-	err = kubectl.Delete(definitionPath)
+	err = kubectl.Delete(definitionsPath)
```
gsantoro, line 50:
```suggestion
	err = kubectl.Delete(s.definitionsPath)
```

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -132,17 +132,20 @@ func (ksd KubernetesServiceDeployer) installCustomDefinitions() error {
 
 var _ ServiceDeployer = new(KubernetesServiceDeployer)
 
-func findKubernetesDefinitions(definitionsDir string) (string, error) {
-	if _, err := os.Stat(definitionsDir); err != nil {
-		return "", errors.Wrapf(err, "can't read definitions directory (path: %s)", definitionsDir)
+func checkKubernetesDefinition(definitionsPath string) (string, error) {
```
gsantoro, line 135: what about only returning the error since the function is called `check...`

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -105,25 +105,25 @@ func (ksd KubernetesServiceDeployer) SetUp(ctxt ServiceContext) (DeployedService
...
+		definitionsPath: ksd.definitionsPath,
 	}, nil
 }
 
 func (ksd KubernetesServiceDeployer) installCustomDefinitions() error {
-	logger.Debugf("install custom Kubernetes definitions (directory: %s)", ksd.definitionsDir)
+	logger.Debugf("install custom Kubernetes definitions (directory: %s)", ksd.definitionsPath)
 
-	definitionPath, err := findKubernetesDefinitions(ksd.definitionsDir)
+	definitionsPath, err := checkKubernetesDefinition(ksd.definitionsPath)
```
gsantoro, line 116: I'll be inclined to only return the error and use the original variable like suggested above.

gsantoro (changes requested):

internal/testrunner/runners/system/servicedeployer/kubernetes.go
constanca-m: Not great. If the directory has the empty file, we don't want to return the path, so have to return an empty string to avoid installing custom resources.

gsantoro: I think I see what you mean. The output of checkKubernetesDefinition has 3 states:
1. kustomization.yaml found, no errors = return (true, nil)
2. empty folder or .empty file found, no errors = return (false, nil)
3. error = same behaviour as before = return (false, nil)

The name of the first return value (now a boolean) can be "found" or something similar.

A better name for the function is then `containsKustomization`

tetianakravchenko (commented):
1. I still don't think that should be enforced usage of `kustomization.yaml`, if manually defined manifests stored in `_dev/deploy/k8s` covers all needs and there is no reason to use kustomization.yml - this should be supported
2. I think you also need to update https://github.com/elastic/elastic-package/tree/main/test/packages/with-kind/kubernetes with accordance with your changes

constanca-m:
> 1. I still don't think that should be enforced usage of `kustomization.yaml`, if manually defined manifests stored in `_dev/deploy/k8s` covers all needs and there is no reason to use kustomization.yml - this should be supported

But what is the reason for that? It seems unnecessary. This is used just for testing, so once there, it doesn't really matter what happens next. And since at this moment, Kubernetes integration is the only one that uses this, it is an easy way to change system testing as we won't need to spread this change to any other package. @tetianakravchenko

internal/testrunner/runners/system/servicedeployer/kubernetes.go
constanca-m: Updated to that.

gsantoro (approved):

tetianakravchenko:
> >        I still don't think that should be enforced usage of kustomization.yaml, if manually defined manifests stored in _dev/deploy/k8s covers all needs and there is no reason to use kustomization.yml - this should be supported
> But what is the reason for that? It seems unnecessary. This is used just for testing, so once there, it doesn't really matter what happens next. And since at this moment, Kubernetes integration is the only one that uses this, it is an easy way to change system testing as we won't need to spread this change to any other package. @tetianakravchenko

yes, Kubernetes integration is the only one that uses it, but I don't think that it implies that this should be enforced for other integrations, that are not using those tests yet (like cloud_defend and cloud_security_posture), I see it more like an alternative

constanca-m:
> yes, Kubernetes integration is the only one that uses it, but I don't think that it implies that this should be enforced for other integrations, that are not using those tests yet (like cloud_defend and cloud_security_posture), I see it more like an alternative

They are using the .empty file so they don't require any changes.

> @constanca-m also wdyt about this? namely content of https://github.com/elastic/elastic-package/tree/main/test/packages/with-kind/kubernetes/_dev/deploy/k8s to align with your changes?

I hadn't notice that, but the datastream being used for testing has the .empty file so we were not testing the custom definitions. I will update it to include the new resources + one more data stream to check the kustomization.yaml file. @tetianakravchenko

tetianakravchenko:
> They are using the .empty file so they don't require any changes.

ok, lets iterate over it

> I hadn't notice that, but the datastream being used for testing has the .empty file so we were not testing the custom definitions. I will update it to include the new resources + one more data stream to check the kustomization.yaml file. @tetianakravchenko

hmm, maybe I misunderstood the purpose of this test folder. But I think it is good to have an example in this repo of the package with `kustomization.yaml`

constanca-m:
> hmm, maybe I misunderstood the purpose of this test folder. But I think it is good to have an example in this repo of the package with `kustomization.yaml`

You got it right, that folder is being used for testing. The tests were not covering the two cases before, but now they are.

test/packages/with-kind/kubernetes/_dev/deploy/k8s/daemonset.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
```
gsantoro, line 3: where is this daemonset.yaml coming from?

constanca-m: We need these resources to check the `state_*` datastreams, but since I only included one for the pod, maybe I should delete all the others? I was trying to follow the same logic as the one that was here before.

tetianakravchenko (approved):

test/packages/with-kind/kubernetes/_dev/deploy/k8s/daemonset.yaml
constanca-m: I am torn between deleting some of the resources because they are useless, or keeping them so they stay the same as in the integrations repo

tetianakravchenko:
@constanca-m please double check with @elastic/ecosystem team regarding failing `buildkite/elastic-package` before merging

test/packages/with-kind/kubernetes/_dev/deploy/k8s/daemonset.yaml
gsantoro: let me see if I understand. 

you are adding a state_pod datastream here since `pod` didn't even need the kube-state-metrics resource that was already defined in those manifests resources (before the kustomization was introduced). those resources in deploy/k8s are in theory used to setup the environment for testing but since we don't check for output against an expected result, they are kind of useless. we only check that the setup didn't have any issues.

I would say, keep them in sync with what we are doing in the integrations repo. it's definitively another PR

constanca-m: Yes, they were useless before because the only data stream defined had the `.empty` file. They are no longer useless.

constanca-m:
> @constanca-m please double check with @elastic/ecosystem team regarding failing `buildkite/elastic-package` before merging

I don't think the error is related to any change of this PR @tetianakravchenko

constanca-m:
@tetianakravchenko @gsantoro The error is more specific than the screenshot yesterday: 

```
Deprecated Environment Variables

The following environment variables have been deprecated, and should no longer be used:

BUILDKITE_PROJECT_SLUG="elastic/elastic-package"
BUILDKITE_PROJECT_PROVIDER="github"
```

But it seems that this should be a different PR. I tried to find these variables specific to this change, but I don't think they exist.

gsantoro:
@constanca-m actually the pipeline is now there at https://buildkite.com/elastic/elastic-package

constanca-m:
> @constanca-m actually the pipeline is now there at https://buildkite.com/elastic/elastic-package

That's for every run, but if you check the error of the specific cause of this PR, that is the log that appears. @gsantoro

mlunadia:
is this linked to [this epic](https://github.com/elastic/observability-dev/issues/2410)?

constanca-m:
> @constanca-m is this linked to [this epic](https://github.com/elastic/observability-dev/issues/2410)?

I wouldn't say so @mlunadia. KSM was already being deployed, this change is just so we don't have to update the version every time we support a new version of K8s.

docs/howto/system_testing.md
```diff
@@ -302,12 +302,11 @@ There are some specific environment variables that should be leveraged to overco
 
 ### Kubernetes service deployer
 
-The Kubernetes service deployer requires the `_dev/deploy/k8s` directory to be present. It can include additional `*.yaml` files to deploy
-custom applications in the Kubernetes cluster (e.g. Nginx deployment). If no resource definitions (`*.yaml` files ) are needed,
-the `_dev/deploy/k8s` directory must contain an `.empty` file (to preserve the `k8s` directory under version control).
+The Kubernetes service deployer requires the `_dev/deploy/k8s` directory to be present. 
+It needs to include either a `kustomization.yaml` file or an `.empty` file. The first is needed to ensure the deployment of each resource,
+while the second is required if no custom resources are needed.
```
jsoriano, line 307: Not sure if I understand this correctly. Plain resources are not going to be supported anymore, and only kustomize can be used now? This is a breaking change, right?

test/packages/with-kind/kubernetes/_dev/deploy/k8s/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+resources:
+  - pv.yaml
+  - pvc.yaml
+  - cronjob.yaml
+  - job.yaml
+  - daemonset.yaml
+  - configmap.yaml
+  - rq.yaml
+  - ss.yaml
+  - github.com/kubernetes/kube-state-metrics?ref=main
```
jsoriano, line 10: Please, use a reference to a released version here to avoid unpredictable builds.
```suggestion
  - github.com/kubernetes/kube-state-metrics?ref=abe3fd3184e16893b5a47196f90a94ed13e1b04d
```
Or a tag.
```suggestion
  - github.com/kubernetes/kube-state-metrics?ref=v2.7.0
```

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -132,15 +132,20 @@ func (ksd KubernetesServiceDeployer) installCustomDefinitions() error {
...
+func containsKustomization(definitionsPath string) (bool, error) {
+	if _, err := os.Stat(definitionsPath); err != nil {
+		return false, errors.Wrapf(err, "can't read definitions directory (path: %s)", definitionsPath)
 	}
-
-	var definitionPaths []string
-	definitionPaths = append(definitionPaths, files...)
-	return definitionPaths, nil
+	// check if kustomization.yaml exists
+	if _, err := os.Stat(filepath.Join(definitionsPath, "kustomization.yaml")); err != nil {
```
jsoriano, line 140: Nit. Please use a constant for `kustomization.yaml` string, specially if it is used in multiple places.

internal/testrunner/runners/system/servicedeployer/kubernetes.go
```diff
@@ -132,15 +132,20 @@ func (ksd KubernetesServiceDeployer) installCustomDefinitions() error {
...
+		// if it does not exist, then the .empty file needs to be present
+		if _, err := os.Stat(filepath.Join(definitionsPath, ".empty")); err != nil {
+			return false, errors.Errorf("kustomization.yaml file is missing (path: %s). Add one or create an .empty file"+
+				" if no custom definitions are required.", definitionsPath)
```
jsoriano, line 144: If we require certain files to be present, it would be better to make these checks on the package-spec.

jsoriano (changes requested):
This looks like a breaking change, and it looks feasible to keep backwards compatibility by checking if kustomize is used or not. Could we do it?

Please correct me if this will work with current packages.

test/packages/with-kind/kubernetes/_dev/deploy/k8s/kustomization.yaml
constanca-m: The idea of using `main` was to make it unnecessary to update KSM version every time. `main` should always be pointing to the latest version.

docs/howto/system_testing.md
constanca-m: It is a breaking change only for Kubernetes. Which will be immediately fixed. Regarding the `.empty` file: it was already required, so there is no change there. I explained better [in this issue](https://github.com/elastic/elastic-package/issues/1102) why this only affects Kubernetes integration.

jsoriano: It is a breaking change in `elastic-package`, that seems that can be prevented without much more effort. I can understand that it probably affects only an integration, but I would really prefer to avoid it.

Accepting the breaking change will complicate updating the integrations repository (elastic-package and the kubernetes integration will need to be updated at the same time, or we need to have CI broken between updates), and it can be a problem if work needs to be done on the kubernetes package with older versions, for backports for example.

test/packages/with-kind/kubernetes/_dev/deploy/k8s/kustomization.yaml
jsoriano: This is a test in `elastic-package` that happens to reference KSM repository, we are not testing the functionality of the kubernetes integration or its integration with KSM here.
Having this reference to main can break elastic-package builds unexpectedly, when it doesn't have any relationship to KSM.

There is no advantage on using main here, and can be a source of problems.

docs/howto/system_testing.md
jsoriano: Btw, following the conversation in the linked issue, it also seems preferred to continue supporting the current method (https://github.com/elastic/elastic-package/issues/1102#issuecomment-1398967317, https://github.com/elastic/elastic-package/issues/1102#issuecomment-1403352269).

constanca-m: > elastic-package and the kubernetes integration will need to be updated at the same time, or we need to have CI broken between updates

That is correct. That's why the plan is to open the PR on Kubernetes integration as soon as this one is merged. It probably won't take more than a few hours to correct that.

> it can be a problem if work needs to be done on the kubernetes package with older versions

I thought there was no backport when it comes to integrations?

gizas: @constanca-m can we enhance the check if kustomisation.yaml exists to use -k, if `.empty`  file exists do nothing otherwise use the -f ? 

So as to keep both manifests and ksutomize?

constanca-m: Yes, that is an easy change

jsoriano: Well, they are only examples of problems. The main thing is that we prefer to avoid breaking changes, unless they have a very good justification. There are good reasons for that. What I miss here is the justification to introduce this breaking change, when there seem to be preferred and reasonable alternatives.

> I thought there was no backport when it comes to integrations?

We prefer to don't have them, yes, but they are [supported](https://github.com/elastic/integrations/blob/main/docs/developer_workflow_bug_fix_older_package_version.md). The thing is that if someone needs to work with an old integration, even if it is only to try to reproduce an issue, and a breaking change has been introduced, they may find problems and require to downgrade their elastic-package installation.

test/packages/with-kind/kubernetes/_dev/deploy/k8s/kustomization.yaml
```diff
@@ -7,4 +7,4 @@ resources:
   - configmap.yaml
   - rq.yaml
   - ss.yaml
-  - github.com/kubernetes/kube-state-metrics?ref=main
+  - github.com/kubernetes/kube-state-metrics?ref=abe3fd3184e16893b5a47196f90a94ed13e1b04d
```
jsoriano, line 10: :+1: thanks

docs/howto/system_testing.md
jsoriano: Race condition sending comments :slightly_smiling_face: 

> can we enhance the check if kustomisation.yaml exists to use -k, if `.empty` file exists do nothing otherwise use the -f ?

> Yes, that is an easy change

This looks great to me.

constanca-m: That makes sense, thank you for explaining. I agree that `-f` should be supported along with `-k` and I will update to make that possible again. Just adding, for future reference, that `-f` may be necessary if users want to create a new metricset and have no intention in using a kustomization file.

test/packages/with-kind/kubernetes/data_stream/state_cronjob/fields/agent.yml
```diff
@@ -0,0 +1,198 @@
+- name: cloud
```
ChrsMark, line 1: Since it's only for testing, could we reduce the amount of fields we have in this file?

jsoriano (commented):
Thanks for the updates to avoid the breaking change!

internal/kubectl/kubectl.go
```diff
@@ -27,9 +30,13 @@ func CurrentContext() (string, error) {
 	return string(bytes.TrimSpace(output)), nil
 }
 
-func modifyKubernetesResources(action string, definitionPaths ...string) ([]byte, error) {
+func modifyKubernetesResources(action string, definitionPaths []string) ([]byte, error) {
```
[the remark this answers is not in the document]
constanca-m: We always have a parameter, so leaving it as optional leads to confusion.

test/packages/with-kind/kubernetes/_dev/deploy/k8s/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+resources:
```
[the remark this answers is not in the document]
constanca-m: Yes, we have one: `state_cronjob`.

test/packages/with-kind/kubernetes/data_stream/state_cronjob/fields/agent.yml
constanca-m: We can, but it was left as it is to be in sync with Kubernetes integration. Should I update it?

jsoriano (approved):
:+1:

PR #1128, opened 2023-02-01
Add variant for Kibana >= 8.7.0 with experimental toggles enabled

jsoriano:
Supersedes https://github.com/elastic/elastic-package/pull/1108.

It adds a new variant for versions of the stack between 8.2 and 8.6, and enables the experimental toggles for Kibana starting on 8.7.

With `elastic-package stack up --version 8.7.0-SNAPSHOT`:

Captura desde 2023-02-01 12-48-30

It adds technical debt but we don't have another way to do it till we templatize these files, as is attempted in https://github.com/elastic/elastic-package/pull/1055.


mrodm (approved):
LGTM

kpollich (approved):
Thanks!

Issue #1116, opened 2023-01-26
Run tests in CI with docker-compose v2


jsoriano:
CI is currently using docker compose v1, but docker compose v2 is better maintained and currently quite mature. [v2 is already the default in Docker for Mac](https://github.com/elastic/elastic-package/issues/1114#issuecomment-1404455407). Last version of docker-compose v1 was released almost two years ago ([v1.29.2](https://github.com/docker/compose/releases/tag/1.29.2)).

Current CI:
```
[2023-01-26T01:42:49.686Z] 2023/01/26 01:42:49 DEBUG running command: /usr/local/bin/docker-compose version --short
[2023-01-26T01:42:50.388Z] 2023/01/26 01:42:50 DEBUG Determined Docker Compose version: 1.25.5, the tool will use Compose V1
```

PR #1130, opened 2023-02-01
Add more stages to main buildkite pipeline

mrodm:
This PR adds into the Buildkite pipeline that migrates [this Jenkins pipeline](https://fleet-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/activity/) the following stages/steps:
- Release step (removed from Jenkinsfile)
- Integration tests added as a group of steps
    - Used same credentials as in Jenkins

Examples tested:
- [x] If unit tests or check static steps fail, integration tests are not run:
no integrations run
- [x] All tests failing are reported using annotation:
failing tests different steps
- [x] Tested re-triggering a building using a comment

Pending
- Coverage reports
- Upload artifacts safely


mrodm:
buildkite test this

.buildkite/scripts/release.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "--- fetching tags"
+# Ensure that tags are present so goreleaser can build the changelog from the last release.
+git rev-parse --is-shallow-repository
+git fetch origin --tags
+
+echo "--- running goreleaser"
+# Run latest version of goreleaser
+# TODO Remove check command once this stage is removed from Jenkinsfile
+# curl -sL https://git.io/goreleaser | bash
+curl -sL https://git.io/goreleaser | bash -s -- check
```
mrodm, line 14: As the release step is performed in Jenkins currently, this just performs a check.

Should we remove the release step from jenkins in this PR and enable it here ?

.buildkite/pipeline.yml
```diff
@@ -1,12 +1,20 @@
+env:
+  SETUP_GVM_VERSION: 'v0.5.0' # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
+  DOCKER_COMPOSE_VERSION: "1.25.5" # "v2.15.1"
```
mrodm, line 3: Using latest docker-compose version (v2.15.1) failed booting up the terraform scenarios (aws and gcp packages), with this error:

```
023/02/07 19:00:45 DEBUG running command: /opt/buildkite-agent/builds/bk-agent-elastic-gcp-1675796088139803659/elastic/elastic-package/bin/docker-compose -f /var/lib/buildkite-agent/.elastic-package/deployer/terraform/terraform-deployer.yml -f /opt/buildkite-agent/builds/bk-agent-elastic-gcp-1675796088139803659/elastic/elastic-package/test/packages/parallel/gcp/data_stream/compute/_dev/deploy/tf/env.yml -p elastic-package-service config
WARN[0000] The "TF_DIR" variable is not set. Defaulting to a blank string.
1 error(s) decoding:
 
* error decoding 'Volumes[0]': invalid spec: :/stage: empty section between colons
2023/02/07 19:00:45 DEBUG waiting for 1s before tearing down...
Error: error running package system tests: could not complete test run: could not setup service: could not get Docker Compose configuration for service: exit status 15
```

This error would be caused by the volume definition in this docker-compose file:
https://github.com/elastic/elastic-package/blob/main/internal/install/_static/terraform_deployer.yml#L15

.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -0,0 +1,79 @@
...
+    test-stack-command-7x
+    test-stack-command-86
+    test-stack-command-8x
+)
+
+for test in ${STACK_COMMAND_TESTS[@]}; do
+    echo "      - label: \":go: Running integration test: ${test}\""
+    echo "        command: ./.buildkite/scripts/integration_tests.sh -t ${test}"
+    echo "        agents:"
+    echo "          provider: \"gcp\""
```
mrodm, line 25: Provider gcp is required to use docker-like commands.

.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,52 @@
...
+        else
+            >&2 echo "Retry $count/$retries exited $exit, no more retries left."
+            return $exit
+        fi
+    done
+    return 0
+}
+
+GCP_SERVICE_ACCOUNT_SECRET_PATH=secret/ci/elastic-elastic-package/gcp-service-account
+AWS_SERVICE_ACCOUNT_SECRET_PATH=secret/ci/elastic-elastic-package/elastic-temp-aws-account-auth
```
mrodm, line 26: Should this be moved to the ci-shared path ?

This credential should be updated.
The advantage of moving it already to the shared path is that no other PR should be required to update this file.

If it is moved, what about this path `kv/ci-shared/platform-ingest/aws-account-auth` ?

.buildkite/pipeline.yml
jsoriano: I think this is because `docker-compose config` didn't require to replace the environment variables in 1.x and it does in 2.x.

I have a fix for this in https://github.com/elastic/elastic-package/pull/1140/files#diff-37d9cb502e75dd46add8cb1fe8a9701aab162575bc068c9891b85ec55a305336L58

But feel free to bring it here if you want to use v2 (and close https://github.com/elastic/elastic-package/issues/1116 if you do :slightly_smiling_face: )

.buildkite/scripts/release.sh
jsoriano: I guess we should move it to buildkite at some moment, so feel free to do it now if easy.

.buildkite/pipeline.yml
mrodm: I would prefer to keep this PR using the same docker-compose version as it runs in Jenkins, to ensure that there are no unexpected changes in behavior.

One note to remark about the tests run in docker-compose v2 is that is much more verbose and the buildkite output per step is truncated.

.buildkite/pipeline.yml
```diff
@@ -19,10 +27,31 @@ steps:
   - wait: ~
```
alexsapran, line 27: Why we are adding this wait step?

mrodm: This wait forces that the two steps defined previously (Run check static and Run unit tests) are finished before executing the following ones.

I wanted to force that way, so in case one of these two steps fail, integration tests are not run. Same behavior as in Jenkins.

.buildkite/scripts/integration_tests.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
```
mrodm, line 1: These files should have a License header ? 
If so, how can it be added ? using golicenser?

.buildkite/scripts/release.sh
mrodm: Ok, I'll remove the release step from Jenkins in favor of this one (removing check flag).

jlind23:
@mrodm I would rather remove the Jenkins job and create follow up issue for Buildkite pipeline otherwise I think we will keep both for a while.

jsoriano:
> in Jenkins, before running the pipeline, the default branch is merged into the changeset pushed, creating a new changeset that will be the one to be tested
> * in Buildkite, the changeset tested is the one pushed to the PR.

I think we want this, otherwise concurrent PRs can break main branches, specially on busy repositories. It is a pity that Buildkite doesn't offer it.

Maybe we can explore the use of Github merge queues (beta) for this https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue.
If I understand well, once a PR is queued, and before merging it, a temporary branch is created with the merge and the required checks executed.
This is enabled in the branch protection rules of a repository.
Captura desde 2023-02-14 10-37-16

We could give it a try.

> in Jenkins, there are coverage reports generated and added comments back to the PR. 
>  * in Buildkite, not able to migrate that feature for now.

Perhaps we can ignore this by now, and have an explicit task on next sprint to explore alternatives. I see this as a nice to have by now.

> Given that, I would keep the Jenkins workflow for now too, WDYT ?

I would only consider a blocker the first point, as this can affect development. For the reports I would be ok if we lose this by now, but have an specific plan to solve it soon.

mrodm:
Currently, testing for the first point this plugin:
https://github.com/seek-oss/github-merged-pr-buildkite-plugin

It would require to change some settings in the buildkite pipeline definition to disable 
`Skip pull requests builds for existing commits`. I think this setting would be important for those PRs created directly in upstream (dependabot).

mrodm:
Following the idea in https://github.com/seek-oss/github-merged-pr-buildkite-plugin, it has been added a new post-checkout hook in the repository (`.buildkite/hooks/post-checkout`) that allows us to have the same behavior as in Jenkins.
After checking out the repository, it performs these steps:
- Fetch latest changes from target branch (main) and checks out that branch
- Creates a new temporal branch based on target branch
- Merges the PR in that temporal branch

That new temporal branch with the merge commit is the one to be tested/run in each step.

This process is repeated for each step in the pipeline.

.buildkite/hooks/pre-command
mrodm: Moved this credential to the ci-shared path

.buildkite/scripts/integration_tests.sh
jsoriano: Yeah, I think they can be kept like this.

.buildkite/pipeline.yml
```diff
@@ -19,10 +27,31 @@ steps:
...
     continue_on_failure: true
 
-  - label: "Junit annotate"
+  - label: ":pipeline: Trigger Integration tests"
+    command: ".buildkite/pipeline.trigger.integration.tests.sh | buildkite-agent pipeline upload"
```
jsoriano, line 31: There is a bit of complexity here. To run integration tests we have this stack of scripts:
* `.buildkite/pipeline.yml`
* `pipeline.trigger.integration.tests.sh`
* `./.buildkite/scripts/integration_tests.sh`
* `make` targets
* `scripts/test-...`

There would be any option to reduce this complexity? Maybe we can remove the make targets and call directly the scripts? Though the make targets can be useful to reproduce locally with the same parameters.

Or most of the things `integration_tests.sh` does is to install things, could this be also done by providing an image with these dependencies?

Food for thought, no need to do anything about this on this PR.

.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,52 @@
...
+# Secrets must be redacted
+# https://buildkite.com/docs/pipelines/managing-log-output#redacted-environment-variables
+
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-package" && "$BUILDKITE_STEP_KEY" == "integration-parallel-gcp" ]]; then
+    export ELASTIC_PACKAGE_GCP_PROJECT_SECRET=$(retry 5 vault read -field projectId ${GCP_SERVICE_ACCOUNT_SECRET_PATH})
+    export ELASTIC_PACKAGE_GCP_CREDENTIALS_SECRET=$(retry 5 vault read -field credentials ${GCP_SERVICE_ACCOUNT_SECRET_PATH})
+
+    # Environment variables required by the service deployer
+    export GOOGLE_CREDENTIALS=${ELASTIC_PACKAGE_GCP_CREDENTIALS_SECRET}
+    export GCP_PROJECT_ID=${ELASTIC_PACKAGE_GCP_PROJECT_SECRET}
```
jsoriano, line 38: Should we set these [`redacted-vars`](https://buildkite.com/docs/agent/v3/configuration#redacted-vars) to hide these variables that doesn't match with the default patterns?

.buildkite/pipeline.yml
```diff
@@ -19,10 +27,31 @@ steps:
   - wait: ~
     continue_on_failure: true
 
-  - label: "Junit annotate"
+  - label: ":pipeline: Trigger Integration tests"
```
jsoriano, line 30: Should we remove integration tests from the Jenkinsfile?

jsoriano (approved):
Overall LGTM, I think that we are increasing the complexity with these buildkite pipelines based on bash, but I guess that we will get used to it as we migrate more things.

.buildkite/hooks/pre-command
mrodm: As the environment variables ELASTIC_PACKAGE_GCP_PROJECT_SECRET and ELASTIC_PACKAGE_GCP_CREDENTIALS_SECRET match the pattern of redacted variables, if other environment variables are built from there, they are also redacted.

I did some tests in this PR https://github.com/elastic/elastic-package/pull/1133
and here it is the output https://buildkite.com/elastic/elastic-package/builds/207#01865643-9d81-42bc-8010-971526776081

It looks like buildkite takes into account the string contained by the environment variables, and any occurrence is redacted.

.buildkite/pipeline.yml
mrodm: At first, in this PR I wanted to check that both Jenkins and Buildkite, had the same results.
I've created this follow-up PR to remove all those stages related to tests once this is merged
https://github.com/elastic/elastic-package/pull/1147

.buildkite/hooks/pre-command
jsoriano: Ah ok, nice then.

.buildkite/pipeline.yml
mrodm: That's true, there are more scripts than before with Jenkins. All that logic was defined inside the same groovy file (`.ci/Jenkinsfile`) i.e. triggering dynamically stages in Jenkins depending on the existing test folders, using specific commands in each one.

> Maybe we can remove the make targets and call directly the scripts? Though the make targets can be useful to reproduce locally with the same parameters.

As you mention, I would keep `make` targets since they are helpful to run and test everything locally.

> Or most of the things integration_tests.sh does is to install things, could this be also done by providing an image with these dependencies?

Those dependencies installed in `integration_tests.sh` probably they could be moved to a custom agent image. It would also make the builds faster since each step would not install everything.

PR #1152, opened 2023-02-21
Increase format version to 2.5.1 for new packages

ruflin:
New packages that were generated, used 2.0.0 by default instead of the most recent format_version of 2.5.1. To make sure new packages have by default access to the most recent features, the format_version should be the latest one.

### DIFF
--- a/internal/packages/archetype/package_manifest.go
+++ b/internal/packages/archetype/package_manifest.go
@@ -4,7 +4,7 @@
 
 package archetype
 
-const packageManifestTemplate = `format_version: 2.3.0
+const packageManifestTemplate = `format_version: 2.5.1
 name: {{.Manifest.Name}}
 title: "{{.Manifest.Title}}"
 version: {{.Manifest.Version}}
--- a/internal/packages/archetype/package_manifest.go
+++ b/internal/packages/archetype/package_manifest.go
@@ -4,7 +4,7 @@
 
 package archetype
 
-const packageManifestTemplate = `format_version: 2.0.0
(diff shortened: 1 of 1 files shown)


ruflin:
Based on https://github.com/elastic/package-spec/tree/v2.5.1 it looks like the package-spec version is already at v2.5.1? Should I update it to this version directly?

Is there a way we could this template up-to-date in an automatic way?

jsoriano:
> Based on https://github.com/elastic/package-spec/tree/v2.5.1 it looks like the package-spec version is already at v2.5.1? Should I update it to this version directly?

Sounds good.

> Is there a way we could this template up-to-date in an automatic way?

I see some options:
* Use the [`spec.VersionsInChangelog()`](https://github.com/elastic/package-spec/blob/ff8286d0c40ad76bb082e9c8ea78f4551c2519c1/spec.go#L49) method to find the most recent version, and use this one. This would be more self-contained, but this may create invalid packages if we introduce breaking changes.
* Similar one, try to obtain the same version as the package-spec in the `go.mod`.
* Create a CI job that triggers on package-spec releases, and opens a PR in `elastic-package` updating the version. This is probably the best option to avoid breaking things unexpectedly.

jsoriano:
Follow up issue created to automate the update https://github.com/elastic/elastic-package/issues/1155.

ruflin:
Thanks for the follow up issue. No strong preference on my end.

I update the PR to 2.5.1.

jsoriano (approved):

PR #1153, opened 2023-02-21
Add more skipped directories when copying package files

mrodm:
Fixes https://github.com/elastic/elastic-package/issues/1154

This PR adds the following folders to be considered skipped when building the package:
- `.git`
- `build`

This is needed when the package files are directly located in the root of the repository.

As it is now, `elastic-package build` command copies all the folders/files except `_dev`. As the `build` folder is created in the root of the repository (same as other package files) it is also copied to the `build` folder again. `.git` folder would be also copied along with the others.

### DIFF
--- a/internal/files/copy.go
+++ b/internal/files/copy.go
@@ -18,7 +18,7 @@ func CopyAll(sourcePath, destinationPath string) error {
 
 // CopyWithoutDev method copies files from the source to the destination, but skips _dev directories and empty folders.
 func CopyWithoutDev(sourcePath, destinationPath string) error {
-	return CopyWithSkipped(sourcePath, destinationPath, []string{"_dev"})
+	return CopyWithSkipped(sourcePath, destinationPath, []string{"_dev", "build", ".git"})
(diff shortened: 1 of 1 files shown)


jsoriano (approved):
:+1:

ruflin (approved):
LGTM. Did run on this branch and now everything works as expected. As side effect seems also to be that now the `build` command is faster which can be explained if less files have to be copoied.

One general issue that seems to persist is that it copies all files in my `.gitignore` files and even adds them to the package. For example my `.idea` files are in there which I didn't intend to share :-)

Lets get this in but I think it is a topic that needs further discussion as packages will accidentally contain files which users did not expect to be in there.

mrodm:
> One general issue that seems to persist is that it copies all files in my .gitignore files and even adds them to the package. For example my .idea files are in there which I didn't intend to share :-)

I've created a new issue for this https://github.com/elastic/elastic-package/issues/1156, feel free to update the description if needed @ruflin .

PR #1159, opened 2023-02-23
Update conditional for release step

mrodm:
This PR updates the filtering in the release step since in the latest attempt to push a tag, that step was not executed.
It looks like build.branch is set to the same value as the tag. At least this is what the environment variables show:
```
BUILDKITE_BRANCH="v0.75.0"
BUILDKITE_TAG="v0.75.0"
```

The condition now is just the value of the `build.tag`.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -43,10 +43,10 @@ pipeline {
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
           }
         }
-        /**
-        Check the source code.
-        */
-        stage('Check it contains Buildkite pipeline') {
+        stage('Check if the PR contains Buildkite pipeline') {
+          when {
+            changeRequest()
+          }
           steps {
             cleanup()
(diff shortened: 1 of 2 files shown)


.ci/Jenkinsfile
```diff
@@ -43,10 +43,10 @@ pipeline {
...
           }
         }
-        /**
-        Check the source code.
-        */
-        stage('Check it contains Buildkite pipeline') {
+        stage('Check if the PR contains Buildkite pipeline') {
+          when {
+            changeRequest()
+          }
```
mrodm, line 49: This stage is intended to check if the PRs are "old", it means if they are based on a commit previous to the buildkite pipeline.

I think it could be just run in case of PRs (`changeRequest()`). The last run when a tag was pushed failed https://fleet-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Felastic-package/detail/v0.75.0/1/pipeline/

jsoriano (approved):

Issue #1084, opened 2022-12-29
Add support to upload packages directly to Kibana/Fleet


ruflin:
In https://github.com/elastic/kibana/issues/70582 the Fleet team is working on adding an API to support uploading packages directly to Kibana. `elastic-package` should have a command available that allows to use this API. The initial idea I had was:

```
elastic-package upload --kibana=http://localhost:5601
```

In case this is run inside a package, the package is built and then uploaded to Kibana. Alternatively if a package zip file already exists, the following could be used:

```
elastic-package upload --path=/foo/bar/package.zip --kibana=http://localhost:5601
```

In this scenario, elastic-package assumes the package inside the zip file is already validated.

jsoriano:
We have the `elastic-package install` command to install packages, should we update this subcommand instead of adding a new one? Having two for the same thing may be confusing.

ruflin:
No strong opinion on my end around naming, the above is not necessarily a proposal on how we should name things and structure the params, much more to explain it.

For everyone curious on what the install command does: 

> The command uses Kibana API to install the package in Kibana. The package must be exposed via the Package Registry.

amitkanfer:
@juliaElastic is this related to https://github.com/elastic/kibana/issues/148599 ?

juliaElastic:
@amitkanfer Yes, this is a planned improvement in `elastic-package` to support the `upload` command. It is for convenience to call the Fleet API.

jsoriano:
Once we have this, https://github.com/elastic/elastic-package/issues/1157 can be a low hanging fruit. https://github.com/elastic/elastic-package/issues/1157 will help with development experience and reliability.

mrodm:
This issue could be done in different steps:
- Add a new `--zip` parameter into the install command that just installs the given zip file
    - https://github.com/elastic/elastic-package/pull/1160
- Update install command to build and use that zip to install (just if Elastic stack >= 8.7.0)

jlind23:
@mrodm If you have could you please think about recording a demo for this? 😄

PR #1160, opened 2023-02-23
Add parameter --zip into install command

mrodm:
Relates https://github.com/elastic/elastic-package/issues/1084

This PR adds a new parameter under install command `--zip`. This parameter requires a path to a zip file (this package must be built using `elastic-package build` command):
```
 $ elastic-package install -h
Use this command to install the package in Kibana.

The command uses Kibana API to install the package in Kibana. The package must be exposed via the Package Registry or built locally in zip format so they can be installed using --zip parameter. Zip packages can be installed directly in Kibana >= 8.7.0.

Context: package

Usage:
  elastic-package install [flags]

Flags:
  -c, --check-condition strings   check if the condition is met for the package, but don't install the package (e.g. kibana.version=7.10.0)
  -h, --help                      help for install
  -R, --root string               root directory of the package
  -z, --zip string                path to the zip package file (*.zip)

Global Flags:
  -v, --verbose   verbose mode
```

This parameter will just be available if the Elastic stack running (kibana) version is >= 8.7.

Example of usage:
- Using Elastic stack 8.7.0-SNAPSHOT
```
 $ elastic-package install --zip /home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.0.6.zip -v
2023/02/23 18:44:59 DEBUG Enable verbose logging
2023/02/23 18:44:59 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
Install zip package: /home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.0.6.zip
2023/02/23 18:44:59 DEBUG GET https://127.0.0.1:5601/api/status
Install the package
2023/02/23 18:44:59 DEBUG POST https://127.0.0.1:5601/api/fleet/epm/packages
Installed assets:
- elastic_package_registry-313c2700-099b-11ed-91b6-3b1f9c2b2771 (type: dashboard)
- metrics-elastic_package_registry.metrics-0.0.6 (type: ingest_pipeline)
- metrics-elastic_package_registry.metrics (type: index_template)
- metrics-elastic_package_registry.metrics@package (type: component_template)
- metrics-elastic_package_registry.metrics@custom (type: component_template)
Done
```
- Using Elastic stack 8.6.0
```
 $ elastic-package install --zip /home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.0.6.zip -v
2023/02/24 10:35:10 DEBUG Enable verbose logging
2023/02/24 10:35:10 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
Install zip package: /home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.0.6.zip
2023/02/24 10:35:10 DEBUG GET https://127.0.0.1:5601/api/status
Error: can't create the package installer: not supported uploading zip packages in Kibana 8.6.0
```
- Set `check-condition` parameter with zip file:
```
 $ elastic-package install --zip build/packages/package_storage_candidate-0.0.1.zip -v --check-condition kibana.version=7.10.0
2023/02/27 17:08:33 DEBUG Enable verbose logging
2023/02/27 17:08:33 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
2023/02/27 17:08:33 DEBUG Reading package manifest from build/packages/package_storage_candidate-0.0.1.zip
Check conditions for package
2023/02/27 17:08:33 DEBUG Verify Kibana version (constraint: ^8.0.0, requirement: 7.10.0)
Error: checking conditions failed: Kibana constraint unsatisfied: [0] 7.10.0 is less than 8.0.0
```

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -26,11 +26,20 @@ func setupInstallCommand() *cobraext.Command {
 	}
 	cmd.Flags().StringSliceP(cobraext.CheckConditionFlagName, "c", nil, cobraext.CheckConditionFlagDescription)
 	cmd.Flags().StringP(cobraext.PackageRootFlagName, cobraext.PackageRootFlagShorthand, "", cobraext.PackageRootFlagDescription)
+	cmd.Flags().StringP(cobraext.InstallZipPackageFlagName, "", "", cobraext.InstallZipPackageFlagDescription)
 
 	return cobraext.NewCommand(cmd, cobraext.ContextPackage)
 }
 
 func installCommandAction(cmd *cobra.Command, _ []string) error {
+	zipPathFile, err := cmd.Flags().GetString(cobraext.InstallZipPackageFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.InstallZipPackageFlagName)
+	}
+	if zipPathFile != "" {
+		return installZipPackage(cmd, zipPathFile)
+	}
+
 	packageRootPath, err := cmd.Flags().GetString(cobraext.PackageRootFlagName)
 	if err != nil {
 		return cobraext.FlagParsingError(err, cobraext.PackageRootFlagName)
@@ -67,6 +76,10 @@ func installCommandAction(cmd *cobra.Command, _ []string) error {
 		return nil
 	}
 
+	return installLocalPackage(cmd, m)
+}
+
+func installLocalPackage(cmd *cobra.Command, m *packages.PackageManifest) error {
 	packageInstaller, err := installer.CreateForManifest(*m)
 	if err != nil {
 		return errors.Wrap(err, "can't create the package installer")
@@ -83,6 +96,13 @@ func installCommandAction(cmd *cobra.Command, _ []string) error {
 	for _, asset := range installedPackage.Assets {
 		cmd.Printf("- %s (type: %s)\n", asset.ID, asset.Type)
 	}
+	cmd.Println("Done")
+	return nil
+}
+
+func installZipPackage(cmd *cobra.Command, zipPath string) error {
+	cmd.Printf("Install zip package: %s\n", zipPath)
+
 	cmd.Println("Done")
 	return nil
 }
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -85,15 +85,7 @@ func installLocalPackage(cmd *cobra.Command, m *packages.PackageManifest) error 
 		return errors.Wrap(err, "can't create the package installer")
 	}
 
-	cmd.Println("Install the package")
-	installedPackage, err := packageInstaller.Install()
-	if err != nil {
-		return errors.Wrap(err, "can't install the package")
-	}
-
-	showAssets(cmd, installedPackage)
-	cmd.Println("Done")
-	return nil
+	return installPackage(cmd, packageInstaller)
 }
 
 func installZipPackage(cmd *cobra.Command, zipPath string) error {
@@ -104,20 +96,20 @@ func installZipPackage(cmd *cobra.Command, zipPath string) error {
 		return errors.Wrap(err, "can't create the package installer")
 	}
 
+	return installPackage(cmd, packageInstaller)
+}
+
+func installPackage(cmd *cobra.Command, packageInstaller installer.Installer) error {
 	cmd.Println("Install the package")
 	installedPackage, err := packageInstaller.Install()
 	if err != nil {
 		return errors.Wrap(err, "can't install the package")
 	}
 
-	showAssets(cmd, installedPackage)
-	cmd.Println("Done")
-	return nil
-}
-
-func showAssets(cmd *cobra.Command, installedPackage *installer.InstalledPackage) {
 	cmd.Println("Installed assets:")
 	for _, asset := range installedPackage.Assets {
 		cmd.Printf("- %s (type: %s)\n", asset.ID, asset.Type)
 	}
+	cmd.Println("Done")
+	return nil
 }
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -92,10 +92,7 @@ func installLocalPackage(cmd *cobra.Command, m *packages.PackageManifest) error 
 		return errors.Wrap(err, "can't install the package")
 	}
 
-	cmd.Println("Installed assets:")
-	for _, asset := range installedPackage.Assets {
-		cmd.Printf("- %s (type: %s)\n", asset.ID, asset.Type)
-	}
+	showAssets(cmd, installedPackage)
 	cmd.Println("Done")
 	return nil
 }
@@ -103,6 +100,24 @@ func installLocalPackage(cmd *cobra.Command, m *packages.PackageManifest) error 
 func installZipPackage(cmd *cobra.Command, zipPath string) error {
 	cmd.Printf("Install zip package: %s\n", zipPath)
 
+	packageInstaller, err := installer.CreateForZip(zipPath)
+	if err != nil {
+		return errors.Wrap(err, "can't create the package installer")
+	}
+
+	installedPackage, err := packageInstaller.Install()
+	if err != nil {
+		return errors.Wrap(err, "can't install the package")
+	}
+
+	showAssets(cmd, installedPackage)
 	cmd.Println("Done")
 	return nil
+}
+
+func showAssets(cmd *cobra.Command, installedPackage *installer.InstalledPackage) {
+	cmd.Println("Installed assets:")
+	for _, asset := range installedPackage.Assets {
+		cmd.Printf("- %s (type: %s)\n", asset.ID, asset.Type)
+	}
 }
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -74,12 +74,12 @@ func installCommandAction(cmd *cobra.Command, _ []string) error {
 		return nil
 	}
 
-	return aInstaller.install(cmd, manifest)
+	return aInstaller.install(cmd, manifest.Name, manifest.Version)
 }
 
 type packageInstaller interface {
 	manifest() (*packages.PackageManifest, error)
-	install(cmd *cobra.Command, manifest *packages.PackageManifest) error
+	install(cmd *cobra.Command, name, version string) error
 }
 
 func newInstaller(zipPath, packageRootPath string) packageInstaller {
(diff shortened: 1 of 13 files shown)


internal/packages/installer/zip_installer.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return nil, err
+	}
+	v, err := semver.NewVersion(kibanaVersion.Number)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Kibana version")
+	}
+	if v.LessThan(semver8_7_0) {
+		return nil, fmt.Errorf("not supported uploading zip packages in Kibana %s", kibanaVersion.Number)
+	}
```
mrodm, line 42: Probably, this check about the Kibana version needs to be moved into `cmd/install.go`. This would be needed to check the version to start using this method by default when running `elastic-package install` in a package directory in case Kibana >= 8.7.0.

internal/packages/installer/zip_installer.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return &zipInstaller{
+		zipPath:      zipPath,
+		kibanaClient: kibanaClient,
+		manifest:     m,
```
mrodm, line 47: This manifest should be kept since it is used by the assets test runner.

internal/packages/packages.go
```diff
@@ -222,6 +225,68 @@ func ReadPackageManifestFromPackageRoot(packageRoot string) (*PackageManifest, e
 	return ReadPackageManifest(filepath.Join(packageRoot, PackageManifestFile))
 }
 
+// ReadPackageManifestFromZipPackage reads and parses the package manifest file for the given zip package.
+func ReadPackageManifestFromZipPackage(zipPackage string) (*PackageManifest, error) {
```
mrodm, line 229: New method to read the Package manifest from a zip file. Needed for two reasons when using a zip file:
- InstalledPackage requires a packages.PackageManifest that it us used in the assets test runner
- `check-condition` parameter requires also the manifest to be able to check the constraints.

internal/kibana/client.go
```diff
@@ -81,22 +83,26 @@ func CertificateAuthority(certificateAuthority string) ClientOption {
...
+	return c.sendRequest(http.MethodGet, resourcePath, nil, map[string]string{"content-type": defaultContentType})
 }
 
 func (c *Client) post(resourcePath string, body []byte) (int, []byte, error) {
-	return c.sendRequest(http.MethodPost, resourcePath, body)
+	return c.postWithContentType(resourcePath, defaultContentType, body)
+}
+
+func (c *Client) postWithContentType(resourcePath, contentTypeHeader string, body []byte) (int, []byte, error) {
+	return c.sendRequest(http.MethodPost, resourcePath, body, map[string]string{"content-type": contentTypeHeader})
```
mrodm, line 94: Any other suggestion to add here the custom content-type header ? @jsoriano 
In case of uploading/installing a zip package that header must be `application/zip`. This header is set [here](https://github.com/elastic/elastic-package/pull/1160/files#diff-5957946cb38e2c6b7a662962ed2a0de756f43e76794e775114ab48f9b65a0e65R44)

kpollich:
cc @ruflin since you were originally inquiring about this behavior 🙂 

This is looking great, @mrodm - let's definitely make a note to send this out to the `elastic-packages` mailing group once it's finalized.

ruflin:
Is there any way to configure a custom Kibana instance? Does not have to be part of this PR but this will become key. Running the stack with elastic-package is for development purpose. But the zip command can be used for any package you have on your disk and push it your Kibana instance for usage.

I expect `--zip` to eventually become the default instead of having to go through the registry and registry is the optional one. Don't know how we will make this change but it hopefully means at some point `--zip` falls away so uploading a package to kibana becomes even simpler / less params.

jsoriano:
> Is there any way to configure a custom Kibana instance?

I guess that `ELASTIC_PACKAGE_KIBANA_HOST` and related environment variables can be used by now.

ruflin:
> I guess that ELASTIC_PACKAGE_KIBANA_HOST and related environment variables can be used by now.

Sounds good, but lets make sure we also test and document it. We eventually have to follow up with "convenience" API I think.

cmd/install.go
```diff
@@ -45,10 +51,19 @@ func installCommandAction(cmd *cobra.Command, _ []string) error {
 			return errors.Wrap(err, "locating package root failed")
 		}
 	}
-
-	m, err := packages.ReadPackageManifestFromPackageRoot(packageRootPath)
-	if err != nil {
-		return errors.Wrapf(err, "reading package manifest failed (path: %s)", packageRootPath)
+	var manifest *packages.PackageManifest
+	if zipPathFile != "" {
```
jsoriano, line 55: Nit. Consider having some kind of factory that provides an object with the methods for each kind of package to install. Or two different methods depending on the method to use. So we avoid these ifs here.
We can also refactor this in a follow up.

scripts/test-install-zip.sh
```diff
@@ -0,0 +1,70 @@
...
+cd $OLDPWD
+
+# Remove unzipped built packages, leave .zip files
+rm -r build/packages/*/
+
+# Install packages and verify
+for zipFile in build/packages/*.zip; do
+  PACKAGE_NAME_VERSION=$(basename ${zipFile} .zip)
+
+  # check that the package is installed
```
jsoriano, line 61: Do this comment belong to the curl below?

internal/packages/installer/installer.go
```diff
@@ -25,20 +30,20 @@ type InstalledPackage struct {
 }
 
 // CreateForManifest function creates a new instance of the installer.
-func CreateForManifest(manifest packages.PackageManifest) (*Installer, error) {
+func CreateForManifest(manifest packages.PackageManifest) (Installer, error) {
```
jsoriano, line 33: Actually this installer doesn't need the whole manifest, right? Only the name and the version.

jsoriano (commented):
Looks good, but I think we should reorganize a bit the code around installers. Responsibilities and interfaces look a bit mixed up.

Added some comments and suggestions.

scripts/test-install-zip.sh
mrodm: Yes, moving it to where it should be, thanks!

internal/packages/installer/zip_installer.go
```diff
@@ -0,0 +1,71 @@
...
+
+type zipInstaller struct {
+	zipPath  string
+	manifest packages.PackageManifest
+
+	kibanaClient *kibana.Client
+}
+
+// CreateForZip function creates a new instance of the installer.
+func CreateForZip(zipPath string, m packages.PackageManifest) (Installer, error) {
```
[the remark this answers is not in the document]
mrodm: Ok ! I'll change both CreateForZip and CreateForManifest functions.
Thanks!

internal/packages/packages.go
```diff
@@ -222,6 +225,68 @@ func ReadPackageManifestFromPackageRoot(packageRoot string) (*PackageManifest, e
...
+		0644,
+	)
+	if err != nil {
+		return err
+	}
+	defer outputFile.Close()
+
+	// elastic-package build command creates a zip that contains all the package files
+	// under a folder named "package-version". Example elastic_package_registry-0.0.6/manifest.yml
+	zipFilePath := fmt.Sprintf("*/%s", sourcePath)
```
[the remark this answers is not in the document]
mrodm: If it is used `fs.Glob` here, later it is still need to look for the specific file inside the `zipReader.File` (of type []*File). Not sure if I'm missing another way of implementing this.

internal/packages/installer/zip_installer.go
mrodm: I wanted to avoid changing the signature of this method (or CreateForManifest) since it is used in testrunners too (assets). But I could take a look at it.

cmd/install.go
mrodm: Added factory, now this method looks like much simpler, thanks!

internal/packages/installer/installer.go
mrodm: Updated both methods `CreateForManifest` and `CreateForZip` to accept name and version instead of the manifest.

internal/kibana/packages.go
```diff
@@ -27,9 +28,32 @@ func (c *Client) InstallPackage(pkg packages.PackageManifest) ([]packages.Asset,
...
+// InstallZipPackage installs the local zip package in Fleet.
+func (c *Client) InstallZipPackage(zipFile string) ([]packages.Asset, error) {
+	path := fmt.Sprintf("%s/epm/packages", FleetAPI)
+
+	fileContents, err := os.ReadFile(zipFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read zip file")
+	}
+
+	req, err := c.newRequest(http.MethodPost, path, fileContents)
```
jsoriano, line 40: It'd be probably better for `newRequest` to receive a reader as body. With `ReadFile` we have to put the whole package in memory. We don't have so big packages now, but we might have soon.
```suggestion
	body, err := os.Open(zipFile)
	if err != nil {
		return nil, errors.Wrap(err, "failed to read zip file")
	}
	defer body.Close() 

	req, err := c.newRequest(http.MethodPost, path, body)
```

`sendRequest` could keep its interface if we don't want to modify the rest of methods, by wrapping the body with a `bytes.NewReader`.
```
func (c *Client) sendRequest(method, resourcePath string, body []byte) (int, []byte, error) {
	request, err := c.newRequest(method, resourcePath, bytes.NewReader(body))
	...
```

internal/kibana/packages.go
```diff
@@ -27,9 +28,32 @@ func (c *Client) InstallPackage(pkg packages.PackageManifest) ([]packages.Asset,
...
 }
 
+// InstallZipPackage installs the local zip package in Fleet.
+func (c *Client) InstallZipPackage(zipFile string) ([]packages.Asset, error) {
+	path := fmt.Sprintf("%s/epm/packages", FleetAPI)
+
+	body, err := os.Open(zipFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read zip file")
+	}
```
jsoriano, line 38: We should close this body.
```suggestion
	}
	defer body.Close()
```

jsoriano (approved):

Issue #1100, opened 2023-01-18
elastic-package starts but does nothing Windows 11


joeperuzzi:
Installing the elastic-package through go and/or running it stand-alone pre-compiled takes CPU cycles but never moves.  No matter what switches are used, the executable doesn't seem to do anything but take CPU cycles.  I've let this run for 20 minutes and nothing happens other than CPU usage, i've disabled all AV (running Elastic Endpoint) as it was causing the Elastic Security process to consume all the CPU and crash explorer.

Currently Running on Windows 11 Pro 10.0.22621 on an Intel i7.  

<img width="926" alt="image" src="https://user-images.githubusercontent.com/122561341/213224426-2181020e-3f73-48c1-8fda-d70736388dce.png">
<img width="467" alt="image" src="https://user-images.githubusercontent.com/122561341/213252499-0bdc462c-5b82-4fc8-8976-b638d99fd253.png">

joeperuzzi:
I executed "go install github.com/elastic/elastic-package@latest" yesterday, the executable will not run to give me a version and the exe properties that is generated from that command does not have a version in its properties.

jsoriano:
@joeperuzzi can you try with one of the released versions? https://github.com/elastic/elastic-package/releases

joeperuzzi:
I tried both 0.70 and 0.71 both have the same outcome as above.

joeperuzzi:
Just as an update, downloaded and tested  v0.74.0 and the issue still exists.

joeperuzzi:
Just as an update, downloaded and tested v0.75.0 and the issue persists 

<img width="672" alt="image" src="https://user-images.githubusercontent.com/122561341/222216622-5a8657f7-79e1-4589-9226-83d3e16f9897.png">

jsoriano:
@joeperuzzi thanks. We haven't been able to reproduce it, so by now you are our only hope to catch this issue :slightly_smiling_face:

Could you try to investigate the process with other tools, like [Process Monitor](https://learn.microsoft.com/en-us/sysinternals/downloads/procmon), to see if you can see something else?

Would you have the chance to try to run another Go program, even if it is just a hello world, to see if it is an issue with Go runtime?

joeperuzzi:
<img width="1128" alt="image" src="https://user-images.githubusercontent.com/122561341/222589694-8bbdc4e6-30f8-4345-a6bb-71224ba2f84c.png">
Seems to be trying to write the c:\.git file and is failing...

[...]

PR #1171, opened 2023-03-03
Fix search of repository root in Windows

jsoriano:
When looking for the repository root in Windows, if the command hasn't been called inside a git repository, the search will loop forever, hanging the process. Same thing happens when looking for package root.

Finalization condition is based on checking all parent directories till root, but root was checked with "/", that will never be the case in Windows.

Changed to take into account the volume name and the OS-specific path separator.

Add step to buildkite pipeline to run unit tests on Windows.

Fixes https://github.com/elastic/elastic-package/issues/1100.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -41,6 +41,8 @@ steps:
         allow_failure: false
       - step: unit-tests-linux
         allow_failure: false
+      - step: unit-tests-windows
+        allow_failure: false
 
   - wait: ~
     continue_on_failure: true
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -26,12 +26,10 @@ steps:
 
   - label: ":go: :windows: Run unit tests"
     key: unit-tests-windows
-    command: "go test ./..."
+    command: ".buildkite/scripts/unit_tests_windows.ps1"
     agents:
-      windows: true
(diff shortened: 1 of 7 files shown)


.buildkite/pipeline.yml
```diff
@@ -24,6 +24,13 @@ steps:
       cpu: "8"
       memory: "4G"
 
+  - label: ":go: :windows: Run unit tests"
+    key: unit-tests-windows
+    command: ".buildkite/scripts/unit_tests_windows.ps1"
+    agents:
+      provider: "gcp"
+      image: "family/ci-windows-2022"
```
mrodm, line 32: Great to have this running also in Windows :+1: 

I would also add this step as a dependency of the step related to integration tests:

```yaml
  - label: ":pipeline: Trigger Integration tests"
    command: ".buildkite/pipeline.trigger.integration.tests.sh | buildkite-agent pipeline upload"
    depends_on:
      - step: check-static
        allow_failure: false
      - step: unit-tests-linux
        allow_failure: false
      - step: unit-tests-windows
        allow_failure: false
```
Doing so, if the unit tests in windows fail, integration tests won't be triggered.

mrodm (commented):
Thanks for adding the step for Windows :+1:

.buildkite/pipeline.yml
jsoriano: Added, thanks :+1:

mrodm (approved):

PR #1172, opened 2023-03-06
Remove notification of the build result in Jenkins

mrodm:
As this pipeline has been migrated to Buildkite, the notification related to build result should be removed.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -67,11 +67,6 @@ pipeline {
       }
     }
   }
-  post {
-    cleanup {
-      notifyBuildResult(prComment: true)
-    }
-  }
 }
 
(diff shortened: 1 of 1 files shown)


mrodm:
> Can we have something similar later with buildkite?

At least with the notifications available in buildkite, there is no option to add a comment with the build status. https://buildkite.com/docs/pipelines/notifications

jsoriano:
> > Can we have something similar later with buildkite?
> 
> At least with the notifications available in buildkite, there is no option to add a comment with the build status. https://buildkite.com/docs/pipelines/notifications

Maybe with [webhooks](https://buildkite.com/docs/pipelines/notifications#webhooks)? :thinking: But it doesn't look very customizable.

PR #1181, opened 2023-03-13
Reference install package howto guide

aspacca:
Let's reference the `elastic-package install` howto guide in the `elastic-package benchmark generate-corpus` one


mrodm (approved):
LGTM!

Issue #1210, opened 2023-04-06
Allow to override GeoLite2 database per profile in compose stack provider


jsoriano:
Before #1140, users could manually modify the stack files to use a real GeoLite database when visually testing integrations.

Add some way to configure this per profile.

PR #1211, opened 2023-04-06
Add setting to configure a directory with geoip databases

jsoriano:
Proposal to solve https://github.com/elastic/elastic-package/issues/1210, and provide an initial framework to have persistent settings per profile.

With this change, to replace GeoIP databases a user needs to follow these steps:
* Create a directory with the GeoIP databases they want to use.
* Create a new profile with `elastic-package profiles create`.
* In the directory of the new profile, create a `config.yml` file with `stack.geoip_dir: "/path/to/geoip_dir"`.
* Use this profile when starting the stack (by running first `elastic-packages profile use`, by using the `--profile` flag of `elastic-package stack up`, or by exporting the `ELASTIC_PACKAGE_PROFILE` environment variable).

Fixes https://github.com/elastic/elastic-package/issues/1210.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ Use this command to add, remove, and manage multiple config profiles.
 	
 Individual user profiles appear in ~/.elastic-package/stack, and contain all the config files needed by the "stack" subcommand. 
 Once a new profile is created, it can be specified with the -p flag, or the ELASTIC_PACKAGE_PROFILE environment variable.
-User profiles are not overwritten on upgrade of elastic-stack, and can be freely modified to allow for different stack configs.
+User profiles can be configured with a "config.yml" file in the profile directory.
 
 ### `elastic-package profiles create`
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -474,6 +474,26 @@ Use this command to print the version of elastic-package that you have installed
...
+You can find the profiles in your system with `elastic-package profiles list`.
+
+You can delete profiles with `elastic-package profiles delete`.
+
+Each profile can have a `config.yml` file that allows to persist configuration settings
+that apply only to commands using this profile.
+
+The following settings are available per profile:
+
+* `stack.geoip_dir` defines a directory with GeoIP databases that can be used by
```
efd6, line 494: It's my understanding from this that the value should point to a directory outside the profile dir. Is that correct? If it is, it probably should be noted here.

README.md
```diff
@@ -474,6 +474,26 @@ Use this command to print the version of elastic-package that you have installed
...
+`elastic-package` uses the "default" profile. Other profiles can be created with the
+`elastic-package profiles create` command. Once a profile is created, it will have its
+own directory inside the elastic-package data directory. Once you have more profiles,
+you can change the default with `elastic-package profiles use`.
+
+You can find the profiles in your system with `elastic-package profiles list`.
+
+You can delete profiles with `elastic-package profiles delete`.
+
+Each profile can have a `config.yml` file that allows to persist configuration settings
```
efd6, line 489: Ergonomically, it would be nice for this to be pre-populated into the directory when the profile is created, probably also with a comment indicating that it persists and may be edited.

README.md
jsoriano: Not mandatory, but definitely recommended, added a line about this. Thanks.

README.md
jsoriano: I have added a `config.yml.example` that users can copy when creating their configurations.

efd6 (approved):
Thanks

mrodm (approved):
:rocket:

PR #1225, opened 2023-04-19
Update Go runtime to 1.20.3

jsoriano:


mrodm (approved):
:+1: 
There is another go.mod file using 1.19 for the script to trigger jenkins builds (.buildkite/scripts/triggerJenkinsJob/go.mod), but this could be updated in another PR I guess.

PR #1226, opened 2023-04-20
Add catalog information to define pipelines as a Backstage Resource

mrodm:
This PR creates the catalog information for the BackStage internal system locally in this repository.
This catalog information includes the same pipelines definitions that were already defined previously for Buildkite. Specifically these two pipelines:
- .buildkite/pipeline.yml
- .buildkite/pipeline.package-storage-publish.yml

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -21,6 +21,7 @@ spec:
       description: ':go: Build and test the elastic-package project'
     spec:
       branch_configuration: "main v0.* v1.*"
+      pipeline_file: ".buildkite/pipeline.yml"
       provider_settings:
         build_pull_request_forks: false
         build_tags: true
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -0,0 +1,79 @@
...
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: elastic-package
+      description: ':go: Build and test the elastic-package project'
+    spec:
+      branch_configuration: "main v0.* v1.*"
+      pipeline_file: ".buildkite/pipeline.yml"
+      provider_settings:
+        build_pull_request_forks: false
```
alexsapran, line 26: should we have this set to true?

alexsapran (approved):

catalog-info.yaml
mrodm: If it is set to true, pull requests created by external contributors would run the pipelines without any check.

Builds from pull requests are created through the [buildkite-pr-bot](https://github.com/elastic/buildkite-pr-bot): https://github.com/elastic/elastic-package/blob/main/.buildkite/pull-requests.json

This bot needs to be set in each repository 
https://docs.elastic.dev/ci/configuring-pull-requests-for-buildkite

jsoriano (approved):

PR #1231, opened 2023-04-24
Add serverless flag to stack up

ruflin:
The idea behind this pull request is to be able to run a `serverless` stack of Elasticsearch with the following command:

```
elastic-package stack up --version=8.8.0-SNAPSHOT -v --serverless
```

I'm opening this PR as draft as I have a few questions on how to implement it best.

### DIFF
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -73,6 +73,11 @@ func setupStackCommand() *cobraext.Command {
 				return cobraext.FlagParsingError(err, cobraext.StackVersionFlagName)
 			}
 
+			/*serverless, err := cmd.Flags().GetString(cobraext.ServerlessFlagName)
+			if err != nil {
+				return cobraext.FlagParsingError(err, cobraext.ServerlessFlagName)
+			}*/
+
 			profile, err := getProfileFlag(cmd)
 			if err != nil {
 				return err
--- a/internal/stack/resources.go
+++ b/internal/stack/resources.go
@@ -47,6 +47,8 @@ const (
 
 	elasticsearchUsername = "elastic"
 	elasticsearchPassword = "changeme"
+
+	serverless = "false"
 )
 
 var (
@@ -117,6 +119,8 @@ func applyResources(profile *profile.Profile, stackVersion string) error {
 
 		"username": elasticsearchUsername,
 		"password": elasticsearchPassword,
+
+		"serverless": serverless,
 
 		"geoip_dir": profile.Config("stack.geoip_dir", "./ingest-geoip"),
 	})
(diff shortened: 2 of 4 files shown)


cmd/stack.go
```diff
@@ -73,6 +73,11 @@ func setupStackCommand() *cobraext.Command {
 				return cobraext.FlagParsingError(err, cobraext.StackVersionFlagName)
 			}
 
+			/*serverless, err := cmd.Flags().GetString(cobraext.ServerlessFlagName)
```
ruflin, line 76: @mrodm @jsoriano Any recommendations on the path to follow here. I'm looking at geoip_dir which is part of a profile. At the same time I don't know if making it part of profiles is overkill? 

As you see further down, it is a variable in the template loaded from the resources. How do I set the resource variable on elastic-stack up in the best way? https://github.com/elastic/elastic-package/pull/1231/files#diff-9fd733fc9d8f10a5c2dd55a062523d3dcb0655bc1db097648b3adabbd3460173R123

internal/stack/resources.go
```diff
@@ -47,6 +47,8 @@ const (
 
 	elasticsearchUsername = "elastic"
 	elasticsearchPassword = "changeme"
+
+	serverless = "false"
```
ruflin, line 51: Setting this to true manually leads to the expected result.

cmd/stack.go
jsoriano: Hey, thanks for adding this. I would add it as a profile setting, in the same way as `geoip_dir`. Maybe in the future we can add a flag to pass this kind of settings through the command line, so we can run commands like `elastic-package stack up -Dstack.serverless=true`.

ruflin: I assume this would mean it can only be used by someone that adds a custom profile to elastic-package? The part I'm worried with this is that the feature will be tricky to use. I want to make it as easy as possible for everyone to test serverless by just adding `--serverless` like other params.

ruflin: If my understand is correct, the path forward is that we have multiple profiles available by default meaning users of elastic-package are likely already familiar by then with profiles. With this in mind, what you suggest makes a lot of sense. So if I want to switch between setups, I just switch the profiles.

jsoriano: We don't ship predefined profiles (apart of the default one), but this would be definitely an option once we have more options.

ruflin:
For reference as this is related: https://github.com/elastic/kibana/pull/155009 I assume we will not only start Elasticsearch in serverless but also Kibana.

sharbuz:
Hi everyone,

ruflin:
Currently closing this as I didn't find the time to work on this. @jsoriano I still think this would be useful.

Issue #1164, opened 2023-02-28
Add end-to-end benchmark command


marc-gr:
Expand the `benchmark` command to add end-to-end capabilities in addition to pipeline ones.

Having a working cluster and a set of agent hosts enrolled, we want to run benchmark scenarios for a given package version and data stream.

Requirements:

- A spec to define benchmark scenarios https://github.com/elastic/package-spec/issues/483

Scope:

- Set up the custom policy
- Start the benchmark services containing the required tooling to deliver the logs
- Run these services
- Collect results
- Stop benchmark services

Not in scope for the command:

- Manage the required infrastructure - it expects a working fleet setup
- ~Generate corpora data on-the-fly - it needs to be pre-generated~

aspacca:
> * Generate corpora data on-the-fly - it needs to be pre-generated

for generating data on-the-fly we are working on https://github.com/elastic/elastic-package/issues/1110

currently in a separated command, I'll move as subcommand of `benchmark`

PR #1232, opened 2023-04-24
Adds system benchmarks command

marc-gr:
This adds the ability to define system (end to end) benchmarks at the package level.

- [x] Refactor `benchrunner` to decouple reports presentation from pipeline benchmarks internals
- [x] Add `benchmark system` command
- [x] Collect indices, disk and ingest metrics from all nodes
- [x] Implement docker input services
- [x] Ability to generate data on the fly using the corpus generator
- [x] Report to STDOUT a summary of the benchmark metrics
- [x]  Report to a file  
- [x] Add documentation
- [x] Add test

Depends on https://github.com/elastic/package-spec/pull/512
Closes https://github.com/elastic/elastic-package/issues/1164

TODO in the future(in no specific order)

- Collect metrics from elastic agents with system package
- Seed generator for consistent data
- Send benchmark runtime metrics to the ES Metricstore
- Add command to compare two benchmark runs
- Support more input service types (like system tests)
- Support benchmark scenarios defined outside of the package scope (for easily running scheduled benchmarks)
- Create benchmarks dashboard (maybe through a benchmark integration)

Example of usage:

With a benchmark scenario defined as

`<package root>/_dev/benchmark/system/100mb-logs-benchmark.yml`
```yml
---
description: Benchmark 100MiB of data ingested
input: filestream
vars: ~
data_stream.name: test
data_stream.vars.paths:
  - "{{SERVICE_LOGS_DIR}}/corpus-*"
warmup_time_period: 10s
corpora.generator.size: 100MiB
corpora.generator.template.path: ./100mb-logs-benchmark/template.log
corpora.generator.config.path: ./100mb-logs-benchmark/config.yml
corpora.generator.fields.path: ./100mb-logs-benchmark/fields.yml
```

We then run:

```
elastic-package benchmark system --benchmark 100mb-logs-benchmark -v
```

To generate a benchmark report:

```
--- Benchmark results for package: system_benchmarks - START ---
╭─────────────────────────────────────────────────────╮
│ info                                                │
├──────────────┬──────────────────────────────────────┤
│ benchmark    │                 100mb-logs-benchmark │
│ description  │    Benchmark 100MiB of data ingested │
│ run ID       │ d2960c04-0028-42c9-bafc-35e599563cb1 │
│ package      │                    system_benchmarks │
│ start ts (s) │                           1682320355 │
│ end ts (s)   │                           1682320355 │
│ duration     │                                 2m3s │
╰──────────────┴──────────────────────────────────────╯
╭───────────────────────────────────────────────────────────────────────╮
│ parameters                                                            │
├─────────────────────────────────┬─────────────────────────────────────┤
│ package version                 │                         999.999.999 │
│ input                           │                          filestream │
│ data_stream.name                │                                test │
│ data_stream.vars.paths          │        [/tmp/service_logs/corpus-*] │
│ warmup time period              │                                 10s │
│ benchmark time period           │                                  0s │
│ wait for data timeout           │                                  0s │
│ corpora.generator.size          │                              100MiB │
│ corpora.generator.template.path │ ./100mb-logs-benchmark/template.log │
│ corpora.generator.template.raw  │                                     │
│ corpora.generator.template.type │                                     │
│ corpora.generator.config.path   │   ./100mb-logs-benchmark/config.yml │
│ corpora.generator.config.raw    │                               map[] │
│ corpora.generator.fields.path   │   ./100mb-logs-benchmark/fields.yml │
│ corpora.generator.fields.raw    │                               map[] │
╰─────────────────────────────────┴─────────────────────────────────────╯
╭───────────────────────╮
│ cluster info          │
├───────┬───────────────┤
│ name  │ elasticsearch │
│ nodes │             1 │
╰───────┴───────────────╯
╭─────────────────────────────────────────────────────────────╮
│ data stream stats                                           │
├────────────────────────────┬────────────────────────────────┤
│ data stream                │ logs-system_benchmarks.test-ep │
│ approx total docs ingested │                         410127 │
│ backing indices            │                              1 │
│ store size bytes           │                      136310570 │
│ maximum ts (ms)            │                  1682320467448 │
╰────────────────────────────┴────────────────────────────────╯
│ disk usage for index .ds-logs-system_ │
│ benchmarks.test-ep-2023.04.22-000001  │
│ (for all fields)                      │
├──────────────────────────────┬────────┤
│ total                        │ 99.8mb │
│ inverted_index.total         │ 31.3mb │
│ inverted_index.stored_fields │ 35.5mb │
│ inverted_index.doc_values    │   30mb │
│ inverted_index.points        │  2.8mb │
│ inverted_index.norms         │     0b │
│ inverted_index.term_vectors  │     0b │
│ inverted_index.knn_vectors   │     0b │
╰──────────────────────────────┴────────╯
╭───────────────────────────────────────────────────────────────────────────────────────────╮
│ pipeline logs-system_benchmarks.test-999.999.999 stats in node Qa9ujRVfQuWhqEESdt6xnw     │
├───────────────────────────────────────────────┬───────────────────────────────────────────┤
│ grok ()                                       │ Count: 407819 | Failed: 0 | Time: 16.615s │
│ user_agent ()                                 │   Count: 407819 | Failed: 0 | Time: 768ms │
│ pipeline (logs-system_benchmarks.test@custom) │    Count: 407819 | Failed: 0 | Time: 59ms │
╰───────────────────────────────────────────────┴───────────────────────────────────────────╯

--- Benchmark results for package: system_benchmarks - END   ---
Done
```

How to run this example locally:

```
$ make install
$ elastic-package stack up -v -d
$ eval "$(elastic-package stack shellinit)"
$ cd test/packages/benchmarks/system_benchmark/
$ elastic-package build
$ elastic-package install
$ elastic-package benchmark system --benchmark 20mb-logs-benchmark -v
```


test/packages/benchmarks/system_benchmark/_dev/benchmark/system/20mb-logs-benchmark.yml
```diff
@@ -0,0 +1,12 @@
+---
```
ruflin, line 1: Why are the benchmarks not already in the `_dev` top level directory? Are these excluded during the build process?

docs/howto/system_benchmarking.md
```diff
@@ -0,0 +1,245 @@
...
+---
+description: Benchmark 100MiB of data ingested
+input: filestream
+vars: ~
+data_stream.name: test
+data_stream.vars.paths:
+  - "{{SERVICE_LOGS_DIR}}/corpus-*"
+warmup_time_period: 10s
+corpora.generator.size: 100MiB
+corpora.generator.template.path: ./100mb-logs-benchmark/template.log
```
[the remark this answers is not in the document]
marc-gr: You mean always look for these files inside a folder named as the benchmark by convention? If that is the case I think we could since would be less error prone I guess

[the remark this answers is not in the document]
ruflin: ++

internal/benchrunner/reporters/reportable.go
```diff
@@ -0,0 +1,84 @@
...
+type Reportable interface {
+	Package() string
+	Report() []byte
+}
+
+var _ Reportable = &Report{}
+
+type Report struct {
+	pkg string
+	p   []byte
```
aspacca, line 17: what's `p` for?
I would expect something more like `r` (report) or `c` (content) :)

internal/benchrunner/runners/system/runner.go
```diff
@@ -0,0 +1,768 @@
...
+	return generator, nil
+}
+
+func (r *runner) getGeneratorConfig() (*config.Config, error) {
+	var (
+		data []byte
+		err  error
+	)
+
+	if r.scenario.Corpora.Generator.Config.Path != "" {
```
aspacca, line 440: nit: here if we have a both a raw config and a config path, the first will override the second.
maybe move to a switch? one of the two will still have priority, or let's validate that only one of the two is set

internal/benchrunner/runners/system/runner.go
```diff
@@ -0,0 +1,768 @@
...
+	return &cfg, nil
+}
+
+func (r *runner) getGeneratorFields() (fields.Fields, error) {
+	var (
+		data []byte
+		err  error
+	)
+
+	if r.scenario.Corpora.Generator.Fields.Path != "" {
```
aspacca, line 473: same as above for fields

internal/benchrunner/runners/system/runner.go
```diff
@@ -0,0 +1,768 @@
...
+	return fields, nil
+}
+
+func (r *runner) getGeneratorTemplate() ([]byte, error) {
+	var (
+		data []byte
+		err  error
+	)
+
+	if r.scenario.Corpora.Generator.Template.Path != "" {
```
aspacca, line 507: same as above for template :)

internal/benchrunner/runners/system/runner.go
```diff
@@ -0,0 +1,768 @@
...
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			return err
+		}
+
+		// TODO: this should be taken care of by the corpus generator tool, once it will be done let's remove this
+		event := bytes.ReplaceAll(buf.Bytes(), []byte("\n"), []byte(""))
```
aspacca, line 549: nit: not sure if `strings.Replacer` might get better performance

docs/howto/system_benchmarking.md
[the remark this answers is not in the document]
aspacca: if we use fixed filenames we loose the possibility to have several scenarios in the same packages (for example in the case of rally benchmarks for different datasets), unless we have dynamically generated filenames

aspacca (commented):
just some minor concerns on how much code we will have to duplicate when will we introduce the rally benchmarks :)

I'd expect a few under `/internal/runners`, what do you think @marc-gr ?

ruflin:
I tried to do some tests on the setup docs provided in the PR description. It seems some file name / paths have changed?  @marc-gr Could you provide a fully working sample?

internal/benchrunner/runners/system/scenario.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/aymerick/raymond"
+	"github.com/elastic/go-ucfg"
+	"github.com/elastic/go-ucfg/yaml"
+
+	"github.com/elastic/elastic-package/internal/benchrunner/runners/system/servicedeployer"
+)
+
+const (
+	devPath = "_dev/benchmark/system"
+)
```
mrodm, line 23: nit
```suggestion
const devPath = "_dev/benchmark/system"
```

internal/benchrunner/runners/system/servicedeployer/compose.go
```diff
@@ -0,0 +1,214 @@
...
+	defer func() {
+		err := files.RemoveContent(s.ctxt.Logs.Folder.Local)
+		if err != nil {
+			logger.Errorf("could not remove the service logs (path: %s)", s.ctxt.Logs.Folder.Local)
+		}
+	}()
+
+	p, err := compose.NewProject(s.project, s.ymlPaths...)
+	if err != nil {
+		return errors.Wrap(err, "could not create Docker Compose project for service")
```
mrodm, line 149: For errors, it could be used `fmt.Errorf` instead of using `errors.Wrap`

```suggestion
		return fmt.Errorf("could not create Docker Compose project for service: %w", err)
```

internal/kibana/policies.go
```diff
@@ -214,3 +214,76 @@ func (c *Client) AddPackageDataStreamToPolicy(r PackageDataStream) error {
 
 	return nil
 }
+
+// PackagePolicy represents an Agent Policy in Fleet.
```
mrodm, line 218: nit
```suggestion
// PackagePolicy represents an Package Policy in Fleet.
```

internal/kibana/outputs.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// CreateOutput persists the given Output in Fleet.
+func (c *Client) CreateOutput(o Output) (*Output, error) {
+	reqBody, err := json.Marshal(o)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not convert output (request) to JSON")
+	}
+
+	statusCode, respBody, err := c.post(fmt.Sprintf("%s/outputs", FleetAPI), reqBody)
```
mrodm, line 32: Are these outputs removed after the benchmark ?

internal/elasticsearch/ingest/diskusage.go
```diff
@@ -0,0 +1,70 @@
...
+	Fields           map[string]DiskUsageStat `json:"fields"`
+}
+
+func GetDiskUsage(esClient *elasticsearch.API, datastream string) (map[string]DiskUsage, error) {
+	resp, err := esClient.Indices.DiskUsage(datastream,
+		esClient.Indices.DiskUsage.WithFlush(true),
+		esClient.Indices.DiskUsage.WithRunExpensiveTasks(true),
+	)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Node Stats API call failed")
```
mrodm, line 51: As this is a new file, could these usages be changed to `fmt.Errorf`?

internal/elasticsearch/ingest/datastreamstats.go
```diff
@@ -0,0 +1,54 @@
...
+	BackingIndices   int    `json:"backing_indices"`
+	StoreSizeBytes   int    `json:"store_size_bytes"`
+	MaximumTimestamp int    `json:"maximum_timestamp"`
+}
+
+func GetDataStreamStats(esClient *elasticsearch.API, datastream string) (*DataStreamStats, error) {
+	req := esClient.Indices.DataStreamsStats.WithName(datastream)
+	resp, err := esClient.Indices.DataStreamsStats(req)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Node Stats API call failed")
```
mrodm, line 32: nit. DataStream Stats API? 
```suggestion
		return nil, fmt.Errorf("failed call to DataStream Stats API: %w", err)
```

internal/elasticsearch/ingest/datastreamstats.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("unexpected response status for Node Stats (%d): %s: %w", resp.StatusCode, resp.Status(), elasticsearch.NewError(body))
+	}
+
+	var statsResponse DataStreamsStats
+	if err = json.Unmarshal(body, &statsResponse); err != nil {
+		return nil, errors.Wrap(err, "error decoding Node Stats response")
+	}
```
mrodm, line 48:
```suggestion
		return nil, fmt.Errorf("unexpected response status for DataStream Stats (%d): %s: %w", resp.StatusCode, resp.Status(), elasticsearch.NewError(body))
	}

	var statsResponse DataStreamsStats
	if err = json.Unmarshal(body, &statsResponse); err != nil {
		return nil, errors.Wrap(err, "error decoding DataStream Stats response")
	}
```

internal/elasticsearch/ingest/diskusage.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read Stats API response body")
+	}
+
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("unexpected response status for Node Stats (%d): %s: %w", resp.StatusCode, resp.Status(), elasticsearch.NewError(body))
```
mrodm, line 61: Should it be Disk Usage Stats (same for the other errors in this method)? 
```suggestion
		return nil, fmt.Errorf("unexpected response status for Disk Usage Stats (%d): %s: %w", resp.StatusCode, resp.Status(), elasticsearch.NewError(body))
```

internal/kibana/outputs.go
marc-gr: I ended up not needing the outputs, will delete the file

cmd/report.go
```diff
@@ -63,9 +63,9 @@ func setupReportsCommand() *cobraext.Command {
 
 func getBenchReportCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "benchmark",
-		Short: "Generate a benchmark report",
-		Long:  "Generate a benchmark report comparing local results against ones from another benchmark run.",
+		Use:   "pipeline-bench",
+		Short: "Generate a pipeline benchmark report",
+		Long:  "Generate a pipeline benchmark report comparing local results against ones from another benchmark run.",
```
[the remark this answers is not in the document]
marc-gr: Initially I intended to add more, but since this is not added in this PR, is something we can discuss how to approach later, will undo this one.

docs/howto/system_benchmarking.md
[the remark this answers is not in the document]
ruflin: I'm ok with this. Lets get a version in and then iterate on it.

cmd/report.go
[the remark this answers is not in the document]
mrodm: I see that `pipeline-bench` was removed but README was not updated with that latest change.

test/packages/benchmarks/system_benchmark/_dev/benchmark/system/20mb-logs-benchmark/config.yml
```diff
@@ -0,0 +1,40 @@
+- name: IP
```
ruflin, line 1: Is it by design that we don't use here names "similar" to ECS like `source.ip`? I'm aware this is Schema A so not ECS yet at the same time using ECS could make it easy to read (if possible). 

This is just an example for a benchmark tests, I'm mostly worried it sets a precedence on how it should be done.

ruflin:
Testing this with the commands provided above and also did some changes to the corpora size for testing. `elastic-package install` is missing in the command list above to make sure the package is setup, but that is only a testing issue.

Overall, this looks great and I think we should get this rather soonish. I did not review the code on my end only if the command works as expected. The following I noticed: It tells me in the report the corpora was generated under `[/tmp/service_logs/corpus-*]` but it seems it is generated under `/Users/ruflin/.elastic-package/tmp/service_logs` (relative path). It seems it also does not clean up the old corpora, so if run it again, it will also ingest the previous ones?

ruflin (approved):

README.md
```diff
@@ -324,11 +336,11 @@ The report will show performance differences between both runs.
 It is formatted as a Markdown Github comment to use as part of the CI results.
 
 
-### `elastic-package report benchmark`
+### `elastic-package report pipeline-bench`
 
 _Context: package_
 
-Generate a benchmark report comparing local results against ones from another benchmark run.
+Generate a pipeline benchmark report comparing local results against ones from another benchmark run.
```
mrodm, line 343: It looks like this docs need to be updated

mrodm (commented):
LGTM
Just a note about it would be needed to re-generate the README, there is still a reference to `pipeline-bench`

Pending CI to be successful run, there are some lint errors related to the error messages:
https://buildkite.com/elastic/elastic-package/builds/968#0188aec7-c0b5-48ca-a612-9c62fe86407a/188-330

test/packages/benchmarks/system_benchmark/_dev/benchmark/system/20mb-logs-benchmark/config.yml
marc-gr: In this case since this represents a Schema A (as it comes from the source), I thought it made more sense to let the field names represent what they are in the original logs, since not always they will even be an ECS to represent the original data (as it comes after transformations, compositions from several fields, etc.). Even though I do not have a strong opinion on this one I think in this case it might not be too troublesome tbh.

marc-gr:
All comments in the PR are addressed now, it relies on some pending spec changed added in https://github.com/elastic/package-spec/pull/526

internal/formatter/yaml_formatter.go
```diff
@@ -31,5 +31,11 @@ func YAMLFormatter(content []byte) ([]byte, bool, error) {
 		return nil, false, errors.Wrap(err, "marshalling YAML node failed")
 	}
 	formatted := b.Bytes()
+
+	// required to preserve yaml files starting with "---" as yaml.Encoding strips them
+	if bytes.HasPrefix(content, []byte("---\n")) {
```
mrodm, line 36: should it be checked that `formatted` does not contain prefix "---\n" too? 

```suggestion
	if !bytes.HasPrefix(formatted, []byte("---\n") && bytes.HasPrefix(content, []byte("---\n")) {
```

marc-gr: From my tests it is always stripped (there the need for this workaround), but makes no harm to add it just in case 👍

mrodm (approved):
Thanks for addressing all the comments!
LGTM

internal/kibana/policies.go
```diff
@@ -9,30 +9,30 @@ import (
...
-	Description string `json:"description"`
-	Namespace   string `json:"namespace"`
-	Revision    int    `json:"revision,omitempty"`
+	ID                 string   `json:"id,omitempty"`
+	Name               string   `json:"name"`
+	Description        string   `json:"description"`
+	Namespace          string   `json:"namespace"`
+	Revision           int      `json:"revision,omitempty"`
+	MonitoringEnabled  []string `json:"monitoring_enabled"`
+	MonitoringOutputID string   `json:"monitoring_output_id"`
```
jsoriano, line 23: @marc-gr it looks like this field is not available in old versions of Kibana, it also seems that it is not used here. Was it needed?

Integrations CI is failing for something related to this field https://github.com/elastic/integrations/pull/6673.

Trying to fix it by adding `omitempty` in https://github.com/elastic/elastic-package/pull/1320.

cc @mrodm

PR #1233, opened 2023-04-24
Use latest docker-compose v2 version

mrodm:
This PR enables docker-compose v2 in CI.
Setting the latest version available at the time of creating this PR (2.17.2)
https://docs.docker.com/compose/release-notes/

Relates https://github.com/elastic/elastic-package/issues/1116

### DIFF
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -179,10 +179,13 @@ func NewProject(name string, paths ...string) (*Project, error) {
 		c.dockerComposeV1 = true
 		return &c, nil
 	}
+
+	versionMessage := fmt.Sprintf("Determined Docker Compose version: %v", ver)
 	if ver.Major() == 1 {
-		logger.Debugf("Determined Docker Compose version: %v, the tool will use Compose V1", ver)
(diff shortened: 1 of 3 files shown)


internal/compose/compose.go
```diff
@@ -179,10 +179,13 @@ func NewProject(name string, paths ...string) (*Project, error) {
...
 		return &c, nil
 	}
+
+	versionMessage := fmt.Sprintf("Determined Docker Compose version: %v", ver)
 	if ver.Major() == 1 {
-		logger.Debugf("Determined Docker Compose version: %v, the tool will use Compose V1", ver)
+		versionMessage = fmt.Sprintf("%s, the tool will use Compose V1", versionMessage)
 		c.dockerComposeV1 = true
 	}
+	logger.Debug(versionMessage)
```
mrodm, line 188: Maybe it can be dismissed `the tool will use Compose V1` sentence and just lave the message `Determined Docker Compose version: %v`?

jsoriano: Though well, it also tells if V1 implementation is being used here.

mrodm: Ok, I'll leave it adding that sentence if docker-compose version is v1.x

PR #1245, opened 2023-04-28
remove mapping to trigger field

gsantoro:
Since we removed the field `faas.trigger` with type nested from ECS definition with this issue https://github.com/elastic/ecs/issues/2192, we also want to remove the dynamic mapping definition.

### DIFF
--- a/internal/builder/testdata/existing.yml
+++ b/internal/builder/testdata/existing.yml
@@ -6,3 +6,10 @@ elasticsearch:
       properties:
         example:
           type: short
+      dynamic_templates:
+        - forwarded_ip_to_ip:
+            mapping:
+                type: ip
+            match: forwarded_ip
+            match_mapping_type: string
+
--- a/internal/builder/testdata/existing.yml
+++ b/internal/builder/testdata/existing.yml
@@ -6,8 +6,3 @@ elasticsearch:
       properties:
         example:
           type: short
-      dynamic_templates:
-        - to_nested:
-            path: trigger
-            mapping:
-              type: nested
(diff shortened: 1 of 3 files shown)


felixbarny (approved):

internal/builder/testdata/existing.yml
```diff
@@ -6,8 +6,3 @@ elasticsearch:
       properties:
         example:
           type: short
-      dynamic_templates:
-        - to_nested:
-            path: trigger
-            mapping:
-              type: nested
```
mrodm, line 13: If this needs to be removed, please add another one. This file is to be used in a test that ensures that dynamic mappings are added without removing the ones defined in this section.

mrodm (changes requested):
Add some element into `dynamic_templates` in the `internal/builder/testdata/existing.yml` file

internal/builder/testdata/existing.yml
```diff
@@ -7,7 +7,9 @@ elasticsearch:
         example:
           type: short
       dynamic_templates:
-        - to_nested:
-            path: trigger
+        - forwarded_ip_to_ip:
             mapping:
-              type: nested
+                type: ip
```
jsoriano, line 12: I think CI complains about some extra space here.
```suggestion
              type: ip
```

internal/builder/testdata/expected.existing.mappings.yml
```diff
@@ -6,10 +6,11 @@ elasticsearch:
         example:
           type: short
       dynamic_templates:
-        - to_nested:
-            path: trigger
+        - forwarded_ip_to_ip:
             mapping:
-              type: nested
+                type: ip
```
jsoriano, line 11: And/or here.
```suggestion
              type: ip
```

mrodm (approved):
Thanks! LGTM!

Issue #582, opened 2021-11-15
[Feature Request] Adding flag to run system tests without cleanup


P1llus:
In certain usecases it would be preferred that we run our system tests without doing the cleanup process at the end.

For example when creating dashboards for a package that we only have local testdata for, that we cannot integrate with properly, ingesting the test data is crucial for dashboard creation.

Adding this issue after a quick discussion with @mtojek on the subject, so we can look into it further.

andrewkroh:
I had been commenting out the code that deletes events in the system test running so I could build dashboards so I think this would be a good feature to add.

I also manually go in to Kibana to grab data for sample_event.json after running the system tests. If there were an option to dump the system test events to file I would use that to find file a suitable event for the `sample_event.json`.

mtojek:
Temporarily you can use the "timeout" flag to defer cleanup of documents.

jsoriano:
> For example when creating dashboards for a package that we only have local testdata for, that we cannot integrate with properly, ingesting the test data is crucial for dashboard creation.

Maybe we should think specifically about the use case of building dashboards. For example, instead of having a flag to avoid cleanup of system tests, we could maybe have a command that starts the service and configures the agent to collect data from it. It could be a subcommand of `elastic-package service`.
This, or other command, could also record the data as fixtures that can be used in the future without needing to start the service.
Having a tool to record and reuse fixtures would also help to use real-world data instead of using the same data used on system tests.

In a more general sense, it may be also nice to add some mechanism to configure generators, that can be used for any purpouse that needs real-like data (sample data for dashboards, performance testing, mocking services for tests...).

> I also manually go in to Kibana to grab data for sample_event.json after running the system tests. If there were an option to dump the system test events to file I would use that to find file a suitable event for the `sample_event.json`.

For this it'd be nice to have something as the conditions mechanism we have in Metricbeat, so events are still collected automatically, and the selection factor can be translated to config or code. Related: https://github.com/elastic/elastic-package/issues/773.

Said that, I am not oposed to have some mechanism to control tear downs :slightly_smiling_face: But for these specific use cases I would prefer to have more specific solutions.

andrewkroh:
> For example, instead of having a flag to avoid cleanup of system tests, we could maybe have a command that starts the service and configures the agent to collect data from it.

I like the idea of a specialized command.

mrodm:
Thinking about options for this issue, I see some different approaches.

Adding a subcommand or parameter under `elastic-package install`:
[...]

Issue #405, opened 2021-07-07
Run system tests with clean data directories


jsoriano:
Beats managed by the agent in system tests can store information in their data directories. This information persists between tests and may affect the result of later executions Tests should be executed on isolation, with clean data directories.

In https://github.com/elastic/integrations/pull/1145 I found that after running system tests several times, one of the data streams started to fail frequently because it couldn't find documents in the data stream. I saw that the filebeat registry contained lots of entries, many of them for the file read by the data stream. Not 100% sure, but I think this might be the cause of the failures, filebeat may not be reading the file because it was already read in the past to a similar offset.

PR #1250, opened 2023-05-05
Add new flags to just setup, teardown or run tests only

mrodm:
Relates #582

This PR is intended to add the required subcommands and/or parameters to allow just run the setup, no-provision or teardown of tests. Focused on system tests.

Added flags:
- `--setup`
    - Creates all the resources , install the package, enrolls the agent
    - It ensures that some docs are stored in ES
    - Writes into `ServiceSetupDir` folder the original policy assigned to the agent as well as the policy created by elastic-package
- `--no-provision`
    - Reads both policies from `ServiceSetupDir`
    - Install the package (in case there are changes locally)
    - Executes all the system tests
    - Does not remove the logs folder to avoid issues in service writing to those files (they should be truncated instead)
- `--tear-down`
    - Reads both policies from `ServiceSetupDir`
    - Runs all the handlers to remove all resources created
    - Stop the service

Pending:
- [x] Add creation of a new folder under the current profile directory for test resources
- [x] Store needed resources between Setup, NoProvision and Teardown in a folder under the current profile (.e.g. `~/.elastic-package/profiles/default/testing/*). Examples of resources to be stored:
    - Original policy assigned before changing it in the setup method, so TearDown step can restore it
      ```shell
       $ ls -l ~/.elastic-package/profiles/default/stack/state/
      total 4
      -rw-r--r-- 1 user user 773 feb 14 17:15 service.json
      ```
- [x] Add methods and CLI parameters to Setup and TearDown for just one config and variant in system test
     - Those new parameters could be added just 
       - if a new method in TestRunner interface like `CanRunSetupTeardownIndependent()` or `CanRunSeparateTests` returns true
        - if the runner implements a new interface (e.g. TestRunnerSetterUp)
          ```
          type TestRunnerSetterUp interface {
           	// Setup installs package and starts the service
           	Configure(TestOptions) error
    
          	// Setup installs package and starts the service
          	Setup(TestOptions) error
          }
          ```
- [x] Ensure this works also for input packages - tested with `sql_input`
- [x] CI step
    - Build step link: https://buildkite.com/elastic/elastic-package/builds/2262#018d6a96-5e74-4e00-8633-e25536c5a1d1/476
- [x] Add new section in docs:
    - Rendered version of the docs: https://github.com/mrodm/elastic-package/blob/add_flag_setup_testrunner/docs/howto/system_testing.md#running-system-tests-without-cleanup-technical-preview

## Checklist
- [x] Test with an integration package (e.g. nginx - integrations)
- [x] Test setting a non-default variant (e.g. mysql percona_8_0_13 - integrations)
- [x] Test with an input package (e.g. sql_input - integrations)
- [x] Test with a package using custom agents (e.g. auditd_manager - elastic-package)
- [x] Test with a package using kind as a service deployer (e.g. kubernetes - elastic-package)

## How to test locally

```
cd /path/integrations/packages/mysql

# Start Elastic stack as usual
elastic-package stack down -v

# Go to a package from integrations repository with variants, taken as example `mysql`
elastic-package test system -v --config-file data_stream/status/_dev/test/system/test-default-config.yml --variant percona_8_0_36 --setup
elastic-package test system -v --no-provision
elastic-package test system -v --tear-down

# Go to a package from integrations repository without variants, taken as example `hashicorp_vault`
elastic-package test system -v --config-file data_stream/audit/_dev/test/system/test-tcp-config.yml --setup

# this command can be executed several times
elastic-package test system -v --no-provision

elastic-package test system -v --tear-down


# Go to an input package, taken as example `sql_input`
elastic-package test system -v --config-file _dev/test/system/test-mysql-config.yml --setup

# this command can be executed several times
elastic-package test system -v --no-provision

elastic-package test system -v --tear-down

elastic-package stack down -v
```


sharbuz:
Hi everyone,

cmd/testrunner.go
```diff
@@ -156,15 +164,27 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 			return fmt.Errorf("cannot determine if package has data streams: %w", err)
 		}
 
+		configFileFlag, _ := cmd.Flags().GetString(cobraext.ConfigFileFlagName)
+		if configFileFlag != "" && !filepath.IsAbs(configFileFlag) {
+			return fmt.Errorf("config file path must be an absolute path: %s", configFileFlag)
+		}
```
jsoriano, line 170: We can obtain here the absolute path if we need it, right? This improves UX.
```suggestion
		if configFileFlag != ""  {
			absPath, err := filepath.Abs(configFileFlag)
			if err != nil {
				return fmt.Errorf("cannot obtain the absolute path for config file path: %s", configFileFlag)
			}
			configFileFlag = absPath
		}
```

cmd/testrunner.go
```diff
@@ -258,6 +289,10 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				ServiceVariant:     variantFlag,
 				WithCoverage:       testCoverage,
 				CoverageType:       testCoverageFormat,
+				ConfigFilePath:     configFileFlag,
+				RunSetup:           runSetup,
```
jsoriano, line 293: `elastic-package service up` fails now after a setup is done, because the container already exists. Should we do something about this?

I see these options:
* Make `elastic-package service up` aware of the service started by `elastic-package test --setup`. And attach to it if started?
* Remove `elastic-package service up`. Its purpose was probably the same as `elastic-package test --setup` has now, but `elastic-package test --setup` is more complete as it also configures the agent, and has a way to tear down. Though it is true that their use cases might be slightly different.
* Keep both commands, but make them explicitly fail if the other one has been used to start a service.

In any case we don't need to solve this now.

docs/howto/system_testing.md
```diff
@@ -599,6 +599,80 @@ A sample workflow would look like:
...
+It's possible also to run these steps independently. For that it is required to set which configuration file (`--config-file`)
+and which variant (`--variant`), if any, is going to be used to start and configure the service for these tests.
+
+Then, each step can be run using one of these flags:
+- Run the setup (`--setup`), after this command is executed:
+    - Service container will be kept running.
+    - Elastic Agent is going to be sending documents to Elasticsearch.
+- Run just the tests (`--no-provision`), after this command is executed:
+    - Service container will not be stopped.
+    - Documents in the Data stream will be deleted, so the tests will use the latest documents sent.
```
jsoriano, line 643: :+1:

jsoriano, line 643: It'd be great to find a way to also reset the agent data, but no need to solve this here, this is not a new issue (https://github.com/elastic/elastic-package/issues/405).

docs/howto/system_testing.md
```diff
@@ -599,6 +599,80 @@ A sample workflow would look like:
...
+Examples:
+
+```shell
+# Start Elastic stack as usual
+elastic-package stack up -v -d
+
+# Testing a package using a configuration file and a variant (e.g. mysql)
+elastic-package test system -v --config-file $(pwd)/data_stream/status/_dev/test/system/test-default-config.yml --variant percona_8_0_36 --setup
+elastic-package test system -v --config-file $(pwd)/data_stream/status/_dev/test/system/test-default-config.yml --variant percona_8_0_36 --no-provision
+elastic-package test system -v --config-file $(pwd)/data_stream/status/_dev/test/system/test-default-config.yml --variant percona_8_0_36 --tear-down
```
jsoriano, line 661: The ice on the cake would be to make it easier to pass the config files, but well automcompletion can make wonders here.

internal/servicedeployer/compose.go
```diff
@@ -25,6 +25,9 @@ type DockerComposeServiceDeployer struct {
 	profile  *profile.Profile
 	ymlPaths []string
 	variant  ServiceVariant
+
+	runTeardown  bool
```
jsoriano, line 29: Nit. In other places we use a different casing, make it consistent.
```suggestion
	runTearDown  bool
```

internal/testrunner/runners/system/runner.go
```diff
@@ -130,10 +142,140 @@ func (r *runner) TestFolderRequired() bool {
...
+	_, err := os.Stat(r.locationManager.ServiceSetupDir())
+	if r.options.RunSetup && !os.IsNotExist(err) {
+		return result.WithError(fmt.Errorf("failed to run --setup, required to tear down previous setup run: %s exists", r.locationManager.ServiceSetupDir()))
+	}
+	if r.options.RunTestsOnly && os.IsNotExist(err) {
+		return result.WithError(fmt.Errorf("failed to run --no-provision, missing service setup folder: %s does not exist", r.locationManager.ServiceSetupDir()))
+	}
+	if r.options.RunTearDown && os.IsNotExist(err) {
+		return result.WithError(fmt.Errorf("failed to run --tear-down, missing service setup folder: %s does not exist", r.locationManager.ServiceSetupDir()))
+	}
```
jsoriano, line 189: Maybe we can provide "higher-level" error messages, the user doesn't really matter about the existence of this file. We can keep the info about the service dir at the debug level.
```suggestion
	_, err := os.Stat(r.locationManager.ServiceSetupDir())
	logger.Debugf("Services setup data exists in %s: %v", r.locationManager.ServiceSetupDir(), !os.IsNotExist(err))
	if r.options.RunSetup && !os.IsNotExist(err) {
		return result.WithError(fmt.Errorf("failed to run --setup, required to tear down previous setup"))
	}
	if r.options.RunTestsOnly && os.IsNotExist(err) {
		return result.WithError(fmt.Errorf("failed to run tests with --no-provision, setup first with --setup"))
	}
	if r.options.RunTearDown && os.IsNotExist(err) {
		return result.WithError(fmt.Errorf("failed to run --tear-down, setup not found"))
	}
```

internal/testrunner/runners/system/runner.go
```diff
@@ -130,10 +142,140 @@ func (r *runner) TestFolderRequired() bool {
...
+		return r.run()
+	}
+
+	result := r.newResult("(init)")
+	if err := r.initRun(); err != nil {
+		return result.WithError(err)
+	}
+
+	if len(r.variants) > 1 {
+		return result.WithError(fmt.Errorf("a variant must be selected or trigger the test in no-variant mode (available variants: %s)", strings.Join(r.variants, ", ")))
```
jsoriano, line 174: Nice :+1:

docs/howto/system_testing.md
```diff
@@ -599,6 +599,80 @@ A sample workflow would look like:
...
+
+Examples:
+
+```shell
+# Start Elastic stack as usual
+elastic-package stack up -v -d
+
+# Testing a package using a configuration file and a variant (e.g. mysql)
+elastic-package test system -v --config-file $(pwd)/data_stream/status/_dev/test/system/test-default-config.yml --variant percona_8_0_36 --setup
+elastic-package test system -v --config-file $(pwd)/data_stream/status/_dev/test/system/test-default-config.yml --variant percona_8_0_36 --no-provision
```
jsoriano, line 660: Is the variant needed to run the test? At this point we are not going to provision anything, I think it'd be better if we would persist the variant info on the setup directory and take it from there. Actually this command fails if you use a different variant.

docs/howto/system_testing.md
jsoriano, line 661: Also for the tear down, could we make it to don't require config file or variant and just tear down whatever was started?

docs/howto/system_testing.md
```diff
@@ -599,6 +599,80 @@ A sample workflow would look like:
 - Run `elastic-package stack up -d -v`
 - Navigate to the package folder in integrations and run `elastic-package test system -v`
 
+### Running system tests without cleanup (technical preview)
```
jsoriano, line 602: Try to print also something about it being on technical preview on the command output when these flags are used.

internal/testrunner/runners/system/runner.go
```diff
@@ -594,42 +785,89 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+			if signal.SIGINT() {
+				return true, errors.New("SIGINT: cancel clearing data")
+			}
+			hits, err := r.getDocs(scenario.dataStream)
+			if err != nil {
+				return false, err
+			}
+			// In this mode, service is still running and the agent is sending documents, so sometimes
+			// cannot be guaranteed to be zero documents
+			return startHits.size() > hits.size(), nil
```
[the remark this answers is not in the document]
jsoriano: Makes sense as everything is running :+1:

jsoriano (commented):
Great job here! I love to be able to run system tests on seconds without needing to start the scenario each time.

Haven't looked too deep into the code yet, adding a first review more based on trying the new flags.

docs/howto/system_testing.md
mrodm: Added a message about the technical preview:

```
 $ elastic-package test system  --config-file data_stream/access/_dev/test/system/test-default-config.yml --setup
Running tests per stages (technical preview)
2024/02/08 11:03:19  INFO Using service variant: ServiceVariant{Name: v1, Env: SERVICE_VERSION=1.19.5}
2024/02/08 11:03:21  INFO License text found in "/home/mariorodriguez/Coding/work/integrations/LICENSE.txt" will be included in package
```

Anything else to add to that message?

jsoriano: Looks good.

docs/howto/system_testing.md
mrodm: Currently, all these operations are tied , config file and variants are required to create the service options and test config.
```go
	serviceOptions, ctxt, err := r.createConfigService(r.variants[0])
	if err != nil {
		return result.WithError(err)
	}

	testConfig, err := newConfig(r.options.ConfigFilePath, ctxt, r.variants[0])
	if err != nil {
		return nil, fmt.Errorf("unable to load system test case file '%s': %w", r.options.ConfigFilePath, err)
	}
	logger.Debugf("Using config: %q", testConfig.Name())
```

And those are used to prepare the scenario to run the requested operation
```go
	scenario, err := r.prepareScenario(testConfig, ctxt, serviceOptions)
```

Not sure how easy or difficult it would be to do that. Maybe I can write also to a file the `serviceOptions` struct and read it from there. This variable is built here:
https://github.com/elastic/elastic-package/blob/c3c9565314a815f81a3ef9dacb4d6f1c0cdc8df5/internal/testrunner/runners/system/runner.go#L253

docs/howto/system_testing.md
mrodm: Trying to do some refactors to get the variant from the service context...

docs/howto/system_testing.md
mrodm: Latest changesets pushed allowed to remove config-file flag for `--no-provision` and `--tear-down` flags.

It required some changes:
- Initialize `locationsManager` in the `init()` function.
- Write the `testConfig` variable to file at the end of the setup process.
    - This allows us to get the variant and config file in `--no-provision` and `--tear-down` modes
- Read one of the files written in the setup dir in `cmd/testrunner.go`, since it is required to know the configuration file path to get the data stream (and test folders).
    - Add a new method to the interface to allow this.

docs/howto/system_testing.md
mrodm: After some changes, I've been able to remove variant flag as a requirement for `--no-provision` and `--tear-down` flags

internal/testrunner/runners/system/runner.go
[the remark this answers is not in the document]
jsoriano: But if this is the only difference, could we move this to a function and parameterize the number of expected documents?

internal/testrunner/runners/system/runner.go
```diff
@@ -722,36 +1044,83 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
 	if waitErr != nil {
-		return result.WithError(waitErr)
+		return nil, waitErr
 	}
 
 	if !passed {
-		return result.WithError(testrunner.ErrTestCaseFailed{Reason: fmt.Sprintf("could not find hits in %s data stream", dataStream)})
+		// return nil, fmt.Errorf("could not find hits in %s data stream", scenario.dataStream)
+		// result.FailureMsg = fmt.Sprintf("could not find hits in %s data stream", dataStream)
+		// return result.WithError(fmt.Errorf("%s", result.FailureMsg))
```
jsoriano, line 1062: Leftovers?

internal/testrunner/runners/system/runner.go
```diff
@@ -722,36 +1044,83 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+			return nil, fmt.Errorf("failed to write origin policy JSON: %w", err)
+		}
+		err = os.WriteFile(filepath.Join(r.locationManager.ServiceSetupDir(), setupAgentFileName), agentBytes, 0644)
+		if err != nil {
+			return nil, fmt.Errorf("failed to write agent JSON: %w", err)
+		}
+		err = os.WriteFile(filepath.Join(r.locationManager.ServiceSetupDir(), setupTestConfigFileName), testConfigBytes, 0644)
+		if err != nil {
+			return nil, fmt.Errorf("failed to write test config JSON: %w", err)
+		}
```
jsoriano, line 1111: Move all this to another function? And is there a reason why we cannot write all this to a single file?

internal/testrunner/runners/system/runner.go
```diff
@@ -1285,7 +1675,7 @@ func writeSampleEvent(path string, doc common.MapStr, specVersion semver.Version
 		return fmt.Errorf("marshalling sample event failed: %w", err)
 	}
 
-	err = os.WriteFile(filepath.Join(path, "sample_event.json"), body, 0644)
+	err = os.WriteFile(filepath.Join(path, "sample_event.json"), body, 0o644)
```
jsoriano, line 1678: Nit. These are equivalent, right? :thinking: I would leave it as it was, as this is the syntax we are using in other places.
```suggestion
	err = os.WriteFile(filepath.Join(path, "sample_event.json"), body, 0644)
```

internal/testrunner/testrunner.go
```diff
@@ -238,12 +246,7 @@ func FindTestFolders(packageRootPath string, dataStreams []string, testType Test
 	folders := make([]TestFolder, len(paths))
 	_, pkg := filepath.Split(packageRootPath)
 	for idx, p := range paths {
-		relP := strings.TrimPrefix(p, packageRootPath)
-		parts := strings.Split(relP, string(filepath.Separator))
-		dataStream := ""
-		if len(parts) >= 3 && parts[1] == "data_stream" {
-			dataStream = parts[2]
-		}
+		dataStream := ExtractDataStreamFromPath(p, packageRootPath)
```
jsoriano, line 249: :+1:

jsoriano (commented):
It works great!

Codewise I am a bit afraid that we are introducing some complexity in an already complex code. I wonder if we could simplify some things.
In any case, I would be fine with merging this mostly as is, and iterate on future refactors.

One thing I would like to review before merging is if we need to keep in files so much info, and in so many files. It looks like too much state and too disperse.

docs/howto/system_testing.md
```diff
@@ -599,6 +599,81 @@ A sample workflow would look like:
...
+        - Uninstall the package.
+    - Stop the service (e.g. container) to be tested.
+    - Wipe all the documents of the data stream.
+
+This process is repeated for each combination of:
+- data stream `D`, if the package is of `integration` type.
+- configuration file defined under `_dev/test/sytem` folder.
+- variant (`_dev/deploy/variants.yml`).
+
+It's possible also to run these steps independently. For that it is required to set which configuration file (`--config-file`)
```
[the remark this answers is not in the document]
mrodm: My idea was to just allow this flag together with --setup. I'll review this.

internal/servicedeployer/compose.go
```diff
@@ -68,9 +73,14 @@ func (d *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
...
+	case d.runTestsOnly:
+		logger.Debug("Skipping removing service logs folder folder %s", outCtxt.Logs.Folder.Local)
+	default:
+		err = files.RemoveContent(outCtxt.Logs.Folder.Local)
+		if err != nil {
+			return nil, fmt.Errorf("removing service logs failed: %w", err)
+		}
 	}
 
 	// Boot up service
```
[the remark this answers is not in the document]
mrodm: I'll check if this can be skipped with `--no-provision`

docs/howto/system_testing.md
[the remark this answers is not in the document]
mrodm: Updated, --config-file and --setup must be used together.

internal/servicedeployer/compose.go
```diff
@@ -68,9 +73,14 @@ func (d *DockerComposeServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedSer
 	}
 
 	// Clean service logs
-	err = files.RemoveContent(outCtxt.Logs.Folder.Local)
-	if err != nil {
-		return nil, fmt.Errorf("removing service logs failed: %w", err)
+	switch {
+	case d.runTestsOnly:
```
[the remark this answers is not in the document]
mrodm: Changed.

internal/servicedeployer/custom_agent.go
```diff
@@ -35,14 +35,19 @@ type CustomAgentDeployer struct {
...
 	dockerComposeFile string
 	stackVersion      string
+
+	runTearDown  bool
+	runTestsOnly bool
 }
 
 // NewCustomAgentDeployer returns a new instance of a deployedCustomAgent.
-func NewCustomAgentDeployer(profile *profile.Profile, dockerComposeFile string, stackVersion string) (*CustomAgentDeployer, error) {
+func NewCustomAgentDeployer(profile *profile.Profile, dockerComposeFile string, stackVersion string, runTearDown, runTestsOnly bool) (*CustomAgentDeployer, error) {
```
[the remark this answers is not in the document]
mrodm: I'll create options for docker and custom agent options, that are the both service deployers that I added support here.

internal/testrunner/runners/system/runner.go
mrodm: Deleted

internal/testrunner/runners/system/runner.go
mrodm: Sure!

I've started to try `nvim` with the LSP configuration, and it does those changes... I need to check how to disable this one specifically.

internal/testrunner/runners/system/runner.go
mrodm: Moved everything to just one file

internal/testrunner/runners/system/runner.go
```diff
@@ -43,10 +43,29 @@ const (
...
+	setupTestConfigFileName = "test-config.json"
 )
 
 func init() {
-	testrunner.RegisterRunner(&runner{})
+	runner, err := NewSystemTestRunner()
+	if err != nil {
+		panic(err)
+	}
+	testrunner.RegisterRunner(runner)
```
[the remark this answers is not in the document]
mrodm: Updated. Now the `init()` has been reverted, and the location manager is initialized both in the `internal/testrunner/runners/system/runner.go` and in the `cmd/testrunner.go`.

To allow being used the same file name in both files, I moved the constant to `internal/testrunner/testrunner.go`. Thereby, cmd/testrunner does not have access to the system runner directly.

internal/testrunner/runners/system/runner.go
```diff
@@ -201,60 +402,83 @@ func (r *runner) newResult(name string) *testrunner.ResultComposer {
...
+		return result.WithError(fmt.Errorf("failed to run tests, required to tear down previous setup run: %s exists", r.locationManager.ServiceSetupDir()))
 	}
 
 	startTesting := time.Now()
-	for _, cfgFile := range cfgFiles {
-		for _, variantName := range r.selectVariants(variantsFile) {
-			partial, err := r.runTestPerVariant(result, locationManager, cfgFile, dataStreamPath, variantName, stackVersion.Version())
+	for _, cfgFile := range r.cfgFiles {
+		for _, variantName := range r.variants {
+			partial, err := r.runTestPerVariant(result, cfgFile, variantName)
```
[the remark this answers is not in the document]
mrodm: Yes, that could be helpful, as you mention, to reduce the time and resources spent in system tests.

Not sure if there would be more changes needed, at least it would be needed to ensure that for every test the (old) documents present in the data stream are deleted before running each one.

internal/testrunner/runners/system/runner.go
```diff
@@ -43,10 +43,29 @@ const (
 
 	allFieldsBody = `{"fields": ["*"]}`
 	DevDeployDir  = "_dev/deploy"
+
+	setupNewPolicyFileName  = "policy-setup.json"
+	setupOrigPolicyFileName = "orig-policy.json"
+	setupAgentFileName      = "agent.json"
+	setupTestConfigFileName = "test-config.json"
```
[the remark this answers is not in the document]
mrodm: Just kept one file for all the information that we currently require:
- current policy
- original policy
- agent data
- configuration file path
- variant name

cmd/testrunner.go
```diff
@@ -156,15 +173,68 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
...
+					}
+					err = json.Unmarshal(contents, &serviceSetupData)
+					if err != nil {
+						return fmt.Errorf("failed to decode service setup data %q: %w", setupDataPath, err)
+					}
+					configFileFlag = serviceSetupData.ConfigFilePath
+				}
+				dataStream := testrunner.ExtractDataStreamFromPath(configFileFlag, packageRootPath)
+				dataStreams = append(dataStreams, dataStream)
+			} else if cmd.Flags().Lookup(cobraext.DataStreamsFlagName) != nil {
```
jsoriano, line 234: Should we mark this flag as mutually exclusive with the others?

internal/testrunner/runners/system/runner.go
```diff
@@ -130,10 +138,172 @@ func (r *runner) TestFolderRequired() bool {
...
+		if len(r.variants) == 0 {
+			logger.Debug("No variant mode")
+		}
+	}
+
+	_, err := os.Stat(r.locationManager.ServiceSetupDir())
+	logger.Debugf("Service setup data exists in %s: %v", r.locationManager.ServiceSetupDir(), !os.IsNotExist(err))
+	if r.options.RunSetup && !os.IsNotExist(err) {
+		return result.WithError(fmt.Errorf("failed to run --setup, required to tear down previous setup"))
+	}
```
jsoriano, line 183: Nit. It'd be nice to make the setup more aware of the state of the current setup, so if the file exists:
* It checks if the setup is actually up, if not, it ignores existing file.
* It checks if the setup was started with the same parameters, then it just finishes, or runs idempotently.
* If the setup is running, and it has different parameters, then we show the current error.

Or maybe we can just try to always run it and trust on operations being idempotent.

But this would add even more complexity, so lets leave this for a future change unless you see a simple and reliable way to do it.

internal/testrunner/runners/system/runner.go
```diff
@@ -73,34 +73,39 @@ type logsByContainer struct {
...
 
 type runner struct {
 	options   testrunner.TestOptions
 	pipelines []ingest.Pipeline
+
+	dataStreamPath  string
+	cfgFiles        []string
+	variants        []string
+	stackVersion    kibana.VersionInfo
+	locationManager *locations.LocationManager
```
jsoriano, line 107: The location manager is only used to obtain the path to the services state dir, maybe we could initiallize the runner directly with the path.
```suggestion
	serviceStatePath string
```

internal/testrunner/runners/system/runner.go
```diff
@@ -130,10 +138,172 @@ func (r *runner) TestFolderRequired() bool {
...
+		serviceSetupPath := filepath.Join(r.locationManager.ServiceSetupDir(), testrunner.ServiceSetupDataFileName)
+		logger.Debugf("Reading test config from file %s", serviceSetupPath)
+		contents, err := os.ReadFile(serviceSetupPath)
+		if err != nil {
+			return result.WithError(fmt.Errorf("failed to read test config %q: %w", serviceSetupPath, err))
+		}
+		err = json.Unmarshal(contents, &serviceSetupData)
+		if err != nil {
+			return result.WithError(fmt.Errorf("failed to decode service options %q: %w", serviceSetupPath, err))
+		}
```
jsoriano, line 202: Nit. Move operations with the service state to their own methods.

internal/testrunner/runners/system/runner.go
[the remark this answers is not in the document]
jsoriano: > for every test the (old) documents present in the data stream are deleted before running each one.

Yes, this should be done independently of the running service.

internal/testrunner/runners/system/runner.go
```diff
@@ -612,59 +837,46 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
-	if err != nil {
-		return result.WithError(fmt.Errorf("error setting log level debug for agent %s: %w", agent.ID, err))
-	}
-	r.resetAgentLogLevelHandler = func() error {
-		logger.Debugf("reassigning original log level %q back to agent...", origLogLevel)
 
-		if err := r.options.KibanaClient.SetAgentLogLevel(agent.ID, origLogLevel); err != nil {
-			return fmt.Errorf("error reassigning original log level to agent: %w", err)
+	switch {
+	case r.options.RunTearDown || r.options.RunTestsOnly:
```
jsoriano, line 871: Nit. if/else?

internal/testrunner/runners/system/runner.go
```diff
@@ -674,31 +886,64 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 		return nil
 	}
 
-	policyWithDataStream, err := r.options.KibanaClient.GetPolicy(policy.ID)
-	if err != nil {
-		return result.WithError(fmt.Errorf("could not read the policy with data stream: %w", err))
+	origAgent := agent
+	origLogLevel := ""
+	switch {
+	case r.options.RunTearDown:
```
jsoriano, line 892: Nit. if/else?

internal/testrunner/runners/system/runner.go
```diff
@@ -674,31 +886,64 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+		if err := r.options.KibanaClient.SetAgentLogLevel(agent.ID, origLogLevel); err != nil {
+			return fmt.Errorf("error reassigning original log level to agent: %w", err)
+		}
+		return nil
+	}
+
+	switch {
+	case r.options.RunTearDown || r.options.RunTestsOnly:
+		logger.Debug("Skip assiging package data stream to agent")
+	default:
```
jsoriano, line 916: Nit. if/else?

jsoriano (changes requested):
Another round of comments. Most of them are about nitpicks or things that we could or should refactor later.

The things that I would like to see addressed:
* There is a panic happening if a variant that doesn't exist is used.
* I am thinking now that we should store the service state in the profile directory, not in the mostly legacy global stack directory. So this doesn't block us in the path to be able to run tests with different profiles independently.

cmd/testrunner.go
```diff
@@ -156,15 +173,68 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
...
+		configFileFlag := ""
+		runSetup := false
+		runTearDown := false
+		runTestsOnly := false
+
+		if runner.CanRunSetupTeardownIndependent() && cmd.Flags().Lookup(cobraext.ConfigFileFlagName) != nil {
+			// not all test types define these flags
+			runSetup, _ = cmd.Flags().GetBool(cobraext.SetupFlagName)
+			runTearDown, _ = cmd.Flags().GetBool(cobraext.TearDownFlagName)
+			runTestsOnly, _ = cmd.Flags().GetBool(cobraext.NoProvisionFlagName)
```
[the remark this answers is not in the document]
mrodm: Added

cmd/testrunner.go
```diff
@@ -156,15 +173,68 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
...
+					fmt.Printf("Reading service setup data from file: %s\n", setupDataPath)
+					contents, err := os.ReadFile(setupDataPath)
+					if err != nil {
+						return fmt.Errorf("failed to read service setup data %q: %w", setupDataPath, err)
+					}
+					err = json.Unmarshal(contents, &serviceSetupData)
+					if err != nil {
+						return fmt.Errorf("failed to decode service setup data %q: %w", setupDataPath, err)
+					}
+					configFileFlag = serviceSetupData.ConfigFilePath
```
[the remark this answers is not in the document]
mrodm: Done.

cmd/testrunner.go
mrodm: Added --data-streams flag to be mutually exclusive wit the flags `--setup`, `--no-provision` and `--tear-down`

internal/testrunner/runners/system/runner.go
```diff
@@ -130,10 +138,172 @@ func (r *runner) TestFolderRequired() bool {
...
+			return result.WithError(fmt.Errorf("failed to read test config %q: %w", serviceSetupPath, err))
+		}
+		err = json.Unmarshal(contents, &serviceSetupData)
+		if err != nil {
+			return result.WithError(fmt.Errorf("failed to decode service options %q: %w", serviceSetupPath, err))
+		}
+	}
+
+	configFile := r.options.ConfigFilePath
+	variant := r.variants[0]
```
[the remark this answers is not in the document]
mrodm: Added a check to ensure that the list of variants retrieved is not empty. That means that no variant defined in the package has matched the one set in the parameter.

This also chagnes how it works for the command `elastic-package test system --variant other -v`. If the variant does not exist in that execution, that execution does not fail but it does not test nothing.

In both cases (with and without flags), a new error would be raised:

```
Error: error running package system tests: could not complete test run: not found variant definition "other"
```

internal/testrunner/runners/system/runner.go
mrodm: `locationManager` is also used to create the service log dir for the agent and the output dir:
https://github.com/elastic/elastic-package/blob/49345174cf32ab05dc8f6d57897a225bd51d7539/internal/testrunner/runners/system/runner.go#L284-L294

So I think it could be kept and maybe just add the suggested field as a new variable without removing the locationManager. WDYT ? @jsoriano 

If we do so, that would help us to remove around 9 or 10 calls to `r.locationManager.ServiceSetupDir()`.  

```go
type runner struct {
	options   testrunner.TestOptions
	pipelines []ingest.Pipeline

	dataStreamPath   string
	cfgFiles         []string
	variants         []string
	stackVersion     kibana.VersionInfo
	locationManager  *locations.LocationManager
	serviceStatePath string
    ...
}
```

internal/testrunner/runners/system/runner.go
jsoriano: Yes, this would mean that a sequence of commands like the one above would work. But agree to leave it for a future change.

cmd/testrunner.go
```diff
@@ -156,15 +173,68 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
...
+					locationManager, err := locations.NewLocationManager()
+					if err != nil {
+						return fmt.Errorf("can't read configuration directory")
+					}
+
+					type setupData struct {
+						ConfigFilePath string `json:"config_file_path"`
+					}
+					var serviceSetupData setupData
+					setupDataPath := filepath.Join(locationManager.ServiceSetupDir(), testrunner.ServiceSetupDataFileName)
```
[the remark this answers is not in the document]
mrodm: I like the idea of moving this to the profiles directory.
I'll try to move this file to:
`.elastic-package/profiles/<profile>/stack/state/service.json`

cmd/testrunner.go
mrodm: Current error is:

```
 $ elastic-package service up 
Boot up the service stack
2024/02/14 16:57:44  INFO Using service variant: ServiceVariant{Name: v1, Env: SERVICE_VERSION=1.19.5}
Error: up command failed: can't set up the service deployer: can't attach service container to the stack network: could not attach container to the stack network (stderr="Error response from daemon: endpoint with name elastic-package-service-nginx-1 already exists in network elastic-package-stack_default\n"): exit status 1
```

It does not give us too much information of the root cause of the issue (the service is already running), but as you mention I would leave it for a future change.

internal/configuration/locations/locations.go
```diff
@@ -35,6 +35,8 @@ var (
 	rallyCorpusDir        = filepath.Join(temporaryDir, "rally_corpus")
 	kubernetesDeployerDir = filepath.Join(deployerDir, "kubernetes")
 	serviceOutputDir      = filepath.Join(temporaryDir, "output")
+
+	serviceSetupDir = filepath.Join(stackDir, "service_setup")
```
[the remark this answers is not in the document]
mrodm: I'll keep the directory in case it is needed to add any other state.
With the latest changes the file path of this file would be:

```
~/.elastic-package/profiles/<profile>/stack/state/service.json
```

internal/testrunner/runners/system/runner.go
mrodm: With the latest changes, (moving this file under the profiles directory), I added two new fields here:
https://github.com/mrodm/elastic-package/blob/baad699d966620fa63514ab5178f133ee51fa405/internal/testrunner/runners/system/runner.go#L109-L110
```
	serviceStateFolder   string
	serviceStateFilePath string
```

There are two since it is also required the folder path to create/delete the whole folder.

To use the same paths in cmd/testrunner.go  a new function has been added in the `internal/testrunner/testrunner.go` to try to avoid duplication:

https://github.com/mrodm/elastic-package/blob/baad699d966620fa63514ab5178f133ee51fa405/internal/testrunner/testrunner.go#L329-L331

jsoriano: > There are two since it is also required the folder path to create/delete the whole folder.

Cannot we use `filepath.Dir(serviceStateFilePath)` to get the folder?

mrodm: > Cannot we use `filepath.Dir(serviceStateFilePath)` to get the folder?

Sure, I'll update the methods using the folder with that call. No need to keep that field in the struct

internal/servicedeployer/factory.go
```diff
@@ -68,8 +87,18 @@ func Factory(options FactoryOptions) (ServiceDeployer, error) {
...
+			DockerComposeFile: customAgentCfgYMLPath,
+			StackVersion:      options.StackVersion,
+			RunTearDown:       options.RunTearDown,
+			RunTestsOnly:      options.RunTestsOnly,
+		}
+		return NewCustomAgentDeployer(opts)
 	case "tf":
+		if options.RunSetup || options.RunTearDown || options.RunTestsOnly {
+			return nil, errors.New("terraform service deployer not supported to run by steps")
+		}
```
mrodm, line 101: For now, this service deployer is not supported. I'd prefer to just leave this option (test flags) for stacks/services started locally for now.

This the error that is going to be shown to the user:

```
2024/02/15 09:44:36 DEBUG setting up service...
Error: error running package system tests: could not complete test run: could not create service runner: terraform service deployer not supported to run by steps
```

jsoriano: :+1: we can add this later

internal/testrunner/runners/system/runner.go
```diff
@@ -995,7 +993,7 @@ func (r *runner) prepareScenario(config *testConfig, ctxt servicedeployer.Servic
 }
 
 func (r *runner) removeServiceStateDir() error {
-	dirPath := r.serviceStateFolder
+	dirPath := filepath.Dir(r.serviceStateFilePath)
 	err := os.RemoveAll(dirPath)
```
jsoriano, line 997: Maybe we should remove the file only, in case we use the directory later for other states. Leaving the directory there should not be a problem.

cmd/testrunner.go
```diff
@@ -88,11 +89,32 @@ func setupTestCommand() *cobraext.Command {
...
+			testTypeCmd.Flags().Bool(cobraext.NoProvisionFlagName, false, cobraext.NoProvisionFlagDescription)
+			testTypeCmd.MarkFlagsMutuallyExclusive(cobraext.SetupFlagName, cobraext.TearDownFlagName, cobraext.NoProvisionFlagName)
+			testTypeCmd.MarkFlagsRequiredTogether(cobraext.ConfigFileFlagName, cobraext.SetupFlagName)
+
+			// config file flag should not be used with tear-down or no-provision flags
+			testTypeCmd.MarkFlagsMutuallyExclusive(cobraext.ConfigFileFlagName, cobraext.TearDownFlagName)
+			testTypeCmd.MarkFlagsMutuallyExclusive(cobraext.ConfigFileFlagName, cobraext.NoProvisionFlagName)
+
+			// variant flag should not be used with tear-down or no-provision flags
+			// cannot be defined here using MarkFlagsMutuallyExclusive as in --config-file
```
jsoriano, line 109: This comment seems a bit lonely around here, not clear to what it refers to. Or is another mutual exclusion missing here?

jsoriano (approved):
I think we can ship it :boat: 
Thanks!

cmd/testrunner.go
mrodm: I wanted to add a note that I could not set the mutually exclusive using `testTypeCmd.MarkFlagsMutuallyExclusive`. Variant flag is defined in the `cmd` variable and not in the `testTypeCmd`. 

This is managed afterwards in the code: 
https://github.com/elastic/elastic-package/pull/1250/files#diff-33701d098205d3537f071235876d903ddbfa851c492917e49852c9212bc1d4fdR316-R320

I'll update this comment with that reference

internal/testrunner/runners/system/runner.go
mrodm: That makes totally sense.
Update this function to just delete the file, and update the tests accordingly.

jsoriano (approved):

PR #1256, opened 2023-05-09
Check logs from elastic-agent in system tests

### LINKED ISSUE #1188: [Testing] Fail test if there are errors in Agent component logs
As a developer I want to be able to detect if there are errors occurring during system tests. These are some scenarios I would like to detected as early as possible:

- An event is failing to index (`elastic-agent-20230313-80.ndjson-{"log.level":"warn","@timestamp":"2023-03-13T15:43:25.350Z","message":"Cannot index event publisher.Event{Content:beat.Event{...`). The system test will pass as long as other events are indexed.
- An input is logging warnings, errors, or deprecation warnings.
- Any panics occurring in the Agent during the system test.

I'd love discuss some options for implementing these types of checks for system tests without making a brittle system.

mrodm:
Closes #1188 

This PR adds into system test a new check to look for errors in container logs. These errors could be specified by means of patterns in the code.

A new parameter `--profile` (or `-p`) has been added into `elastic-package test` so it can be retrieved the profile information, required to dump the logs of the services.

This adds a new test result whose Test Name is `(<service> logs)` under the test type `system`.

Example of output setting fake patterns:
```
 $ elastic-package test system -v
...
FAILURE DETAILS:
elastic_package_registry/metrics (elastic-agent logs):
[0] found error "New State ID is AbF6RZN5"
[1] found error "New State ID is kTKFMOWJ"
elastic_package_registry/metrics (package-registry logs):
[0] found error "GET / HTTP/2.0"
[1] found error "GET /package/elastic_package_registry/0.0.6 HTTP/1.1"
[2] found error "GET /package/elastic_package_registry/0.0.6/ HTTP/1.1"
[3] found error "GET /search HTTP/1.1"


╭──────────────────────────┬─────────────┬───────────┬─────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────┬───────────────╮
│ PACKAGE                  │ DATA STREAM │ TEST TYPE │ TEST NAME               │ RESULT                                                                                │  TIME ELAPSED │
├──────────────────────────┼─────────────┼───────────┼─────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────┼───────────────┤
│ elastic_package_registry │ metrics     │ system    │ default                 │ PASS                                                                                  │ 22.331858707s │
│ elastic_package_registry │ metrics     │ system    │ (elastic-agent logs)    │ FAIL: test case failed: one or more errors found while examining elastic-agent.log    │    4.287049ms │
│ elastic_package_registry │ metrics     │ system    │ (package-registry logs) │ FAIL: test case failed: one or more errors found while examining package-registry.log │    85.89657ms │
╰──────────────────────────┴─────────────┴───────────┴─────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────┴───────────────╯
--- Test results for package: elastic_package_registry - END   ---
Done
```

## How to test this locally

In a given package, run the following commands:
```shell
 $ elastic-package build -v
 # start the stack with the required version (e.g. 8.0.0)
 $ elastic-package stack up -v -d --version <version>
 $ elastic-package test system -v
 $ elastic-package stack down
```


cmd/profiles.go
```diff
@@ -235,3 +237,57 @@ func availableProfilesAsAList() ([]string, error) {
 
 	return profileNames, nil
 }
+
+func getProfileFlag(cmd *cobra.Command) (*profile.Profile, error) {
```
mrodm, line 241: Moved this function and `getProviderFromProfile` function from `cmd/stack.go` into this one (cmd/profiles.go), since they look more related to profiles. But it's true that in this file it is not used...

This function will be used also in `cmd/testrunner.go` to get/load the profile to be used in tests.

internal/testrunner/runners/system/runner.go
```diff
@@ -912,3 +944,45 @@ func (r *runner) generateTestResult(docs []common.MapStr) error {
 
 	return nil
 }
+
+func (r *runner) anyErrorMessages(profile *profile.Profile, serviceName string) error {
```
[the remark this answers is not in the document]
jsoriano: Btw, if this is a method of the `runner` struct, it has the profile in `r.options.Profile`, right?

internal/testrunner/runners/system/runner.go
```diff
@@ -51,6 +54,13 @@ const (
 	waitForDataDefaultTimeout = 10 * time.Minute
 )
 
+var errorPatterns = map[string][]string{
+	"elastic-agent": []string{
+		"Cannot index event publisher.Event",
+		"(panic|runtime error)",
```
jsoriano, line 60: Nit. Pre-compile regular expressions here.
```suggestion
		regexp.MustCompile("Cannot index event publisher.Event"),
		regexp.MustCompile("(panic|runtime error)"),
```

internal/testrunner/runners/system/runner.go
```diff
@@ -912,3 +944,45 @@ func (r *runner) generateTestResult(docs []common.MapStr) error {
...
+	if err != nil {
+		return err
+	}
+
+	var multiErr multierror.Error
+
+	for _, pattern := range errorPatterns[serviceName] {
+		r := regexp.MustCompile(pattern)
+
+		for _, log := range logs {
```
jsoriano, line 974: Is it possible that the logs may be poluted by messages of initialization or previous tests? Should we take into account the timestamps, to select logs happened during the execution of the test?

jsoriano (commented):
Approach looks good in general, added some observations and nitpicking.

internal/testrunner/runners/system/runner.go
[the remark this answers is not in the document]
mrodm: > Btw, if this is a method of the `runner` struct, it has the profile in `r.options.Profile`, right?

Removing, I was already using the profile from `r.options`

internal/testrunner/runners/system/runner.go
mrodm: That's true, it could happen if the stack is kept between different runs of tests.
Added logic to just process the logs after the starting time of the system test execution.

internal/testrunner/runners/system/runner.go
```diff
@@ -912,3 +944,45 @@ func (r *runner) generateTestResult(docs []common.MapStr) error {
...
+		return nil
+	}
+
+	tempDir, err := os.MkdirTemp("", "test-system-")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tempDir)
+
+	logs, err := stack.ParseLogs(stack.ParseLogsOptions{
```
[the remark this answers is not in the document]
mrodm: Updated to not keep a list of log messages. They are now processed directly using a function passed as a parameter

cmd/testrunner.go
```diff
@@ -71,6 +72,7 @@ func setupTestCommand() *cobraext.Command {
 	cmd.PersistentFlags().BoolP(cobraext.TestCoverageFlagName, "", false, cobraext.TestCoverageFlagDescription)
 	cmd.PersistentFlags().DurationP(cobraext.DeferCleanupFlagName, "", 0, cobraext.DeferCleanupFlagDescription)
 	cmd.PersistentFlags().String(cobraext.VariantFlagName, "", cobraext.VariantFlagDescription)
+	cmd.PersistentFlags().StringP(cobraext.ProfileFlagName, "p", "", fmt.Sprintf(cobraext.ProfileFlagDescription, install.ProfileNameEnvVar))
```
mrodm, line 75: Added profile parameter into `elastic-package test` command

mrodm:
According to the output in https://github.com/elastic/integrations/pull/6133, a new version of `elastic-package` containing these changes will make fail the system tests in PRs for `system` and `elasticsearch` packages (and daily jobs in integrations).

Would that be a problem? or would it be expected? cc @jlind23 @kpollich @jsoriano

mrodm:
> @mrodm I don't think it is a problem, it means that this packages might have to be fixed. Do you know what the problem is? Could it tests that on purpose logs errors?

I don't know the reason of these errors nor the root cause of them

jlind23:
As soon as we merge we should ask the teams to look at these. You can already ping them btw

internal/testrunner/runners/system/runner.go
```diff
@@ -51,6 +53,40 @@ const (
...
+			// 	excludes: []*regexp.Regexp{
+			// 		regexp.MustCompile("is unahorized API key id"),
+			// 	},
+			// },
+			// logsRegexp{
+			// 	includes: regexp.MustCompile("->HEALTHY"),
+			// 	excludes: []*regexp.Regexp{
+			// 		regexp.MustCompile(`Healthy$`),
+			// 	},
+			// },
```
mrodm, line 82: to be deleted

internal/testrunner/runners/system/runner.go
```diff
@@ -182,6 +219,45 @@ func (r *runner) run() (results []testrunner.TestResult, err error) {
 			}
 		}
 	}
+
+	// check agent logs
```
jsoriano, line 223: Nit. Consider moving this to another function. This may also help to add tests for this.

jsoriano (commented):
This is looking good, could we add some tests that cover the pattern matching logic?

cmd/stack.go
```diff
@@ -239,12 +238,12 @@ func setupStackCommand() *cobraext.Command {
...
 		Short: "Show status of the stack services",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			profile, err := getProfileFlag(cmd)
+			profile, err := cobraext.GetProfileFlag(cmd)
 			if err != nil {
 				return err
 			}
 
-			provider, err := getProviderFromProfile(cmd, profile, false)
+			provider, err := cobraext.GetProviderFromProfile(cmd, profile, false)
```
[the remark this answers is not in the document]
mrodm: I'll like the second approach adding stack in the name. I'll update it.

internal/stack/parselogs.go
```diff
@@ -0,0 +1,72 @@
...
+
+		var log LogLine
+		err := json.Unmarshal([]byte(messageLog), &log)
+		if err != nil {
+			// there are logs that are just plain text in these logs
+			log.Message = strings.TrimSpace(messageLog)
+		}
+
+		// There could be valid messages with just plain text without timestamp
+		if !startProcessing && log.Timestamp.Before(options.StartTime) {
```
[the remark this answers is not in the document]
mrodm: Until now, I haven't noticed any issue when adding this check.

internal/testrunner/runners/system/runner.go
mrodm: Done. I've added tests for this new function. That made me realize that the ordering of logs was not respected when they were shown.

mrodm:
> @mrodm @jsoriano are we good with merging this or is there still some work to do?

@jlind23 there is still a final review pending, but I hope it shouldn't be pending too many changes.
But, considering that this PR would be approved, I have some concerns on merging this.

There are still three issues pending to be fixed (one was already fixed):
- https://github.com/elastic/integrations/issues/6178
- https://github.com/elastic/integrations/issues/6230
- https://github.com/elastic/integrations/issues/6294

If this PR is merged, then any attempt to update elastic-package on integrations repository would be blocked (there would be failures) and even future elastic-package releases.

If that following PR in integrations is merged (those failures are ignored), then all those packages affected would be failing and they could not release any new version of the package.

What should we do here? @jsoriano @jlind23

jlind23:
@mrodm in order to move forward can't we mark these 3 failing packages as an exception until it's fixed? If we wait more some new failures can be introduced in the mean time

mrodm:
> @mrodm in order to move forward can't we mark these 3 failing packages as an exception until it's fixed? If we wait more some new failures can be introduced in the mean time

Not sure how to filter that, packages could fail for other reasons too.

An option (not the best one), they could ask us to merge those PRs...

jsoriano:
> > @mrodm in order to move forward can't we mark these 3 failing packages as an exception until it's fixed? If we wait more some new failures can be introduced in the mean time
> 
> Not sure how to filter that, packages could fail for other reasons too.
> 
> An option (not the best one), they could ask us to merge those PRs...

It is possible to skip system tests (see [docs](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md#test-case-definition)). We could add something like this to their test definition when updating elastic-package in integrations:
```
skip:
  link: https://github.com/elastic/integrations/issues/6178
  reason: "This integration produces events that cannot be indexed"
```
Each one with its link, and include a comment on these issues about the test being skipped.

jsoriano (approved):
:+1:

mrodm:
> It is possible to skip system tests (see [docs](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md#test-case-definition)). We could add something like this to their test definition when updating elastic-package in integrations:
> 
> ```
> skip:
>   link: https://github.com/elastic/integrations/issues/6178
>   reason: "This integration produces events that cannot be indexed"
> ```
> 
> Each one with its link, and include a comment on these issues about the test being skipped.

Thanks @jsoriano !
Then, it is good to be merged now, and we can add those skips in the test system configuration :+1:

PR #1267, opened 2023-05-18
Unit test corpusgenerator

aspacca:
We'll soon release a new version of the corpus generator tool

the the signature of the function used from the dependency didn't change:
new
```
func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields, totEvents uint64) (*GeneratorWithTextTemplate, error) {
```
v0.5.0
```
func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields, totSize uint64) (*GeneratorWithTextTemplate, error) {
```

just the last `uint64` has a different meaning, so nothing will break: just the number of events generated will be different

here we add a unit test to check the expected number of events so when dependabot will create the PR for the bump will fail


mrodm (approved):
LGTM! Thanks for adding this unit test :+1:

Issue #703, opened 2022-02-16
Make terraform outputs available as variables in tests


andrewkroh:
This is a feature request to make the data from `terraform outputs -json` available in system tests. An example use case is a test that creates an S3 bucket and SQS topic and needs information about those resources in the test case.

```hcl
# _dev/deploy/tf/outputs.tf

output "queue_url" {
  value = aws_sqs_queue.test-queue.url
}
```

And then be able to reference those values in some way in test parameters.

```yaml
# _dev/test/system/test-awss3-config.yml

input: aws-s3
data_stream:
  vars:
    access_key_id: "{{{ AWS_ACCESS_KEY_ID }}}"
    secret_access_key: "{{{ AWS_SECRET_ACCESS_KEY }}}"
    queue_url: "{{{ TF_OUTPUT_queue_url }}}"
```

cmacknz:
Take a look at Localstack and see if it will cover this without needing real AWS resources https://docs.localstack.cloud/user-guide/aws/s3/

andrewkroh:
@bhapas has been doing some experimentation. A few key learning points have been:

- Using localstack to emulate AWS S3 and SQS is working as expected for testing purposes.
- Terraform can be used provision the the S3 bucket, SQS queue, and enable `ObjectCreated` notifications in Localstack.
- However, `elastic-package` does not allow using both the Terraform service deployer together with the Docker deployer. We need to be able to use them in combination. We need to system test to be able to start Localstack in docker and then run the Terraform deployer.
- The Filebeat aws-s3 input is not working with the format of the `queue_url` from Localstack. It has some implicit expectations[^1] about the format of the hostname in URL. And `http://localstack:4566/000000000000/s3-event-notification-queue` is not satisfying them.

[^1]: https://github.com/elastic/beats/blob/a927de0fb2463273dbb0126fd9ebdd296a9a26a9/x-pack/filebeat/input/awss3/input.go#L303-L316

bhapas:
The primary goal of this issue was to be able to enable terraform outputs for system test config. An additional issue has been created to track the testing against localstack - https://github.com/elastic/beats/issues/35758 and related issue #1302 to support creation of localstack service along with terraform service.

bhapas:
A bug was identified in the solution and fixed with #1299

PR #1272, opened 2023-05-22
Add Terraform Outputs to ServiceContext to use them in agent policy test config

bhapas:
## What does this PR do

This PR enables the terraform service deployer to generate terraform outputs and use them in the system test config to add terraform related data on agent policy config.

## Why is this important

This enables the integrations developers to test `aws-s3` input to use the terraform generated values like SQS `queue_url` in the system test config to run an agent to poll notifications from the underlying S3 bucket against the SQS Queue.

## How to test this PR

A system test package `aws_logs` been created under `test/packages/parallel`.

- Build elastic-package using `make install`
- Navigate to `test/packages/parallel/aws_logs`
- export `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` into the env variables
- Run `elastic-package install -v`
- Run `elastic-package test -v`

## Screenshots
The `queue_url` is generated by terraform and added into agent policy config using handlebars template in system test config.

Screenshot 2023-05-24 at 13 47 14

## Test Outputs

``` json
{
  "queue_url": {
    "sensitive": false,
    "type": "string",
    "value": "https://sqs.us-east-1.amazonaws.com/<ACCOUNT_ID>/elastic-package-aws-logs-queue-<TEST_RUN_ID>"
  },
  "triangle_output": {
    "sensitive": false,
    "type": [
      "object",
      {
        "description": "string",
        "s_one": "number",
        "s_three": "number",
        "s_two": "number"
      }
    ],
    "value": {
      "description": "this is a triangle",
      "s_one": 1,
      "s_three": 2.5,
      "s_two": 2.5
    }
  }
}
```
To access the variables in agent config
``` yaml
queue_url: {{TF_OUTPUT_queue_url}}

# Accessing complex Object outputs
test_variable_one: {{TF_OUTPUT_triangle_output.s_one}}
test_variable_two: {{TF_OUTPUT_triangle_output.s_two}}
test_variable_three: {{TF_OUTPUT_triangle_output.s_three}}
```

## Related Issues

- Relates #703


internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -42,6 +44,36 @@ type TerraformServiceDeployer struct {
...
+	}
+
+	// Add the terraform outputs to custom properties with the key appeneded by "TF_OUTPUT_"
+	// The values are converted safely to strings
+	for k := range payload {
+		valueMap := payload[k]
+		customProps[terraformOutputPrefix+k] = fmt.Sprint(valueMap["value"])
+	}
+
+	logger.Debugf("Custom properties %s", customProps)
```
bhapas, line 72:
```suggestion
	
```

Shall be removed

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -42,6 +44,36 @@ type TerraformServiceDeployer struct {
 	definitionsDir string
 }
 
+// addTerraformOutputs method reads the terraform outputs generated in the json format and
+// adds them to the custom properties of ServiceContext and can be used in the handlebars template
+// like `{{TF_OUTPUT_queue_url}}` where `queue_url` is the output configured
+func addTerraformOutputs(customProps map[string]interface{}) error {
```
andrewkroh, line 50: Can you add a unit test for the function. If this accepted a file path as an argument then you could more easily construct a unit test for it.

andrewkroh (commented):
Nice. It wasn't too complicated to add. Most of the work was in testing.

docs/howto/system_testing.md
```diff
@@ -262,6 +262,37 @@ data "aws_ami" "latest-amzn" {
 
 Notice the use of the `TEST_RUN_ID` variable. It contains a unique ID, which can help differentiate resources created in potential concurrent test runs.
 
+#### Terraform Outputs
+
+The outputs generated by the terraform service deployer can be accessed in the system test config using handlebars template.
+Ex:
```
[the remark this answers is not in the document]
bhapas: Not really :-)

docs/howto/system_testing.md
```diff
@@ -262,6 +262,37 @@ data "aws_ami" "latest-amzn" {
 
 Notice the use of the `TEST_RUN_ID` variable. It contains a unique ID, which can help differentiate resources created in potential concurrent test runs.
 
+#### Terraform Outputs
+
+The outputs generated by the terraform service deployer can be accessed in the system test config using handlebars template.
```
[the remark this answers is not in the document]
bhapas: Yes..
## Test Outputs

``` json
{
  "queue_url": {
    "sensitive": false,
    "type": "string",
    "value": "https://sqs.us-east-1.amazonaws.com/<ACCOUNT_ID>/elastic-package-aws-logs-queue-<TEST_RUN_ID>"
  },
  "triangle_output": {
    "sensitive": false,
    "type": [
      "object",
      {
        "description": "string",
        "s_one": "number",
        "s_three": "number",
        "s_two": "number"
      }
    ],
    "value": {
      "description": "this is a triangle",
      "s_one": 1,
      "s_three": 2.5,
      "s_two": 2.5
    }
  }
}
```
Accessing in config would be
``` yaml
queue_url: {{TF_OUTPUT_queue_url}}

# Accessing complex Object outputs
test_variable_one: {{TF_OUTPUT_triangle_output.s_one}}
test_variable_two: {{TF_OUTPUT_triangle_output.s_two}}
test_variable_three: {{TF_OUTPUT_triangle_output.s_three}}
```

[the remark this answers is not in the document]
bhapas: Yes... This should work now

internal/testrunner/runners/system/servicedeployer/terraform.go
bhapas: Fixed..

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -42,6 +44,36 @@ type TerraformServiceDeployer struct {
...
+	// Read the `output.json` file where terraform outputs are generated
+	content, err := os.ReadFile("/tmp/output.json")
+	if err != nil {
+		return errors.Wrap(err, "Unable to read the file output.json")
+	}
+
+	// Unmarshall the data into `payload`
+	logger.Debug("Unmarshalling terraform output json")
+	var payload map[string]map[string]interface{}
```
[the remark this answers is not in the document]
bhapas: Somehow it was not possible to handle complex outputs using mapstr.Flatten [function](https://github.com/elastic/elastic-agent-libs/blob/bb53a3b792218bf6f302c9f55736d5746b8e644b/mapstr/mapstr.go#L266-L277)

So handled in a nicer way using the same interface

internal/testrunner/runners/system/servicedeployer/terraform_test.go
```diff
@@ -0,0 +1,132 @@
...
+package servicedeployer
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFindPolicyTemplateForInput(t *testing.T) {
```
bhapas, line 14:
```suggestion
func TestAddTerraformOutputs(t *testing.T) {
```

internal/testrunner/runners/system/servicedeployer/terraform_test.go
```diff
@@ -0,0 +1,132 @@
...
+						"s_one": 1,
+						"s_three": 2.5,
+						"s_two": 2.5
+					  }
+					}
+				}`,
+			),
+			expectedProps: map[string]interface{}{
+				"TF_OUTPUT_queue_url":                   "https://sqs.us-east-1.amazonaws.com/1234654/elastic-package-aws-logs-queue-someId",
+				"TF_OUTPUT_triangle_output.s_one":       1.0,
```
andrewkroh, line 103: Why are these expected to be flattened? Handlebars supports[^1][^2] addressing nested values through its path expressions. So in theory if you write `{{TF_OUTPUT_triangle_output.s_one}}` then it should return the nested value without needed to store it as flattened. If Handlebars is not working as expected then I would like to try to solve that issue.

[^1]: https://handlebarsjs.com/guide/expressions.html#path-expressions
[^2]: https://github.com/aymerick/raymond/blob/b565731e1464263de0bda75f2e45d97b54b60110/handlebars/basic_test.go#L385-L386

This is what I was thinking...

```diff
diff --git a/internal/testrunner/runners/system/servicedeployer/terraform.go b/internal/testrunner/runners/system/servicedeployer/terraform.go
index 07322e5..5dd2911 100644
--- a/internal/testrunner/runners/system/servicedeployer/terraform.go
+++ b/internal/testrunner/runners/system/servicedeployer/terraform.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/pkg/errors"
 
-	"github.com/elastic/elastic-agent-libs/mapstr"
 	"github.com/elastic/go-resource"
 
 	"github.com/elastic/elastic-package/internal/compose"
@@ -59,25 +58,24 @@ func addTerraformOutputs(outCtxt ServiceContext) error {
 	// Unmarshall the data into `terraformOutputs`
 	logger.Debug("Unmarshalling terraform output json")
 
-	var terraformOutputs map[string]interface{}
-	err = json.Unmarshal(content, &terraformOutputs)
-	if err != nil {
-		return fmt.Errorf("error during json Unmarshal %w", err)
+	// https://github.com/hashicorp/terraform/blob/v1.4.6/internal/command/views/output.go#L217-L222
+	type OutputMeta struct {
+		Value interface{} `json:"value"`
 	}
+	var outputMetas map[string]OutputMeta
 
-	// Flatten if there are any objects in the output json
-	outputs := (mapstr.M).Flatten(terraformOutputs)
-	customProps := outCtxt.CustomProperties
+	if err = json.Unmarshal(content, &outputMetas); err != nil {
+		return fmt.Errorf("error during json Unmarshal %w", err)
+	}
 
-	// Considering only values from the output json
-	// https://github.com/hashicorp/terraform/blob/v1.4.6/internal/command/views/output.go#L217-L222
-	// Add the terraform outputs to custom properties with the key appeneded by "TF_OUTPUT_"
-	for k, v := range outputs {
-		if strings.Contains(k, ".value") {
-			newKey := strings.Replace(k, ".value", "", 1)
-			customProps[terraformOutputPrefix+newKey] = v
-		}
+	if len(outputMetas) > 0 && outCtxt.CustomProperties == nil {
+		outCtxt.CustomProperties = make(map[string]any, len(outputMetas))
 	}
+	// Prefix variables names with TF_OUTPUT_
+	for k, om := range outputMetas {
+		outCtxt.CustomProperties[terraformOutputPrefix+k] = om.Value
+	}
+
 	return nil
 }
 
diff --git a/internal/testrunner/runners/system/servicedeployer/terraform_test.go b/internal/testrunner/runners/system/servicedeployer/terraform_test.go
index 435038d..117c162 100644
--- a/internal/testrunner/runners/system/servicedeployer/terraform_test.go
+++ b/internal/testrunner/runners/system/servicedeployer/terraform_test.go
@@ -99,11 +99,13 @@ func TestFindPolicyTemplateForInput(t *testing.T) {
 				}`,
 			),
 			expectedProps: map[string]interface{}{
-				"TF_OUTPUT_queue_url":                   "https://sqs.us-east-1.amazonaws.com/1234654/elastic-package-aws-logs-queue-someId",
-				"TF_OUTPUT_triangle_output.s_one":       1.0,
-				"TF_OUTPUT_triangle_output.s_two":       2.5,
-				"TF_OUTPUT_triangle_output.s_three":     2.5,
-				"TF_OUTPUT_triangle_output.description": "this is a triangle",
+				"TF_OUTPUT_queue_url": "https://sqs.us-east-1.amazonaws.com/1234654/elastic-package-aws-logs-queue-someId",
+				"TF_OUTPUT_triangle_output": map[string]any{
+					"description": "this is a triangle",
+					"s_one":       1.0,
+					"s_two":       2.5,
+					"s_three":     2.5,
+				},
 			},
 		},
 	}
```

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -178,6 +178,12 @@ func (s *dockerComposeDeployedService) TearDown() error {
 	}); err != nil {
 		return errors.Wrap(err, "could not shut down service using Docker Compose")
 	}
+
+	// Remove the outputs generated by the service container
+	if err := os.RemoveAll("/tmp/" + s.ctxt.Test.RunID); err != nil {
```
andrewkroh, line 183: I suggest moving this into the `defer` block up near L160 so that this is cleaned up even on error.

internal/testrunner/runners/system/servicedeployer/terraform_env.go
```diff
@@ -25,6 +28,9 @@ func (tsd TerraformServiceDeployer) buildTerraformExecutorEnvironment(ctxt Servi
 	vars[tfTestRunID] = ctxt.Test.RunID
 	vars[tfDir] = tsd.definitionsDir
 
+	os.MkdirAll(outputDir+ctxt.Test.RunID, os.ModePerm)
```
andrewkroh, line 31: Can you store this directory path within the `ServiceContext`. I see this path being recreated in a multiple places and it would more clear and less prone to future errors if it were constructed and stored in just one place.

This is missing an error check.

Why `os.ModePerm`? Can we use `0o755`?

internal/testrunner/runners/system/servicedeployer/terraform_env.go
```diff
@@ -7,14 +7,17 @@ package servicedeployer
...
 	"strings"
 
 	"github.com/elastic/elastic-package/internal/compose"
 )
 
 const (
 	tfDir       = "TF_DIR"
+	tfOutputDir = "TF_OUTPUT_DIR"
 	tfTestRunID = "TF_VAR_TEST_RUN_ID"
+	outputDir   = "/tmp/"
```
andrewkroh, line 20: Prefer `os.TempDir()` to avoid any OS specific paths.

internal/testrunner/runners/system/servicedeployer/terraform_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	t.Parallel()
+	for _, tc := range testCases {
+		tc := tc
+
+		t.Run(tc.testName, func(t *testing.T) {
+			tc.ctxt.CustomProperties = make(map[string]interface{})
+
+			os.Mkdir("/tmp/"+tc.runId, os.ModePerm)
```
andrewkroh, line 118: You can use https://pkg.go.dev/testing#B.TempDir and it will take care of cleaning up after the test is done.

internal/testrunner/runners/system/servicedeployer/terraform_test.go
```diff
@@ -0,0 +1,132 @@
...
+			file, _ := os.OpenFile("/tmp/"+tc.runId+"/tfOutputValues.json", os.O_CREATE|os.O_WRONLY, 0777)
+			file.Write(tc.content)
```
andrewkroh, line 120: Use https://pkg.go.dev/os#WriteFile. And check for errors.

internal/testrunner/runners/system/servicedeployer/terraform_test.go
bhapas: Used https://pkg.go.dev/testing@go1.20.4#T.TempDir instead

internal/testrunner/runners/system/servicedeployer/terraform_test.go
bhapas: Looks like Handlebars is handling this already.. Do not have to Flatten stuff

internal/testrunner/runners/system/servicedeployer/terraform_env.go
bhapas: @andrewkroh This seems to be an issue in CI. The buildkite user does not have permissions to create new directories in `os.TempDir()`. Using `build` directory instead?

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -117,10 +158,28 @@ func (tsd TerraformServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedServic
...
+	}
 	service.ctxt = outCtxt
 	return &service, nil
 }
 
+func createOutputDir(runId string) (string, error) {
+	buildDir, err := builder.BuildDirectory()
+	if err != nil {
+		return "", fmt.Errorf("locating build directory failed %w", err)
+	}
```
mrodm, line 174: Instead of using `builder.BuildDirectory()`, it would be better to add a new method into the locationManager specific for this use case.

It should be something like:
https://github.com/elastic/elastic-package/blob/59e4fa3fd20e417fe712b9b08c1c717ac1f6ad05/internal/configuration/locations/locations.go#L89

In that way, that location would be under the stack path.

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -117,10 +158,28 @@ func (tsd TerraformServiceDeployer) SetUp(inCtxt ServiceContext) (DeployedServic
...
+
+	od := filepath.Join(buildDir + runId)
+	if err := os.MkdirAll(od, 0755); err != nil {
+		return "", fmt.Errorf("failed to create output directory: %w", err)
+	}
+	return od, nil
```
mrodm, line 180: nit. instead of concatenating the strings in the filepath.Join, let the method to concatenate both to ensure everything is ok.

And I would use complete words for the variables:

```suggestion
	outputDir := filepath.Join(buildDir, runId)
	if err := os.MkdirAll(outputDir, 0755); err != nil {
		return "", fmt.Errorf("failed to create output directory: %w", err)
	}
	return outputDir, nil
```

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -42,6 +46,38 @@ type TerraformServiceDeployer struct {
...
+		return fmt.Errorf("error during json Unmarshal %w", err)
+	}
+
+	if len(terraformOutputs) > 0 && outCtxt.CustomProperties == nil {
+		outCtxt.CustomProperties = make(map[string]any, len(terraformOutputs))
+	}
+	// Prefix variables names with TF_OUTPUT_
+	for k, outputs := range terraformOutputs {
+		outCtxt.CustomProperties[terraformOutputPrefix+k] = outputs.Value
+	}
```
mrodm, line 77: Maybe it can be return early if there are no terraformOutputs

```suggestion
	if len(terraformOutputs) == 0 {
		return nil
	}
	if outCtxt.CustomProperties == nil {
		outCtxt.CustomProperties = make(map[string]any, len(terraformOutputs))
	}
	// Prefix variables names with TF_OUTPUT_
	for k, outputs := range terraformOutputs {
		outCtxt.CustomProperties[terraformOutputPrefix+k] = outputs.Value
	}
```

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -42,6 +46,38 @@ type TerraformServiceDeployer struct {
 	definitionsDir string
 }
 
+// addTerraformOutputs method reads the terraform outputs generated in the json format and
+// adds them to the custom properties of ServiceContext and can be used in the handlebars template
+// like `{{TF_OUTPUT_queue_url}}` where `queue_url` is the output configured
+func addTerraformOutputs(outCtxt ServiceContext) error {
+	// Read the `output.json` file where terraform outputs are generated
+	content, err := os.ReadFile(outCtxt.OutputDir + terraformOutputJsonFile)
```
mrodm, line 54: It's better to use filepath.Join to ensure the paths are generated correctly.

```suggestion
	terraformOutputFile := filepath.Join(outCtxt.OutputDir, terraformOutputJsonFile)
	content, err := os.ReadFile(terraformOutputFile)
```

I think doing so, maybe this variable `terraformOutputJsonFile` needs to remove the `/` at the beginning (to be tested):
```
    terraformOutputJsonFile = "tfOutputValues.json"
```

internal/testrunner/runners/system/servicedeployer/compose.go
```diff
@@ -158,6 +158,10 @@ func (s *dockerComposeDeployedService) TearDown() error {
 		if err != nil {
 			logger.Errorf("could not remove the service logs (path: %s)", s.ctxt.Logs.Folder.Local)
 		}
+		// Remove the outputs generated by the service container
+		if err = os.RemoveAll(s.ctxt.OutputDir); err != nil {
+			logger.Errorf("could not remove the temporary output files %w", err)
+		}
```
mrodm, line 164: Should this be run just when `s.ctxt.OutputDir` is different from empty string ?

bhapas: If it is an empty string it does nothing 
https://github.com/golang/go/blob/094c75219a160be1667c9757f363ffad8926632b/src/os/removeall_at.go#L15-L20

internal/testrunner/runners/system/servicedeployer/terraform.go
bhapas: There is not any notable computing that is happening if `terraformOutputs` is empty.

But yeah , May be more modular and cleaner.

internal/testrunner/runners/system/servicedeployer/terraform_env.go
bhapas: Changed this as per suggestion [here](https://github.com/elastic/elastic-package/pull/1272#discussion_r1211694536)

andrewkroh (approved):
LGTM

mrodm (approved):
LGTM :+1:

PR #1273, opened 2023-05-22
Some UI improvements around current profile

jsoriano:
`elastic-package profiles use` can be used to select the profile that will be used by default. This change includes some UI improvements around it.

* Current profile is shown now in `elastic-package profiles list`.
* If current profile is deleted, current profile is replaced by the default profile, that cannot be removed.
* A confirmation message is shown now after changing the current profile.

How to test
---
* Create a new profile (`elastic-package profiles create test`).
* List profiles, and check that you can see what is the current profile (`elastic-package profiles list`).
* Select the created profile as the current one (`elastic-package profiles use test`).
* List profiles, and check that you can see that your profile is the current profile (`elastic-package profiles list`).
* Delete the current profile (`elastic-package profiles delete test`). 
* List profiles, and check that the default profile is the current profile (`elastic-package profiles list`).

### DIFF
--- a/cmd/profiles.go
+++ b/cmd/profiles.go
@@ -63,9 +63,9 @@ User profiles can be configured with a "config.yml" file in the profile director
 			}
 
 			if fromName == "" {
-				fmt.Printf("Created profile %s.\n", newProfileName)
+				fmt.Printf("Created profile %q.\n", newProfileName)
 			} else {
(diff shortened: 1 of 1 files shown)


mrodm (approved):
:rocket: Tested locally :+1:

PR #1281, opened 2023-05-26
Explicitly install packages on system tests

jsoriano:
This will allow to earlier detect if one is not testing the version they currently have in the working copy.

From 8.7.0 version, it builds and install the zip package. On older versions it installs whatever is in the configured package registry. In that case, if the version of the package is not available in the repository, test execution will fail.

Also removes uses of `errors.Wrap` in the file.

It won't be tested in CI till stack version is updated to >= 8.7.0.


internal/testrunner/runners/system/runner.go
```diff
@@ -519,14 +529,50 @@ func checkEnrolledAgents(client *kibana.Client, ctxt servicedeployer.ServiceCont
...
-		return nil, errors.Wrap(err, "agent enrollment failed")
+		return nil, fmt.Errorf("agent enrollment failed: %w", err)
 	}
 	if !enrolled {
 		return nil, errors.New("no agent enrolled in time")
 	}
 	return agents, nil
 }
 
+func installPackage(client *kibana.Client, manifest *packages.PackageManifest, rootPath string) error {
```
mrodm, line 540: This looks similar to the factory created in `cmd/install.go` file. Do you think it would be possible to reuse some of the code? Maybe the installer code: 
- for local packages: internal/packages/installer/installer.go
- for zip packages: internal/packages/installer/zip_installer.go

https://github.com/elastic/elastic-package/tree/main/internal/packages/installer

internal/testrunner/runners/system/runner.go
```diff
@@ -328,12 +330,20 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+		return result.WithError(fmt.Errorf("can't create Kibana client: %w", err))
+	}
+
+	// Install the package before creating the policy, so we control exactly what is being
+	// installed.
+	logger.Debug("Installing package...")
+	err = installPackage(kib, pkgManifest, r.options.PackageRootPath)
+	if err != nil {
+		return result.WithError(fmt.Errorf("unable to install package: %w", err))
 	}
```
mrodm, line 347: Should it be uninstalled the package using the kibana client at the end of the test run too?

internal/testrunner/runners/system/runner.go
jsoriano: I have merged both codes in a single factory in the `installer` package, let me know what you think.

I am also passing the kibana client from the callers, I would like to have more explicit initializations of the clients we use. This is something I have WIP to be able to get the settings from the profile in this branch https://github.com/elastic/elastic-package/compare/main...jsoriano:elastic-package:clients-from-profile?expand=1.

internal/testrunner/runners/system/runner.go
jsoriano: Added. Good point.

cmd/install.go
```diff
@@ -60,23 +51,27 @@ func installCommandAction(cmd *cobra.Command, _ []string) error {
...
+		var found bool
+		var err error
+		packageRootPath, found, err = packages.FindPackageRoot()
+		if !found {
+			return errors.New("package root not found")
+		}
+		if err != nil {
+			return errors.Wrap(err, "locating package root failed")
+		}
 	}
```
mrodm, line 65: As it is now the tests, if it is used the zip parameter this could be executed outside of the package folder:
- https://github.com/elastic/elastic-package/blob/main/scripts/test-install-zip.sh#L119

Because of that, it is possible that it is not found the package root.

Should it be added another test to install zip directly from the package folder?

internal/packages/installer/factory.go
```diff
@@ -0,0 +1,99 @@
...
+// If a zip path is given, this package is validated and installed as is. This fails on versions
+// of Kibana lower than 8.7.0.
+// When no zip is given, package is built as zip and installed if version is at least 8.7.0,
+// or from the package registry otherwise.
+func NewForPackage(options Options) (Installer, error) {
+	if options.Kibana == nil {
+		return nil, errors.New("missing kibana client")
+	}
+	if options.RootPath == "" && options.ZipPath == "" {
+		return nil, errors.New("missing package root path and pre-built zip package")
```
mrodm, line 49: As setting one of those is valid, I would update the error message to be "or"

```suggestion
		return nil, errors.New("missing package root path or pre-built zip package")
```

internal/testrunner/runners/system/runner.go
mrodm: > I have merged both codes in a single factory in the installer package, let me know what you think.

It looks perfect! :muscle: thanks for doing this change!


> I am also passing the kibana client from the callers, I would like to have more explicit initializations of the clients we use. This is something I have WIP to be able to get the settings from the profile in this branch https://github.com/elastic/elastic-package/compare/main...jsoriano:elastic-package:clients-from-profile?expand=1.

Ok! that would also help us to add more tests if the kibana client is set as a parameter :+1:

cmd/install.go
mrodm: I see that you already fixed this :+1:

mrodm (approved):
Thanks! This will make easier testing changes while developing new version of packages. No need to build and restart package-registry container to get the new version of the package, nice!

Just left one minor suggestion for an error message.

internal/testrunner/runners/system/runner.go
```diff
@@ -395,12 +403,35 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+		SkipValidation: true,
+	})
+	if err != nil {
+		return result.WithError(fmt.Errorf("failed to initialize package installer: %v", err))
+	}
+	_, err = installer.Install()
+	if err != nil {
+		return result.WithError(fmt.Errorf("failed to install package: %v", err))
+	}
+	r.deletePackageHandler = func() error {
```
jsoriano, line 429: I couldn't just use `defer` here, I needed to use a handler to play well with the tear downs here.

mrodm (approved):
:+1:

PR #1274, opened 2023-05-22
Update default Elastic Stack to 8.7.1

jsoriano:

PR #1283, opened 2023-05-30
Use fields if synthetic mode is enabled

### LINKED ISSUE #1253: Support fields validation when synthetic source is used
Current fields validation is based on `_source`, but when using synthetic source, the synthesized `_source` may look different to the original one, what is causing issues like [this one](https://github.com/elastic/integrations/issues/6099) when using TSDB.

We should improve fields validation on system tests to support cases like this one, maybe basing it on the fields API.

We have a [similar issue with runtime fields](https://github.com/elastic/elastic-package/issues/1229), where fields are not in the source.

mrodm:
Closes #1253

This PR adds the support to check `fields` key from the documents retrieved from the Elasticsearch instead of `_source`.
For that, for each system test it is checked whether or not the `synthetic mode` is enabled:
- This value is retrieved querying to the Component templates API

In case that data stream (component template) has synthetic enabled, there are some changes required:
- validation in system tests:
    - disabled Normalization
        - This normalization was introduced in #765
        - If synthetic is used, all elements in `fields` are arrays and this validation cannot be ensured.
    - update validation about dataset
        - all the elements in `fields` are arrays, and that's need to be taken into account when checking for exact values
- update how sample event files are generated (so static tests keep working):
    - when `synthetics` source mode is enabled, this file is generated using the `fields` key from the documents with some modifications
        - if normalization is not set and there is just one element in the field, change the field to be just an scalar value.
        - if normalization is set, keep the the field as array

A test package has been included to test the synthetic mode:
- `sample_event.json` is written using the `fields` key from the document
- until 8.7.1 is not used as default version, synthetics is not going to be used

As part of this PR:
- remove uses of errors.Wrap in the `internal/fields/validate.go`.
- add a simple test for GetValue method from common.MapStr.

## Considerations

As synthetic source mode is enabled (as part of the time series), `_source` field gets built from the documents and the fields on the document could be differnt.

For that, in case `synthetic` source mode is enabled, it is used the `fields` key from the documents retrieved from Elasticsearch instead of the `_source` field.

**This makes that `sample_event.json` files must be re-generated for those data streams.**

Packages (data streams) that do not use this mode are tested using the same `_source` field. No changes for them.

## How to test

For this it can be used the `apache_tomcat` from the [integrations repository](https://github.com/elastic/integrations)

1. Build elastic-package from this branch
2. Go to the `apache_tomcat` package folder
3. Update the manifest of the `cache` datastream to set the `time_series` mode. Add at the end of the manifest this definition:
   ```yaml
   elasticsearch:
     index_mode: time_series
4. Run the following commands:
   ```shell
   elastic-package stack down
   elastic-package build -v
   elastic-package stack up -v -d --version 8.7.1
   # regenerate sample_event.json files for the desired data streams
   elastic-package test system -v -g --data-streams cache,requests
   elastic-package test -v
   ```


internal/testrunner/runners/system/runner.go
```diff
@@ -228,11 +229,85 @@ func createTestRunID() string {
...
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&results); err != nil {
+		return false, fmt.Errorf("could not decode search results response: %w", err)
+	}
+
+	if len(results.ComponentTemplates) == 0 {
+		logger.Debugf("no component template found for data stream %s", dataStream)
+		return false, nil
+	}
```
mrodm, line 274: In current elastic stack version, there is no component template for `sql_input` package (input) for the Elastic stack version started (8.6.2).
 
In that scenario, I assume that `synthetic` is not enabled.

mrodm:
The test package fails with these errors:
```
synthetic_mode/synthetic Verify sample_event.json:
[0] field "event.dataset" should have value "synthetic_mode.synthetic", it has "[synthetic_mode.synthetic]"
[1] field "data_stream.dataset" should have value "synthetic_mode.synthetic", it has "[synthetic_mode.synthetic]"
```

Probably it needs to wait until elastic-package runs by default 8.7.1 version so the synthetic source mode is detected.

jlind23:
> Probably it needs to wait until elastic-package runs by default 8.7.1 version so the synthetic source mode is detected.
Shall we update the stack version being used then?

mrodm:
> > Probably it needs to wait until elastic-package runs by default 8.7.1 version so the synthetic source mode is detected.
> 
> Shall we update the stack version being used then?

It is being done in this PR https://github.com/elastic/elastic-package/pull/1274
We could wait for this.

jlind23:
> It is being done in this PR https://github.com/elastic/elastic-package/pull/1274
We could wait for this.

I guess we are good now :)

mrodm:
> > It is being done in this PR #1274
> > We could wait for this.
> 
> I guess we are good now :)

There are still errors to be fixed in field validations when running static tests:
```
Run static tests for the package
--- Test results for package: synthetic_mode - START ---
FAILURE DETAILS:
synthetic_mode/synthetic Verify sample_event.json:
[0] field "event.dataset" should have value "synthetic_mode.synthetic", it has "[synthetic_mode.synthetic]"
[1] field "data_stream.dataset" should have value "synthetic_mode.synthetic", it has "[synthetic_mode.synthetic]"


╭────────────────┬─────────────┬───────────┬──────────────────────────┬────────────────────────────────────────────┬──────────────╮
│ PACKAGE        │ DATA STREAM │ TEST TYPE │ TEST NAME                │ RESULT                                     │ TIME ELAPSED │
├────────────────┼─────────────┼───────────┼──────────────────────────┼────────────────────────────────────────────┼──────────────┤
│ synthetic_mode │ synthetic   │ static    │ Verify sample_event.json │ FAIL: one or more errors found in document │  53.778195ms │
╰────────────────┴─────────────┴───────────┴──────────────────────────┴────────────────────────────────────────────┴──────────────╯
```

Trying to find a way to solve that yet.

jlind23:
@mrodm can I also mark this as closing https://github.com/elastic/integrations/issues/6099 ?

mrodm:
> @mrodm can I also mark this as closing [elastic/integrations#6099](https://github.com/elastic/integrations/issues/6099) ?

@jlind23 Not sure if it could be considered closed. If the current approach is approved, some changes in that integration would be needed so its tests are successful:
- at least, it would be needed to update `sample_event.json` file for the data streams that require `synthetic` mode.

internal/testrunner/runners/system/runner.go
```diff
@@ -533,6 +616,14 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
 		return result.WithError(fmt.Errorf("%s", result.FailureMsg))
 	}
 
+	syntheticEnabled, err := r.isSyntheticsEnabled(dataStream, componentTemplatePackage)
+	if err != nil {
+		return result.WithError(err)
```
jsoriano, line 621: Nit. Add some context on this error.
```suggestion
		return result.WithError(fmt.Errorf("failed to check if synthetic source is enabled", err))
```

internal/testrunner/runners/system/runner.go
```diff
@@ -295,11 +296,85 @@ func createTestRunID() string {
...
+	Source []common.MapStr `json:"_source"`
+	Fields []common.MapStr `json:"fields"`
+}
+
+func (h hits) getDocs(syntheticsEnabled bool) []common.MapStr {
+	if syntheticsEnabled {
+		return h.Fields
+	}
+	return h.Source
+}
```
jsoriano, line 365: This is probably fine for synthetics, but I wonder if we should actually merge fields and source, to support also the runtime cases. But we can think about this when supporting tests with runtime fields.

jsoriano (commented):
Good job here, there are quite some tricky cases.

I wonder if we can make this code less dependant on synthetic mode being enabled or not. For example we could try to always merge the fields in the source and the fields in `fields`. And assume that we have to disable normalization validation for all fields coming from `fields`. Then maybe we wouldn't even need to check if synthetic is enabled. But we can revisit this later when supporting runtime fields.

Something I think we have to confirm in any case before merging this is the format of the sample event, objects used to be expanded, but they are not here for the synthetic mode.

internal/fields/validate.go
```diff
@@ -273,9 +282,28 @@ func (v *Validator) validateDocumentValues(body common.MapStr) multierror.Error
...
+				}
+				if err == nil && len(vals) != 1 {
+					ok = false
+				}
+				if err == nil && len(vals) == 1 {
+					str = vals[0]
+				}
+			} else {
+				str, ok = value.(string)
+			}
```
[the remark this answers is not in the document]
mrodm: Better, with that function. Thanks!

internal/testrunner/runners/system/runner.go
```diff
@@ -295,11 +296,85 @@ func createTestRunID() string {
...
-func (r *runner) getDocs(dataStream string) ([]common.MapStr, error) {
+func buildAllFieldsBody() io.Reader {
+	var b strings.Builder
+	b.WriteString("\n")
+	b.WriteString(`{
+	  "fields": ["*"]
+	}`)
+	b.WriteString("\n")
+	return strings.NewReader(b.String())
+}
```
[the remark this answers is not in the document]
mrodm: Tested as:
```
const allFieldsBody = `{"fields": ["*"]}`
```

and it works, no need to have that function for just that. Updated :+1: 
Thanks!

test/packages/other/synthetic_mode/data_stream/synthetic/sample_event.json
```diff
@@ -0,0 +1,75 @@
+{
+    "@timestamp": "2023-06-02T15:21:39.436Z",
+    "agent.ephemeral_id": "d5a061f5-0aac-4c11-8496-7f2c42f2393c",
+    "agent.id": "bc2c6d4c-79fd-4368-ae8b-540913be01fe",
```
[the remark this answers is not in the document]
mrodm: I did some quick tests using `go-ucfg`, and I found that there are some issues when there are multi-fields.

For instance, in the test package (that uses `import_mappings`), these two fields exist in the `fields` key in the document:
```
                    "host.os.name.text": [
                        "Ubuntu"
                    ],
                    "host.os.name": [
                        "Ubuntu"
                    ],
```
and that could not be loaded into a JSON. It fails with this error:
> duplicate key accessing 'host.os.name'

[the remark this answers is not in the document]
mrodm: Added a new method to expand the keys to write the sample_event.json keeping the same format as before.

This is executed just when synthetic is enabled.

internal/fields/validate.go
```diff
@@ -345,18 +370,109 @@ func (v *Validator) validateScalarElement(key string, val interface{}, doc commo
...
+		}
+
+		// Possible errors found but not limited to those
+		// - expected map but type is string
+		// - expected map but type is []interface{}
+		if strings.HasPrefix(err.Error(), "expected map but type is") {
+			logger.Warnf("not able to add key %s, is this a multifield?: %s", k, err)
+			continue
+		}
+		return nil, fmt.Errorf("not added key %s with value %s: %w", k, value, err)
```
mrodm, line 472: Examples found in my local tests:

```
2023/06/07 10:41:26  WARN not able to add key host.os.name.text, is this a multifield?: expected map but type is string
2023/06/07 10:41:26  WARN not able to add key metricset.name.text, is this a multifield?: expected map but type is []interface {}
```

As there could be other keys for the multifield a part from `text` I didn't want to hardcode that

test/packages/other/synthetic_mode/data_stream/synthetic/sample_event.json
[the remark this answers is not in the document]
mrodm: Example of file generated from the fields found in the `fields` key of the document:


```json
{
    "@timestamp": "2023-06-07T08:30:02.333Z",
    "agent": {
        "ephemeral_id": "296ade19-2a9f-47f6-95f9-dcfa40559fd0",
        "id": "86e09a11-1c59-4cb9-9eda-2b025a06143e",
        "name": "docker-fleet-agent",
        "type": "metricbeat",
        "version": "8.7.1"
    },
    "data_stream": {
        "dataset": "synthetic_mode.synthetic",
        "namespace": "ep",
        "type": "metrics"
    },
    "ecs": {
        "version": "8.7.0"
    },
    "elastic_agent": {
        "id": [
            "86e09a11-1c59-4cb9-9eda-2b025a06143e"
        ],
        "snapshot": [
            false
        ],
        "version": [
            "8.7.1"
        ]
    },
    "event": {
        "agent_id_status": "verified",
        "category": [
            "web"
        ],
        "dataset": "synthetic_mode.synthetic",
        "duration": 7261724,
        "ingested": "2023-06-07T08:30:06Z",
        "kind": "metric",
        "module": "synthetic_mode",
        "type": [
            "info"
        ]
    },
    "host": {
        "architecture": "x86_64",
        "containerized": [
            false
        ],
        "hostname": "docker-fleet-agent",
        "id": "fd2c4b0943e444508c12855a04d117c7",
        "ip": [
            "172.26.0.7"
        ],
        "mac": [
            "02-42-AC-1A-00-07"
        ],
        "name": "docker-fleet-agent",
        "os": {
            "codename": [
                "focal"
            ],
            "family": "debian",
            "kernel": "5.19.0-42-generic",
            "name": "Ubuntu",
            "platform": "ubuntu",
            "type": "linux",
            "version": "20.04.6 LTS (Focal Fossa)"
        }
    },
    "metricset": {
        "name": [
            "collector"
        ],
        "period": [
            10000
        ]
    },
    "prometheus": {
        "labels": {
            "branch": "HEAD",
            "goversion": "go1.18.3",
            "instance": "elastic-package-service-synthetic_mode-1:9090",
            "job": "prometheus",
            "revision": "d7e7b8e04b5ecdc1dd153534ba376a622b72741b",
            "version": "2.36.2"
        },
        "prometheus_build_info": {
            "value": 1
        }
    },
    "service": {
        "address": "http://elastic-package-service-synthetic_mode-1:9090/metrics",
        "type": "prometheus"
    }
}
```

internal/fields/validate.go
jsoriano: I would say that these errors about "expected map" are expected with multi-fields, I think we can log them at the debug level, or don't log anything, but don't log them at the warn level.

test/packages/other/synthetic_mode/data_stream/synthetic/sample_event.json
```diff
@@ -0,0 +1,94 @@
...
+    "elastic_agent": {
+        "id": [
+            "86e09a11-1c59-4cb9-9eda-2b025a06143e"
+        ],
+        "snapshot": [
+            false
+        ],
+        "version": [
+            "8.7.1"
+        ]
```
jsoriano, line 27: Umm, why are these values left as arrays? I don't think they have any normalization.

Is it because they don't have a definition (from [here](https://github.com/elastic/elastic-package/pull/1283/files#diff-36d325d58b4ea8957abc63576d2e04d30ec8e58435ba06d442cb7923af063ad3R392))? Maybe for these cases it is better then that we use the single value when the array has only one element, and the array if there are more elements.

internal/fields/validate.go
```diff
@@ -345,18 +370,109 @@ func (v *Validator) validateScalarElement(key string, val interface{}, doc commo
...
+				for _, normalize := range definition.Normalize {
+					switch normalize {
+					case "array":
+						shouldBeArray = true
+					}
+					if shouldBeArray {
+						break
+					}
+				}
+			}
```
jsoriano, line 417: Maybe something like this to prefer single values over arrays for fields without definition.
```suggestion
			shouldBeArray := false
			if definition != nil {
				if !v.disabledDependencyManagement && definition.External != "" {
					def, err := v.FieldDependencyManager.ImportField(definition.External, key)
					if err != nil {
						return nil, fmt.Errorf("can't import field (field: %s): %w", key, err)
					}
					definition = &def
				}

				// normalization should just be checked if synthetic source is enabled and the
				// spec version of this package is >= 2.0.0
				if v.disabledNormalization && !v.specVersion.LessThan(semver2_0_0) {
					for _, normalize := range definition.Normalize {
						switch normalize {
						case "array":
							shouldBeArray = true
						}
						if shouldBeArray {
							break
						}
					}
				}
			}
```

jsoriano: Nit. And consider moving the code to check if it should be an array to its own function.

jsoriano (commented):
Nice, the sample document looks better now, added a couple of suggestions more about it.

test/packages/other/synthetic_mode/data_stream/synthetic/sample_event.json
mrodm: Exactly, as those fields do not have any definition they are kept as they are in the document (`fields` key)

PR #1285, opened 2023-05-30
Add PR id to bucket URL

mrodm:
This PR reviews the configuration of the main pipeline:

1. adds PR number/id to the bucket URL so it make it easy to look for the right artifacts.
2. `artifact_paths` keys has been reviewed to ensure all the artifacts are uploaded as they were in Jenkins.
3. This PR also updates some function names in the shell scripts to match the same style in all files.

### URLs for artifacts in Bucket

Currently, the bucket URLs where like:
- `gs://<bucket_url>/buildkite/elastic-package/<build_id>`

With this PR these URLs would take into account the PR number or the branch (as it was in Jenkins):
- In case of pull request: `gs://<bucket_url>/buildkite/elastic-package/PR-<pr_number>-<build_id>`
- In case of branch or tag: `gs://<bucket_url>/buildkite/elastic-package/<branch/tag>-<build_id>`

### Artifacts stored in Buildkite

`artifact_paths` has been reviewed to ensure all the artifacts are uploaded as they were in Jenkins.
- **Exception**: those logs (artifacts) won't be updated for packages under `parallel` folder.

### DIFF
--- a/.buildkite/pipeline.trigger.integration.tests.sh
+++ b/.buildkite/pipeline.trigger.integration.tests.sh
@@ -42,9 +42,8 @@ for test in ${CHECK_PACKAGES_TESTS[@]}; do
     echo "          provider: \"gcp\""
     echo "        artifact_paths:"
     echo "          - build/test-results/*.xml"
-    echo "          - build/elastic-stack-dump/stack/check-*/logs/*.log"
-    echo "          - build/elastic-stack-dump/stack/check-*/logs/fleet-server-internal/**/*"
-    echo "          - build/elastic-stack-status/*/*"
+    echo "          - build/elastic-stack-dump/check-*/logs/*.log"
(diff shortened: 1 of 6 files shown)


.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -62,8 +61,6 @@ for package in $(find . -maxdepth 1 -mindepth 1 -type d) ; do
     echo "          provider: \"gcp\""
     echo "        artifact_paths:"
     echo "          - build/test-results/*.xml"
-    echo "          - build/elastic-stack-dump/stack/check-*/logs/*.log"
-    echo "          - build/elastic-stack-dump/stack/check-*/logs/fleet-server-internal/**/*"
```
mrodm, line 66: Following the approach of not uploading logs in these packages,  instead of fixing the path, I prefer to remove those·

jsoriano (approved):

PR #1286, opened 2023-05-30
internal/testrunner/runners/pipeline: clarify warning of error.message type

efd6:
The current warning is unclear and confusing when a []any containing only strings is held in error.message. encoding/json will store an array of string as []any, so when we see that check each element for stringness.

Relates to https://github.com/elastic/integrations/pull/6258#issuecomment-1569066252.

Please take a look.


jsoriano (approved):
Looks good. Thanks!

PR #1289, opened 2023-05-31
Accept newer revisions when waiting for policies to be assigned

jsoriano:
When tearing down tests, we try to reassign the previous policy. If it was modified and we also expect the revision to be the same, it will never succeed, it will wait till timeout.

### DIFF
--- a/internal/kibana/agents.go
+++ b/internal/kibana/agents.go
@@ -16,7 +16,10 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
-var waitForPolicyAssignedTimeout = 10 * time.Minute
+var (
+	waitForPolicyAssignedTimeout     = 10 * time.Minute
+	waitForPolicyAssignedRetryPeriod = 2 * time.Second
+)
 
 // Agent represents an Elastic Agent enrolled with fleet.
 type Agent struct {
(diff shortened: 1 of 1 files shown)


internal/kibana/agents.go
```diff
@@ -99,13 +102,19 @@ func (c *Client) waitUntilPolicyAssigned(a Agent, p Policy) error {
 		}
 		logger.Debugf("Agent data: %s", agent.String())
 
-		if agent.PolicyID == p.ID && agent.PolicyRevision == p.Revision {
+		if agent.PolicyID == p.ID && agent.PolicyRevision >= p.Revision {
```
mrodm, line 105: In my latest runs of elastic-package test in different packages locally, I didn't see the message (error?) of this timeout.

Is the Revision number always increased ?

jsoriano: I have seen this when trying https://github.com/elastic/elastic-package/pull/1274 locally. The Agent has the policy assigned, but it has a newer revision, then it is never satisfied and the cleanup is interrupted with a timeout error after a long wait. Not sure though how I reached this situation, but it happened to me at least a couple of times.

In this situation the only options are to accept the newer policy or to fail fast, and I think that accepting the policy as is now is the best option of these two. There does not seem to be any way to assign an specific revision of a policy.

jsoriano: > There does not seem to be any way to assign an specific revision of a policy.

There isn't, checked in https://petstore.swagger.io/?url=https://raw.githubusercontent.com/elastic/kibana/8.7/x-pack/plugins/fleet/common/openapi/bundled.json#/default/reassign-agent

jsoriano: > Is the Revision number always increased ?

I am hoping so :smiley:

mrodm: Perfect! Good catch!

mrodm (approved):

PR #1290, opened 2023-06-01
Change order PATH folders

mrodm:
This PR changes the order of folders when setting the PATH environment variable.


mrodm:
> Was it causing issues somewhere?

Not now, but just to ensure that commands from the system have precedence.

Issue #1282, opened 2023-05-29
Elastic Stack Up not working on Windows or Ubuntu WSL


nicpenning:
I am wondering if something is amiss with the docker compose file.

When trying to follow these docs: https://github.com/elastic/integrations/blob/main/docs/developer_workflow_design_build_test_integration.md

And trying to run this command: `elastic-package stack up -v -d`

It just gets stuck at Build 0%

But trying to run the commands manually to start the stack up I am seeing this error:

`service "fleet-server" has neither an image nor a build context specified: invalid compose project`

Is it possible fleet was not included in the stack or should I be doing something to prevent fleet from being deployed to simply walk through building an integration?

I will post more details when I'm at a computer. 
```[tasklist]
### Tasks
- [ ] Add some mention to having a working docker installation to the Getting Started guide.
- [ ] Throw an error instead of hanging forever when building images if docker is not available.
```

jsoriano:
Hey Nic, thanks for reporting your issue.

What versions of elastic-package, docker and docker-compose are you using?

nicpenning:
Looks like docker-compose: v2.18.1 and elastic-package: b57cc2d (which I believe is v0.80.0, but unsure).

Method for getting elastic-package was:

```
git clone https://github.com/elastic/elastic-package.git
make build
```

```
Welcome to Ubuntu 22.04.2 LTS (GNU/Linux 5.10.102.1-microsoft-standard-WSL2 x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

napsta@el33t-b00k-1:~$ docker-compose --version
Docker Compose version v2.18.1
napsta@el33t-b00k-1:~$ cd elastic/elastic-package/
napsta@el33t-b00k-1:~/elastic/elastic-package$ ./elastic-package version
elastic-package version-hash b57cc2d (build time: 2023-05-28T17:50:42-05:00)
```

This is my issue:
(Building 0.0s is just stuck and never progresses and you can't control-C out of it, you have to end the process/terminal).
```
napsta@el33t-b00k-1:~/elastic/elastic-package$ ./elastic-package stack up -v
2023/05/31 21:35:32 DEBUG Enable verbose logging
2023/05/31 21:35:32 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
Boot up the Elastic stack
[...]

PR #1294, opened 2023-06-02
Initialize default reports directory only when needed

jsoriano:
`report` subcommand had some code to initialize its defaults for the output directory. This code was executed during the initialization of the subcommand, that is allways executed.

When running any command out of a package directory, this initialization fails, as it depends on the package directory, and it produces an error message.

Move this initialization to the code that gets the parameter in the `report` subcommand, so it is only executed there, and only if the user has not provided a value for the flag.

This was reported in https://github.com/elastic/elastic-package/issues/1282#issuecomment-1573067350.

### DIFF
--- a/cmd/report.go
+++ b/cmd/report.go
@@ -18,7 +18,9 @@ import (
 	"github.com/elastic/elastic-package/internal/reportgenerator/outputs"
 )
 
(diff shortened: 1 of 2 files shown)


mrodm (approved):
:+1:

PR #1297, opened 2023-06-07
Skipped malformed docker-compose log lines

### LINKED ISSUE #1296: f1106b9b causes build failures due to valid log lines
Since f1106b9b valid builds can fail with 
```
Error: error running package system tests: could not complete test run: check log messages failed: malformed docker-compose log line
```
With judicious addition of information to the error message:
```diff
diff --git a/internal/stack/parselogs.go b/internal/stack/parselogs.go
index 6b8557e..a2d62b6 100644
--- a/internal/stack/parselogs.go
+++ b/internal/stack/parselogs.go
@@ -41,7 +41,7 @@ func ParseLogs(options ParseLogsOptions, process func(log LogLine) error) error
                messageSlice := strings.SplitN(line, "|", 2)

                if len(messageSlice) != 2 {
-                       return fmt.Errorf("malformed docker-compose log line")
+                       return fmt.Errorf("malformed docker-compose log line: %q", line)
                }

                messageLog := messageSlice[1]
```
we see
```
Error: error running package system tests: could not complete test run: check log messages failed: malformed docker-compose log line: "Attaching to elastic-package-stack_elastic-agent_1"
```
Replacing the error return with a `continue` allows completion of the assessment of log lines that appears to be the intention of #1256.

/cc @mrodm

mrodm:
Closes #1296 

When running `elastic-package` along with docker-compose v1.* , the first line of the `docker-compose logs` command is always like: `Attaching to <container>`. For instance:
```
```

This message is not actual log and `elastic-package test system` was failing due to that. Expected messages should be like:
```
elastic-package-stack-elastic-agent-1  | {"log.level":"info","@timestamp":"2023-06-07T13:27:48.548Z","message":"Non-zero metrics in the last 30s","component":{"binary":"metricbeat","dataset":"elastic_agent.metricbeat","id":"beat/metrics-monitoring","type":"beat/metrics"},"log":{"source":"beat/metrics-monitoring"},"log.logger":"monitoring","log.origin":{"file.line":187,"file.name":"log/log.go"},"service.name":"metricbeat",...,"ecs.version":"1.6.0"}
```

This PR changes this behaviour to avoid raising an error and instead it logs a debug message. Those log lines that are not log-like messages are skipped.

### DIFF
--- a/internal/stack/parselogs.go
+++ b/internal/stack/parselogs.go
@@ -56,6 +56,8 @@ func ParseLogs(options ParseLogsOptions, process func(log LogLine) error) error 
 		}
 
 		// There could be valid messages with just plain text without timestamp
+		// and therefore not processed, cannot be ensured in which timestamp they
+		// were generated
 		if !startProcessing && log.Timestamp.Before(options.StartTime) {
 			continue
 		}
--- a/internal/stack/parselogs.go
+++ b/internal/stack/parselogs.go
@@ -7,10 +7,11 @@ package stack
 import (
 	"bufio"
 	"encoding/json"
-	"fmt"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/elastic/elastic-package/internal/logger"
 )
 
 type ParseLogsOptions struct {
@@ -41,7 +42,8 @@ func ParseLogs(options ParseLogsOptions, process func(log LogLine) error) error 
 		messageSlice := strings.SplitN(line, "|", 2)
 
 		if len(messageSlice) != 2 {
-			return fmt.Errorf("malformed docker-compose log line")
+			logger.Debugf("skipped malformed docker-compose log line: %s", line)
+			continue
 		}
 
 		messageLog := messageSlice[1]


jlind23 (approved):

internal/stack/parselogs.go
```diff
@@ -41,7 +42,8 @@ func ParseLogs(options ParseLogsOptions, process func(log LogLine) error) error
 		messageSlice := strings.SplitN(line, "|", 2)
 
 		if len(messageSlice) != 2 {
-			return fmt.Errorf("malformed docker-compose log line")
+			logger.Debugf("skipped malformed docker-compose log line: %s", line)
```
jsoriano, line 45: Should we check if the malformed message is actually "Attaching to..." to avoid silently ignoring messages here?

jsoriano (approved):

internal/stack/parselogs.go
mrodm: It could be checked that to narrow the skipped messages.
So for any other unexpected message, should we return error as it was before? @jsoriano

efd6: What is the concern with missing unknown log format lines? ISTM that given that the vetting done later depends on knowing what we are looking for, unknown things will always be at risk of being missed.

efd6 (approved):
Thanks

mrodm:
buildkite test this

jsoriano (approved):

PR #1298, opened 2023-06-08
Validate JSON output file before unmarshall

### LINKED ISSUE #1299: Terraform outputs in aws package make the test flaky
The terraform outputs in `aws` package as added [here](https://github.com/bhapas/elastic-package/commit/f8b3c8c6adbd730788d034f1665b925bc794627a#diff-56559e2b5a4a5f1a24f2de5a2bd7cf849068b89034b7bd05b13d42d75e30a19fR34) have flaky tests on CI.

The outputs are not generated sometimes and it results in the flaky test

Sample logs:
```
2023/06/08 11:43:57 DEBUG Unmarshalling terraform output json
2023/06/08 11:43:57 DEBUG **********JSON CONTENT**************** {
  "instance_id": {
    "sensitive": false,
    "type": "string",
    "value": "i-somerandomid"
  }
}
```

```
2023/06/08 12:22:06 DEBUG Unmarshalling terraform output json
2023/06/08 12:22:06 DEBUG **********JSON CONTENT****************
2023/06/08 12:22:06 DEBUG waiting for 1s before tearing down...
Error: error running package system tests: could not complete test run: could not setup service: could not handle terraform output error during json Unmarshal unexpected end of JSON input
+ cleanup
+ r=1
```

bhapas:
The terraform output file sometimes is empty and json [unmarshall](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/servicedeployer/terraform.go#L67) throws an error.

The fix is to validate the json file before unmarshalling it.

- Fixes #1299

### DIFF
--- a/internal/testrunner/runners/system/servicedeployer/terraform.go
+++ b/internal/testrunner/runners/system/servicedeployer/terraform.go
@@ -64,8 +64,14 @@ func addTerraformOutputs(outCtxt ServiceContext) error {
 	// Unmarshall the data into `terraformOutputs`
 	logger.Debug("Unmarshalling terraform output json")
 	var terraformOutputs map[string]OutputMeta
-	if err = json.Unmarshal(content, &terraformOutputs); err != nil {
-		return fmt.Errorf("error during json Unmarshal %w", err)
+
+	if json.Valid(content) {
+		if err = json.Unmarshal(content, &terraformOutputs); err != nil {
+			return fmt.Errorf("error during json Unmarshal %w", err)
+		}
+	} else {
+		logger.Error("Invalid Json Content")
+		return nil
 	}
 
 	if len(terraformOutputs) == 0 {
--- a/internal/testrunner/runners/system/servicedeployer/terraform.go
+++ b/internal/testrunner/runners/system/servicedeployer/terraform.go
@@ -65,13 +65,13 @@ func addTerraformOutputs(outCtxt ServiceContext) error {
 	logger.Debug("Unmarshalling terraform output json")
 	var terraformOutputs map[string]OutputMeta
 
-	if json.Valid(content) {
-		if err = json.Unmarshal(content, &terraformOutputs); err != nil {
-			return fmt.Errorf("error during json Unmarshal %w", err)
-		}
-	} else {
-		logger.Error("Invalid Json Content")
+	if !json.Valid(content) {
+		logger.Debug("Invalid Json content in the terraform output file, skipped creating outputs")
 		return nil
+	}
+
+	if err = json.Unmarshal(content, &terraformOutputs); err != nil {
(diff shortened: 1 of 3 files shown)


internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -64,8 +64,14 @@ func addTerraformOutputs(outCtxt ServiceContext) error {
...
-	if err = json.Unmarshal(content, &terraformOutputs); err != nil {
-		return fmt.Errorf("error during json Unmarshal %w", err)
+
+	if json.Valid(content) {
+		if err = json.Unmarshal(content, &terraformOutputs); err != nil {
+			return fmt.Errorf("error during json Unmarshal %w", err)
+		}
+	} else {
+		logger.Error("Invalid Json Content")
+		return nil
```
mrodm, line 74: this could be rewriting to exit earlier.

Would it make sense to just log a Debug message?

```suggestion
    if !json.Valid(content) {
        logger.Error("Invalid JSON Content")
        return nil
    }
	if err = json.Unmarshal(content, &terraformOutputs); err != nil {
		return fmt.Errorf("error during json Unmarshal %w", err)
```

internal/testrunner/runners/system/servicedeployer/terraform_test.go
```diff
@@ -121,7 +143,8 @@ func TestAddTerraformOutputs(t *testing.T) {
 			}
 
 			// Test that the terraform output values are generated correctly
-			addTerraformOutputs(tc.ctxt)
+			err := addTerraformOutputs(tc.ctxt)
+			assert.Equal(t, err, nil)
```
mrodm, line 147: nit. there is a specific method to test the errors:

```suggestion
			assert.NoError(t, err)
```

or maybe it can be required to not be an error there 

```suggestion
			require.NoError(t, err)
```

internal/testrunner/runners/system/servicedeployer/terraform.go
```diff
@@ -64,6 +64,12 @@ func addTerraformOutputs(outCtxt ServiceContext) error {
 	// Unmarshall the data into `terraformOutputs`
 	logger.Debug("Unmarshalling terraform output json")
 	var terraformOutputs map[string]OutputMeta
+
+	if !json.Valid(content) {
+		logger.Debug("Invalid Json content in the terraform output file, skipped creating outputs")
+		return nil
+	}
```
mrodm, line 71: What happens if `aws_logs` has an invalid json here? Would the test package fail if that happens ?

bhapas: The system test fails on the package if there is an invalid json because the output is not available in the system test config

mrodm (approved):

PR #1315, opened 2023-06-20
Skip error uninstalling system package

mrodm:
This PR has two goals:
- skip errors when `system` package is uninstalled
- fix error management in tear down process

Currently, `system` package is being used by default as part of another agent policy. Because of that, when system or asset tests finish and try to uninstall the package, if this package is  `system`, this fails with this error:

> [2023-06-19T09:20:47.479Z] Error: error running package system tests: could not complete test run: failed to tear down runner: failed to uninstall package: can't remove the package: could not remove package; API status code = 400; response body = {"statusCode":400,"error":"Bad Request","message":"unable to remove package with existing package policy(s) in use by agent(s)"}

As part of this PR, `asset` tests has also been checked and fixed since this test runner also performs the uninstallation of the package.


mrodm:
> Created or updated PR in integrations repostiory to test this vesrion. Check [elastic/integrations#6619](https://github.com/elastic/integrations/pull/6619)

`system` package has been fixed

internal/testrunner/runners/system/runner.go
```diff
@@ -496,10 +496,25 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+			return nil
 		}
+		// by default system package is part of an agent policy and it cannot be uninstalled
+		// https://github.com/elastic/elastic-package/blob/5f65dc29811c57454bc7142aaf73725b6d4dc8e6/internal/stack/_static/kibana.yml.tmpl#L62
+		switch pkgManifest.Name {
+		case "system":
+			logger.Debugf("failed to uninstall package %q: %s", pkgManifest.Name, err.Error())
+		default:
+			logger.Warnf("failed to uninstall package %q: %s", pkgManifest.Name, err.Error())
+		}
```
mrodm, line 509: Should we also show a Debug message in case `system` package fails ? It will always fail)

If not this code could be just changed to be something like:

```go
		// by default system package is part of an agent policy and it cannot be uninstalled
		// https://github.com/elastic/elastic-package/blob/5f65dc29811c57454bc7142aaf73725b6d4dc8e6/internal/stack/_static/kibana.yml.tmpl#L62
		if err != nil && pkgManifest.Name != "system" {
			logger.Warnf("failed to uninstall package %q: %s", pkgManifest.Name, err.Error())
		}
		return nil
```

jsoriano: Yeah, let's log the error only for other packages.

internal/testrunner/runners/asset/runner.go
```diff
@@ -92,9 +96,18 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
...
+
 		logger.Debug("removing package...")
-		if err := packageInstaller.Uninstall(); err != nil {
-			return errors.Wrap(err, "error cleaning up package")
+		err = packageInstaller.Uninstall()
+
+		// by default system package is part of an agent policy and it cannot be uninstalled
+		// https://github.com/elastic/elastic-package/blob/5f65dc29811c57454bc7142aaf73725b6d4dc8e6/internal/stack/_static/kibana.yml.tmpl#L62
+		if err != nil && pkgManifest.Name != "system" {
+			logger.Warnf("failed to uninstall package %q: %s", pkgManifest.Name, err.Error())
```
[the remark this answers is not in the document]
mrodm: Added comment to explain it https://github.com/elastic/elastic-package/pull/1315/commits/6e0af5e22b72d2b9f7e285301a0a290376fd18ac

jsoriano (approved):

PR #1332, opened 2023-06-29
Re-wrap errors

jsoriano:
Use standard library for errors wrapping.

Most code changes done with `github.com/xdg-go/go-rewrap-errors` and `goimports`.

Steps followed to prepare this PR:
* `find . -name "*.go" -exec go run github.com/xdg-go/go-rewrap-errors@latest -w {} \;`
* `find . -name "*.go" -exec goimports -w {} \;`
* `go mod tidy`
* Manually replaced uses of `log.Fatal(fmt.Errorf` with `log.Fatalf(...` (not really related, but took the opportunity to fix this).
* `make check-static`, and fix linting issues related to error messages.
* Fix case with nil error where `errors.Wrap` would have returned nil (https://github.com/elastic/elastic-package/pull/1332/commits/6993819765ad15966fcc03c796fb37d1aa5a8b32).


mrodm (approved):
Nice to have all these occurrences replaced, thanks!

internal/packages/changelog/yaml.go
```diff
@@ -100,19 +103,19 @@ func SetManifestVersion(d []byte, version string) ([]byte, error) {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to decode manifest")
+		return nil, fmt.Errorf("failed to decode manifest: %w", err)
 	}
 
 	// Manifest is a document, with a single element, that should be a map.
 	if len(node.Content) == 0 || node.Content[0].Kind != yaml.MappingNode {
-		return nil, errors.Wrap(err, "unexpected manifest content")
+		return nil, fmt.Errorf("unexpected manifest content: %w", err)
 	}
```
efd6, line 112: This is not right; err is nill here but this returns a non-nil error referring to err.

jsoriano: Thanks, good catch.

PR #1333, opened 2023-06-29
[system testing] Add provision for system testing negative / false-positive scenarios in packages

### LINKED ISSUE #1308: Reduce false negatives with assert.hit_count
When the `assert.hit_count` condition is satisfied the test completes. But if more data is ingested after the condition is satisfied then the test will not catch it leading to a false negative result.

To avoid this problem the test case runner could perform a one or more additional hit count checks after some delay to help ensure that the test fails when more data is ingested than is expected.

bhapas:
This PR changes the way the `system testing` asserts at the number of hits in ElasticSearch when a system test is run.

Earlier, The testrunner exited when `enough` docs are found as per the config and there were more docs that got ingested after `assertHitCount` condition satisfied.

Now, The testrunner tries to get all the docs before checking for the `assertHitCount` and fails if the docs are more/less than the configured count

The `retryTicker` in the `waitUntilTrue` function is changed to `5 seconds` meaning , the retry interval for each poll to ES has been  changed to `5 seconds` and so if for `10 seconds`, there is no change in number of docs , the test passes.

### How to test

Run `PACKAGE_TEST_TYPE=false_positives PACKAGE_UNDER_TEST=httpjson_false_positive_asserts ./scripts/test-check-packages-false-positives.sh`

The `test-check-packages.sh` file has been modified to catch a failure from the new package `httpjson_false_positive_asserts` and a `testcase failure` is considered as `success scenario`

- Fixes #1308


jsoriano:
test integrations

scripts/test-check-false-positives.sh
```diff
@@ -0,0 +1,68 @@
...
+  # Clean used resources
+  for d in test/packages/${PACKAGE_TEST_TYPE:-false_positives}/${PACKAGE_UNDER_TEST:-httpjson_false_positive_asserts}/; do
+    (
+      cd $d
+      elastic-package clean -v
+    )
+  done
+
+  # This is a false positive scenario and tests that the test case failure is a success scenario  
+  if [ "${PACKAGE_TEST_TYPE:-false_positives}" == "false_positives" ] && [ "${PACKAGE_UNDER_TEST:-httpjson_false_positive_asserts}" == "httpjson_false_positive_asserts" ]; then
```
jsoriano, line 23: Nice to have a test for false positive cases, thanks! Not needed to change it here by now, but would it be possible to reuse this script to be able to add more test packages?

internal/testrunner/runners/system/runner.go
```diff
@@ -981,7 +987,7 @@ func waitUntilTrue(fn func() (bool, error), timeout time.Duration) (bool, error)
 	timeoutTicker := time.NewTicker(timeout)
 	defer timeoutTicker.Stop()
 
-	retryTicker := time.NewTicker(1 * time.Second)
+	retryTicker := time.NewTicker(5 * time.Second)
```
jsoriano, line 990: `waitUntilTrue` is used in more places and this change will affect them, could we maybe add the additional wait in the function to check the hits?

internal/testrunner/runners/system/runner.go
```diff
@@ -620,20 +620,26 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
 
 		var err error
 		hits, err = r.getDocs(dataStream)
+		if hits.size() == 0 {
+			return false, err
+		}
 
-		if config.Assert.HitCount > 0 {
-			return hits.size() >= config.Assert.HitCount, err
+		ret := hits.size() == oldHits
```
jsoriano, line 635: This is going to make all tests to wait till timeout or till the count of documents is stable, isn't it? Could we continue taking into acount if `config.Assert.HitCount` is configured so we don't extend the wait for the rest of cases?

Maybe to solve this issue it is enough with adding a sleep when `hits.size() >= config.Assert.HitCount`?

jsoriano (changes requested):
Nice to have the first negative test :slightly_smiling_face:

I think these changes are going to affect the timings of all tests, and not only the ones with hit count assertions, could we change it so these changes only affect tests with hit count assertions?

mrodm:
test integrations

scripts/test-check-false-positives.sh
bhapas: Yes. It should be possible to add more packages under `PACKAGE_TEST_TYPE` `false_positives`

scripts/test-check-false-positives.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# Boot up the stack
+elastic-package stack up -d -v
+
+elastic-package stack status
+
+# Run package tests
+eval "$(elastic-package stack shellinit)"
+
+for d in test/packages/${PACKAGE_TEST_TYPE:-false_positives}/${PACKAGE_UNDER_TEST:-httpjson_false_positive_asserts}/; do
```
mrodm, line 59: In `pipeline.trigger.integration.tests.sh` script, PACKAGE_UNDER_TEST is not being set for the loop where this package has been added.

If it is needed, it must be added the `-p` parameter here in this line

https://github.com/elastic/elastic-package/pull/1333/files#diff-4e2e09cf3f17805ad30f828b15ea267ba7a186ff3f5129b9ecba76ef7317718bR41

Currently, when PACKAGE_UNDER_TEST is being set is because it is required/needed to just execute one package per stage in Jenkins, as it is done for the test packages under `test/packages/parallel`. These packages have special requirements (e.g. logs are updated to a special artifact buckets).

If it is needed to execute every package under `false_positives` in a different stage, it could be added a new loop there in that script for this false_positives packages. Maybe something like:

```bash
pushd test/packages/false_positives > /dev/null
for package in $(find . -maxdepth 1 -mindepth 1 -type d) ; do
    package_name=$(basename ${package})
    echo "      - label: \":go: Running integration test: ${package_name}\""
    echo "        key: \"integration-false_positives-${package_name}\""
    echo "        command: ./.buildkite/scripts/integration_tests.sh -t test-check-packages-false-positives -p ${package_name}"
    echo "        agents:"
    echo "          provider: \"gcp\""
    echo "        artifact_paths:"
    echo "          - build/test-results/*.xml"
    echo "          - build/elastic-stack-dump/check-*/logs/*.log"
    echo "          - build/elastic-stack-dump/check-*/logs/fleet-server-internal/**/*"
done
popd > /dev/null
```

internal/testrunner/runners/system/runner.go
bhapas: Breaking this down into scenarios :

### HitCount configured
- Number of docs less than the count in `n` iterations - Timeout happens as the `true` condition is never satisfied - Test fails
- Number of docs equal in 2 successive iterations with a gap of 5 seconds - No more  docs to ingest and then verify if the count is matching - Catching false positives
- Number of docs never equal - Some issue in the test case causing the docs ingestion to go into infinite loop - Timeout and Test fails

### HitCount Not configured
- Docs ingestion complete - Meaning the `hits.size()` is same in 2 consecutive iterations with 1 second each - Adds up 1 second on each test case but provides us with data if we were to validate something else.
- Docs ingestion incomplete - Just picking up those docs which are ingested in the first iteration and validating against them

I think it should not add up a lot of time for each run , as I see the test suites are running in parallel , but we shall have the data ingested for any other validation - Just to avoid testing against incomplete data - WDYT?

internal/testrunner/runners/system/runner.go
bhapas: Will send in a `retryTimeout` from the calling function

bhapas: Or Maybe a `time.sleep(4 * time.Second)` if config for `assert.hitCount` exists

scripts/test-check-false-positives.sh
bhapas: These `false_positive` tests should just run as `System tests` for any other package, just that the outcome is inversed.

I have add the `PACKAGE_UNDER_TEST` in https://github.com/elastic/elastic-package/pull/1333/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R81 

and it seems to run as a separate suite atleast. Not sure if it would be necessary to run each package separately? Do you think the the testrunner exits on execution failure / the script failure? Like `exit 1 from test execution`  or `exit 1 from script execution` 


Or Did I misunderstand ?

Screenshot 2023-07-03 at 11 48 14

bhapas: I think I got your point now.. 

But as it is now , I don't see a need to execute each package in a separate buildKite stage. As they are just another system tests.. And I think the value of `PACKAGE_UNDER_TEST` is correctly taken in each execution as per https://buildkite.com/elastic/elastic-package/builds/1099#01890bc4-ce48-4b8b-9638-8ae44f58e102/131-2282

bhapas:
test integrations

internal/testrunner/runners/system/runner.go
```diff
@@ -629,11 +630,21 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
+				return false, err
+			}
+
+			ret := hits.size() == oldHits
+			if !ret {
+				oldHits = hits.size()
+				time.Sleep(4 * time.Second)
+			}
 
+			return ret, err
```
jsoriano, line 643: In this case, should we wait for the hits size to be greater than `config.Assert.HitCount`? Otherwise, if the time between the first and last event takes more than 5 secons, it won't wait for the last hits.
```suggestion
			if hits.size() < config.Assert.HitCount {
				return false, err
			}		
			time.Sleep(4 * time.Second)
			return true, err
```

bhapas: In this case if the 2 docs are ingested in a span of 5 seconds in 2 iterations , and `assert.hit_count: 2` is set in the config , this code returns true in the first round after it matches `2` , but total docs ingested are 4.

bhapas: May be can be mitigated this way - WDYT?

```diff
diff --git a/internal/testrunner/runners/system/runner.go b/internal/testrunner/runners/system/runner.go
index cc47cc9..a77af87 100644
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -634,6 +634,10 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
                       if config.Assert.HitCount > 0 {
			if hits.size() == 0 {
				return false, err
			}
 
+                       if hits.size() < config.Assert.HitCount {
+                               return false, err
+                       }
+
                       ret := hits.size() == oldHits
			if !ret {
				oldHits = hits.size()
				time.Sleep(4 * time.Second)
			}

			return ret, err
```

jsoriano: The problem I see is that we are not waiting for the number of events to be at least the one we expect. So it may happen that the wait finishes, the count is not the one we expect yet, because we haven't waited enough, and then the later assertion fails, when it should have passed.

This implementation is waiting to have at least one event, and that the count of events is stable during at least 5 seconds. But I think it should also wait till the count of events is at least `assert.hit_count`, or timeout if it never is.

Imagine an scenario with `assert.hit_count: 2`. One event arrives during the first second, second one arrives in ten seconds.
* One event is ingested.
* One iteration happens, oldHits is zero, current hits is one, we set oldHits to one, wait 4 seconds and retry.
* Second iteration happens, oldHits is one, current hits is still one, we return true, the wait finishes.
* Assertion fails because there is only one hit. But it shouldn't because another one is going to arrive in ten seconds, way before the timeout.

I think that this wait should never return true if `config.Assert.HitCount` is set and the current count of hints is below this expected count.

bhapas: > The problem I see is that we are not waiting for the number of events to be at least the one we expect. So it may happen that the wait finishes, the count is not the one we expect yet, because we haven't waited enough, and then the later assertion fails, when it should have passed.

Resolved this in the latest comment - https://github.com/elastic/elastic-package/pull/1333#discussion_r1250752641

scripts/test-check-false-positives.sh
bhapas: Added another CI step

internal/testrunner/runners/system/runner.go
jsoriano: Ah yes, sorry this would work :slightly_smiling_face: I was writing my comment at the same time and didn't see it.

bhapas: Pushed the changes now :-)

internal/testrunner/runners/system/runner.go
```diff
@@ -629,11 +630,25 @@ func (r *runner) runTest(config *testConfig, ctxt servicedeployer.ServiceContext
...
 		if config.Assert.HitCount > 0 {
-			return hits.size() >= config.Assert.HitCount, err
-		}
+			if hits.size() == 0 {
+				return false, err
+			}
+
+			if hits.size() < config.Assert.HitCount {
+				return false, err
+			}
```
jsoriano, line 639: Nit. First if is redundant.
```suggestion
			if hits.size() < config.Assert.HitCount {
				return false, err
			}
```

.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -49,6 +49,22 @@ for test in ${CHECK_PACKAGES_TESTS[@]}; do
...
+    package_name=$(basename ${package})
+    echo "      - label: \":go: Running integration test: ${package_name}\""
+    echo "        key: \"integration-false_positives-${package_name}\""
+    echo "        command: ./.buildkite/scripts/integration_tests.sh -t test-check-packages-false-positives -p ${package_name}"
+    echo "        agents:"
+    echo "          provider: \"gcp\""
+    echo "        artifact_paths:"
+    echo "          - build/test-results/*.xml"
+    echo "          - build/elastic-stack-dump/check-*/logs/*.log"
+    echo "          - build/elastic-stack-dump/check-*/logs/fleet-server-internal/**/*"
```
mrodm, line 63: Should we do as in `parallel` packages and upload logs to the internal bucket ? cc @jsoriano

.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -49,6 +49,22 @@ for test in ${CHECK_PACKAGES_TESTS[@]}; do
     fi
 done
 
+pushd test/packages/false_positives > /dev/null
+for package in $(find . -maxdepth 1 -mindepth 1 -type d) ; do
+    package_name=$(basename ${package})
+    echo "      - label: \":go: Running integration test: ${package_name}\""
```
mrodm, line 55: Should it be indicated here that is from `false positive`? 

```suggestion
    echo "      - label: \":go: Running integration test (false positive): ${package_name}\""
```

.buildkite/pipeline.trigger.integration.tests.sh
jsoriano: Sorry, I missed this comment.

I would do the same as in `parallel`, so we don't forget to do it here if we want to avoid risks. I guess that in `parallel` we are doing this to avoid leaking secrets related to cloud integrations or the terraform test runner.

bhapas: @jsoriano Does this add any value? I know why we did it in `parallel` , as we used real cloud instances and secrets. These are mere mocked tests , or?

jsoriano: Tests packages including terraform scenarios are actually executed.

bhapas: Done.. Pushed

bhapas:
test integrations

scripts/test-check-false-positives.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+cleanup() {
+  r=$?
+
+  # Dump stack logs
+  elastic-package stack dump -v --output "build/elastic-stack-dump/check-${PACKAGE_UNDER_TEST:-${PACKAGE_TEST_TYPE:-httpjson_false_positive_asserts}}"
```
jsoriano, line 9: Should this have been changed after this comment? https://github.com/elastic/elastic-package/pull/1333/files#r1250740254
```suggestion
  elastic-package stack dump -v --output "build/elastic-stack-dump/check-${PACKAGE_UNDER_TEST:-${PACKAGE_TEST_TYPE:-*}}"
```

jsoriano (commented):
LGTM, let's wait for a final review by Mario to the CI scripts :innocent: 

Thanks!

scripts/test-check-false-positives.sh
bhapas: Instead each package is run in a separate small step in parallel

bhapas: @mrodm Is this the same as we talked or?

.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -49,6 +49,24 @@ for test in ${CHECK_PACKAGES_TESTS[@]}; do
...
+    echo "        key: \"integration-false_positives-${package_name}\""
+    echo "        command: ./.buildkite/scripts/integration_tests.sh -t test-check-packages-false-positives -p ${package_name}"
+    echo "        env:"
+    echo "          UPLOAD_SAFE_LOGS: 1"
+    echo "        agents:"
+    echo "          provider: \"gcp\""
+    echo "        artifact_paths:"
+    echo "          - build/test-results/*.xml"
+    echo "          - build/elastic-stack-dump/check-*/logs/*.log"
+    echo "          - build/elastic-stack-dump/check-*/logs/fleet-server-internal/**/*"
```
mrodm, line 65: According to this thread https://github.com/elastic/elastic-package/pull/1333/files#r1250786653, let's just remove these two from `artifact_paths` and just keep the result xml files, as it happens for `parallel` test packages:
 ```suggestion
```

File: scripts/test-check-false-positives.sh (hunk 9 above)
Comment (mrodm) [reply scripts/test-check-false-positives.sh:9]: @jsoriano is right, in this script the default value for `PACKAGE_UNDER_TEST` should be `*` instead of the `httpjson_false_positive_asserts` package. In that regard, this script would be similar to `scripts/test-check-packages.sh`.

File: scripts/test-check-false-positives.sh [hunk 11]
```diff
@@ -0,0 +1,68 @@
...
+
+  # Take down the stack
+  elastic-package stack down -v
+
+  # Clean used resources
+  for d in test/packages/${PACKAGE_TEST_TYPE:-false_positives}/${PACKAGE_UNDER_TEST:-httpjson_false_positive_asserts}/; do
```
mrodm, line 15:
```suggestion
  for d in test/packages/${PACKAGE_TEST_TYPE:-false_positives}/${PACKAGE_UNDER_TEST:-*}/; do
```

scripts/test-check-false-positives.sh
```diff
@@ -0,0 +1,68 @@
...
+  exit $r
+}
+
+trap cleanup EXIT
+
+export ELASTIC_PACKAGE_LINKS_FILE_PATH="$(pwd)/scripts/links_table.yml"
+
+OLDPWD=$PWD
+# Build/check packages
+for d in test/packages/${PACKAGE_TEST_TYPE:-false_positives}/${PACKAGE_UNDER_TEST:-httpjson_false_positive_asserts}/; do
```
mrodm, line 40:
```suggestion
for d in test/packages/${PACKAGE_TEST_TYPE:-false_positives}/${PACKAGE_UNDER_TEST:-*}/; do
```

scripts/test-check-false-positives.sh
mrodm, line 59:
```suggestion
for d in test/packages/${PACKAGE_TEST_TYPE:-false_positives}/${PACKAGE_UNDER_TEST:-*}/; do
```

scripts/test-check-false-positives.sh
mrodm, line 23: Would it be enough to just check for `PACKAGE_TESAT_TYPE` being `false_positives` ? 

That would mean that every test package located under `false_positives`folder would follow these  requirements (so that it needs to fail to be considered succeeded).

```suggestion
  if [ "${PACKAGE_TEST_TYPE:-false_positives}" == "false_positives" ]; then
```

test/packages/false_positives/httpjson_false_positive_asserts/_dev/build/build.yml
```diff
@@ -0,0 +1,3 @@
+dependencies:
+  ecs:
+    reference: git@8.8
```
mrodm, line 3: Let's just use tag here. 
There were some issues related to use branches instead of tags

```suggestion
    reference: git@v8.8.0
```

Probably, it results in some failures in the README

scripts/test-check-false-positives.sh
bhapas: DOne

bhapas:
test integrations

.buildkite/scripts/integration_tests.sh
```diff
@@ -24,6 +24,7 @@ source .buildkite/scripts/install_deps.sh
 source .buildkite/scripts/tooling.sh
 
 PARALLEL_TARGET="test-check-packages-parallel"
+FALSE_POSITIVES_TARGET="test-check-packages-false_positives"
```
mrodm, line 27:
```suggestion
FALSE_POSITIVES_TARGET="test-check-packages-false-positives"
```

scripts/test-check-false-positives.sh
```diff
@@ -0,0 +1,68 @@
...
+    )
+  done
+
+  # This is a false positive scenario and tests that the test case failure is a success scenario  
+  if [ "${PACKAGE_TEST_TYPE:-false_positives}" == "false_positives" ]; then
+    if [ $r == 1 ]; then
+        exit 0
+      elif [ $r == 0 ]; then
+        exit 1
+    fi
```
mrodm, line 28: Currently, with these changes even if the new step runs successfully, there are some JUnit annotations because there were some expected tests failing.

To avoid showing those Junit annotations in Buildkite, I think one option that could we do here is to delete the `build/test-results/*.xml` files in the cleanup function if "$r == 1" . WDYT ? @bhapas 

Junit annotations with errors

Also, in case there were no errors, some echo message could be added there to explain why it is failing.


```suggestion
    if [ $r == 1 ]; then
      rm build/test-results/*.xml
      exit 0
    elif [ $r == 0 ]; then
      echo "Expected to fail tests, but there was none failing"
      exit 1
    fi
```

cc @jsoriano

scripts/test-check-false-positives.sh
```diff
@@ -0,0 +1,70 @@
...
+      cd $d
+      elastic-package clean -v
+    )
+  done
+
+  # This is a false positive scenario and tests that the test case failure is a success scenario  
+  if [ "${PACKAGE_TEST_TYPE:-false_positives}" == "false_positives" ]; then
+    if [ $r == 1 ]; then
+      rm build/test-results/*.xml
+      exit 0
```
mrodm, line 26: Just wondering if this could be leveraged to check specifically which tests are failing.

It could happen that some other test is failing and not the expected one. Or another scenario it would be that more tests are failing than expected.

I see that the resulting xml is like this 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="system" tests="1" failures="1">
    <!--test suite for system tests-->
    <testcase name="system test: pagination" classname="httpjson_false_positive_asserts.generic" time="28.841040253">
      <failure>observed hit count 4 did not match expected hit count 2</failure>
    </testcase>
  </testsuite>
</testsuites>
```

Maybe we could add another file next to the folder `test/packages/false_positives/httpjson_false_positive_asserts`, something like:
`test/packages/false_positives/httpjson_false_positive_asserts-expected_errors.xml` or similar

and compare somehow those two xmls.

There would be differences because I see there is a field `time="XX.XX"`. Could it be checked at least the failures lines ? Or a line by line comparison with regexes ? 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="system" tests="1" failures="1">
    <!--test suite for system tests-->
    <testcase name="system test: pagination" classname="httpjson_false_positive_asserts.generic" time="*">
      <failure>observed hit count 4 did not match expected hit count 2</failure>
    </testcase>
  </testsuite>
</testsuites>
```

Just testing with the contents of the xml:
- a new file could be added that contains the expected errors:
    - it can contain regexes and double quotes must be escaped.
  ```
  <testcase name=\"system test: pagination\" classname=\"httpjson_false_positive_asserts.generic\" time=\".*\"> * <failure>observed hit count 4 did not match expected hit count 2</failure>
  ```
- Then in this `scripts/test-check-packages-false-positives` it could be added the comparison line by line to check for these errors

mrodm: In this commit, I pushed an approach to deal with this
https://github.com/elastic/elastic-package/pull/1333/commits/51e33f4d8e93fc18b7eae030911167fb0f150f55

Let me know what you think @jsoriano @bhapas

mrodm: That changeset is supposed to fail the build (https://buildkite.com/elastic/elastic-package/builds/1142), I will do a couple of tests more
- same number of errors, but different one
- expected errors match with the actual errors

jsoriano: LGTM. Maybe at some point we should migrate these scripts to go, they are starting to have some complexity.

mrodm: First build with the expected output
- Error: There are unexpected errors in httpjson_false_positive_asserts

bhapas: Test Expected Error Scenarios:

- [ ]  Number of expected and actual errors are not equal
- [ ]  Number of expected and actual errors are equal but different contents in the actual errors than expected
- [ ]  The Expected and actual errors are same

mrodm (approved):
:+1: Thanks for adding this negative tests @bhapas!

docs/howto/system_testing.md
```diff
@@ -568,6 +568,23 @@ to indexing generated data from the integration's data streams into Elasticsearc
...
+- The negative / false-positive test packages are added under `test/packages/false_positives`
+- It is required to have a file `<package_name>.expected_errors` with the lines needed for every package added under `test/packages/false_positives`.
+- One line per error, taking into account that all `\n` were removed, meaning it is just one line for everything.
+- As it is used `grep` with `-E` some kind of regexes can be used.
+
+Example `expected_errors` file content:
+
+```xml
+<testcase name=\"system test: pagination\" classname=\"httpjson_false_positive_asserts.generic\" time=\".*\"> * <failure>observed hit count 4 did not match expected hit count 2</failure>
+```
```
jsoriano, line 586: Thanks for adding docs for this too!

jsoriano (approved):

Issue #1334, opened 2023-06-29
Error: undefined field for ECS multi-field mapping


ritalwar:
Getting "field process.name.text is undefined" error for status data stream of MongoDB package.

This field is defined as a multi_field in the ECS mapping for the "process.name" field. I have included some screenshots for better understanding of the issue.

Note: I am enabling TSDB for this package.

![image](https://github.com/elastic/elastic-package/assets/102972658/9ba7b123-c2da-48e6-95de-88dd50db58ae)
![image](https://github.com/elastic/elastic-package/assets/102972658/d4886f4d-ce05-43f9-9be8-ee2737fbdf4e)
![image](https://github.com/elastic/elastic-package/assets/102972658/8903e1fb-792d-4c68-b0e4-4f55c95da209)

PR #1335, opened 2023-06-29
Fix validation of multi-fields of external fields

jsoriano:
Refactor validator so it preloads external fields instead of loading them on demand when requested. This allows to have a single implementation of external fields resolution.

Regarding multi-fields in external fields, it was not possible to validate them before because it was not possible to differentiate them from undefined fields without previously resolving the parent. This would require to resolve the parent for all unknown fields that were potential children of external fields. Now they are preloaded, so they can be directly validated.

There was also an error in dependency management that was being ignored to help on testing. This error is reported now, and tests depending on this explicitly disable dependency management to avoid it.

The status data stream of the mongodb package from https://github.com/elastic/integrations/pull/6692 is included here as a test case for time series and multi-fields. This is required because enabling synthetic source is the only way we have now to find multi-fields in test documents.

Fix https://github.com/elastic/elastic-package/issues/1334.


jsoriano:
test integrations

jsoriano:
test integrations

internal/docs/exported_fields.go
```diff
@@ -23,15 +23,17 @@ type fieldsTableRecord struct {
 var escaper = strings.NewReplacer("*", "\\*", "{", "\\{", "}", "\\}", "<", "\\<", ">", "\\>")
 
 func renderExportedFields(fieldsParentDir string) (string, error) {
-	validator, err := fields.CreateValidatorForDirectory(fieldsParentDir)
+	injectOptions := fields.InjectFieldsOptions{
+		// Keep External parameter when rendering fields, so we can render
+		// documentation for empty groups imported from ECS, for backwards compatibility.
+		KeepExternal: true,
```
jsoriano, line 29: This parameter for fields injection can be a bit difficult to follow. Its only purpose is to mimic a collateral behaviour we had with lazy resolution of external fields: In some places we were skipping group fields when they were empty. But this was not happening with external fields, because on these checks the field was still unresolved.

We may argue that this behaviour is unexpected and we should fix it instead, because it treats fields on different ways depending if they are external or not. But fixing this would break CI builds of packages, because it changes what fields are rendered.

This flag helps keeping the old behaviour for rendered documentation.:
* When rendering documentation, we keep external, so we can keep the old behaviour.
* In other cases, such as when resolving external fields in package builds, we remove the "external" parameter once the field is injected, so it is not written to fields files. On these cases we may be removing empty groups from built fields files, but this is actually good because they don't generate any meaningful mapping.

jsoriano:
test integrations

jsoriano:
Something is happening with buildkite :thinking:

mrodm:
buildkite test this

mrodm:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

internal/fields/validate.go
```diff
@@ -161,51 +170,56 @@ func createValidatorForDirectoryAndPackageRoot(fieldsParentDir string, finder pa
...
-	// This is also useful for testing purposes, where we don't have a real package, but just "fields" directory. The package root is always absent.
-	if !found {
-		logger.Debug("Package root not found, dependency management will be disabled.")
-		v.disabledDependencyManagement = true
-		return v, nil
+		return nil, fmt.Errorf("can't load fields from directory (path: %s): %w", fieldsDir, err)
 	}
 
-	bm, ok, err := buildmanifest.ReadBuildManifest(packageRoot)
+	v.Schema = append(fields, v.Schema...)
```
mrodm, line 194: At this point v.Schema would just contain the fields retrieved from the ECS reference?

mrodm (approved):
Great job fixing this and supporting backwards compatibility! :muscle:

internal/fields/validate.go
jsoriano: Correct, it would contain the ECS schema fields when import all ECS schema fields is enabled.

jlind23:
@mrodm as @jsoriano is out for now, can we please make this land?

mrodm:
> @mrodm as @jsoriano is out for now, can we please make this land?

@jlind23 sure, I'll merge this PR since all these changes look good to me.
To release a new elastic-package version, I would wait for @jsoriano last check, in case it is better to test or add something else first.

jlind23:
@mrodm sounds good to me, thanks.

PR #1337, opened 2023-06-30
Fix error when detecting an invalid manifest after updating changelog

jsoriano:
Reported in https://github.com/elastic/elastic-package/pull/1332#discussion_r1247181453, thanks @efd6!


internal/packages/changelog/yaml.go
```diff
@@ -108,7 +108,7 @@ func SetManifestVersion(d []byte, version string) ([]byte, error) {
 
 	// Manifest is a document, with a single element, that should be a map.
 	if len(node.Content) == 0 || node.Content[0].Kind != yaml.MappingNode {
-		return nil, fmt.Errorf("unexpected manifest content: %w", err)
+		return nil, errors.New("unexpected manifest content")
```
efd6, line 111:
```suggestion
		return nil, errors.New("unexpected manifest content: not a map")
```

efd6 (approved):

PR #1340, opened 2023-07-04
System benchmarks improvements

marc-gr:
This PR adds several new functionalities to the system benchmarks:

- Support for `k8s` and `tf` service deployers
- Support for service deployer variants
- Support to set an external metricstore to send metrics during the benchmark
- Support to reindex the benchmark data to the metricstore
- Support custom `policy_template` setting


marc-gr:
test integrations

sharbuz:
Hi everyone,

jsoriano:
test integrations

internal/benchrunner/runners/system/metrics.go
```diff
@@ -126,24 +153,98 @@ func (c *collector) collect() {
...
+	if c.msapi == nil {
+		return
+	}
+	for _, e := range events {
+		reqBody := bytes.NewReader(e)
+		resp, err := c.msapi.Index(c.indexName(), reqBody)
+		if err != nil {
+			logger.Debugf("error indexing event: %v", err)
+			continue
+		}
```
jsoriano, line 169: Do we need to close this body?
```suggestion
		}
		defer resp.Body.Close()
```

And take into account that we are keeping here a response per event. The body should be probably closed on each iteration.

internal/benchrunner/runners/system/metrics.go
```diff
@@ -247,3 +352,46 @@ func (c *collector) collectTotalHits() int {
 	}
 	return totalHits
 }
+
+func (c *collector) createEventsFromMetrics(m metrics) [][]byte {
+	dsEvent := struct {
+		Ts int64 `json:"@timestamp"`
```
jsoriano, line 358: Nit.
```suggestion
		Timestamp int64 `json:"@timestamp"`
```

internal/benchrunner/runners/system/metrics.go
```diff
@@ -126,24 +153,98 @@ func (c *collector) collect() {
...
+func (c *collector) createMetricsIndex() {
+	if c.msapi == nil {
+		return
+	}
+	reader := bytes.NewReader(
+		[]byte(`{
+			"settings": {
+				"number_of_replicas": 0
+			},
+			"mappings": {
```
jsoriano, line 191: Should we have a more specific mapping? Are all possible fields integer numbers and strings?

docs/howto/system_benchmarking.md
```diff
@@ -243,3 +401,20 @@ Finally, when you are done running the benchmark, bring down the Elastic Stack.
...
+export ELASTIC_PACKAGE_ESMETRICSTORE_USERNAME=elastic
+export ELASTIC_PACKAGE_ESMETRICSTORE_PASSWORD=changeme
+export ELASTIC_PACKAGE_ESMETRICSTORE_CA_CERT="$HOME/.elastic-package/profiles/default/certs/ca-cert.pem"
+```
+
+The only one that is optional is `ELASTIC_PACKAGE_ESMETRICSTORE_CA_CERT`.
+
+When these are detected, metrics will be automatically collected every second and sent to a new index called `bench-metrics-{dataset}-{testRunID}"`.
+
+Additionally, if the `reindex-to-metricstore` flag is used, the data generated during the benchmark will be sent to the metricstore into an index called `bench-reindex-{datastream}-{testRunID}` for further analysis. The events will be enriched with metadata related to the benchmark run.
```
jsoriano, line 420: Could we give information about the format of the stored data, or how could a user leverage it?

internal/benchrunner/runners/system/metrics.go
```diff
@@ -193,12 +295,11 @@ func (c *collector) waitUntilReady() {
 
 readyLoop:
 	for {
-		if signal.SIGINT() {
-			logger.Debug("SIGINT: cancel metrics collection")
+		select {
+		case <-c.stopC:
```
jsoriano, line 299: Is `stopC` also closed on SIGINT?

internal/benchrunner/runners/system/metrics.go
```diff
@@ -208,9 +309,13 @@ readyLoop:
 		}
 	}
 
-	if c.warmupD > 0 {
-		logger.Debugf("waiting %s for warmup period", c.warmupD)
-		<-time.After(c.warmupD)
+	if c.scenario.WarmupTimePeriod > 0 {
+		logger.Debugf("waiting %s for warmup period", c.scenario.WarmupTimePeriod)
+		select {
+		case <-c.stopC:
```
jsoriano, line 315: :+1:

docs/howto/system_benchmarking.md
jsoriano, line 420: Could this metricstore be used to compare benchmarks instead of using the reports on disk files?

internal/benchrunner/runners/system/runner.go
```diff
@@ -691,6 +708,159 @@ func (r *runner) enrollAgents() error {
...
+	if !r.options.ReindexData {
+		return nil
+	}
+	if r.options.ESMetricsAPI == nil {
+		return errors.New("the option to reindex data is set, but the metricstore was not initialized")
+	}
+
+	logger.Debug("starting reindexing of data...")
+
+	logger.Debug("gettings orignal mappings...")
```
jsoriano, line 721:
```suggestion
	logger.Debug("getting orignal mappings...")
```

internal/benchrunner/runners/system/runner.go
```diff
@@ -691,6 +708,159 @@ func (r *runner) enrollAgents() error {
...
+
+		hits, found := sr["hits"].(map[string]interface{})["hits"].([]interface{})
+		if !found || len(hits) == 0 {
+			break
+		}
+
+		var bulkBodyBuilder strings.Builder
+		for _, hit := range hits {
+			bulkBodyBuilder.WriteString(fmt.Sprintf("{\"index\":{\"_index\":\"%s\",\"_id\":\"%s\"}}\n", indexName, hit.(map[string]interface{})["_id"]))
+			enriched := r.enrichEventWithBenchmarkMetadata(hit.(map[string]interface{})["_source"].(map[string]interface{}))
```
jsoriano, line 812: Nit. Could we make these conversions safer? They could panic if something is not a map here.

internal/benchrunner/runners/system/runner.go
```diff
@@ -691,6 +708,159 @@ func (r *runner) enrollAgents() error {
 	return nil
 }
 
+func (r *runner) reindexData() error {
```
jsoriano, line 711: Nit. Could you add a comment about what this method does?

internal/benchrunner/runners/system/runner.go
```diff
@@ -691,6 +708,159 @@ func (r *runner) enrollAgents() error {
...
+			"settings": {"number_of_replicas":0},
+			"mappings": %s
+		}`, mapping)),
+	)
+
+	indexName := fmt.Sprintf("bench-reindex-%s-%s", r.runtimeDataStream, r.ctxt.Test.RunID)
+
+	logger.Debugf("creating %s index in metricstore...", indexName)
+
+	createRes, err := r.options.ESMetricsAPI.Indices.Create(
```
jsoriano, line 764: How is this index different to the one created in `createMetricsIndex`? Both are created in the metrics ES.

internal/servicedeployer/factory.go
```diff
@@ -11,12 +11,17 @@ import (
 	"path/filepath"
```
jsoriano, line 11: Nice to reuse the same service deployers :+1:

internal/servicedeployer/factory.go
```diff
@@ -67,28 +75,27 @@ func Factory(options FactoryOptions) (ServiceDeployer, error) {
...
-	packageDevDeployPath := filepath.Join(options.PackageRootPath, devDeployDir)
-	_, err = os.Stat(packageDevDeployPath)
-	if err == nil {
+	packageDevDeployPath := filepath.Join(options.PackageRootPath, options.DevDeployDir)
+	if _, err := os.Stat(packageDevDeployPath); err == nil {
 		return packageDevDeployPath, nil
 	} else if !errors.Is(err, os.ErrNotExist) {
 		return "", fmt.Errorf("stat failed for package (path: %s): %w", packageDevDeployPath, err)
 	}
-	return "", fmt.Errorf("\"%s\" directory doesn't exist", devDeployDir)
```
jsoriano, line 85: I think this was intentional, backslashes in Windows paths are escaped by `%q`, but are not by `%s`.

jsoriano (commented):
Nice refactors and new features. Added some small comments and nitpicks.

I wonder if we could document the metricstore a bit more, it is not clear to me when a user would use it and how.

It would be also good to split this PR into multiple ones, at least one for the metricstore and another one for the service deployers. Wdyt? Please don't do it if it would be a significative effort to split them now.

docs/howto/system_benchmarking.md
```diff
@@ -13,15 +13,15 @@ Conceptually, running a system benchmark involves the following steps:
 1. Create a benchmark policy that configures a single data stream for a single package.
 1. Assign the policy to the enrolled Agent(s).
 1. Metrics collections from the cluster starts. (**TODO**: record metrics from all Elastic Agents involved using the `system` integration.)
-1. **TODO**: Send the collected metrics to the ES Metricstore if set.
+1. Send the collected metrics to the ES Metricstore if set.
 1. Generate data if configured (it uses the [corpus-generator-rool](https://github.com/elastic/elastic-integration-corpus-generator-tool))
```
ebeahan, line 17:
```suggestion
1. Generate data if configured (it uses the [corpus-generator-tool](https://github.com/elastic/elastic-integration-corpus-generator-tool))
```

scripts/test-check-packages.sh
```diff
@@ -85,10 +85,9 @@ for d in test/packages/${PACKAGE_TEST_TYPE:-other}/${PACKAGE_UNDER_TEST:-*}/; do
           --old ${OLDPWD}/build/benchmark-results-old \
           --threshold 1 --report-output-path="${OLDPWD}/build/benchreport"
       fi
-      # FIXME: running system benchmark in package "system_benchmark" fails with panic
-      # if [ "${package_to_test}" == "system_benchmark" ]; then
-      #   elastic-package benchmark system --benchmark logs-benchmark -v --defer-cleanup 1s
-      # fi
+      if [ "${package_to_test}" == "system_benchmark" ]; then
+        elastic-package benchmark system --benchmark logs-benchmark -v --defer-cleanup 1s
+      fi
```
mrodm, line 90: Nice ! 
I think it can be removed from this command the `--defer-cleanup` parameter. Maybe , is it just there because of copying from the test command ?

```suggestion
      if [ "${package_to_test}" == "system_benchmark" ]; then
        elastic-package benchmark system --benchmark logs-benchmark -v
      fi
```

internal/benchrunner/runners/system/runner.go
```diff
@@ -691,6 +708,159 @@ func (r *runner) enrollAgents() error {
...
+		}
+
+		res, err = r.options.ESAPI.Scroll(
+			r.options.ESAPI.Scroll.WithScrollID(scrollId),
+			r.options.ESAPI.Scroll.WithScroll(time.Minute),
+		)
+		if err != nil {
+			return fmt.Errorf("error executing scroll: %s", err)
+		}
+		defer res.Body.Close()
```
mrodm, line 840: Should this be executed at this moment instead of defering it at the end of the function ? 

```suggestion
		res.Body.Close()
```

internal/benchrunner/runners/system/metrics.go
marc-gr: Yes, since when SIGINT happens runner ends gracefully, and the collector `stop` function is always executed, either after collection or on early return on `runner.Run`

internal/benchrunner/runners/system/runner.go
marc-gr: Metricstore by default will store metrics data from the benchmark and optionally reindexed data, to keep these 2 data types separated, we create this second index using the same mappings as the original ones from the benchmark.

docs/howto/system_benchmarking.md
marc-gr: Correct, in a following PR, when the comparison functionality is added, will be able to do it from both places.

scripts/test-check-packages.sh
marc-gr: It is there on purpose. I think it might be good to debug the benchmark setup in some circumstances since it let you prevent tearing down services and cleaning up the data.

internal/benchrunner/runners/system/metrics.go
marc-gr: all values should be properly mapped with this one as there are only integers, strings and booleans in the metrics

marc-gr:
test integrations

marc-gr:
test integrations

marc-gr:
test

cmd/benchmark.go
```diff
@@ -375,3 +400,30 @@ func generateDataStreamCorpusCommandAction(cmd *cobra.Command, _ []string) error
...
+		logger.Debugf("can't initialize metricstore, missing environment configuration")
+		return nil, nil
+	}
+
+	esClient, err := elasticsearch.NewClient(
+		elasticsearch.OptionWithAddress(address),
+		elasticsearch.OptionWithUsername(user),
+		elasticsearch.OptionWithPassword(pass),
+		elasticsearch.OptionWithCertificateAuthority(cacert),
+	)
```
mrodm, line 419: I think this client should be created with `stack.NewElasticsearchClient` method instead of the elasticsearch one, so it follows the same approach as in other modules. 
```suggestion
	esClient, err := stack.NewElasticsearchClient(
		elasticsearch.OptionWithAddress(address),
		elasticsearch.OptionWithUsername(user),
		elasticsearch.OptionWithPassword(pass),
		elasticsearch.OptionWithCertificateAuthority(cacert),
	)
```

jsoriano (approved):

Issue #1348, opened 2023-07-10
"experimentalDataStreamSettings" configuration not allowed in 8.10.0-SNAPSHOT


mrodm:
Relates https://github.com/elastic/integrations/pull/6843

When running `elastic-package stack up -v -d --version 8.10.0-SNAPSHOT` the following error makes kibana container fail:

```
[2023-07-10T14:46:15.574+00:00][FATAL][root] Error: [config validation of [xpack.fleet].enableExperimental]: [experimentalDataStreamSettings] is not allowed. Allowed values are: createPackagePolicyMultiPageLayout, packageVerification, showDevtoolsRequest, diagnosticFileUploadEnabled, displayAgentMetrics, showIntegrationsSubcategories, agentFqdnMode, showExperimentalShipperOptions, agentTamperProtectionEnabled, secretsStorage
...
 FATAL  Error: [config validation of [xpack.fleet].enableExperimental]: [experimentalDataStreamSettings] is not allowed. Allowed values are: createPackagePolicyMultiPageLayout, packageVerification, showDevtoolsRequest, diagnosticFileUploadEnabled, displayAgentMetrics, showIntegrationsSubcategories, agentFqdnMode, showExperimentalShipperOptions, agentTamperProtectionEnabled, secretsStorage
```

Related Jenkins job: https://fleet-ci.elastic.co/blue/organizations/jenkins/Ingest-manager%2Fintegrations/detail/PR-6843/4/pipeline/

PR #1350, opened 2023-07-10
[Fleet] Fix version constraint for experimentalDataStreamSettings

jillguyonnet:
Fixes https://github.com/elastic/elastic-package/issues/1348

The Kibana version constraint of the `experimentalDataStreamSettings` was amended in https://github.com/elastic/elastic-package/pull/1328 to version > 8.7.0 and < 8.10.0. However, this fails for version 8.10.0-SNAPSHOT which is lower than 8.10.0.

This PR fixes the constraint and adds a unit test to the `TestSemverLessThan` function.


mrodm (approved):
Thanks! :+1:

Issue #1354, opened 2023-07-12
export dashboards issue


agithomas:
Observed with versions : 8.8.0, 8.8.2
Package used : Memcached

Steps to reproduce the issue: 
1. Start the stack with the command `elastic-package stack up --version 8.8.2 -v -d`
2. Add a Memcached integration
3. Export the dashboard with the command `elastic-package export dashboards`. Choose `[Metrics Memcached] Overview`
4. Update the changelog , manifest.yml file to next version for memached. Run - `elastic-package build; elastic-package stack up -v -d --services package-registry`
5. Go to kibana -> Integration package -> upgrade Memcached package
6. You will notice the error as below
<img width="344" alt="image" src="https://github.com/elastic/elastic-package/assets/101976829/eccf39a1-d487-49de-94ed-a03e4aa80830">

agithomas:
> @agithomas can you try to install the package with `elastic-package install` and check the error reported there?

Please find the error below

```
agikthomas@service-integration-dev-idc-1:~/go/src/github.com/elastic/integrations/packages/memcached$ elastic-package install memcached
Install the package
Error: can't install the package: can't install the package: could not install package; API status code = 500; response body = {"statusCode":500,"error":"Internal Server Error","message":"Migration function for version 7.13.1 threw an error"}
```

mrodm:
I was able to reproduce this error, when that happened this error was in Kibana container:
`Caused by:
TypeError: Cannot read properties of undefined (reading 'layers')`

this is the whole stack trace:
```
[2023-07-14T10:37:17.805+00:00][ERROR][savedobjects-service] TypeError: Cannot read properties of undefined (reading 'layers')
    at commonRenameOperationsForFormula (/usr/share/kibana/node_modules/@kbn/lens-plugin/server/migrations/common_migrations.js:29:78)
    at Object.7.13.1 (/usr/share/kibana/node_modules/@kbn/lens-plugin/server/embeddable/make_lens_embeddable_factory.js:25:91)
    at migrateFn (/usr/share/kibana/node_modules/@kbn/embeddable-plugin/common/lib/migrate.js:23:48)
    at migrations.<computed> (/usr/share/kibana/node_modules/@kbn/embeddable-plugin/common/lib/get_all_migrations.js:34:10)
    at /usr/share/kibana/node_modules/@kbn/dashboard-plugin/server/dashboard_saved_object/migrations/migrate_by_value_dashboard_panels.js:50:29
[...]

PR #1359, opened 2023-07-17
Prevent exporting dashboards from 8.8

jsoriano:
Dashboards exported from Kibana 8.8 may not be installed till Kibana 8.10 due to an issue with migrations.

See:
- Issue report: https://github.com/elastic/elastic-package/issues/1354
- Fix in Fleet: https://github.com/elastic/kibana/pull/161969

Closes https://github.com/elastic/elastic-package/issues/1354

### DIFF
--- a/internal/export/dashboards.go
+++ b/internal/export/dashboards.go
@@ -6,9 +6,12 @@ package export
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/Masterminds/semver/v3"
 
 	"github.com/elastic/elastic-package/internal/common"
 	"github.com/elastic/elastic-package/internal/kibana"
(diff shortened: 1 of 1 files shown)


internal/export/dashboards.go
```diff
@@ -51,6 +61,23 @@ func Dashboards(kibanaClient *kibana.Client, dashboardsIDs []string) error {
...
 }
 
+func checkKibanaVersion(info kibana.VersionInfo) error {
+	version, err := semver.NewVersion(info.Number)
+	if err != nil {
+		return fmt.Errorf("cannot parse version %s: %w", info.Number, err)
+	}
+
+	minVersion := semver.MustParse("8.8.0")
+	maxVersion := semver.MustParse("8.10.0")
```
mrodm, line 71: This depends on what value contains `version` variable.

If this variable contains prerelease tags, `8.10.0-SNAPSHOT` would not be allowed to export dashboards:

https://go.dev/play/p/HZD1ehU46FI

mrodm (commented):
LGTM
Just a doubt if it is used an Elastic stack with a SNAPSHOT version.

internal/export/dashboards.go
jsoriano: Tested with 8.10.0-SNAPSHOT and it works as expected.

mrodm (approved):
:+1:

PR #1372, opened 2023-07-26
Support routing_rules.yml and ingest pipeline test for reroute processor

kaiyan-sheng:
## What does this PR do?

This PR does two things: 
1. Support `routing_rules.yml` file when running pipeline test using command `elastic-package test pipeline -d <datastream> --generate -v`.

`routing_rules.yml` now can be read and take into account when running the test pipeline command. The file is read as yml and convert into reroute processors. Then these reroute processors are attached under the `processors` section from the ingest pipeline.

For example: 
```
- source_dataset: awsfirehose.log
  rules:
    - target_dataset: aws.cloudtrail
      if: ctx['aws.cloudwatch.log_stream'].contains('CloudTrail')
      namespace:
        - "{{labels.data_stream.namespace}}"
        - default
```

`routing_rules.yml` looks like the example above will be translated into:
```
- reroute:
      tag: awsfirehose.log
      if: ctx['aws.cloudwatch.log_stream'].contains('CloudTrail')
      dataset: aws.cloudtrail
      namespace: ["{{labels.data_stream.namespace}}", default]
``` 

2. Support ingest pipeline test for reroute processors when they are added under `elasticsearch/ingest_pipeline` directly.

- This PR added the index name for testing based on the dataset so we don't see error like this:

```
[0] unexpected pipeline error: invalid data stream name: [_index]; must follow naming scheme <type>-<dataset>-<namespace>
```

- When checking the value for `data_stream.dataset`, it's read from the reroute processors for the new target dataset. Because reroute processor will change the value of this field.

I've been testing this PR locally with https://github.com/elastic/integrations/pull/7146. After `make build` elastic package locally, I ran `elastic-package test pipeline -d log --generate -v` and under `integrations/packages/awsfirehose/data_stream/log/_dev/test/pipeline/test-cloudtrail.json-expected.json`, you can see the `data_stream.dataset` value changed to `aws.cloudtrail` and `data_stream.namespace` changed to `default`. These are the two fields I defined under `reroute` processor:
```
  - reroute:
      if: ctx['aws.cloudwatch.log_stream'].contains('CloudTrail')
      dataset: aws.cloudtrail
      namespace: default
```

Same thing works for defining it under `routing_rules.yml`
```
- source_dataset: awsfirehose.log
  rules:
    - target_dataset: aws.cloudtrail
      if: ctx['aws.cloudwatch.log_stream'].contains('CloudTrail')
      namespace:
        - "{{labels.data_stream.namespace}}"
        - default
```


internal/elasticsearch/ingest/pipeline.go
```diff
@@ -74,6 +75,7 @@ func SimulatePipeline(api *elasticsearch.API, pipelineName string, events []json
 	var request simulatePipelineRequest
 	for _, event := range events {
 		request.Docs = append(request.Docs, pipelineDocument{
+			Index:  "logs-generic-default",
```
felixbarny, line 78: If the `event` contains `data_stream.*` values, it would be ideal to use that as the index name as that would simulate the reality more closely.

internal/fields/validate.go
```diff
@@ -320,6 +320,11 @@ func (v *Validator) validateDocumentValues(body common.MapStr) multierror.Error
 				continue
 			}
 
+			// skip checking `data_stream.dataset` because reroute processor will change this field value
```
felixbarny, line 323: Note that the reroute processor also changes `event.dataset` so that it matches `data_stream.dataset`.

internal/fields/validate.go
```diff
@@ -320,6 +320,11 @@ func (v *Validator) validateDocumentValues(body common.MapStr) multierror.Error
...
+			if datasetField == "data_stream.dataset" {
```
felixbarny, line 324: Rather than skipping the validation, would it make sense to set up the validator with the new expected dataset instead? Disclaimer that I'm not familiar with the code base.

jsoriano:
> This PR is to support ingest pipeline test for reroute processor.

Hey @kaiyan-sheng,

This PR adds support to test the reroute processors defined on ingest pipelines, right? But the recommended way to add routing rules is to use the `routing_rules.yml` file (as defined in https://github.com/elastic/package-spec/issues/514).

I think that appart of these changes, for pipeline tests, we would need to make `elastic-package` to install the routing rules in a way that they apply when using the simulate API.

It would be also great to add a test package here.

jsoriano:
Also, by reading the `routing_rules.yml` file, tests might know on what datasets data stream to look for the data.

felixbarny:
> This PR adds support to test the reroute processors defined on ingest pipelines, right? But the recommended way to add routing rules is to use the routing_rules.yml file (as defined in https://github.com/elastic/package-spec/issues/514).

The routing rules in `routing_rules.yml` will be translated into reroute processors that are added in the package's pipeline. So in theory, they should be tested by `test pipeline` command. If not, that's an issue we should fix.

jsoriano:
> > This PR adds support to test the reroute processors defined on ingest pipelines, right? But the recommended way to add routing rules is to use the routing_rules.yml file (as defined in [elastic/package-spec#514](https://github.com/elastic/package-spec/issues/514)).
> 
> The routing rules in `routing_rules.yml` will be translated into reroute processors that are added in the package's pipeline. So in theory, they should be tested by `test pipeline` command. If not, that's an issue we should fix.

Yes, they are translated to reroute processors when installed through Fleet, but I think that currently `elastic-package test pipelines` only installs the pipelines, it doesn't install the package, so it is probably not installing the routing rules. We should probably change this.

kaiyan-sheng:
@jsoriano Yep it only looks for the yml files under `elasticsearch/ingest_pipeline` for pipelines right now.  Can I keep this PR for only supporting ingest pipelines for reroute processors and then have a separate PR for supporting `routing_rules.yml`?

zmoog:
More than happy to help with adding support for the `routing-rules.yml` to the pipeline testing; I also need this for the Kubernetes integration. I'm available for both implementation and review.

internal/elasticsearch/ingest/datastream.go
```diff
@@ -16,12 +16,25 @@ import (
...
 var ingestPipelineTag = regexp.MustCompile(`{{\s*IngestPipeline.+}}`)
 
+type Rule struct {
+	TargetDataset interface{}   `yaml:"target_dataset"`
+	If            interface{}   `yaml:"if"`
+	Namespace     []interface{} `yaml:"namespace"`
+}
+
+type SourceData struct {
+	SourceDataset interface{} `yaml:"source_dataset"`
```
zmoog, line 34: According to the the package-spec 2.9.0, the [source_dataset](https://github.com/elastic/package-spec/blob/067a0e5656a1603d03ff7f7589c1d9517becca3d/spec/integration/data_stream/routing_rules.spec.yml#L64-L68) is a `string`. Do you think we use the `string` type here, or there's a reason to use `inerface{}`?

internal/elasticsearch/ingest/datastream.go
```diff
@@ -16,12 +16,25 @@ import (
...
 	"github.com/elastic/elastic-package/internal/elasticsearch"
 	"github.com/elastic/elastic-package/internal/packages"
 )
 
 var ingestPipelineTag = regexp.MustCompile(`{{\s*IngestPipeline.+}}`)
 
+type Rule struct {
+	TargetDataset interface{}   `yaml:"target_dataset"`
+	If            interface{}   `yaml:"if"`
+	Namespace     []interface{} `yaml:"namespace"`
```
zmoog, line 30: According the the package-spec 2.9.0:

- [target_dataset](https://github.com/elastic/package-spec/blob/067a0e5656a1603d03ff7f7589c1d9517becca3d/spec/integration/data_stream/routing_rules.spec.yml#L11-L27) and [namespace](https://github.com/elastic/package-spec/blob/067a0e5656a1603d03ff7f7589c1d9517becca3d/spec/integration/data_stream/routing_rules.spec.yml#L34-L51) can both be `string` or an array of `string`.
- [if](https://github.com/elastic/package-spec/blob/067a0e5656a1603d03ff7f7589c1d9517becca3d/spec/integration/data_stream/routing_rules.spec.yml#L28-L33) can only be a `string`

Maybe we can unmarshall them using a dedicated struct like:

```golang
type StringOrArray []string

func (s *StringOrArray) UnmarshalYAML(value *yaml.Node) error {
	var multi []string
	err := value.Decode(&multi)
	if err != nil {
		var single string
		err := value.Decode(&single)
		if err != nil {
			return err
		}
		*s = []string{single}
	} else {
		*s = multi
	}
	return nil
}

type Rule struct {
	TargetDataset StringOrArray `yaml:"target_dataset"`
	Namespace     StringOrArray `yaml:"namespace"`
	If            string        `yaml:"if"`
}
```

internal/elasticsearch/ingest/datastream.go
```diff
@@ -75,12 +92,43 @@ func loadIngestPipelineFiles(dataStreamPath string, nonce int64) ([]Pipeline, er
 			Path:    path,
 			Name:    getPipelineNameWithNonce(name[:strings.Index(name, ".")], nonce),
 			Format:  filepath.Ext(path)[1:],
-			Content: c,
+			Content: append(c, routingPipeline...),
```
zmoog, line 95: I think we should add the `reroute` processor to the default pipeline only.

For example, here are the pipeline source and what Fleet generated during the package install:

Source:

```yaml
description: Pipeline for Kubernetes container logs
processors:
  - set:
      field: service.name
      copy_from: kubernetes.labels.app_kubernetes_io/name
      ignore_empty_value: true
on_failure:
  - set:
      field: event.kind
      value: pipeline_error
  - append:
      field: error.message
      value: '{{{ _ingest.on_failure_message }}}'

```

Generated:

```json
{
  "logs-kubernetes.container_logs-1.46.0": {
    "description": "Pipeline for Kubernetes container logs",
    "processors": [
      {
        "set": {
          "field": "service.name",
          "copy_from": "kubernetes.labels.app_kubernetes_io/name",
          "ignore_empty_value": true
        }
      },
      {
        "pipeline": {
          "name": "logs-kubernetes.container_logs@custom",
          "ignore_missing_pipeline": true
        }
      },
      {
        "reroute": {
          "tag": "kubernetes.container_logs",
          "dataset": [
            "{{kubernetes.labels.elastic_co/dataset}}",
            "{{data_stream.dataset}}",
            "kubernetes.container_logs"
          ],
          "namespace": [
            "{{kubernetes.labels.elastic_co/namespace}}",
            "{{data_stream.namespace}}",
            "default"
          ],
          "if": "ctx?.kubernetes?.labels != null"
        }
      }
    ],
    "on_failure": [
      {
        "set": {
          "field": "event.kind",
          "value": "pipeline_error"
        }
      },
      {
        "append": {
          "field": "error.message",
          "value": "{{{ _ingest.on_failure_message }}}"
        }
      }
    ],
    "_meta": {
      "managed_by": "fleet",
      "managed": true,
      "package": {
        "name": "kubernetes"
      }
    }
  }
}
```

internal/elasticsearch/ingest/datastream.go
```diff
@@ -75,12 +92,43 @@ func loadIngestPipelineFiles(dataStreamPath string, nonce int64) ([]Pipeline, er
...
+	constructPipeline := ""
+	for _, srcData := range data {
+		for _, rule := range srcData.Rules {
+			rerouteProcessor := "  - reroute:\n"
+			rerouteProcessor += "      tag: " + srcData.SourceDataset.(string) + "\n"
+			rerouteProcessor += "      if: " + rule.If.(string) + "\n"
+			//TODO: deal with multiple datasets?
+			rerouteProcessor += "      dataset: " + rule.TargetDataset.(string) + "\n"
+			//TODO: deal with multiple namespaces?
+			rerouteProcessor += "      namespace: " + rule.Namespace[0].(string) + "\n"
```
zmoog, line 125: We must support multiple datasets and namespaces.

internal/elasticsearch/ingest/datastream.go
zmoog: I ran a quick test, and Fleet added the reroute processor at the end of the processor list for the default pipeline only.

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: Yep thats why I added the TODOs later. I'm trying to figure out what's the use case for an array of namespaces and target_dataset.

zmoog: > Yep thats why I added the TODOs later. 

Oh, god it! The question marks deceived me 😇 


> I'm trying to figure out the use case for an array of namespaces and target_dataset.

They can be pretty handy. The processor uses the first dataset and namespace with an actual value in the array, like the patterns in the grok. 

For example, in the Kubernetes use case:

```yaml
# Route container logs entries to the correct dataset and namespace based on the pod labels
- source_dataset: kubernetes.container_logs
  rules:
    - target_dataset:
        - "{{kubernetes.labels.elastic_co/dataset}}"
        - "{{data_stream.dataset}}"
        - kubernetes.container_logs
      namespace:
        - "{{kubernetes.labels.elastic_co/namespace}}"
        - "{{data_stream.namespace}}"
        - default
      if: "ctx?.kubernetes?.labels != null"
```

For the namespace, it will use the `{{kubernetes.labels.elastic_co/namespace}}` field if it exists, or try the next one until the literal value `default`.

kaiyan-sheng: done 😬 Thanks!!!

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: yep added!

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -108,6 +108,18 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 		expectedDataset = pkgManifest.Name + "." + r.options.TestFolder.DataStream
 	}
 
+	// when reroute processors are used, expectedDataset should be set to "" to avoid checking the dataset fields
+	for _, pipeline := range r.pipelines {
+		if processors, err := pipeline.Processors(); err == nil {
+			for _, p := range processors {
+				if p.Type == "reroute" {
+					expectedDataset = ""
```
jsoriano, line 116: Should the expected dataset be one of the indicated in the reroute processor?

internal/elasticsearch/ingest/datastream.go
```diff
@@ -70,6 +89,27 @@ func loadIngestPipelineFiles(dataStreamPath string, nonce int64) ([]Pipeline, er
...
+
+		// Unmarshal the YAML data into a ESIngestPipeline struct
+		var esPipeline ESIngestPipeline
+		err = yaml.Unmarshal(c, &esPipeline)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal ingest pipeline YAML data (path: %s): %w", path, err)
+		}
+
+		// read routing_rules.yml and convert it into reroute processors in ingest pipeline
+		rerouteProcessors, err := loadRoutingRuleFile(dataStreamPath)
```
jsoriano, line 101: Could you please add a test package that has a routing rule file?

internal/elasticsearch/ingest/datastream.go
```diff
@@ -81,6 +121,73 @@ func loadIngestPipelineFiles(dataStreamPath string, nonce int64) ([]Pipeline, er
...
+		for _, rule := range srcData.Rules {
+			var td []string
+			switch rule.TargetDataset.(type) {
+			case string:
+				td = []string{rule.TargetDataset.(string)}
+			case []string:
+				td = rule.TargetDataset.([]string)
+			case []interface{}:
+				for _, n := range rule.TargetDataset.([]interface{}) {
+					td = append(td, n.(string))
```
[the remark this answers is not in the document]
kaiyan-sheng: Good point! Added! I also added a default for the switch case to catch other misformats.

internal/testrunner/runners/pipeline/runner.go
kaiyan-sheng: Yes! Just changed the code to get expectedDataset from the reroute processors. Thanks!

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: Yes thank you!! I added a package called awsfirehose with a routing rule file, how does this test package work?

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: Thanks @zmoog ! Does that mean if we have both `default.yml` and `third-party.yml` (for example) and `routing_rules.yml`, the reroute processors will only be attached to default.yml processors not the `third-party.yml` ?

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -103,9 +111,27 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
...
+			return nil, fmt.Errorf("unmarshalling ingest pipeline content failed: %w", err)
+		}
+		for _, processor := range esIngestPipeline.Processors {
+			if reroute, ok := processor["reroute"]; ok {
+				expectedDatasets = append(expectedDatasets, reroute.Dataset...)
+			}
+		}
+	}
+
+	if expectedDatasets == nil {
```
jsoriano, line 129: Nit. Just in case it is initialized in some refactor.
```suggestion
	if len(expectedDatasets) == 0 {
```

test/packages/parallel/awsfirehose/data_stream/log/_dev/test/pipeline/test-cloudtrail.json-expected.json
```diff
@@ -0,0 +1,28 @@
+{
```
jsoriano, line 1: I guess that with the reroute processor it can happen that events are written into multiple data streams, should we take this into account and maybe have multiple expectation files, one per data stream?

jsoriano (commented):
Thanks for adding tests! It looks like expectation files need to be regenerated.

test/packages/parallel/awsfirehose/data_stream/log/_dev/test/pipeline/test-cloudtrail.json-expected.json
felixbarny: Even with the reroute processor, a single create document request will result in that document being stored in at most one data stream.

jsoriano: Yes, I was wondering if we want to check that each document has been stored in the expected data stream.

felixbarny: Ah, I guess I misunderstood you then. Different documents can definitely be routed to different data streams and I agree we should test that. In some sense, we're already testing it by asserting on the `data_stream` fields. Is there more you had in mind, such as checking the `_index`? That would be a bit more challenging as the `_index` field is not part of the document itself.

kaiyan-sheng: Hmm even though dataset and namespace can be a list of strings, my understanding is it's either using the value of a field or the given value. For example: 
```
{
      "reroute": {
        "tag": "mydata",
        "dataset": [
          "{{container.name}}",
          "mydata"
        ]
      }
    }
```
If the `container.name` field exists in the document, it will be routed — otherwise it falls back to `mydata`.

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: Yep I just tested with `third-party.yml` in Kibana and `routing_rules.yml` only got attached to the `default.yml` pipeline. Will make sure this is the behavior with this PR! Thank you!

test/packages/parallel/awsfirehose/data_stream/log/_dev/test/pipeline/test-cloudtrail.json-expected.json
felixbarny: Yes, that's correct. For an individual document, there’s only one target data stream. The multiple values are fallbacks. But different documents can have different data streams, depending on which values are present and which if condition in the reroute processors match.

felixbarny: I think it would be good to add another test document here where the condition (`if: ctx['aws.cloudwatch.log_stream'].contains('CloudTrail')`) does not match so that the document just lands in the `"awsfirehose.log"` data stream.

internal/elasticsearch/ingest/datastream.go
zmoog: Yeah, exactly! Kibana only injects the rules into the default pipeline (defined in the `default.yml` file). 

I'm sorry, but somehow I missed your comment from Aug 1.

internal/testrunner/runners/system/runner.go
```diff
@@ -15,6 +15,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/elastic/elastic-package/internal/elasticsearch/ingest"
+
+	"gopkg.in/yaml.v3"
+
```
jsoriano, line 21: Nit. Imports order.
```suggestion

	"gopkg.in/yaml.v3"

	"github.com/elastic/elastic-package/internal/elasticsearch/ingest"
```

jsoriano (approved):
Thanks!

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: Done :) Good catch! Thanks!

internal/elasticsearch/ingest/datastream.go
```diff
@@ -16,11 +16,41 @@ import (
...
+type Rule struct {
+	TargetDataset interface{} `yaml:"target_dataset"`
+	If            string      `yaml:"if"`
+	Namespace     interface{} `yaml:"namespace"`
+}
+
+type SourceData struct {
+	SourceDataset string `yaml:"source_dataset"`
+	Rules         []Rule `yaml:"rules"`
+}
```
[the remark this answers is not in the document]
kaiyan-sheng: Good point! Done! Thank you for the review!

PR #1380, opened 2023-08-03
[system testing] Add docker network to stack to communicate with service deployers

bhapas:
This PR adds a docker network in the `elastic-package stack up` docker-compose so that any service deployer services can communicate with the stack from other docker container groups.

An example scenario would be :  A `localstack` container should communicate with `terraform service deployer` to write resources in localstack and in turn the `elastic-package-stack` needs to communicate with localstack to read SQS messages / S3 objects.

Hence, a network is created when the `elastic-package stack up` is run and any services that need to communicate in the group shall connect to this network. The docker-compose version is upgraded to `3.5` as well.


bhapas:
test integrations

PR #1385, opened 2023-08-08
Keep all fields of pipelines when adding reroute rules

jsoriano:
Avoid modifying pipelines when unmarshalling them to add reroute rules. For that,
unmarshal them to a generic map, instead to a struct with limited fields.


jsoriano:
test integrations

jsoriano:
Merging as failures seen in https://github.com/elastic/integrations/pull/7312 don't seem to be related.

Issue #1389, opened 2023-08-11
routing_rules.yml should allow no namespace and no target_dataset


kaiyan-sheng:
Currently, we dont allow `routing_rules.yml` not to specify a namespace or target_dataset. But in reroute processor, namespace and dataset are both optional fields.

PR #1391, opened 2023-08-11
Support routing_rules.yml with optional config options

kaiyan-sheng:
`source_dataset`, `if` condition, `target_dataset` and `namespace` are all optional according to reroute processor documentation. This PR is to add support in elastic-package for routing_rules.yml to have optional configs. 

closes https://github.com/elastic/elastic-package/issues/1389

### DIFF
--- a/internal/elasticsearch/ingest/datastream.go
+++ b/internal/elasticsearch/ingest/datastream.go
@@ -28,9 +28,9 @@ var (
 )
 
 type Rule struct {
-	TargetDataset interface{} `yaml:"target_dataset"`
-	If            string      `yaml:"if"`
-	Namespace     interface{} `yaml:"namespace"`
+	TargetDataset *interface{} `yaml:"target_dataset"`
+	If            *string      `yaml:"if"`
+	Namespace     *interface{} `yaml:"namespace"`
 }
 
 type RoutingRule struct {
@@ -192,20 +192,27 @@ func loadRoutingRuleFile(dataStreamPath string) ([]map[string]interface{}, error
 			}
 
 			processor := make(map[string]interface{})
-			processor["reroute"] = RerouteProcessor{
+			rp := RerouteProcessor{
 				Tag:       r.SourceDataset,
-				If:        rule.If,
 				Dataset:   td,
 				Namespace: ns,
 			}
+			if rule.If != nil {
+				rp.If = *rule.If
+			}
+			processor["reroute"] = rp
 			rerouteProcessors = append(rerouteProcessors, processor)
 		}
 	}
 	return rerouteProcessors, nil
 }
 
-func convertValue(value interface{}, label string) ([]string, error) {
-	switch value := value.(type) {
+func convertValue(originalValue *interface{}, label string) ([]string, error) {
+	if originalValue == nil {
+		return nil, nil
+	}
+
+	switch value := (*originalValue).(type) {
 	case string:
 		return []string{value}, nil
 	case []string:
--- a/internal/elasticsearch/ingest/datastream.go
+++ b/internal/elasticsearch/ingest/datastream.go
@@ -208,10 +208,6 @@ func loadRoutingRuleFile(dataStreamPath string) ([]map[string]interface{}, error
 }
 
 func convertValue(originalValue interface{}, label string) ([]string, error) {
-	if originalValue == nil {
-		return nil, nil
-	}
-
 	switch value := originalValue.(type) {
 	case string:
 		return []string{value}, nil
@@ -227,6 +223,8 @@ func convertValue(originalValue interface{}, label string) ([]string, error) {
 			}
 		}
 		return result, nil
+	case nil:
+		return nil, nil
 	default:
 		return nil, fmt.Errorf("%s in routing_rules.yml has to be a string or an array of strings: %v", label, value)
 	}
--- a/internal/elasticsearch/ingest/datastream.go
+++ b/internal/elasticsearch/ingest/datastream.go
@@ -192,23 +192,20 @@ func loadRoutingRuleFile(dataStreamPath string) ([]map[string]interface{}, error
 			}
 
 			processor := make(map[string]interface{})
-			rp := RerouteProcessor{
+			processor["reroute"] = RerouteProcessor{
 				Tag:       r.SourceDataset,
+				If:        rule.If,
 				Dataset:   td,
 				Namespace: ns,
 			}
-			if rule.If != "" {
-				rp.If = rule.If
-			}
-			processor["reroute"] = rp
 			rerouteProcessors = append(rerouteProcessors, processor)
 		}
 	}
 	return rerouteProcessors, nil
 }
 
-func convertValue(originalValue interface{}, label string) ([]string, error) {
-	switch value := originalValue.(type) {
+func convertValue(value interface{}, label string) ([]string, error) {
+	switch value := value.(type) {
 	case string:
 		return []string{value}, nil
 	case []string:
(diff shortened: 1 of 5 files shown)


internal/elasticsearch/ingest/datastream.go
```diff
@@ -28,9 +28,9 @@ var (
 )
 
 type Rule struct {
-	TargetDataset interface{} `yaml:"target_dataset"`
-	If            string      `yaml:"if"`
-	Namespace     interface{} `yaml:"namespace"`
+	TargetDataset *interface{} `yaml:"target_dataset"`
+	If            *string      `yaml:"if"`
+	Namespace     *interface{} `yaml:"namespace"`
```
jsoriano, line 33: An `interface{}` can be checked directly against nil, we wouldn't need a pointer to it.
```suggestion
	TargetDataset interface{} `yaml:"target_dataset"`
	If            *string      `yaml:"if"`
	Namespace     interface{} `yaml:"namespace"`
```

Or would it be possible to use omitempty instead?
```suggestion
	TargetDataset interface{} `yaml:"target_dataset,omitempty"`
	If            string      `yaml:"if,omitempty"`
	Namespace     interface{} `yaml:"namespace,omitempty"`
```

internal/elasticsearch/testdata/routing_rules/good/no_target_dataset_namespace_if/routing_rules.yml
```diff
@@ -0,0 +1,11 @@
+- source_dataset: missing_target_dataset
+  rules:
+    - if: ctx['aws.cloudwatch.log_stream'].contains('Test1')
+      namespace: default
```
jsoriano, line 4: What would be the purpose of this reroute rule? Only to reroute to the data stream of another namespace? Would this make sense in a package?

kaiyan-sheng: Yeah when `dataset` is not given, the original dataset part of the index name will be the fallback. For this case, it will just reroute to a different namespace. 
I'm only adding these cases for testing since `dataset` is not required by [reroute processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/reroute-processor.html). Probably this is not a common use case. But for reroute processors to not have a given namespace, which will fallback to the old namespace, is common.

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: I guess it's about when `convertValue` is called and when field is empty it doesn't fall into any of the switch cases. So now since I added a check in `convertValue` for `if originalValue == nil`, it's all good :)

internal/elasticsearch/ingest/datastream.go
```diff
@@ -192,20 +192,27 @@ func loadRoutingRuleFile(dataStreamPath string) ([]map[string]interface{}, error
...
 }
 
-func convertValue(value interface{}, label string) ([]string, error) {
-	switch value := value.(type) {
+func convertValue(originalValue interface{}, label string) ([]string, error) {
+	if originalValue == nil {
+		return nil, nil
+	}
+
+	switch value := originalValue.(type) {
```
jsoriano, line 215: Nit. Make this a case in the switch?
```suggestion
	switch value := originalValue.(type) {
	case nil:
		return nil, nil
```

internal/elasticsearch/testdata/routing_rules/good/no_target_dataset_namespace_if/routing_rules.yml
jsoriano: Yes, I see the use case in a general sense of the reroute processor. But would a reroute rule without a target dataset be included in a package? Do you have an example for this?
I wonder if we should allow this in packages.

internal/elasticsearch/ingest/datastream.go
kaiyan-sheng: Oh yes what am I thinking?! Thanks!!

internal/elasticsearch/testdata/routing_rules/good/no_target_dataset_namespace_if/routing_rules.yml
kaiyan-sheng: Yeah for inside a package, I don't have any use case for without a target dataset. @felixbarny Do you know any example for this?

felixbarny: It should even be allowed to create a completely empty reroute processor to route according to the dataset and namespace values in the document. For example, when the data_stream.* values have been modified via a `set` processor.

jsoriano: Btw, @mrodm pointed me that these properties are mandatory now in the spec. If we want any of them to be actually optional we would need to change the spec first.
https://github.com/elastic/package-spec/blob/32eb741e298a29a505029e324b2e7c50e3031cef/spec/integration/data_stream/routing_rules.spec.yml#L52-L55

felixbarny: > @felixbarny would such a rule be distributed in a package?

It might be, yes. I don't think it makes sense to restrict what's possible to do with packages. We might not need it... until we do need it. I think the package spec spec should go out of the way and not be too restrictive.

jsoriano: I would prefer to have a use case for this before relaxing the check. In the meantime this is protecting developers from forgetting to set options that look needed in all the known cases.

Users can still use custom processing to add any routing rule they want.

kaiyan-sheng: Thank you @jsoriano @felixbarny for your input!! For my use case, we definitely don't need to specify the `namespace` part. While we are thinking about use cases for optional `target_dataset`, should we first make the change in the spec and here to make `namespace` optional?

jsoriano: Yep, namespace can make more sense as optional in rules distributed in packages. @kaiyan-sheng would you like to open the PR to change this in package spec?

kaiyan-sheng: Yes I can do that :) Thanks!

kaiyan-sheng: @jsoriano The PR for package spec is merged. Do we need to wait for the new spec version before moving forward on this PR?

jsoriano: Well, CI in this PR is not failing, I guess we can go on, though it won't be possible to use this in packages till all pieces are released.

jsoriano (commented):
Sorry! I see I have a pending review :facepalm:

internal/elasticsearch/ingest/datastream.go
```diff
@@ -192,20 +192,23 @@ func loadRoutingRuleFile(dataStreamPath string) ([]map[string]interface{}, error
...
-			processor["reroute"] = RerouteProcessor{
+			rp := RerouteProcessor{
 				Tag:       r.SourceDataset,
-				If:        rule.If,
 				Dataset:   td,
 				Namespace: ns,
 			}
+			if rule.If != "" {
+				rp.If = rule.If
+			}
```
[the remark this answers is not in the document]
kaiyan-sheng: yep if is still required... sorry this needs to be removed!

jsoriano (approved):

PR #1393, opened 2023-08-14
Render Mustache expressions in routing rules datasets values

### LINKED ISSUE #1388: Pipeline test fails when target datasets in routing rules contain an expression
elastic-package version: `0.85.0`

Given the following `routing_rules.yml` file:

```yaml
- source_dataset: kubernetes.container_logs
  rules:
    - target_dataset:
        - "{{kubernetes.labels.elastic_co/dataset}}"
        - "{{data_stream.dataset}}"
        - kubernetes.container_logs
      namespace:
        - "{{kubernetes.labels.elastic_co/namespace}}"
        - "{{data_stream.namespace}}"
        - default
      if: "ctx?.kubernetes?.labels != null"
```

and `test-common-config.yml` file:

```yaml
fields:
  data_stream:
    dataset: kubernetes.container_logs
    namespace: default
  kubernetes:
    labels:
      elastic_co/dataset: kubernetes.container_logs.nginx
      elastic_co/namespace: nginx
      app_kubernetes_io/version: "v0.1.0"
      app_kubernetes_io/name: "myservice"
```

and `test-nginx.log` file:

```text
2023/07/25 15:24:11 [notice] 1#1: start worker process 33
2023/07/25 15:24:11 [notice] 1#1: start worker process 34
2023/07/25 15:24:11 [notice] 1#1: start worker process 35
2023/07/25 15:24:11 [notice] 1#1: using the "epoll" event method
```

When I run `elastic-package test pipeline -v -g` command I get the following error:

```shell
2023/08/10 19:23:46 DEBUG Enable verbose logging
--- Test results for package: kubernetes - START ---
FAILURE DETAILS:
kubernetes/container_logs test-nginx.log:
[0] field "data_stream.dataset" should have value in ["{{kubernetes.labels.elastic_co/dataset}}" "{{data_stream.dataset}}" "kubernetes.container_logs"], it has "kubernetes.container_logs.nginx"


╭────────────┬────────────────┬───────────┬────────────────┬─────────────────────────────────────────────────────────────────────────────┬──────────────╮
│ PACKAGE    │ DATA STREAM    │ TEST TYPE │ TEST NAME      │ RESULT                                                                      │ TIME ELAPSED │
├────────────┼────────────────┼───────────┼────────────────┼─────────────────────────────────────────────────────────────────────────────┼──────────────┤
│ kubernetes │ container_logs │ pipeline  │ test-nginx.log │ FAIL: test case failed: one or more problems with fields found in documents │      2.617ms │
╰────────────┴────────────────┴───────────┴────────────────┴─────────────────────────────────────────────────────────────────────────────┴──────────────╯
--- Test results for package: kubernetes - END   ---
Done
Error: one or more test cases failed
```

It seems elastic-package is interpreting the `target_dataset` values as literals and not expressions, so it's not evaluating the `"{{kubernetes.labels.elastic_co/dataset}}"` expression in the into its actual value `kubernetes.container_logs.nginx`.

zmoog:
### Motivation

Datasets values in the `routing_rules.yml` file can contain mustache expressions like:

    {{kubernetes.labels.elastic_co/dataset}}

However, the current test implementation in `elastic-package` use these values as static strings, so tests fail on integrations using datasets with mustache expressions like the Kubernetes integration.

### Change description

This PR add the rendering mustache expressions in datasets values using the github.com/cbroglie/mustache library.

### Additional Notes

Closes #1388 
Unblocks elastic/integrations#7118

### Reviewer checklist

* [x] PR address a single concern.
* [x] PR title and description are appropriately filled.
* [x] Changes will be merged in `main.`
* [x] Changes are covered by tests.
* [x] Logging is meaningful in case of troubleshooting.
* [ ] ~~Docs are updated (at least the `README.md`, if needed).~~
* [x] History is clean, commit messages are meaningful (see `CONTRIBUTING.md`) and are well-formatted.


internal/fields/validate.go
```diff
@@ -320,8 +322,15 @@ func (v *Validator) validateDocumentValues(body common.MapStr) multierror.Error
 				continue
 			}
 
+			var actualExpectedDatasets []string
+			for _, dataset := range v.expectedDatasets {
+				renderedDataset := mustache.Render(dataset, body)
```
jsoriano, line 327: Could you add a test case that would leverage this rendering? A good place may be here https://github.com/elastic/elastic-package/blob/main/test/packages/parallel/awsfirehose/data_stream/log/routing_rules.yml

cc @kaiyan-sheng

zmoog: Yep, definitely!

zmoog: @jsoriano I added an integration test to check the expression rendering, but the test is [failing](https://buildkite.com/elastic/elastic-package/builds/1289#0189fedd-3a84-43a8-8258-d6fa49c837cc). What's the best way to run this locally?

zmoog: The closest match I could find is the following:

```shell
PACKAGE_UNDER_TEST=kubernetes PACKAGE_TEST_TYPE=parallel ./scripts/test-check-packages.sh 
```

However, the previous command runs fine on my machine. 

There's something else, I'm looking into this.

zmoog: Nevermind! 

I didn't run `elastic-package check` in the kubernetes test package. My bad! 🤦

kaiyan-sheng:
@zmoog Thanks for making the fix 😬 @jsoriano Do you mind having a separate test package `kubernetes` here or you prefer we add this test case into the existing `aws` test package?

zmoog:
> Thanks for making the fix 😬 @jsoriano Do you mind having a separate test package `kubernetes` here or you prefer we add this test case into the existing `aws` test package?

I was also wondering about adding a Kubernetes test vs. using the existing Firehose one. 

I added a Kubernetes one mainly because I had clear real-world examples for this scenario and less on Firehose. But I am willing to consolidate in one if you think it's a better option.

jsoriano:
There is already a kubernetes test package, under [`test/packages/with-kind`](https://github.com/elastic/elastic-package/tree/d369b9ef4137e5dbfe2d8183f226a663ec7d312a/test/packages/with-kind/kubernetes).

I am fine with adding this case to the firehose or the existing kubernetes package, but please don't add another kubernetes test package.

zmoog:
> There is already a kubernetes test package, under [`test/packages/with-kind`](https://github.com/elastic/elastic-package/tree/d369b9ef4137e5dbfe2d8183f226a663ec7d312a/test/packages/with-kind/kubernetes).

Oh, thanks! I missed this one, my bed.

> I am fine with adding this case to the firehose or the existing kubernetes package, but please don't add another kubernetes test package.

Got it! I'm moving my test case into one of the existing packages.

zmoog:
@jsoriano, I added the simple case possible to the firehose integration.

kaiyan-sheng (approved):
LGTM! The test case for Cloudtrail is not real but enough for testing.

zmoog:
> LGTM! The test case for Cloudtrail is not real but enough for testing.

Yeah, it's functional for testing, but it could carry better semantics. Let me see if I can find a better test case.

zmoog:
Pushed a more realistic test case, but still not great. 

The main problem is the firehose event does not contain a lot of field values I can use for a dataset or namespace, due to [data stream fields name restrictions](https://www.elastic.co/guide/en/ecs/master/ecs-data_stream.html).

jsoriano:
test integrations

test/packages/parallel/awsfirehose/data_stream/log/routing_rules.yml
```diff
@@ -1,6 +1,11 @@
 - source_dataset: awsfirehose.log
   rules:
+    - target_dataset:
+        - "{{cloud.account.id}}"
```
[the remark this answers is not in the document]
zmoog: In general, expressions in the routing rules enable the rule to pick a dataset or namespace value for a field in the log event.

For example, the Kubernetes integration picks the dataset value from the `{{kubernetes.labels.elastic_co/dataset}}` field.

Unfortunately, the Firehose event does not offer a lot of fields with values compatible with the data stream naming [restrictions](https://www.elastic.co/guide/en/ecs/master/ecs-data_stream.html).

So, I imagined a scenario where users want to partition the log events into multiple AWS account-based data streams using `{{cloud.account.id}}`. But it's a bit confusing.

The other option is to add an extra field in the log event, for example, `{{labels.elastic_co/dataset}}`, and use this field in the rule. It does not exist in the Firehose realm but is less confusing overall. WDYT?

PR #1374, opened 2023-07-28
Add serverless provider

mrodm:
This PR adds a new provider to create Serverless projects.

It requires to set these two environment variables :
- EC_API_KEY
- EC_HOST

Notes:
- In pipeline tests, Geo IP field keys have been skipped
    - It cannot be set a specific GeoIP database to make the test stable.
- A local agent is started locally (using docker-compose). This agent is configured and connected to the new project.
    - As done locally, if elastic-agent starts to fail, it is retried once

Tested with the following packages from [integrations repository](https://github.com/elastic/integrations):
- elastic_package_registry
- nginx
- apache
- haproxy
- system
- mysql
- squid
- gcp
- tomcat
- zookeeper
- winlog
- zoom
- sql_input
- redis
- aws (just asset and pipeline tests)

## How to test
```shell
export EC_API_KEY="xx"
# if not defined EC_HOST, it will be used the default value
export EC_HOST="xx"

elastic-package profiles create serverless --from default
elastic-package profiles use serverless

# check that this new profile is selected
elastic-package profiles list

# Boot up a new project and check its status
elastic-package stack up -v -d --version 8.9.0 --provider serverless
elastic-package stack status

# from the package folder install or test
elastic-package install
elastic-package test -v

# Delete the project
elastic-package install down -v
```

PR #1230, opened 2023-04-21
Add stack provider for Elastic Cloud

jsoriano:
Last piece of https://github.com/elastic/elastic-package/pull/1055.

Testing pending.

PR #1411, opened 2023-08-28
Add flag for user-defined parameters in stack up

jsoriano:
Allow to define additional parameters for `elastic-package stack up` from the command line.

Stack providers can access these settings using `profile.Config()`, they have to provide their own handling and persistence if required.

If the user wants to persist the setting, they need to write it into the `config.yml` file in the profile.

For example this could be used by the serverless stack provider (https://github.com/elastic/elastic-package/pull/1374) to select the project type:
```
elastic-package stack up --provider serverless -U stack.serverless.project_type=observability
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,6 +461,8 @@ To expose local packages in the Package Registry, build them first and boot up t
 
 For details on how to connect the service with the Elastic stack, see the [service command](https://github.com/elastic/elastic-package/blob/main/README.md#elastic-package-service).
 
+You can customize your stack using profile settings, see [Elastic Package profiles](https://github.com/elastic/elastic-package/blob/main/README.md#elastic-package-profiles-1) section. These settings can be also overriden with the --parameter flag. Settings configured this way are not persisted.
+
 ### `elastic-package stack update`
 
 _Context: global_
--- a/cmd/stack.go
+++ b/cmd/stack.go
@@ -86,14 +86,20 @@ func setupStackCommand() *cobraext.Command {
 				return err
 			}
 
+			userParameters, err := cobraext.GetStackUserParameterFlags(cmd)
+			if err != nil {
+				return err
+			}
+
 			cmd.Printf("Using profile %s.\n", profile.ProfilePath)
 			cmd.Println(`Remember to load stack environment variables using 'eval "$(elastic-package stack shellinit)"'.`)
 			err = provider.BootUp(stack.Options{
-				DaemonMode:   daemonMode,
-				StackVersion: stackVersion,
-				Services:     services,
-				Profile:      profile,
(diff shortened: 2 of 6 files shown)


jsoriano:
Moving back to draft. Wondering if this should be directly stored in the profile configuration, and providers would read it from there, as done in https://github.com/elastic/elastic-package/pull/1230/files#diff-1ddc0b543f2deb096c99ee164380726eb20a9728c1ebebaaa8ae1088a9ec5854R147.

mrodm:
> Moving back to draft. Wondering if this should be directly stored in the profile configuration, and providers would read it from there, as done in https://github.com/elastic/elastic-package/pull/1230/files#diff-1ddc0b543f2deb096c99ee164380726eb20a9728c1ebebaaa8ae1088a9ec5854R147.

In this other PR, I followed the same approach to get some parameters (e.g. project type) from the profile configuration:
https://github.com/elastic/elastic-package/pull/1374/files#diff-b4efb1ac558d15d3077525cb9be59dce12ee723a4749253e9ca7cc860419f1f8R193-R194

If it is done so, would it mean that a user parameter introduced in the command line is going to be re-used in the following executions? if the user does not overwrite it again
```shell
elastic-package stack up --provider serverless -U serverless.project_type=security
elastic-package stack down

# this execution would also keep the same value for the parameter `serverless.project_type` from the previous run (security)
elastic-package stack up --provider serverless
```

It looks like that would match the way of working of `provider` parameter. Once set in the CLI, it would be kept (stored) until it is overwritten.

jsoriano:
The thing is that as it is implemented now:
* `<profile>/config.yml` is intended for user configuration, this is read with `profile.Config(...)` I don't think `elastic-package` should modify this file, as is intended for users.
* `<profile>/stack/config.json` contains configuration about the started stack. This file should be written only by `elastic-package stack up`, and read by any command that needs to access or do anything with the stack.

So if we add settings with CLI flags, where do we persist them?
* In `<profile>/config.yml` we would be modifying a file intended for users. If we do this here we would have to try to keep the format, maybe adding the flag overrides at the end. It would persist between executions of stack up.
* In `<profile>/stack/config.json` could make more sense, but the providers would need to take care of reading it, and persisting it when needed. It wouldn't persist between executions of stack up.
* In some other place.

I am more leaned to the second option, but this may complicate a bit configuration management in stack providers. Maybe we can provide helpers for that.

Another option may be to offer a subcommand to manage profile config, instead of adding a flag, and let it manage the profile `config.yml`. This would be similar to how `git config` command manages git configuration, but we would also need to try to keep format.

mrodm (approved):
:+1:

PR #1415, opened 2023-08-29
Review positional arguments in commands

mrodm:
This PR reviews usage of positional arguments in all commands and sub-commands.
Currently, if a positional argument is set in the command line, this parameter is ignored. I think that could be misleading.

This PR enforces to not use positional arguments if they are not needed.
As an example:

- Use of a positional argument where it is not used by the command:
```shell
 $ elastic-package clean aa
Clean used resources
Build resources removed: /home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry
Temporary service logs removed: /home/mariorodriguez/.elastic-package/tmp/service_logs
Done
``` 
- With this PR, that would raise an error:
```shell
 $ elastic-package clean aa
Error: unknown command "aa" for "elastic-package clean"
```

In case of `elastic-package test` and `elastic-package benchmark` commands the same positional arguments has been kept. Keeping that would show a more specific error for those commands.

Example:
- Current error shown:
```shell
 $ elastic-package test aaa
Run test suite for the package
Error: unsupported test type: aaa
```
- If it is used `cobra.NoArgs`:
```shell
 $ elastic-package test aaa
Error: unknown command "aaa" for "elastic-package test"
```

### DIFF
--- a/cmd/benchmark.go
+++ b/cmd/benchmark.go
@@ -80,6 +80,7 @@ func getPipelineCommand() *cobra.Command {
 		Use:   "pipeline",
 		Short: "Run pipeline benchmarks",
 		Long:  "Run pipeline benchmarks for the package",
+		Args:  cobra.NoArgs,
 		RunE:  pipelineCommandAction,
 	}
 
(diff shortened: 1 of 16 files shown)


jsoriano (approved):
Good one! :rocket:

PR #1417, opened 2023-08-29
Avoid requesting latest github release in every command

### LINKED ISSUE #871: elastic-package exhausts Github API rate-limit
It appears that elastic-package makes HTTP requests to Github [every time](https://github.com/elastic/elastic-package/blob/e21049424392c77427bb36d3b06b536c72050cb7/cmd/root.go#L48) it executes. This quickly exhausts the Github API request limit for my IP if I run 60 commands in an hour. Perhaps this check cache the result to `~/.elastic-package` and check for new versions less frequently. Also an env var to disable the check entirely would be welcomed (e.g. `ELASTIC_PACKAGE_UPDATE_CHECK_DISABLED=1`)

`
2022/06/27 19:25:36 DEBUG Error: cant't check latest release, GET https://api.github.com/repos/elastic/elastic-package/releases/latest: 403 API rate limit exceeded for 333.333.333.333. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) [rate reset in 19m47s]
`

From https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-personal-accounts:

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the person making requests.

mrodm:
Fixes #871 

This PR avoids doing for every command execution the latest release published to Github.

It introduces two mechanisms for that:
- defining `ELASTIC_PACKAGE_CHECK_UPDATE_DISABLED` environment variable with value "true" or "1"
    - adding this env. var avoids completely this check.
- adds a cache file where this request is stored. This data would be requested again just after the expiration time
    - In this cache file, it is stored the timestamp where the request was done
    - By default, the expiration time is 30 minutes

A new step has been added to CI to ensure this is working.


internal/version/check_update.go
```diff
@@ -6,18 +6,37 @@ package version
...
 	"github.com/elastic/elastic-package/internal/github"
 	"github.com/elastic/elastic-package/internal/logger"
 )
 
 const (
 	repositoryOwner = "elastic"
 	repositoryName  = "elastic-package"
+
+	latestVersionFile    = "latestVersion"
+	defaultCacheDuration = "30m"
```
jsoriano, line 29: Define it directly as duration?
```suggestion
	defaultCacheDuration = 30 * time.Minute
```

internal/version/check_update.go
```diff
@@ -26,31 +45,102 @@ func CheckUpdate() {
...
 		logger.Debugf("Error: can't parse current version tag, %v", err)
 		return
 	}
 
 	if currentVersion.LessThan(releaseVersion) {
-		logger.Infof("New version is available - %s. Download from: %s", *release.TagName, *release.HTMLURL)
+		logger.Infof("New version is available - %s. Download from: %s", release.TagName, release.HtmlURL)
+	}
+
+	if err := writeLatestReleaseToCache(release); err != nil {
```
jsoriano, line 96: The file is always written even if the content is not expired?

internal/version/check_update.go
```diff
@@ -26,31 +45,102 @@ func CheckUpdate() {
...
+	}
+	err = os.WriteFile(latestVersionPath, contents, 0644)
+	if err != nil {
+		return fmt.Errorf("writing file failed (path: %s): %w", latestVersionPath, err)
+	}
+
+	return nil
+}
+
+func checkCachedLatestVersion(expiration time.Duration) (*versionLatest, bool, error) {
```
jsoriano, line 121: Nit. Split this method in one to load the latest version, and another one to check it?
```suggestion
func loadCachedLatestVersion() (*versionLatest, error) {
    ...
}

func checkLatestVersion(latest *versionLatest, expiration time.Duration) bool {
    return latest.Timestamp.Before(expiration)
}
```

internal/version/check_update.go
mrodm: That's true, I've checked that scenario and it will return earlier, thanks!

jsoriano (approved):

Issue #653, opened 2022-01-19
Make sure that elastic-package is not using deprecated Fleet API


mtojek:
This issue is an outcome of https://github.com/elastic/kibana/issues/121485, which removes some API in 9.0. We need to make sure that we don't use them.

So far I can see that we use `/epm/packages/{pkgkey}`, which is planned for removal.

PR #1420, opened 2023-08-29
Stop using deprecated Fleet APIs

jsoriano:
Stop using Fleet APIs deprecated since 8.0 (https://www.elastic.co/guide/en/kibana/master/release-notes-8.0.0-rc1.html#deprecations-8.0.0-rc1).

The only one that seems we are using is:
>  Use `/epm/packages/{packageName}/{version}` instead of `/epm/packages/{pkgkey}`.

Kibana client retrieves its version from the status API on initialization.

Fixes https://github.com/elastic/elastic-package/issues/653


juliaElastic (approved):
LGTM

internal/kibana/test/httptest.go
```diff
@@ -54,11 +57,11 @@ func pathForURL(url string) string {
 }
 
 func recordRequest(t *testing.T, r *http.Request, path string) {
-	client, err := NewClient()
+	client, err := stack.NewKibanaClient()
```
mrodm, line 60: Good catch!

internal/kibana/test/httptest.go
```diff
@@ -54,11 +57,11 @@ func pathForURL(url string) string {
...
 	require.NoError(t, err)
 
 	t.Logf("Recording %s in %s", r.URL.RequestURI(), path)
-	status, respBody, err := client.get(r.URL.RequestURI())
+	status, respBody, err := client.SendRequest(http.MethodGet, r.URL.RequestURI(), nil)
```
mrodm, line 64: Why not doing public get, post, put and delete methods?

mrodm (approved):
:+1:

internal/kibana/test/httptest.go
jsoriano: I don't have a strong preference, I found better to expose the most generic one. Would you prefer to expose the ones per method?

PR #1426, opened 2023-08-31
Set archetype format version to the latest available spec version

### LINKED ISSUE #1155: Automate format version update in package archetype
`elastic-package create package` creates a new package based on an embedded archetype. This archetype has a hard-coded `format_version` that we have to remember to manually update now.

Create some automation so this format version gets updated as new versions of the package spec are released.

Some options are discussed in https://github.com/elastic/elastic-package/pull/1152#issuecomment-1438684198.

jsoriano:
Fill the spec version when creating packages with `elastic-package create package` using
the latest available version in the Package Spec.

Review related tests to don't depend on the current directory.

Closes #1155


mrodm (approved):
Nice! No more PRs to update the spec version here :muscle:

PR #1428, opened 2023-08-31
Load stack configuration from profile

jsoriano:
Commands that require an stack to connect will try to connect with the
stack of the current profile by default. This behaviour can be overridden
by setting the usual environment variables provided by `elastic-package shellinit`.

With this change `elastic-package shellinit` is not necessary when working
with stacks managed by `elastic-package`. `shellinit` is still useful to export
the configuration for other commands or scripts.
The environment variables can also be used to work with stacks not managed
by `elastic-package`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,7 +437,20 @@ Dump stack data for debug purposes.
 
 _Context: global_
 
-Export environment variables.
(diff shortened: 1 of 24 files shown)


mrodm (approved):
:rocket:

PR #1436, opened 2023-09-04
Fix parsing of install response

jsoriano:
New and old APIs for packages include the assets in different fields. Old API uses `response`, while new API uses `items`.

### DIFF
--- a/internal/kibana/packages.go
+++ b/internal/kibana/packages.go
@@ -76,12 +76,20 @@ func processResults(action string, statusCode int, respBody []byte) ([]packages.
 	}
 
 	var resp struct {
-		Assets []packages.Asset `json:"response"`
+		// Assets are here when old API is used.
+		Response []packages.Asset `json:"response"`
+
+		// Assets are here when new API is used.
+		Items []packages.Asset `json:"items"`
 	}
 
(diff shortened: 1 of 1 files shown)


jsoriano:
test integrations

internal/kibana/packages.go
```diff
@@ -76,12 +76,20 @@ func processResults(action string, statusCode int, respBody []byte) ([]packages.
 	}
 
 	var resp struct {
-		Assets []packages.Asset `json:"response"`
+		// Assets are here when old API is used.
+		Response []packages.Asset `json:"response"`
+
+		// Assets are here when new API is used.
+		Items []packages.Asset `json:"items"`
```
mrodm, line 83: Just a small suggestion. But maybe it's not needed because it's already in the `epmPackageURL`. 

```suggestion
		// Assets are here when old API (kibana < 8.0) is used.
		Response []packages.Asset `json:"response"`

		// Assets are here when new API (kibana >= 8.0) is used.
		Items []packages.Asset `json:"items"`
```

mrodm (approved):
:+1:

internal/kibana/packages.go
jsoriano: :+1:

PR #1437, opened 2023-09-08
Fix tests around pipelines with custom dataset

ruflin:
The dataset value in the manifest.yml file can be overwritten to allow dataset names which are not `{package}.{dirName}`. One example of this can be found here: https://github.com/elastic/integrations/pull/7670 This is already in use today and is used by some packages. Fleet installs all the assets as expected.

When working on https://github.com/elastic/integrations/pull/7670, CI is failing. The reason is that the pipeline check did not take the custom dataset config into account. The reason why this didn't show up earlier is because the check for the template already has a similar condition in place. Likely all previous packages that used this feature did only have templates and no pipelines.

### DIFF
--- a/internal/packages/assets.go
+++ b/internal/packages/assets.go
@@ -135,9 +135,14 @@ func loadElasticsearchAssets(pkgRootPath string) ([]Asset, error) {
 				continue // ingest pipeline is not defined
 			}
 
+			// If no dataset value is set in the manifest, it falls back to {packageName}.{dirName}
+			if dsManifest.Dataset == "" {
+				dsManifest.Dataset = fmt.Sprintf("%s.%s", pkgManifest.Name, dsManifest.Name)
+			}
+
 			ingestPipelineName := dsManifest.GetPipelineNameOrDefault()
 			if ingestPipelineName == defaultPipelineName {
-				ingestPipelineName = fmt.Sprintf("%s-%s.%s-%s", dsManifest.Type, pkgManifest.Name, dsManifest.Name, pkgManifest.Version)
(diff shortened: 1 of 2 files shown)


tommyers-elastic (approved):
LGTM

ruflin:
I'm wondering if the failure that CI hit is in any way related?

```
Error: error running package system tests: could not complete test run: could not find hits in metrics-aws.ec2_metrics-ep data stream
```

The following can be seen before in the logs which makes me suspicious:

```
WARN[0000] The "AWS_SESSION_TOKEN" variable is not set. Defaulting to a blank string.
WARN[0000] The "AWS_PROFILE" variable is not set. Defaulting to a blank string.
```

What does elastic-package exactly test around aws.ec2 metrics?

ruflin:
I did rerun the job and seems it was just flaky.

ruflin:
This PR currently blocks https://github.com/elastic/integrations/pull/7670 @jsoriano As soon as this PR is in, will CI in https://github.com/elastic/integrations/pull/7670 pick up the new version or will we need to first release a new version?

mrodm:
> This PR currently blocks [elastic/integrations#7670](https://github.com/elastic/integrations/pull/7670) @jsoriano As soon as this PR is in, will CI in [elastic/integrations#7670](https://github.com/elastic/integrations/pull/7670) pick up the new version or will we need to first release a new version?

It's needed a new elastic-package release in order to be used by integrations. Once create the release dependabot creates the required PR.

mrodm (commented):
LGTM

Could you update the test package `awsfirehouse` to use that custom dataset as in the integrations PR ? In that way, this change could also be tested here in elastic-package @ruflin

PR #1440, opened 2023-09-11
Avoid dotted fields when creating new package

mrodm:
This PR avoids creating package manifest using dotted fields.
These fields would not be allowed in package-spec v3

Relates https://github.com/elastic/package-spec/issues/538


jsoriano (approved):
:+1:

Issue #1435, opened 2023-09-04
Disallow importing reusable ECS fields at the root level when not allowed


jsoriano:
Elastic Package is currently allowing to import reusable ECS fields at the root level when they are not allowed. This happens for example with `geo` fields, that shouldn't be at the root level, but nested under some specific groups, see https://www.elastic.co/guide/en/ecs/8.9/ecs-geo.html#_field_reuse_10

Conditionally remove the logic for this, so packages using format version v3 or greater cannot import these fields this way. It should still be supported for older packages to avoid breaking changes.
Packages that need to use these fields at the root level will need to copy the resolved definition.

PR #1441, opened 2023-09-12
Disallow importing ECS fields root level

jsoriano:
Elastic Package should not import reusable ECS fields that are
defined at the top level, but are not intended to be reusable there.
Fail in that case for packages starting on Package Spec v3.

~~Test package requires a release of Package Spec v3.~~ Package Spec v3 not needed for this test because field import fails before running validations.

Closes https://github.com/elastic/elastic-package/issues/1435.


scripts/test-check-false-positives.sh
```diff
@@ -19,54 +19,74 @@ cleanup() {
...
   (
-    cd $d
-    elastic-package check -v
+    cd $package_root
+    elastic-package build 2>&1 | tee $output_file || true # Ignore errors here
   )
-done
-cd -
+
+  diff -w -u $expected_build_output $output_file || (
```
mrodm, line 82: Probably this needs to be prefixed by $package_root. Since the $output_file is generated after changing to that directory in that `( )` block.

```suggestion
  diff -w -u $expected_build_output $package_root/$output_file || (
```

jsoriano: Yes, this is why I was using an absolute path. I have added an `mkdir -p` to ensure that the parent directory exists.

mrodm: Ok, dismiss this suggestion
just show this previous error:

> tee: /opt/buildkite-agent/builds/bk-agent-prod-gcp-1694523471797672347/elastic/elastic-package/build/elastic-package-output: No such file or directory

mrodm (approved):
Nice ! 
:rocket:

PR #1443, opened 2023-09-12
Use slices to check if a slice contains a string

jsoriano:


mrodm (approved):
Perfect! cleaning :broom: :smile:

Issue #1439, opened 2023-09-11
Possible regression in field values validation


ebeahan:
## Summary

In `elastic-package` v0.84.0+, `elastic-package test pipeline -v` no longer produces errors due if ECS field value fails validation. 

The same package and version (in the below example, `cisco_asa` version 2.21.0) tested using v0.83.2 does produce validation errors.

### v0.83.2
--------

```bash
$ go install github.com/elastic/elastic-package@v0.83.2

$ elastic-package version
2023/09/11 12:14:48  INFO New version is available - v0.86.1. Download from: https://github.com/elastic/elastic-package/releases/tag/v0.86.1
elastic-package v0.83.2 version-hash undefined (build time: unknown)

$ elastic-package test pipeline -v
2023/09/11 12:14:58  WARN CommitHash is undefined, in both /Users/ericbeahan/.elastic-package/version and the compiled binary, config may be out of date.
2023/09/11 12:14:58 DEBUG Enable verbose logging
2023/09/11 12:14:58  INFO New version is available - v0.86.1. Download from: https://github.com/elastic/elastic-package/releases/tag/v0.86.1
Run pipeline tests for the package
--- Test results for package: cisco_asa - START ---
FAILURE DETAILS:
cisco_asa/log test-additional-messages.log:
[0] parsing field value failed: field "event.type" value "change" is not one of the expected values (access, allowed, connection, denied, end, info, protocol, start) for any of the values of "event.category" (network)
[1] parsing field value failed: field "event.type" value "deletion" is not one of the expected values (access, allowed, connection, denied, end, info, protocol, start) for any of the values of "event.category" (network)
[2] parsing field value failed: field "event.type" value "error" is not one of the expected values (access, allowed, connection, denied, end, info, protocol, start) for any of the values of "event.category" (network)
cisco_asa/log test-anyconnect-messages.log:
[0] parsing field value failed: field "event.type" value "error" is not one of the expected values (access, allowed, connection, denied, end, info, protocol, start) for any of the values of "event.category" (network)
cisco_asa/log test-sample.log:
[0] parsing field value failed: field "event.type" value "deletion" is not one of the expected values (access, allowed, connection, denied, end, info, protocol, start) for any of the values of "event.category" (network)


╭───────────┬─────────────┬───────────┬──────────────────────────────┬─────────────────────────────────────────────────────────────────────────────┬──────────────╮
[...]

PR #1452, opened 2023-09-13
Fix validations based on fields imported from ECS

jsoriano:
We inject fields from ECS on different scenarios: when generating documentation, when building packages, or when validating documents in tests.

On #1335 we did a refactor around this, to fix some validation issues, in a way that more code is reused between all these uses. After this change, validation used field definitions that include the already resolved external fields, but we weren't including there information that was used by validators, what included:
* Validation of expected values.
* Validation of allowed values.

So these validations haven't been executed since then.

Tests have been included to try to avoid regressions related to this in the future.

Fix https://github.com/elastic/elastic-package/issues/1439


internal/fields/dependency_manager.go
```diff
@@ -318,17 +328,28 @@ func transformImportedField(fd FieldDefinition) common.MapStr {
...
+		}
+
+		if len(fd.AllowedValues) > 0 {
+			m["allowed_values"] = fd.AllowedValues
+		}
+
+		if len(fd.ExpectedValues) > 0 {
+			m["expected_values"] = fd.ExpectedValues
+		}
+	}
```
mrodm, line 352: This means that these fields (normalize, allowed_values, expected_values) won't be added to each field definition when the package are built, is that right?

internal/fields/validate.go
```diff
@@ -163,6 +163,8 @@ func CreateValidatorForDirectory(fieldsParentDir string, opts ...ValidatorOption
 
 func createValidatorForDirectoryAndPackageRoot(fieldsParentDir string, finder packageRootFinder, opts ...ValidatorOption) (v *Validator, err error) {
 	v = new(Validator)
+	// In validator, inject fields with settings used for validation, such as `allowed_values`.
+	v.injectFieldsOptions.IncludeValidationSettings = true
```
mrodm, line 167: If it is set to true here, every run of validation would use those fields, even in built packages ?

jsoriano:
test integrations

internal/fields/validate.go
mrodm: Talked offline, and this should be just used for validators. Built packages do not use this one.

internal/fields/dependency_manager.go
jsoriano: Yes, this is not used in Fleet or Elasticsearch.

mrodm (approved):
Thanks!!

Issue #320, opened 2021-04-13
Pipeline test file generation escapes special characters in golden files generation


P1llus:
When creating new expected.json files from raw log files, it seems that certain characters are HTML escaped, adding in unicode characters in the generated output.

I would like to know is this is the intended results, as it differs from how the output looked like in beat modules test file generation (which used python).

An example would be zeek logs, here characters like `<>` is turned into ` \u003c` and `\u003e`.

Which would turn this example raw event:

```
        "zeek.session_id": "CPRLCB4eWHdjP852Bk",
        "zeek.sip.call_id": "8694cd7e-976e4fc3-d76f6e38@172.16.133.19",
        "zeek.sip.request.body_length": 0,
        "zeek.sip.request.from": "\"AppNeta\" <sip:116954_Boston6@newyork.voip.ms>",
        "zeek.sip.request.path": [
            "SIP/2.0/UDP 172.16.133.19:5060"
        ],
        "zeek.sip.request.to": "<sip:116954_Boston6@newyork.voip.ms>",
        "zeek.sip.response.body_length": 0,
        "zeek.sip.response.from": "\"AppNeta\" <sip:116954_Boston6@newyork.voip.ms>",
        "zeek.sip.response.path": [
            "SIP/2.0/UDP 172.16.133.19:5060"
        ],
        "zeek.sip.response.to": "<sip:116954_Boston6@newyork.voip.ms>;tag=as023f66a5",
        "zeek.sip.sequence.method": "REGISTER",
        "zeek.sip.sequence.number": "4127",
        "zeek.sip.status.code": 401,
        "zeek.sip.status.msg": "Unauthorized",
        "zeek.sip.transaction_depth": 0,
        "zeek.sip.uri": "sip:newyork.voip.ms:5060",
        "zeek.sip.user_agent": "PolycomSoundStationIP-SSIP_5000-UA/3.2.4.0267"
```

To this:
```
"zeek": {
                "session_id": "CPRLCB4eWHdjP852Bk",
                "sip": {
                    "request": {
                        "path": [
                            "SIP/2.0/UDP 172.16.133.19:5060"
                        ],
                        "from": "\"AppNeta\" \u003csip:116954_Boston6@newyork.voip.ms\u003e",
                        "to": "\u003csip:116954_Boston6@newyork.voip.ms\u003e",
                        "body_length": 0
                    },
                    "sequence": {
                        "method": "REGISTER",
                        "number": "4127"
                    },
                    "response": {
[...]

PR #1464, opened 2023-09-21
Avoid encoding html entities in JSON documents

jsoriano:
Avoid encoding HTML entities in JSON documents generated by tests.

The main JSON "Marshaling" helpers also encode HTML entities to reduce
the risk of injections through JSON documents embedded in HTML.
In documents generated by package tests, we want to have the raw document,
as found in Elasticsearch.

This change changes the rendering behaviour for packages starting on 2.12.0,
so these entities are not encoded.

Fix https://github.com/elastic/elastic-package/issues/320.

### DIFF
--- a/internal/docs/readme_test.go
+++ b/internal/docs/readme_test.go
@@ -9,9 +9,10 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/elastic/elastic-package/internal/packages"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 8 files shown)


jsoriano:
test integrations

jsoriano:
test integrations

mrodm (approved):
:rocket: 

> This change changes the rendering behaviour for packages starting on 2.12.0,
so these entities are not encoded.

Ok, perfect!

PR #1487, opened 2023-10-04
Review buildkite scripts cleanup process

mrodm:


.buildkite/scripts/integration_tests.sh
```diff
@@ -87,12 +95,11 @@ upload_safe_logs() {
         return
     fi
 
-    local gsUtilLocation=$(google_cloud_auth_safe_logs)
+    google_cloud_auth_safe_logs
```
mrodm, line 98: Calling this function as `$(google_cloud_auth_safe_logs)` make that the variable that was exported in that function (GOOGLE_APPLICATION_CREDENTIALS) was not available for the rest of the function. It seems that it creates a subshell and that variable does not exist outside of it.

.buildkite/scripts/integration_tests.sh
```diff
@@ -67,14 +77,12 @@ if [[ "${TARGET}" == "" ]]; then
 fi
 
 google_cloud_auth_safe_logs() {
-    local gsUtilLocation=$(mktemp -d -p . -t ${TMP_FOLDER_TEMPLATE})
+    local gsUtilLocation=$(mktemp -d -p ${WORKSPACE} -t ${TMP_FOLDER_TEMPLATE})
```
mrodm, line 80: Ensure that the temporal folder is created under WORKSPACE

.buildkite/scripts/integration_tests.sh
```diff
@@ -5,10 +5,20 @@ WORKSPACE="$(pwd)"
...
+    fi
+
     echo "Deleting temporal files..."
     cd ${WORKSPACE}
     rm -rf "${TMP_FOLDER_TEMPLATE_BASE}.*"
     echo "Done."
+
+    exit $error_code
 }
 trap cleanup EXIT
```
mrodm, line 23: I didn't want to move this function to `tooling.sh` to avoid more refactors in this PR.

Each pipeline (script) has its own cleanup and the one from `release` pipeline is different.

.buildkite/scripts/integration_tests.sh
```diff
@@ -5,10 +5,20 @@ WORKSPACE="$(pwd)"
 TMP_FOLDER_TEMPLATE_BASE="tmp.elastic-package"
 
 cleanup() {
+    local error_code=$?
+
+    if [ $error_code != 0 ] ; then
+        if [ -f ${GOOGLE_APPLICATION_CREDENTIALS+x} ]; then
+             google_cloud_logout_active_account
+        fi
+    fi
```
mrodm, line 14: Ensure that in case of failure, the active account is logged out.

.buildkite/scripts/integration_tests.sh
mrodm: I tried to take into account this by using this [parameter expansion](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) `${var+x}` 

```shell
 if [ -n "${GOOGLE_APPLICATION_CREDENTIALS+x}" ]; then
```

In that case, if the variable is not defined the logout is not executed.

I think that it could be a good approach, since that variable is set by the `google_cloud_auth` just after the login command

https://github.com/elastic/elastic-package/blob/f1b90c7227f090da3c0a4dfb6c6711d67084d90a/.buildkite/scripts/tooling.sh#L23-L25

In [this commit](https://github.com/elastic/elastic-package/pull/1487/commits/34c03753e623f17c35fc5011c11293e3e36b17c6) I added two conditions that the variable is defined and the file pointing that variable exists. Maybe it can be just leave that the variable is defined. If the variable is defined, run the logout function

@sharbuz WDYT ?

.buildkite/scripts/integration_tests.sh
sharbuz: perfect! we check the variable and if it exists we check the file!

.buildkite/hooks/pre-exit
```diff
@@ -0,0 +1,21 @@
...
+    unset GOOGLE_CREDENTIALS
+    unset GCP_PROJECT_ID
+fi
+
+
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-package" && ("$BUILDKITE_STEP_KEY" == "integration-parallel-aws" || "$BUILDKITE_STEP_KEY" == "integration-parallel-aws_logs") ]]; then
+    unset ELASTIC_PACKAGE_AWS_ACCESS_KEY
+    unset ELASTIC_PACKAGE_AWS_SECRET_KEY
+    unset AWS_ACCESS_KEY_ID
+    unset AWS_SECRET_ACCESS_KEY
```
sharbuz, line 20: Can we update the `unset_secrets` func to unset `_KEY` and `_KEY_ID` secrets to exclude whole this condition in the `pre-exit` hook?

.buildkite/scripts/signAndPublishPackage.sh
```diff
@@ -8,10 +8,21 @@ source .buildkite/scripts/install_deps.sh
...
 cleanup() {
+    local error_code=$?
+
+    if [ $error_code != 0 ] ; then
+        # if variable is defined, run the logout
+        if [ -n "${GOOGLE_APPLICATION_CREDENTIALS+x}" ]; then
+             google_cloud_logout_active_account
+        fi
+    fi
+
```
sharbuz, line 19: Perhaps it's better to move the `cleanup` func into `tooling.sh` script to avoid duplicating the code in the `signAndPublishPackage.sh` and `integration_tests.sh`. @mrodm WDYT?

mrodm: I wanted to avoid that in this PR at least, I added a comment about this in this other thread:
https://github.com/elastic/elastic-package/pull/1487#discussion_r1347228719

.buildkite/hooks/pre-exit
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+source .buildkite/scripts/tooling.sh
+
+set -euo pipefail
+
+unset_secrets
+
+# unset the environment variables not managged by "unset_secrets"
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "elastic-package" && "$BUILDKITE_STEP_KEY" == "integration-parallel-gcp" ]]; then
```
jsoriano, line 10: There would be any problem with unsetting all these variables always?
```suggestion
```

jsoriano (approved):

.buildkite/hooks/pre-exit
mrodm: mmm yes , and it probably it would be safer.
Executing those unset commands even if the variable does not exist should not cause any issue.
I'll change it.

Issue #1488, opened 2023-10-05
Field validation failing for array of objects


andrewkroh:
I'm trying to address validation errors for a field that was declared as `type: array`. This array contains a list of objects. The fields within these objects are fully specified in the mapping. I thought I should be able to remove the field entry where `type: array` is used ([like this](https://github.com/andrewkroh/integrations/commit/8053d45d51ddd8a44e9db618169cc6af85ba3b13)). Given how Elasticsearch flattens arrays of objects, the intermediate objects should have been implicit (e.g. for a field `foo.bar` we don't require a `foo` entry). But when I do this the pipeline tests begin failing with `field "powershell.command.invocation_details" is undefined`.

Another potential fix would be to explicitly declare the `invocation_details` as `type: object`, but this solution would run afoul with https://github.com/elastic/package-spec/pull/628. I could also make this `nested`, but I don't need that capability.

I think our goal with the validation is to ensure that we have mappings for all of the leaf fields. To meet that goal I think elastic-package needs to flatten arrays of objects [like Elasticsearch does](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested-arrays-flattening-objects) before applying the field validation.

What would you recommend to make this pipeline test field validation pass?

- Definition: https://github.com/elastic/integrations/blob/6719d691048bfd9803c98cd406beb51ae3653fac/packages/windows/data_stream/powershell/fields/fields.yml#L41-L63

- Sample Data: https://github.com/elastic/integrations/blob/75c6fada34f9c4f5b3b4fd81fceb574b8462dc54/packages/windows/data_stream/powershell_operational/_dev/test/pipeline/test-events.json-expected.json#L80-L82

-----

This is a simplified example of the issue.

```yaml
---
- name: foo.bar
  type: keyword
```

```json
{
  "foo": [
    {
      "bar": "baz"
    },
    {
      "bar": "baz"
    }
  ]
}
```

jsoriano:
> What would you recommend to make this pipeline test field validation pass?

I think that the only option for this case would be to define `invocation_details` as a `group` and define the individual fields inside.

Using arrays of objects this way is actually [tested](https://github.com/elastic/elastic-package/blob/29ded9d537b0fc2f9f62b66859cbb607726741fc/internal/fields/validate_test.go#L594), but not recommended, and its validation is not implemented.

I will try to slip in validation for these cases for Package Spec v3.

PR #1489, opened 2023-10-05
Validate arrays of objects

jsoriano:
Add validation for arrays of objects, considering them as valid when they are inside `nested` or `group`.

When `group` is used, a filterable error is given, mentioning that `nested` should be used instead on this case.

Fix https://github.com/elastic/elastic-package/issues/1488

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -36,7 +36,7 @@ var (
 
 	defaultExternal = "ecs"
 
-	arrayOfObjectsErr = errors.New("array of objects not used as nested type can lead to unexpected results")
+	errArrayOfObjects = errors.New("array of objects not used as nested type can lead to unexpected results")
 )
 
 // Validator is responsible for fields validation.
@@ -796,7 +796,7 @@ func (v *Validator) parseSingleElementValue(key string, definition FieldDefiniti
 			}
 			errs := v.validateMapElement(key, common.MapStr(val), doc)
 			if definition.Type == "group" {
-				errs = append(errs, arrayOfObjectsErr)
+				errs = append(errs, errArrayOfObjects)
 			}
 			if len(errs) == 0 {
 				return nil
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -22,12 +22,13 @@ import (
 	"github.com/cbroglie/mustache"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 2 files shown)


internal/fields/validate.go
```diff
@@ -27,13 +27,20 @@ import (
 	"github.com/elastic/elastic-package/internal/multierror"
 	"github.com/elastic/elastic-package/internal/packages"
 	"github.com/elastic/elastic-package/internal/packages/buildmanifest"
+	"github.com/elastic/package-spec/v2/code/go/pkg/specerrors"
 )
 
+// EPF - Elastic Package Fields [validation]
+const ArrayOfObjectsErrorCode = "EPF00001"
```
jsoriano, line 34: @mrodm do we have a place where we are already defining Elastic Package errors?

mrodm: Currently, there is no definition of errors in Elastic Package. Until now, it was just getting the errors from package-spec validate method and return the errors filtered (if any).

Should this be added into a new `internal/errors` package ? Or maybe another naming to not have collisions with standard errors package (e.g. validationerrors? I don't like neither too much this).

internal/fields/validate.go
```diff
@@ -783,11 +791,22 @@ func (v *Validator) parseSingleElementValue(key string, definition FieldDefiniti
...
 		case map[string]interface{}:
-			// TODO: This is probably an element from an array of objects,
+			// This is probably an element from an array of objects,
 			// even if not recommended, it should be validated.
+			if v.specVersion.LessThan(semver3_0_0) {
+				break
+			}
+			errs := v.validateMapElement(key, common.MapStr(val), doc)
+			if definition.Type == "group" {
+				errs = append(errs, arrayOfObjectsErr)
```
mrodm, line 804: I don't think these errors could be filtered as it is the code.

The methods that get the errors and filter are just for the errors from package-spec. This is in `internal/validation` package. For instance
https://github.com/elastic/elastic-package/blob/29ded9d537b0fc2f9f62b66859cbb607726741fc/internal/validation/validation.go#L26

internal/fields/validate.go
jsoriano: Removed by now.

internal/fields/validate.go
jsoriano: Removed by now.

mrodm (approved):

Issue #1474, opened 2023-09-27
Update system benchmark to use `totEvents` and drop deprecated `size` field from `package-spec`


aspacca:
In `package-spec`, `corpora.generator.size` field for the system benchmark scenario was deprecated in favor of `corpora.generator.totEvents` ([see source](https://github.com/aspacca/package-spec/blob/main/spec/integration/_dev/benchmark/system.scenario.spec.yml#L62-L71))

We should update the [code defining the yaml schema](https://github.com/elastic/elastic-package/blob/main/internal/benchrunner/runners/system/scenario.go#L52-L57) to use the new field and change in current artifacts the field from `size` to `totEvents` and set the relevant equivalent values

Contextually to this change we should update to [latest version of corpora generator](https://github.com/elastic/elastic-integration-corpus-generator-tool/releases/tag/v0.6.0)

PR #1491, opened 2023-10-06
Bump corpus generator v0.6.0

aspacca:
resolves manually conflicts at https://github.com/elastic/elastic-package/pull/1478
fixes https://github.com/elastic/elastic-package/issues/1474

@marc-gr do we use elsewhere the `.size/.tot_events` package specs?
I could not find any in integrations, but in case there are there, we should adapt a different strategy


aspacca:
@mrodm we still had `size` required in the `package-spec`: I've changed it in https://github.com/elastic/package-spec/pull/639

it would be better to include in v3.0.0 of the spec

since the only use of `size` is currently only in the `elastic-package` and `package-spec` itself if think it's safe to not have a code handling the migration, what do you think?

PR #1492, opened 2023-10-06
Fix import of object type from ECS

jsoriano:
Not including object type in object fields can lead to ambiguous interpretations
when installing the mappings. ECS fields of type object include this information,
import it for fields when available.

Object fields must have an object type in Package Spec v3, so this change is
required to import objects from ECS without workarounds.


juliaElastic (approved):
LGTM

PR #1493, opened 2023-10-06
Allow to override ECS objects to group or nested

jsoriano:
Package Spec v3 is more strict with the validation of objects. They
cannot be used now just to group fields, as is done in ECS.

When it is used only to group fields, the developers needs to decide
if it should be interpreted by Fleet as a `group`, so no mapping is installed
for the field itself, but for its subfields, or as a `nested`, so the object
is installed as `nested`, usually intended for arrays of objects.

Needed for https://github.com/elastic/integrations/pull/8115.


kpollich (approved):
LGTM 🚀

PR #1495, opened 2023-10-06
Revert requirement of using nested type for arrays of objects

jsoriano:
Revert change introduced in https://github.com/elastic/elastic-package/pull/1489 to require `nested` type for arrays of objects, as it is causing many conflicts with existing packages.


jsoriano:
Testing the change in https://github.com/elastic/integrations/pull/8115

jlind23 (approved):

PR #1498, opened 2023-10-09
Support validation of arrays of arrays of nested objects

jsoriano:
Handle https://github.com/elastic/integrations/pull/8115#issuecomment-1753082456, where nested objects are stored inside an array of arrays.


jsoriano:
Going back to draft by now in case this is not needed.

mrodm (approved):
LGTM

PR #1501, opened 2023-10-11
Accept null as array

jsoriano:
Accept `null` in test documents for values when an array of objects is expected.

### DIFF
--- a/internal/fields/validate_test.go
+++ b/internal/fields/validate_test.go
@@ -736,6 +736,21 @@ func Test_parseElementValue(t *testing.T) {
 				}
 			},
 		},
+		{
+			key:   "null_array",
+			value: nil,
+			definition: FieldDefinition{
+				Name: "null_array",
+				Type: "group",
+				Fields: []FieldDefinition{
+					{
+						Name: "id",
+						Type: "keyword",
+					},
+				},
+			},
+			specVersion: *semver3_0_1,
+		},
 	} {
 
 		t.Run(test.key, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/fields/validate_test.go
```diff
@@ -736,6 +736,21 @@ func Test_parseElementValue(t *testing.T) {
...
+			value: nil,
+			definition: FieldDefinition{
+				Name: "null_array",
+				Type: "group",
+				Fields: []FieldDefinition{
+					{
+						Name: "id",
+						Type: "keyword",
+					},
+				},
```
mrodm, line 750: Should this be an empty array ?

jsoriano: An empty array would work. The problem solved here is when a field defined as group or nested is `null` in a document, like `onepassword.details` here https://github.com/elastic/integrations/pull/8115/commits/09a01f29c32e6a872f03070d712b5c0e83b0d5d0#diff-76a51622c000706ccf925266e2e1255f954f7b55380135342a9d51aa3b81b5cdL55

mrodm: Ok, I misunderstood the `definition` field. I thought it was resulting definition built.

Just to be sure, it's being compared the value nil of a field `null_array` against that definition/schema in `definition`. Would that be right?

jsoriano: Yes, this would be like validating a document with this JSON content:
```
{
  "null_array": null
}
```

Validation fails without this change.

mrodm: Thanks!

jsoriano:
Failure in AWS test package seems unrelated, merging.

PR #1506, opened 2023-10-16
Apply automatic formatting only to manifests

jsoriano:
Modifies automation introduced in https://github.com/elastic/elastic-package/pull/1485 so it is only applied to manifests by now.

Applying it to other files like transforms or test configuration files may lead to obscure errors.

### DIFF
--- a/internal/formatter/yaml_formatter.go
+++ b/internal/formatter/yaml_formatter.go
@@ -9,18 +9,17 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/Masterminds/semver/v3"
 	"gopkg.in/yaml.v3"
 )
 
 // YAMLFormatter is responsible for formatting the given YAML input.
 type YAMLFormatter struct {
-	specVersion semver.Version
+	keysWithDotsAction int
 }
 
-func NewYAMLFormatter(specVersion semver.Version) *YAMLFormatter {
+func NewYAMLFormatter(keysWithDotsAction int) *YAMLFormatter {
 	return &YAMLFormatter{
-		specVersion: specVersion,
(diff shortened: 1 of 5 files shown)


internal/formatter/yaml_formatter.go
```diff
@@ -126,3 +134,24 @@ func mergeNodes(node *yaml.Node) {
 
 	node.Content = node.Content[:k]
 }
+
+func quoteKeysWithDot(node *yaml.Node) {
```
jsoriano, line 138: Started applying this to test files, but this could also have unexpected results.

This is not used now, if we finally decide not to use it I will remove it.

internal/formatter/yaml_formatter.go
```diff
@@ -55,6 +52,17 @@ func (f *YAMLFormatter) Format(content []byte) ([]byte, bool, error) {
 	return formatted, string(content) == string(formatted), nil
 }
 
+func applyActionOnKeysWithDots(node *yaml.Node, action int) {
+	switch action {
+	case KeysWithDotActionNested:
+		extendNestedObjects(node)
```
mrodm, line 58: Should it be checked that spec version is ^3.0.0 to apply this format ?

internal/formatter/yaml_formatter.go
mrodm: Not very confident to apply this transformation. Maybe just in transform.yml files ?

jsoriano: Yeah, after some tests I think this one can be a bit misleading, I am more leaned to remove it, and apply only the automation for nested objects in manifests.

jsoriano: Removed

PR #1518, opened 2023-10-19
Update var used in google_cloud_auth

mrodm:
This PR updates the variable used in `google_cloud_auth` to use the local variable


jsoriano (approved):
No idea what it was doing before, but change looks better :D

PR #1521, opened 2023-10-23
Skip by default compareResults in pipeline tests for serverless

mrodm:
This PR skips the compare results function for serverless by default. This function is in charge of comparing the documents received by elasticsearch with the samples. The issue is in Serverless, since there are fields related to Geo IP database that cannot be checked properly.

It also adds a new environment variable ELASTIC_PACKAGE_SERVERLESS_PIPELINE_TEST_ENABLE_COMPARE_RESULTS. This new environment variable can be defined to force to run this `compareResults` function.

When `elastic-package test pipeline` is executed with a local stack, there are no changes. The complete pipeline tests are run.

Examples to execute with this new environment variable:

```shell
 $ elastic-package stack up -v -d --version 8.10.4 --provider serverless

 $ ELASTIC_PACKAGE_SERVERLESS_PIPELINE_TEST_ENABLE_COMPARE_RESULTS=true elastic-package test pipeline -v

 $ ELASTIC_PACKAGE_SERVERLESS_PIPELINE_TEST_ENABLE_COMPARE_RESULTS=false elastic-package test pipeline -v
```

### DIFF
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -79,7 +79,7 @@ func (r *runner) Run(options testrunner.TestOptions) ([]testrunner.TestResult, e
 		r.runCompareResults = true
 
 		v, ok := os.LookupEnv(serverlessDisableCompareResults)
-		if ok && strings.ToLower(v) != "false" {
+		if ok && strings.ToLower(v) == "true" {
 			r.runCompareResults = false
 		}
 	}
--- a/internal/testrunner/runners/pipeline/runner.go
+++ b/internal/testrunner/runners/pipeline/runner.go
@@ -35,7 +35,7 @@ const (
 )
 
 var (
-	serverlessEnableCompareResults = environment.WithElasticPackagePrefix("SERVERLESS_PIPELINE_TEST_ENABLE_COMPARE_RESULTS")
+	serverlessDisableCompareResults = environment.WithElasticPackagePrefix("SERVERLESS_PIPELINE_TEST_DISABLE_COMPARE_RESULTS")
 )
 
 type runner struct {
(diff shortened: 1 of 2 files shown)


internal/testrunner/runners/pipeline/runner.go
```diff
@@ -33,9 +34,15 @@ const (
 	TestType testrunner.TestType = "pipeline"
 )
 
+var (
+	serverlessEnableCompareResults = environment.WithElasticPackagePrefix("SERVERLESS_PIPELINE_TEST_ENABLE_COMPARE_RESULTS")
```
jsoriano, line 38: Add somewhere a TODO comment mentioning that this is a temporary workaround till we have something better for deterministic geoip in serverless.

jsoriano (approved):
:+1:

internal/testrunner/runners/pipeline/runner.go
mrodm: Updated the TODO message here: https://github.com/elastic/elastic-package/pull/1521/commits/6b35523e5203354b5b6db1304031eece5bd21967

jsoriano (approved):
:+1:

internal/testrunner/runners/pipeline/runner.go
jsoriano, line 38: I am thinking now that maybe we should leave this enabled by default, so when running locally, by default we see the errors. And we can disable in the CI jobs where we want to use this. WDYT?
```suggestion
	serverlessEnableCompareResults = environment.WithElasticPackagePrefix("SERVERLESS_PIPELINE_TEST_DISABLE_COMPARE_RESULTS")
```

jsoriano (approved):
:+1:

Issue #1475, opened 2023-09-27
Add support for rally benchmark


aspacca:
Similarly to system benchmark we will add a command to run rally benchmark based on the [`package-spec` definition](https://github.com/elastic/package-spec/blob/main/spec/integration/_dev/benchmark/rally.scenario.spec.yml)

aspacca:
fixed in #1522

PR #1522, opened 2023-10-24
Rally benchmark

aspacca:
adding command to run rally benchmark

- some repetition of the code between `system` and `rally`: it is no 100% the same, not sure if trying to factorize it and move to common
- require esrally installed on the host, available from `$PATH`
- missing docs

```
$ pwd
elastic-package/test/packages/benchmarks/rally_benchmark
$ elastic-package benchmark rally --benchmark logs-benchmark
Run rally benchmarks for the package
--- Benchmark results for package: rally_benchmarks - START ---
╭──────────────────────────────────────────────────────────────────────────────────────────────────╮
│ info                                                                                             │
├────────────────────────┬─────────────────────────────────────────────────────────────────────────┤
│ benchmark              │                                                          logs-benchmark │
│ description            │                                         Benchmark 20000 events ingested │
│ run ID                 │                                    85c5f94b-d53a-4b76-b897-25a5e8df6f22 │
│ package                │                                                        rally_benchmarks │
│ start ts (s)           │                                                              1698117981 │
│ end ts (s)             │                                                              1698118001 │
│ duration               │                                                                     20s │
│ generated corpora file │                   ~/.elastic-package/tmp/rally_corpus/corpus-1222483366 │
╰────────────────────────┴─────────────────────────────────────────────────────────────────────────╯
╭────────────────────────────────────────────────────────────────────╮
│ parameters                                                         │
├─────────────────────────────────┬──────────────────────────────────┤
│ package version                 │                      999.999.999 │
│ input                           │                       filestream │
│ data_stream.name                │                           testds │
│ data_stream.vars.paths          │                     [dummy path] │
│ warmup time period              │                              10s │
│ corpora.generator.total_events  │                            20000 │
│ corpora.generator.template.path │ ./logs-benchmark/template.ndjson │
│ corpora.generator.template.raw  │                                  │
│ corpora.generator.template.type │                           gotext │
│ corpora.generator.config.path   │      ./logs-benchmark/config.yml │
│ corpora.generator.config.raw    │                            map[] │
│ corpora.generator.fields.path   │      ./logs-benchmark/fields.yml │
│ corpora.generator.fields.raw    │                            map[] │
╰─────────────────────────────────┴──────────────────────────────────╯
╭───────────────────────╮
│ cluster info          │
├───────┬───────────────┤
│ name  │ elasticsearch │
│ nodes │             1 │
╰───────┴───────────────╯
╭──────────────────────────────────────────────────────────────╮
│ data stream stats                                            │
├────────────────────────────┬─────────────────────────────────┤
│ data stream                │ logs-rally_benchmarks.testds-ep │
│ approx total docs ingested │                               0 │
│ backing indices            │                               1 │
│ store size bytes           │                             269 │
│ maximum ts (ms)            │                               0 │
╰────────────────────────────┴─────────────────────────────────╯
╭────────────────────────────────────╮
│ disk usage for index .ds-logs-rall │
│ y_benchmarks.testds-ep-2023.10.23- │
│ 000001 (for all fields)            │
├──────────────────────────────┬─────┤
│ total                        │ 0 B │
│ inverted_index.total         │ 0 B │
│ inverted_index.stored_fields │ 0 B │
│ inverted_index.doc_values    │ 0 B │
│ inverted_index.points        │ 0 B │
│ inverted_index.norms         │ 0 B │
│ inverted_index.term_vectors  │ 0 B │
│ inverted_index.knn_vectors   │ 0 B │
╰──────────────────────────────┴─────╯
╭─────────────────────────────────────────────────────────────────────────────────────────╮
│ pipeline logs-rally_benchmarks.testds-999.999.999 stats in node 0VW3SV6bRRue_LPWFWiOtQ  │
├────────────────────────────────────────────────┬────────────────────────────────────────┤
│ Totals                                         │ Count: 20000 | Failed: 0 | Time: 498ms │
│ grok ()                                        │ Count: 20000 | Failed: 0 | Time: 455ms │
│ user_agent ()                                  │  Count: 20000 | Failed: 0 | Time: 24ms │
│ pipeline (logs-rally_benchmarks.testds@custom) │   Count: 20000 | Failed: 0 | Time: 3ms │
╰────────────────────────────────────────────────┴────────────────────────────────────────╯
╭────────────────────────────────────────────────────────────────────────────────────────────╮
│ rally stats                                                                                │
├────────────────────────────────────────────────────────────────┬───────────────────────────┤
│ Cumulative indexing time of primary shards                     │     2.374416666666667 min │
│ Min cumulative indexing time across primary shards             │                     0 min │
│ Median cumulative indexing time across primary shards          │   0.02196666666666667 min │
│ Max cumulative indexing time across primary shards             │   0.35408333333333336 min │
│ Cumulative indexing throttle time of primary shards            │                     0 min │
│ Min cumulative indexing throttle time across primary shards    │                     0 min │
│ Median cumulative indexing throttle time across primary shards │                     0 min │
│ Max cumulative indexing throttle time across primary shards    │                     0 min │
│ Cumulative merge time of primary shards                        │    0.3455833333333333 min │
│ Cumulative merge count of primary shards                       │                       577 │
│ Min cumulative merge time across primary shards                │                     0 min │
│ Median cumulative merge time across primary shards             │ 0.0030833333333333333 min │
│ Max cumulative merge time across primary shards                │   0.06146666666666667 min │
│ Cumulative merge throttle time of primary shards               │                     0 min │
│ Min cumulative merge throttle time across primary shards       │                     0 min │
│ Median cumulative merge throttle time across primary shards    │                     0 min │
│ Max cumulative merge throttle time across primary shards       │                     0 min │
│ Cumulative refresh time of primary shards                      │    0.2704333333333333 min │
│ Cumulative refresh count of primary shards                     │                     18634 │
│ Min cumulative refresh time across primary shards              │                     0 min │
│ Median cumulative refresh time across primary shards           │  0.005966666666666666 min │
│ Max cumulative refresh time across primary shards              │  0.018783333333333332 min │
│ Cumulative flush time of primary shards                        │    11.530216666666666 min │
│ Cumulative flush count of primary shards                       │                     18178 │
│ Min cumulative flush time across primary shards                │ 6.666666666666667e-05 min │
│ Median cumulative flush time across primary shards             │   0.16653333333333334 min │
│ Max cumulative flush time across primary shards                │    0.7956666666666667 min │
│ Total Young Gen GC time                                        │                   0.017 s │
│ Total Young Gen GC count                                       │                         3 │
│ Total Old Gen GC time                                          │                       0 s │
│ Total Old Gen GC count                                         │                         0 │
│ Store size                                                     │    0.06993633136153221 GB │
│ Translog size                                                  │  0.0001253066584467888 GB │
│ Heap used for segments                                         │                      0 MB │
│ Heap used for doc values                                       │                      0 MB │
│ Heap used for terms                                            │                      0 MB │
│ Heap used for norms                                            │                      0 MB │
│ Heap used for points                                           │                      0 MB │
│ Heap used for stored fields                                    │                      0 MB │
│ Segment count                                                  │                       497 │
│ Total Ingest Pipeline count                                    │                     20026 │
│ Total Ingest Pipeline time                                     │                   0.741 s │
│ Total Ingest Pipeline failed                                   │                         0 │
│ Min Throughput                                                 │           62131.30 docs/s │
│ Mean Throughput                                                │           62131.30 docs/s │
│ Median Throughput                                              │           62131.30 docs/s │
│ Max Throughput                                                 │           62131.30 docs/s │
│ 50th percentile latency                                        │     253.73077099999898 ms │
│ 100th percentile latency                                       │     269.62587499999916 ms │
│ 50th percentile service time                                   │     253.73077099999898 ms │
│ 100th percentile service time                                  │     269.62587499999916 ms │
│ error rate                                                     │                  100.00 % │
╰────────────────────────────────────────────────────────────────┴───────────────────────────╯

--- Benchmark results for package: rally_benchmarks - END   ---
Done

```

@marc-gr: I have a doubt about warm-up period. it is run in a goroutine in the setup method, rally will be executed before that. I'm not sure how does it works for system benchmark. what is it supposed to achieve the warm up period?


aspacca:
closes #1475

README.md
```diff
@@ -153,6 +153,12 @@ These benchmarks allow you to benchmark any Ingest Node Pipelines defined by you
 
 For details on how to configure pipeline benchmarks for a package, review the [HOWTO guide](./docs/howto/pipeline_benchmarking.md).
 
+#### Rally Benchmarks
+
+These benchmarks allow you to benchmark an integration corpus with rally.
+
+For details on how to configure rally benchmarks for a package, review the [HOWTO guide](./docs/howto/rally_benchmarking.md).
```
ruflin, line 160: Is this link correct? Trying to find it in this PR.

aspacca: the link is correct. but I haven't yet written the docs :)

ruflin: Can we add these to the PR? I was looking for these as I hit some issues testing the PR. Will comment on more on the issue I hit soon.

internal/benchrunner/runners/rally/report.go
```diff
@@ -0,0 +1,239 @@
...
+
+func reportJSONFormat(r *report) ([]byte, error) {
+	b, err := json.MarshalIndent(r, "", "\t")
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
+}
+
+func reportHumanFormat(r *report) []byte {
```
ruflin, line 99: Trying to get my head around how this report exactly works. What happens if rally adds new fields to the report, does it mean we have to update our code? Does rally itself already have some tooling / code in place to output it into a human readable format?

ruflin:
I'm running these commands to do some testing. Few findings:


* `elastic-package build` fails with `package with GA version (999.999.999) is using an unreleased version of the spec (3.0.1-next) (PSR00001)`. Interestingly enough the rally benchmark runs. Did it install the package?
* Keep generated rally track: Is there an option that I can keep the generated rally track file? I checked the directory that was referenced but it seems it is empty after the run. I assume it cleans it up? It would be nice to be able to generate only the rally track so others can run it later or we can run it multiple times for comparison. This would also help to review the rally track itself. I see you have "defer-cleanup" but that only seems temporary?
* esrally installation: If rally is not there, can we ask users to run `pip3 install esrally` or point them to the docs directly?

aspacca:
> * `elastic-package build` fails with `package with GA version (999.999.999) is using an unreleased version of the spec (3.0.1-next) (PSR00001)`. Interestingly enough the rally benchmark runs. Did it install the package?

I was not aware of this: during my test I used `go run main.go`
Before merging the PR the version of the spec will be `3.0.1` (most likely), ie: a released version of the spec

> * Keep generated rally track: Is there an option that I can keep the generated rally track file? I checked the directory that was referenced but it seems it is empty after the run. I assume it cleans it up? It would be nice to be able to generate only the rally track so others can run it later or we can run it multiple times for comparison. This would also help to review the rally track itself. I see you have "defer-cleanup" but that only seems temporary?

We have a separated command for generating rally track without running the benchmark: `elastic-package benchmark generate-corpus --rally-track-output-dir`. currently this command still uses the generator assets from the generator repo, but I was planning to refactor it to use the assets in the package. maybe it's better to drop at all the command and add an option to save the track/do not run rally directly in the new command of this PR

> * esrally installation: If rally is not there, can we ask users to run `pip3 install esrally` or point them to the docs directly?

👍

ruflin:
> We have a separated command for generating rally track without running the benchmark: elastic-package benchmark generate-corpus --rally-track-output-dir. currently this command still uses the generator assets from the generator repo, but I was planning to refactor it to use the assets in the package. maybe it's better to drop at all the command and add an option to save the track/do not run rally directly in the new command of this PR

As a first step, lets make sure it all uses the assets from the package. I agree, unifying the commands might make sense but we could also solve it with docs pointing to it. Key is that the outcome is the same, meaning if I run the rally track "live" or generate the corpus, same data is in.

marc-gr:
> @marc-gr: I have a doubt about warm-up period. it is run in a goroutine in the setup method, rally will be executed before that. I'm not sure how does it works for system benchmark. what is it supposed to achieve the warm up period?

In your case I think this comes builtin with rally itself, the intention of this is to defer metric collection until the warm up period ends, so that time is not going to be taken into account for reporting.

aspacca:
> As a first step, lets make sure it all uses the assets from the package. I agree, unifying the commands might make sense but we could also solve it with docs pointing to it.

it was in end faster to add saving the rally track and a dry run in this PR than refactor the previous command: I will remove it in a next PR

> Key is that the outcome is the same, meaning if I run the rally track "live" or generate the corpus, same data is in.

it will not be exactly the same data: meaning that if I generate the corpus multiple time (running rally) the data will have the same "shape" (cardinality, range, etc), but different randomized value because seeds and time values are based on "now".

the generator tool already has the option to pass a seed and "now" from command line, let me know if you want to add to the command in elastic-package as well

README.md
aspacca: yes, it's planned to add to this PR: in the description I listed `missing docs` :)

internal/benchrunner/runners/rally/report.go
aspacca: this is basically a copy of the report that @marc-gr created for system benchmark.
I just added the stats from rally

rally outputs to stdout in human readable format by default, but the output is mixed with the rest, so I save a report from rally in csv format, and dump here with our pretty table :)

since the source is a csv no matter if rally adds new fields, we'll get them with no change required, as long as they are saved in the rally csv repo

aspacca:
> In your case I think this comes builtin with rally itself, the intention of this is to defer metric collection until the warm up period ends, so that time is not going to be taken into account for reporting.

ok, it's a warm up period for metrics collection, that makes sense as it is now then.
I thought it was a warmup for starting the benchmark :)

ruflin:
> it will not be exactly the same data: meaning that if I generate the corpus multiple time (running rally) the data will have the same "shape" (cardinality, range, etc), but different randomized value because seeds and time values are based on "now".

Yep, that is expected.

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
```diff
@@ -0,0 +1,20 @@
+---
+description: Benchmark 20000 events ingested
+input: filestream
```
ruflin, line 3: Is the input config needed?

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
```diff
@@ -0,0 +1,20 @@
...
+data_stream:
+  name: testds
+  vars:
+    paths:
+      - "dummy path"
+warmup_time_period: 10s
+corpora:
+  generator:
+    total_events: 20000
+    template:
```
ruflin, line 14: I'm looking at the templates from kubernetes.container: https://github.com/elastic/elastic-integration-corpus-generator-tool/tree/main/assets/templates/kubernetes.container There are two `.tpl` files. How does this translate to this structure? 

There is also configs and config, is that a cleanup?

aspacca: > I'm looking at the templates from kubernetes.container: https://github.com/elastic/elastic-integration-corpus-generator-tool/tree/main/assets/templates/kubernetes.container There are two `.tpl` files. How does this translate to this structure?

the two template are the same. one "pretty-printed" with new lines. we will use only the latter once migrating it to the package (currently `elastic-package` removes the newline, but we'll later do that directly in the corpus generator)
 
> There is also configs and config, is that a cleanup?

I see only `configs.yaml`. there will be only one config file per benchmark. how do we name it it's not relevant, as long as we refer it with the proper name in the benchmark scenario, and while having some naming convention might be useful :)

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
aspacca: unluckily yes :(
I was hoping to be able to get rid of it, and hardcode in the code an input type that requires no vars.

the problem is that the package and policy will be installed, before running rally, and they must match an input defined by the package I'm going to benchmark, otherwise the policy will be rejected.

I'm trying to get rid of it, but it requires installing a common input for every package that's not part of the original package

ruflin: Is this the issue around "each package need requires to have a policy"? @tommyers-elastic Is this still required? How did you deal with it in the context of firehose? It seems https://github.com/elastic/integrations/pull/7146 does not contain an input? 

It is also not clear to me, why it is needed here? What exactly is triggered when installing the package? Do you create a policy or only install the package? I would assume for only installing the package, this is not needed?

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
ruflin: ++ on having a convention. If we have the convention and someone follows it, it means also the config could be left out by default eventually. Something we can iterate on.

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
ruflin: I looked at what happens during the rally run and it sets up a policy. I don't think this should happen, it should only install the package itself, no policy setup. Basically what `elastic-package install` does. This should also remove the requirement for an input.

aspacca: > I looked at what happens during the rally run and it sets up a policy. I don't think this should happen, it should only install the package itself, no policy setup. Basically what `elastic-package install` does. This should also remove the requirement for an input.

indeed! I've realized it as well and remove everything: I've just pushed the commit :)

ruflin: It seems the input config is still here? Is this now still needed?

test/packages/benchmarks/rally_benchmark/data_stream/testds/agent/stream/filestream.yml.hbs
```diff
@@ -0,0 +1,4 @@
+paths:
```
ruflin, line 1: As discussed above, seems we don't need all the input configs. Even though it does not matter for the test run, lets keep the sample / test package as minimal as possible and remove everything related to inputs.

test/packages/benchmarks/rally_benchmark/data_stream/testds/manifest.yml
```diff
@@ -0,0 +1,16 @@
+title: Test
+release: experimental
+type: logs
+streams:
```
ruflin, line 4: See above, I hope this is not required.

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
ruflin: Running it again, it seems now it does not install the package at all anymore? We need the package for the pipeline and templates.

internal/benchrunner/runners/rally/runner.go
```diff
@@ -0,0 +1,816 @@
...
+	} else {
+		logger.Debugf("found %d hits in %s data stream", numHits, dataStream)
+	}
+
+	return numHits, nil
+}
+
+func waitUntilTrue(fn func() (bool, error), timeout time.Duration) (bool, error) {
+	timeoutTicker := time.NewTicker(timeout)
+	defer timeoutTicker.Stop()
```
jsoriano, line 787: Nit. This one doesn't need to be a ticker.
```suggestion
	timeout := time.NewTimer(timeout)
	defer timeout.Stop()
```

internal/benchrunner/runners/rally/scenario.go
```diff
@@ -0,0 +1,115 @@
...
+
+type corporaConfig struct {
+	Raw  map[string]interface{} `config:"raw" json:"raw"`
+	Path string                 `config:"path" json:"path"`
+}
+
+type corporaFields struct {
+	Raw  map[string]interface{} `config:"raw" json:"raw"`
+	Path string                 `config:"path" json:"path"`
+}
```
jsoriano, line 61: Nit. These two structs are the same?

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
```diff
@@ -0,0 +1,17 @@
...
+      type: gotext
+      path: ./logs-benchmark/template.ndjson
+    config:
+      path: ./logs-benchmark/config.yml
+    fields:
+      path: ./logs-benchmark/fields.yml
```
jsoriano, line 17: It looks like all the files are known and the only variable is the dir, should we require only the directory name?
```suggestion
    config:
      template_type: gotext
      dir: ./logs-benchmark
```

Or we want to support to reuse files between different benchmarks?

README.md
jsoriano: Will give it a try when there are docs :slightly_smiling_face:

jsoriano (commented):
Thanks! Will give it a try when there are docs.

internal/benchrunner/runners/rally/metrics.go
```diff
@@ -0,0 +1,370 @@
...
+			case <-c.tick.C:
+				once.Do(func() {
+					c.waitUntilReady()
+					c.startIngestMetrics = c.collectIngestMetrics()
+					c.startTotalHits = c.collectTotalHits()
+					c.startMetrics = c.collect()
+					c.publish(c.createEventsFromMetrics(c.startMetrics))
+				})
+				m := c.collect()
+				c.publish(c.createEventsFromMetrics(m))
```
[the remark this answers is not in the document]
aspacca: @marc-gr can you please answer @jsoriano questions? :)

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
aspacca: this is still an open discussion: the spec for system scenario was defined this way. also in the spec we can either have a path to a file or raw inline content. I'd keep it this way, even considering changing it will be another one breaking change :)

aspacca:
@jsoriano I've addressed all your comments but a few about `metrics.go`: this is just a copy of the code from @marc-gr and I think it would be better for him to address in a separated PR. in `metrics.go` I "fixed" the code as you requested only where the behavior won't change.

please, let me know if you are satisfied with the outcome. @ruflin the same for you

once the code it is finalised I will add the docs :)

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
aspacca: sorry. there was some leftover.

about the package not installed, according to [fleet api](https://www.elastic.co/guide/en/fleet/current/fleet-api-docs.html) I don't see a way to install a package without creating a policy.
but then we are back to the beginning where the policy needs to match the inputs of the package.
I guess for firehose integration it works without input because the package has no input

ruflin: I think the API call you are looking for is here: https://www.elastic.co/guide/en/fleet/current/fleet-apis.html#installPackage

Any package can be just installed without policy, policy is just "another" asset.

ruflin: @jsoriano I wonder if we need to rely on the package registry at all here for installation or could instead us `elastic-package install` command so there is no dependency on the registry. Maybe that already happens?

jsoriano: Yes, for versions of Kibana that support it, `elastic-package install` installs packages without using the registry, and without requiring a policy.

You can find here how the system test runner installs packages to leverage the best available method: https://github.com/elastic/elastic-package/blob/a7e19f9823f38bff7822a4bd3c61f5f61d4bb81a/internal/testrunner/runners/system/runner.go#L490

test/packages/benchmarks/rally_benchmark/_dev/benchmark/rally/logs-benchmark.yml
jsoriano: Oh ok, the raw and the possibility of reusing files can be good reasons to keep it like this.

internal/benchrunner/runners/rally/runner.go
```diff
@@ -263,7 +276,27 @@ func (r *runner) run() (report reporters.Reportable, err error) {
 		return nil, fmt.Errorf("can't reindex data: %w", err)
 	}
 
-	return createReport(r.options.BenchName, r.corporaFile, r.scenario, msum, rallyStats)
+	return createReport(r.options.BenchName, r.corpusFile, r.scenario, msum, rallyStats)
+}
+
+func (r *runner) installPackage(packageName, packageVersion string) error {
```
jsoriano, line 282: As [commented in other thread](https://github.com/elastic/elastic-package/pull/1522#discussion_r1378669668), consider installing packages as system tests do to avoid depending on the package registry when not needded: https://github.com/elastic/elastic-package/blob/a7e19f9823f38bff7822a4bd3c61f5f61d4bb81a/internal/testrunner/runners/system/runner.go#L490

internal/corpusgenerator/rally.go
```diff
@@ -54,12 +55,12 @@ func GenerateRallyTrack(dataStream string, corpusFile *os.File, corpusDocsCount
 
 	parsedTpl, err := t.Delims("[[", "]]").Parse(rallyTrackTemplate)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("erro while parsing rally track template: %w", err)
```
jsoriano, line 58:
```suggestion
		return nil, fmt.Errorf("error while parsing rally track template: %w", err)
```

internal/benchrunner/runners/rally/runner.go
```diff
@@ -0,0 +1,855 @@
...
+				return errors.Join(os.Remove(persistedRallyTrack), err)
+			}
+
+			logger.Infof("rally track and corpus saved at: %s", r.options.RallyTrackOutputDir)
+			return nil
+		}
+	}
+
+	r.clearCorporaHandler = func() error {
+		return errors.Join(os.Remove(r.corpusFile), os.Remove(r.reportFile), os.Remove(r.trackFile))
```
jsoriano, line 546: Nit.
```suggestion
		return errors.Join(
			os.Remove(r.corpusFile),
			os.Remove(r.reportFile),
			os.Remove(r.trackFile),
		)
```

jsoriano (commented):
> this is just a copy of the code from @marc-gr and I think it would be better for him to address in a separated PR. in `metrics.go` I "fixed" the code as you requested only where the behavior won't change.
> 
> please, let me know if you are satisfied with the outcome.

Ok, code-wise we can continue as it is now, but please follow up on the pending comments. I think that at least error handling should be revisited.

> once the code it is finalised I will add the docs :)

Yes please :slightly_smiling_face:

ruflin:
I did run the command on my end again and adjusted the config with the following:

```
corpora:
  generator:
    total_events: 900000
````

Few findings:

Before the warmup period ended, I saw several times the following message (debug mode):

```
2023/11/01 16:13:53 DEBUG unexpected response status for DataStream Stats (404): 404 Not Found: elasticsearch error (type=index_not_found_exception): no such index [logs-rally_benchmarks.testds-ep]
Root cause:
[
  {
    "type": "index_not_found_exception",
    "reason": "no such index [logs-rally_benchmarks.testds-ep]",
    "position": {
      "offset": 0,
      "start": 0,
      "end": 0
    }
  }
]
2023/11/01 16:13:53 DEBUG waiting 10s for warmup period
```
As no data was shipped yet, I would argue this is expected. Is this a rally error? Do we need the warmup period for this or is it just an example?

The full benchmark result looked as below (collapsed). The total ingest pipeline count seems to (almost) accurate:

```
│ Total Ingest Pipeline count                                    │                     900074 │
```

Where is this diff coming from?

Where I got curious is with the following line:

```
│ Totals                                         │ Count: 444885 | Failed: 0 | Time: 14.803s │
```
What happened to all the other docs, seems like this is half of the docs? 

This gets me to the approx ingested docs:

```
│ approx total docs ingested │                          518521 │
```
I expected this to be identical to the number I specified in the track itself?

Where are these diffs coming from?

<details>

<summary>Test run results</summary>

```
--- Benchmark results for package: rally_benchmarks - START ---
╭────────────────────────────────────────────────────────────────────────────────────────────╮
│ info                                                                                       │
├────────────────────────┬───────────────────────────────────────────────────────────────────┤
│ benchmark              │                                                    logs-benchmark │
│ description            │                                   Benchmark 20000 events ingested │
│ run ID                 │                              9fd8b742-3260-4c4f-be04-eae781ffaafa │
│ package                │                                                  rally_benchmarks │
│ start ts (s)           │                                                        1698851644 │
│ end ts (s)             │                                                        1698851663 │
│ duration               │                                                               19s │
│ generated corpora file │ /Users/ruflin/.elastic-package/tmp/rally_corpus/corpus-3307614978 │
╰────────────────────────┴───────────────────────────────────────────────────────────────────╯
╭────────────────────────────────────────────────────────────────────╮
│ parameters                                                         │
├─────────────────────────────────┬──────────────────────────────────┤
│ package version                 │                      999.999.999 │
│ data_stream.name                │                           testds │
│ warmup time period              │                              10s │
│ corpora.generator.total_events  │                           900000 │
│ corpora.generator.template.path │ ./logs-benchmark/template.ndjson │
│ corpora.generator.template.raw  │                                  │
│ corpora.generator.template.type │                           gotext │
│ corpora.generator.config.path   │      ./logs-benchmark/config.yml │
│ corpora.generator.config.raw    │                            map[] │
│ corpora.generator.fields.path   │      ./logs-benchmark/fields.yml │
│ corpora.generator.fields.raw    │                            map[] │
╰─────────────────────────────────┴──────────────────────────────────╯
╭───────────────────────╮
│ cluster info          │
├───────┬───────────────┤
│ name  │ elasticsearch │
│ nodes │             1 │
╰───────┴───────────────╯
╭──────────────────────────────────────────────────────────────╮
│ data stream stats                                            │
├────────────────────────────┬─────────────────────────────────┤
│ data stream                │ logs-rally_benchmarks.testds-ep │
│ approx total docs ingested │                          518521 │
│ backing indices            │                               1 │
│ store size bytes           │                       353266325 │
│ maximum ts (ms)            │                   1698851597975 │
╰────────────────────────────┴─────────────────────────────────╯
│ disk usage for index .ds-logs-rally_b │
│ enchmarks.testds-ep-2023.11.01-000001 │
│ (for all fields)                      │
├──────────────────────────────┬────────┤
│ total                        │ 264 MB │
│ inverted_index.total         │  86 MB │
│ inverted_index.stored_fields │  96 MB │
│ inverted_index.doc_values    │  72 MB │
│ inverted_index.points        │ 9.4 MB │
│ inverted_index.norms         │    0 B │
│ inverted_index.term_vectors  │    0 B │
│ inverted_index.knn_vectors   │    0 B │
╰──────────────────────────────┴────────╯
╭────────────────────────────────────────────────────────────────────────────────────────────╮
│ pipeline logs-rally_benchmarks.testds-999.999.999 stats in node qWQdhtF3RmWdrTDSuLRyNA     │
├────────────────────────────────────────────────┬───────────────────────────────────────────┤
│ Totals                                         │ Count: 444885 | Failed: 0 | Time: 14.803s │
│ grok ()                                        │ Count: 444885 | Failed: 0 | Time: 13.416s │
│ user_agent ()                                  │   Count: 444885 | Failed: 0 | Time: 884ms │
│ pipeline (logs-rally_benchmarks.testds@custom) │    Count: 444885 | Failed: 0 | Time: 56ms │
╰────────────────────────────────────────────────┴───────────────────────────────────────────╯
╭─────────────────────────────────────────────────────────────────────────────────────────────╮
│ rally stats                                                                                 │
├────────────────────────────────────────────────────────────────┬────────────────────────────┤
│ Cumulative indexing time of primary shards                     │     1.3116166666666667 min │
│ Min cumulative indexing time across primary shards             │                      0 min │
│ Median cumulative indexing time across primary shards          │ 0.00016666666666666666 min │
│ Max cumulative indexing time across primary shards             │     1.2954333333333332 min │
│ Cumulative indexing throttle time of primary shards            │                      0 min │
│ Min cumulative indexing throttle time across primary shards    │                      0 min │
│ Median cumulative indexing throttle time across primary shards │                      0 min │
│ Max cumulative indexing throttle time across primary shards    │                      0 min │
│ Cumulative merge time of primary shards                        │    0.10246666666666666 min │
│ Cumulative merge count of primary shards                       │                          1 │
│ Min cumulative merge time across primary shards                │                      0 min │
│ Median cumulative merge time across primary shards             │                      0 min │
│ Max cumulative merge time across primary shards                │    0.10246666666666666 min │
│ Cumulative merge throttle time of primary shards               │   0.029366666666666666 min │
│ Min cumulative merge throttle time across primary shards       │                      0 min │
│ Median cumulative merge throttle time across primary shards    │                      0 min │
│ Max cumulative merge throttle time across primary shards       │   0.029366666666666666 min │
│ Cumulative refresh time of primary shards                      │                0.13535 min │
│ Cumulative refresh count of primary shards                     │                        326 │
│ Min cumulative refresh time across primary shards              │                      0 min │
│ Median cumulative refresh time across primary shards           │                 0.0002 min │
│ Max cumulative refresh time across primary shards              │                0.11655 min │
│ Cumulative flush time of primary shards                        │    0.12921666666666667 min │
│ Cumulative flush count of primary shards                       │                        160 │
│ Min cumulative flush time across primary shards                │                      0 min │
│ Median cumulative flush time across primary shards             │  0.0016833333333333333 min │
│ Max cumulative flush time across primary shards                │                 0.0176 min │
│ Total Young Gen GC time                                        │                    1.141 s │
│ Total Young Gen GC count                                       │                        136 │
│ Total Old Gen GC time                                          │                        0 s │
│ Total Old Gen GC count                                         │                          0 │
│ Store size                                                     │      0.3340198630467057 GB │
│ Translog size                                                  │       1.074068870395422 GB │
│ Heap used for segments                                         │                       0 MB │
│ Heap used for doc values                                       │                       0 MB │
│ Heap used for terms                                            │                       0 MB │
│ Heap used for norms                                            │                       0 MB │
│ Heap used for points                                           │                       0 MB │
│ Heap used for stored fields                                    │                       0 MB │
│ Segment count                                                  │                        260 │
│ Total Ingest Pipeline count                                    │                     900074 │
│ Total Ingest Pipeline time                                     │                   51.357 s │
│ Total Ingest Pipeline failed                                   │                          0 │
│ Min Throughput                                                 │             2077.70 docs/s │
│ Mean Throughput                                                │            28063.49 docs/s │
│ Median Throughput                                              │            32122.07 docs/s │
│ Max Throughput                                                 │            38684.85 docs/s │
│ 50th percentile latency                                        │       749.7723335109185 ms │
│ 90th percentile latency                                        │      1241.7405874119147 ms │
│ 99th percentile latency                                        │      2332.3510095642996 ms │
│ 100th percentile latency                                       │       2345.935666002333 ms │
│ 50th percentile service time                                   │       749.7723335109185 ms │
│ 90th percentile service time                                   │      1241.7405874119147 ms │
│ 99th percentile service time                                   │      2332.3510095642996 ms │
│ 100th percentile service time                                  │       2345.935666002333 ms │
│ error rate                                                     │                     0.00 % │
╰────────────────────────────────────────────────────────────────┴────────────────────────────╯

--- Benchmark results for package: rally_benchmarks - END   ---
Done
```
</details>

File: internal/benchrunner/runners/rally/runner.go (hunk 11 above)
Comment (aspacca) [reply internal/benchrunner/runners/rally/runner.go]: done

**Comment by aspacca** [comment]
> As no data was shipped yet, I would argue this is expected. Is this a rally error? Do we need the warmup period for this or is it just an example?

yes, it is expected, I'd remove the logging there, since it only brings noise

> The full benchmark result looked as below (collapsed). The total ingest pipeline count seems to (almost) accurate:
> 
> ```
> │ Total Ingest Pipeline count                                    │                     900074 │
> ```
> 
> Where is this diff coming from?

This stats comes directly from rally, I checked and we generate the exact amoung of events we request. So not sure

> This gets me to the approx ingested docs:
> 
> ```
> │ approx total docs ingested │                          518521 │
> ```
> 
> I expected this to be identical to the number I specified in the track itself?
> 
> Where are these diffs coming from?

"approx total docs ingested" is the diff between the count api results before running rally, and after running rally
could you check the debug logs like the followings:
```
2023/11/02 10:19:57 DEBUG found 0 hits in logs-rally_benchmarks.testds-ep data stream
2023/11/02 10:20:44 DEBUG found 900000 hits in logs-rally_benchmarks.testds-ep data stream
```

File: internal/benchrunner/runners/rally/metrics.go [hunk 14]
```diff
@@ -0,0 +1,370 @@
...
+		}
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			logger.Errorf("failed to read index response body: %v", err)
+		}
+		resp.Body.Close()
+
+		if resp.StatusCode != 201 {
+			logger.Errorf("error indexing event (%d): %s: %v", resp.StatusCode, resp.Status(), elasticsearch.NewError(body))
```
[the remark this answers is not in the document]
aspacca: it might be a transient error.
to my understanding publishing events to the metricsstore is best effort: we should not stop the benchmark if it fails
but we do log the error, so the user is aware of that

internal/benchrunner/runners/rally/metrics.go
```diff
@@ -0,0 +1,370 @@
...
+		c.metricsAPI.Indices.Create.WithBody(reader),
+	)
+	if err != nil {
+		logger.Debugf("could not create index: %v", err)
+		return
+	}
+	createRes.Body.Close()
+
+	if createRes.IsError() {
+		logger.Debug("got a response error while creating index")
```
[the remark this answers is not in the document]
aspacca: I don't think the error should be returned, but yes, let's use a `logger.Errorf`

internal/benchrunner/runners/rally/metrics.go
```diff
@@ -0,0 +1,370 @@
...
+		select {
+		case <-c.stopC:
+			return
+		case <-waitTick.C:
+		}
+		dsstats, err := ingest.GetDataStreamStats(c.esAPI, c.datastream)
+		if err != nil {
+			logger.Debug(err)
+		}
+		if dsstats != nil {
```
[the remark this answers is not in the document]
aspacca: yes, the method is `waitUntilReady`, that will return either when there is at least one document in the datastream, or when `stopC` is closed (ie: rally finished running)

internal/benchrunner/runners/rally/metrics.go
```diff
@@ -0,0 +1,370 @@
...
+	c.endIngestMetrics = c.collectIngestMetrics()
+	c.diskUsage = c.collectDiskUsage()
+	c.endTotalHits = c.collectTotalHits()
+	c.endMetrics = c.collect()
+}
+
+func (c *collector) collectTotalHits() int {
+	totalHits, err := getTotalHits(c.esAPI, c.datastream)
+	if err != nil {
+		logger.Debugf("could not total hits: %w", err)
```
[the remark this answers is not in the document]
aspacca: it's tricky: the stats that we really care about are the one from rally. we add some extra stats (the same as collected by system benchmark): the name of the metrics where we show this potential zero value is `approx total docs ingested` :)

ruflin:
The part that makes me nervous with the above diffs is that I'm worried that some of the ingestion might not work or similar. For example also in your initial example, ingested docs is 0: `approx total docs ingested │                               0 │`. Maybe these stats are not always up-to-date and have a delay? I wonder if refresh or something similar would help? https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html It seems rally stats are correct but node stats are not? I'll also check the logs.

aspacca:
> The part that makes me nervous with the above diffs is that I'm worried that some of the ingestion might not work or similar. For example also in your initial example, ingested docs is 0: `approx total docs ingested │ 0 │`.

That's correct:
```
error rate                                                     │                  100.00 %
```

nothing was ingested :)

> Maybe these stats are not always up-to-date and have a delay? I wonder if refresh or something similar would help? https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html It seems rally stats are correct but node stats are not? I'll also check the logs.

I will execute a refresh on the index, that should do

ruflin:
I did another run and the reported numbers are too low for example for ingest pipeline stats. But I did a follow up check for `GET /_nodes/stats` manually and there the results is correct. It seems like we have a delay / caching issue, lets see if refresh solves it. I'm not too optimistic around refresh as this only refreshes the index, likely not the node stats. @marc-gr Did you encounter a similar issue on your benchmarks?

aspacca:
@ruflin I now refresh the index before collecting stats
also I removed polling the hits while rally is running, since this will compete for resources

docs/howto/rally_benchmarking.md
```diff
@@ -0,0 +1,268 @@
...
+Example:
+
+`logs-benchmark.yml`
+```yaml
+---
+---
+description: Benchmark 20000 events ingested
+data_stream:
+   name: testds
+warmup_time_period: 10s
```
ruflin, line 63: As we don't have an elastic-agent shipping data, what data is there that could come in during the warm up period? Do we need this?

aspacca: with my last commit we wait for this period after we setup the benchmark: ie deleting all existing documents in the data stream. then we collect the stats before running rally, refreshing the index. so yes, I'm ok to remove as well since it make it simpler and lower latency

docs/howto/rally_benchmarking.md
```diff
@@ -0,0 +1,268 @@
...
+You can see a sample collected metric [here](./sample_metric.json)
+
+Additionally, if the `reindex-to-metricstore` flag is used, the data generated during the benchmark will be sent to the metricstore into an index called `bench-reindex-{datastream}-{testRunID}` for further analysis. The events will be enriched with metadata related to the benchmark run.
+
+## Persisting rally tracks and dry-run
+
+If the `rally-track-output-dir` flag is used, the track and the corpus generated during the benchmark will be saved in the directory passed as value of the flag.
+Additionally, if the `dry-run` flag is used, the command will exits before running `esrally`.
+If both the flags above are used at the same time, the command will just generate the track and corpus and save them, without running any benchmark.
+If the `dry-run` flag only is used, the command will just wipe the data stream returning no report.
```
ruflin, line 268: The rally track directory contains a few files. Can you add some short details for each file. The most important part I'm looking for: What is the rally track file that someone could copy and eventually run with esrally seperately if they want. The manual command to do it would be nice a addition.

internal/cobraext/flags.go
```diff
@@ -53,6 +53,12 @@ const (
 	BenchWithTestSamplesFlagName        = "use-test-samples"
 	BenchWithTestSamplesFlagDescription = "use test samples for the benchmarks"
 
+	BenchCorpusRallyTrackOutputDirFlagName        = "rally-track-output-dir"
+	BenchCorpusRallyTrackOutputDirFlagDescription = "output dir of the rally track: if present the command will save the generated rally track"
+
+	BenchCorpusRallyDryRunFlagName        = "dry-run"
+	BenchCorpusRallyDryRunFlagDescription = "Do not run rally bust just generate the rally track"
```
ruflin, line 60:
```suggestion
	BenchCorpusRallyDryRunFlagDescription = "Do not run rally but just generate the rally track"
```

ruflin:
If @jsoriano agrees, lets get this PR in rather soonish and then add the corpus templates to some of the integration packages. I expect that this will also provide us some more feedback for iterating on the command itself. Also get it in the hands of the rest of the team to start playing with it.

Can we treat the command "beta" for now so we can still make breaking changes to it?

aspacca:
> If @jsoriano agrees, lets get this PR in rather soonish

we must merge https://github.com/elastic/package-spec/pull/653 in order to pass CI

jsoriano (commented):
Ok to merge this and iterate, once the package-spec change is released.

ruflin (approved):

aspacca:
> Ok to merge this and iterate, once the package-spec change is released.

before merging this I have to revert the last two commits, but without package-spec released it won't pass CI.
do I miss anything? :)

ruflin:
I just merged the package-spec PR. My assumption is that we would stay on a commit reference but would move it over to the one in package-spec which is merged now instead of a release. And then as soon as a new package-spec is out, we update.

PR #1526, opened 2023-10-27
Include elastic-api-version header in Kibana requests

jsoriano:
Installing zips seems to require elastic-api-version header starting on Kibana 8.12.

### DIFF
--- a/internal/kibana/packages.go
+++ b/internal/kibana/packages.go
@@ -41,6 +41,7 @@ func (c *Client) InstallZipPackage(zipFile string) ([]packages.Asset, error) {
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/zip")
+	req.Header.Add("elastic-api-version", "2023-10-31")
 
 	statusCode, respBody, err := c.doRequest(req)
 	if err != nil {


jsoriano:
test integrations

mrodm:
Latest comment to test integrations didn't trigger the build because of this error:
> go.mod:206: replace github.com/jsoriano/elastic-package: version "b55f0d6933bf71c5fa4d22f6a4137cc5f9cd3119" invalid: unknown revision b55f0d6933bf71c5fa4d22f6a4137cc5f9cd3119

mrodm:
test integrations

internal/kibana/packages.go
```diff
@@ -41,6 +41,7 @@ func (c *Client) InstallZipPackage(zipFile string) ([]packages.Asset, error) {
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/zip")
+	req.Header.Add("elastic-api-version", "2023-10-31")
```
mrodm, line 44: Should this header be added for all other requests performed by elastic-package (dashboards, status...)?

Maybe, it can be added (or moved?) here for all of them:
https://github.com/elastic/elastic-package/blob/689e77a11c2f534aad75f3f1f3bdc7dd375eee72/internal/kibana/client.go#L153

jsoriano: Not sure, it only seems to be required here, maybe this API has been GA'ed on this version?

mrodm: Ok. I guess we could keep it here, and move it to the `newRequest()` function if there are new other API requests failing with the same error.

mrodm (approved):

PR #1533, opened 2023-11-03
Ignore error log when a component goes to a degraded state

jsoriano:
This error is unexpectedly happening with 8.11 stacks, but is a temporary issue, component recovers after some seconds. The issue is under investigation.


kpollich (approved):

mrodm (approved):
LGTM

PR #1538, opened 2023-11-07
Generate test GPG keys when running the tests

jsoriano:


scripts/gpgkey/main.go
```diff
@@ -0,0 +1,82 @@
...
+	rsaKey, err = rsaKey.Lock(passphrase)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	rsaKeyArmor, err := rsaKey.Armor()
+	if err != nil {
+		log.Fatal(err)
+	}
+	rsaPublicKeyArmor, err := rsaKey.GetArmoredPublicKey()
```
mrodm, line 54: Not familiar with this library,
should this be using `rsaKeyArmor` instead of `rsaKey`?

```suggestion
	rsaPublicKeyArmor, err := rsaKeyArmor.GetArmoredPublicKey()
```

jsoriano: I don't think so. `rsaKey` contains information about both the private key and its public key. "Armoring" here only refers to getting each one of the keys in PEM format, the format that elastic-package expects to find.

jsoriano: `rsaKeyArmor` is a string.

mrodm: Ah Ok!
Thanks!

mrodm (approved):
:rocket:

Issue #1537, opened 2023-11-07
[meta] Improve system tests robustness


marc-gr:
Several integrations that rely on external APIs use `stream` as their test services to build mocks and run system tests.

There are some blind spots when it comes to detecting config errors that could lead to wrong pagination, infinite pagination loops, etc. mainly caused by the inability of `elastic-package` to detect anything else than `is there any data?` in the data stream or `is there the expected number of events?` if `assert.hit_count` is set, which makes things better but is not enough.

As part of an effort to improve system tests robustness some things could be improved:

- [ ] Add an option to `data_stream/_dev/test/system/config.yml` to fail the tests if the test services exit unsuccessfully. https://github.com/elastic/elastic-package/pull/1540
- [x] https://github.com/elastic/stream/issues/61
- [x] https://github.com/elastic/stream/issues/62

Once this is done we can leverage these new behavior for system tests that require it.

PR #1540, opened 2023-11-07
Add ignore_service_error flag to selectively ignore test service errored containers

marc-gr:
Adds a flag to selectively fail if a test service container exits with an error.

Relates https://github.com/elastic/elastic-package/issues/1537


internal/compose/compose.go
```diff
@@ -318,8 +318,7 @@ func (p *Project) Logs(opts CommandOptions) ([]byte, error) {
 func (p *Project) WaitForHealthy(opts CommandOptions) error {
 	// Read container IDs
 	args := p.baseArgs()
-	args = append(args, "ps")
-	args = append(args, "-q")
+	args = append(args, "ps", "-a", "-q")
```
marc-gr, line 321: I encountered false positive failures for service containers that successfully finished but did so before reaching this point, so they would not be listed without `-a`. Since `exited but with 0` is considered healthy down the line, this is a requirement for these containers to show.

jsoriano:
test integrations

marc-gr:
test integrations

marc-gr:
test integrations

marc-gr:
@jsoriano the failing ones in the test PR seem to be unrelated if I am not mistaken

jsoriano (approved):
> @jsoriano the failing ones in the test PR seem to be unrelated if I am not mistaken

Good, let's change the default behaviour then to be on the safe side. Thanks for checking!

Issue #1486, opened 2023-10-04
Handle unexpected status codes in HTTP responses


jsoriano:
At least the kibana client, and at least when exporting dashboards, is not handling error status codes. For example it receives 400 errors when querying not available endpoints, and it tries to continue, producing unrelated failures later.

Review HTTP requests done by elastic-package and check that they are handling unexpected status codes.

PR #1549, opened 2023-11-08
Review handling of status code in HTTP requests

jsoriano:
In a number of places we were not handling HTTP status codes, what may hide issues
or complicate debugging.
Also ensure that body is closed in all cases.

Fixes https://github.com/elastic/elastic-package/issues/1486


internal/benchrunner/runners/rally/metrics.go
```diff
@@ -287,9 +286,14 @@ func (c *collector) collectDiskUsage() map[string]ingest.DiskUsage {
...
-	_, err := c.esAPI.Indices.Refresh(c.esAPI.Indices.Refresh.WithIndex(c.datastream))
+	resp, err := c.esAPI.Indices.Refresh(c.esAPI.Indices.Refresh.WithIndex(c.datastream))
 	if err != nil {
-		logger.Errorf("unable to refresh data stream at the end of rally run")
+		logger.Errorf("unable to refresh data stream at the end of rally run: %s", err)
+		return
+	}
+	defer resp.Body.Close()
+	if resp.IsError() {
+		logger.Errorf("unable to refresh data stream at the end of rally run: %s", resp)
```
mrodm, line 296: Should it be used `resp.String()` as here ? 
https://github.com/elastic/elastic-package/pull/1549/files#diff-e0b74fb572545d017ff634116e9791993aa6840936640ece5aa5ffa9549de539R348

internal/testrunner/runners/system/runner.go
```diff
@@ -1150,9 +1159,15 @@ func (r *runner) previewTransform(transformId string) ([]common.MapStr, error) {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to delete data stream docs: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Not found error is fine here, this means that data was already not there.
+	if resp.IsError() && resp.StatusCode != http.StatusNotFound {
+		return fmt.Errorf("failed to delete data stream docs for data stream %s: %s", dataStream, resp.String())
 	}
```
mrodm, line 1171: Should it be managed the other cases of `resp.IsError()` ?

internal/benchrunner/runners/rally/metrics.go
jsoriano: Indeed, good catch :+1:

internal/benchrunner/runners/rally/runner.go
```diff
@@ -336,10 +337,17 @@ func (r *runner) collectAndSummarizeMetrics() (*metricsSummary, error) {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to delete data stream docs for data stream %s: %w", dataStream, err)
 	}
+	defer resp.Body.Close()
+
+	// Not found error is fine here, this means that data was already not there.
+	if resp.IsError() && resp.StatusCode != http.StatusNotFound {
+		return fmt.Errorf("failed to delete data stream docs for data stream %s: %s", dataStream, resp.String())
+	}
```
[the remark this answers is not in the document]
jsoriano: Current code doesn't return an error on `http.StatusNotFound`, and we want to return an error in the rest of cases.

internal/testrunner/runners/system/runner.go
jsoriano: 503 needed to be handled, because of https://github.com/elastic/elasticsearch/issues/65846. Other errors should be just returned.

jsoriano: Ah no, 503 needed to be handled when searching in `getDocs`. Here I think that we don't need to handle other cases.

internal/testrunner/runners/system/runner.go
```diff
@@ -396,6 +401,20 @@ func (r *runner) getDocs(dataStream string) (*hits, error) {
...
 	defer resp.Body.Close()
 
+	if resp.StatusCode == http.StatusNotFound {
+		// No docs yet.
+		return &hits{}, nil
+	}
+	if resp.StatusCode == http.StatusServiceUnavailable && strings.Contains(resp.String(), "no_shard_available_exception") {
```
jsoriano, line 408: FTR, this is the error I saw that made me to handle this case:
```
Error: error running package system tests: could not complete test run: failed to search docs for data stream metrics-apache.status-ep: [503 Service Unavailable] {"error":{"root_cause":[{"type":"no_shard_available_action_exception","reason":null}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":".ds-metrics-apache.status-ep-2023.11.09-000001","node":null,"reason":{"type":"no_shard_available_action_exception","reason":null}}]},"status":503}
```

internal/benchrunner/runners/rally/runner.go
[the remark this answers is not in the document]
jsoriano: I have found that we can explicitly ignore errors on not found indexes, I am adding this option so we don't need special handling for 404s.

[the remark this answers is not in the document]
mrodm: Even better adding that option :+1:

mrodm (approved):
LGTM 
Should we run a `test integrations` in this PR before merging ?

jsoriano:
test integrations

internal/benchrunner/runners/rally/runner.go
```diff
@@ -752,40 +756,66 @@ func (r *runner) reindexData() error {
...
+		r.options.ESAPI.Scroll.WithScrollID(sr.ScrollID),
+		r.options.ESAPI.Scroll.WithScroll(time.Minute),
+	)
+	if err != nil {
+		return fmt.Errorf("error executing scroll: %s", err)
+	}
+	if resp.IsError() {
+		return fmt.Errorf("error executing scroll: %s", resp.String())
+	}
+	resp.Body.Close()
```
jsoriano, line 817: This close should be before the previous if, I will move this.

jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
I had to do some fixes after testing with integrations, it seems to be fine now @mrodm could you re-review? Thanks!

mrodm (approved):
:rocket:

PR #1550, opened 2023-11-08
Add support for ML models in elastic-package dump

jsoriano:
Some packages can include ML models, include them in dumps when available.


internal/dump/testdata/elasticsearch-8-dga-dump-all/ml_models/dga_1611725_2.0.json
```diff
@@ -0,0 +1,259 @@
...
+  "model_id": "dga_1611725_2.0",
+  "model_type": "tree_ensemble",
+  "created_by": "api_user",
+  "version": "8.9.0",
+  "create_time": 1699476967200,
+  "model_size_bytes": 246784104,
+  "estimated_operations": 0,
+  "license_level": "platinum",
+  "description": "Model used to detect domain generation algorithm (DGA) activity in your network data.",
+  "compressed_definition": "//REDACTED//",
```
jsoriano, line 11: Manually removed to avoid having several megabytes of binaries here.

internal/elasticsearch/test/httptest.go
```diff
@@ -61,7 +62,11 @@ func recordRequest(t *testing.T, r *http.Request, path string) {
 	require.NoError(t, err)
 
 	t.Logf("Recording %s in %s", r.URL.Path, path)
-	req, err := http.NewRequest(r.Method, r.URL.Path, nil)
+	var recordURL url.URL
+	recordURL.Path = r.URL.Path
+	recordURL.RawQuery = r.URL.RawQuery
+
+	req, err := http.NewRequest(r.Method, recordURL.String(), nil)
```
jsoriano, line 69: Fix to correctly record responses of urls with query parameters. I can move this small change to a different PR if needed.

mrodm (approved):

internal/elasticsearch/test/httptest.go
jsoriano: Yes, IIRC the idea is to remove anything that is not needed from the original request. Hosts, authentication and so on are provided by the elasticsearch client.
This has limitations, as we have found here. If we had things based on headers it would fail too. Also if we would like to capture requests with payload.
We may continue improving it as we need it.

PR #1551, opened 2023-11-09
Move service status check just after data collection.

marc-gr:
This way if no data is collected we know if it was due to the test service exiting. Otherwise the specific error gets hidden by the no data collected one.


jsoriano (approved):
:+1:

PR #1555, opened 2023-11-14
Add template for tsdb

aspacca:
Adding an index template that overrides `index.time_series.start_time/end_time` in order to fix the following error when running rally benchmark with more than 2 hours of events:
```
HTTP status: 400, message: the document timestamp [2023-12-04T07:27:45.000Z] is outside of ranges of currently writable indices [[2023-11-08T05:13:28.000Z,2023-11-08T09:13:28.000Z]
```

see more context at: https://github.com/elastic/integrations/issues/8431

The template is added directly in the rally track since the track can be persisted, and we want to be able to replay it directly with rally with the same corpus that has more than 2hours of event


ruflin:
You mention that the template is part of the rally track. Does this mean elastic-package installs the package and then rally follows up with the component template? If yes, that is pretty nice as it makes sure the rally track has all the "custom" info bundled with it.

internal/benchrunner/runners/rally/runner.go
```diff
@@ -45,48 +45,97 @@ import (
...
+            "operation": {
+                "operation-type": "bulk",
+                "bulk-size": {{bulk_size | default(5000)}},
+                "ingest-percentage": {{ingest_percentage | default(100)}}
+            },
+            "clients": {{bulk_indexing_clients | default(8)}}
+        },
+        {
+            "operation": {
+                "operation-type": "delete-composable-template",
```
ruflin, line 94: Nice, it even does the cleanup.

internal/benchrunner/runners/rally/runner.go
```diff
@@ -961,14 +1072,17 @@ func generateRallyTrack(dataStream string, corpusFile *os.File, corpusDocsCount
...
 	templateData := map[string]any{
-		"DataStream":        dataStream,
-		"CorpusFilename":    filepath.Base(corpusFile.Name()),
-		"CorpusDocsCount":   corpusDocsCount,
-		"CorpusSizeInBytes": corpusSizeInBytes,
+		"DataStream":         dataStream,
+		"CorpusFilename":     filepath.Base(corpusFile.Name()),
+		"CorpusDocsCount":    corpusDocsCount,
+		"CorpusSizeInBytes":  corpusSizeInBytes,
+		"ComposableTemplate": dataStream,
```
ruflin, line 1080: I expected this to be call `@custom` or similar. Have not run the code on my end yet, can you share what the exact name of this composable template is that is loaded?

aspacca: it is something like: `metrics-kubernetes.pod-ep` (the same name as the data stream)

`metrics-kubernetes.pod@custom` is managed by the integration package, and it was not enough to there the `time_series.start_time/end_time` and `routing_path`, not sure why.

I ended up calling the simulate index template APi for `metrics-kubernetes.pod` and generate a copy of the template just with different `time_series.start_time/end_time` and higher priority, to apply to index pattern `metrics-kubernetes.pod-ep`

internal/benchrunner/runners/rally/runner.go
```diff
@@ -45,48 +45,97 @@ import (
...
 
 const (
-	// RallyCorpusAgentDir is folder path where rally corporsa files produced by the service
+	// RallyCorpusAgentDir is folder path where rally corpora files produced by the service
 	// are stored on the Rally container's filesystem.
 	RallyCorpusAgentDir = "/tmp/rally_corpus"
 
-	// BenchType defining rally benchmark
-	BenchType benchrunner.Type = "rally"
+	rallyTrackTemplateForTSDB = `{% import "rally.helpers" as rally with context %}
```
jsoriano, line 52: Nit. Consider moving this template to its own file and embed it here.

internal/benchrunner/runners/rally/runner.go
```diff
@@ -274,6 +341,43 @@ func (r *runner) setUp() error {
 	return nil
 }
 
+func (r *runner) extractSimulatedTemplate(indexTemplate string) (string, error) {
+	simulateTemplate, err := r.options.ESAPI.Indices.SimulateTemplate(r.options.ESAPI.Indices.SimulateTemplate.WithName(indexTemplate))
+	if err != nil {
+		return "", fmt.Errorf("error simulating template from composable template: %s: %w", indexTemplate, err)
+	}
```
jsoriano, line 348: Remember to close the body.
```suggestion
	}
	defer simulateTemplate.Body.Close()
```

internal/benchrunner/runners/rally/runner.go
```diff
@@ -274,6 +341,43 @@ func (r *runner) setUp() error {
...
+
```
jsoriano, line 349: Check also the status code of the response.
```suggestion

	if  simulateTemplate.IsError() {
		return "", fmt.Errorf("error simulating template from composable template: %s: %s", indexTemplate, simulateTemplate.String())
	}
```

internal/benchrunner/runners/rally/runner.go
```diff
@@ -274,6 +341,43 @@ func (r *runner) setUp() error {
...
+	if err != nil {
+		return "", fmt.Errorf("error unmarshaling simulated template from composable template: %s: %w", indexTemplate, err)
+	}
+
+	simulatedTemplate["priority"] = 1000
+	simulatedTemplate["index_patterns"] = []string{indexTemplate + "-ep"}
+
+	indexTimeSeries := map[string]interface{}{
+		"start_time": "2000-01-01T00:00:00Z",
+		"end_time":   "2099-12-31T23:59:59Z",
```
jsoriano, line 366: Could this be too broad? Could we calculate something more limited based on the current time?

internal/benchrunner/runners/rally/runner.go
```diff
@@ -274,6 +341,43 @@ func (r *runner) setUp() error {
...
+	}
+
+	simulatedTemplate["template"].(map[string]interface{})["settings"].(map[string]interface{})["index"].(map[string]interface{})["time_series"] = indexTimeSeries
```
jsoriano, line 369: This will panic if a template doesn't already have time series settings. Could we make this code safer?

internal/benchrunner/runners/rally/runner.go
```diff
@@ -274,6 +341,43 @@ func (r *runner) setUp() error {
...
+
+	delete(simulatedTemplate, "overlapping")
```
jsoriano, line 371: Nit. Please add a comment explaining why we delete this field.

internal/benchrunner/runners/rally/runner.go
```diff
@@ -274,6 +341,43 @@ func (r *runner) setUp() error {
...
+
+	newTemplate, err := json.Marshal(simulatedTemplate)
```
jsoriano, line 373: Nit. This will encode html entities, please check if this can be a problem here. If it is an issue here you can use a formatter like this one https://github.com/elastic/elastic-package/blob/main/internal/formatter/json_formatter.go#L65.

internal/benchrunner/runners/rally/runner.go
ruflin: I'm worried we modify here the package instead of just extending it. `metrics-kubernetes.pod@custom` is not managed by the integration package, it is only referenced and the goal is to use it exactly for customisations. 

Will read up on https://github.com/elastic/integrations/issues/8431#issuecomment-1807882262 to better understand what the issue there is.

ruflin: In case we move forward with the proposed approach, could this have any sideffects? For example, the same dataset is used with newer versions of the package? So far, this meant the newer package and all the new templates are installed. But with the solution we have here it means always the old "templates" are installed. Or does it still mostly work, as all the "important" bits are in the components which are referenced?

aspacca: > I'm worried we modify here the package instead of just extending it. `metrics-kubernetes.pod@custom` is not managed by the integration package, it is only referenced and the goal is to use it exactly for customisations.

I looked in the Kibana UI and `metrics-kubernetes.pod@custom` is marked as a managed template.
I will try another time to add the minimum settings required for overriding `time_series.start_time/end_time` in `metrics-kubernetes.pod@custom` and try to understand why it does not work.
From my point of view we are not exactly modifying the package, rather overriding for `metrics-kubernetes.pod-ep` the templates from the package with another separated template that's a copy/merge of all the ones from the package.



> In case we move forward with the proposed approach, could this have any sideffects? For example, the same dataset is used with newer versions of the package? So far, this meant the newer package and all the new templates are installed. But with the solution we have here it means always the old "templates" are installed. Or does it still mostly work, as all the "important" bits are in the components which are referenced?

In case of a new version of the package, the template in the rally track might be obsolete. You should regenerate the track in order to be up to date. If you run the benchmark directly from `elastic-package` the template will be always up to date with the package in the local folder we install as zip.

Note that the above will be true even if we change only `metrics-kubernetes.pod@custom`, as long as soon we have a new package with changes in that template.

I'm indeed a little confused by how `metrics-kubernetes.pod@custom` is supposed to work.
What if we run `elastic-package benchmark rally` on cluster where `metrics-kubernetes.pod@custom` was already (well :)) customized?

with the current solution we will keep the customization in the overriding template.
If we use `metrics-kubernetes.pod@custom` I think we should to the same, adding the settings for time_series.start_time/end_time` and finally cleaning it.
Let me add that cleaning up an "ephemeral" template like the one in the current solution might be easier to mantain that cleaning up `metrics-kubernetes.pod@custom` :)

internal/benchrunner/runners/rally/runner.go
aspacca: AFAIK the simulated template for TSDB will always have time series settings

internal/benchrunner/runners/rally/runner.go
aspacca: @ruflin Indeed I've realized that if we create an index template with the same content in the rally track, we can replay the track directly with rally even without installing the package.
if we use `metrics-kubernetes.pod@custom` before replaying the rally track we should also install the package, that would anyway be true since we'd loose also ingestion pipelines

aspacca: > `metrics-kubernetes.pod@custom` is not managed by the integration package

```
{"component_templates":[{"name":"metrics-kubernetes.pod@custom","component_template":{"template":{"settings":{}},"_meta":{"package":{"name":"kubernetes"},"managed_by":"fleet","managed":true}}}]}
```

ruflin: That `@custom` is marked as managed but is not seems to have some confusion recently. The is an optional template, meaning it is only used if it is there. @kpollich Pinging you here as the confusion around `@custom` has popped up in other palces too.

> @ruflin Indeed I've realized that if we create an index template with the same content in the rally track, we can replay the track directly with rally even without installing the package.

I would assume the template you export is only the one that links to all the component templates, so you still needed the component templates from the package?

kpollich: Thanks for the ping, @ruflin. Taking a look at the [code path](https://github.com/elastic/kibana/blob/2a8a68c1f8d14074d23ca9feef118aa172191bfe/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/install.ts#L400C19-L406) in Fleet, we never removed the logic to create an empty `@custom` component template here. Since `ignore_missing_templates` exists now we should probably remove this. 

I created https://github.com/elastic/kibana/issues/171540 to capture the removal.

aspacca: > I would assume the template you export is only the one that links to all the component templates, so you still needed the component templates from the package?

I simulate the template `metrics-kubernetes.pod` template, that's composed of all the templates from the package, ending up with a single index temple body that's equivalent to the one that would be used for `metrics-kubernetes.pod-ep` data stream if the package would be installed. it will indeed bring as well any customisation made in `metrics-kubernetes.pod@custom` at the time of the simulation, if any. that's the reason why I don't need the component template from the package.

The simulate API is used to "expand" any implicit of default setting, like indeed the one for `index.routing_path` :)

I've pushed a commit with customising instead (well) `metrics-kubernetes.pod@custom`: I still need to run the simulate API for the reason just above.

The clean up in this case will update back `metrics-kubernetes.pod@custom` with the content before it had increasing the time series start/end time slot.

beware that this approach does not work because a validation issue from ES that complains about an empty `index.routing_path` (that's not) on the final merged `metrics-kubernetes.pod` template

aspacca:
@ruflin I've saved locally the https://github.com/elastic/elastic-package/pull/1555/commits/b460ca8a4777cc9f5180e10ffb3bf6d1cf214d60 commit

I will revert it from the current PR since this solution is not working due to a bug in ES and keep for the moment the solution with the copied simulated template

internal/benchrunner/runners/rally/runner.go
ruflin: It is currently what is used by the ES team in rally tracks. Long term TSDB plans to remove this limitation so we can get rid of this hack again.

internal/benchrunner/runners/rally/runner.go
ruflin: How does the discussion we had yesterday around exporting "all" Elasticsearch assets to the rally track change this discussion?

internal/benchrunner/runners/rally/runner.go
```diff
@@ -274,6 +341,47 @@ func (r *runner) setUp() error {
 	return nil
 }
 
+func (r *runner) extractSimulatedTemplate(indexTemplate string) (string, error) {
+	simulateTemplate, err := r.options.ESAPI.Indices.SimulateTemplate(r.options.ESAPI.Indices.SimulateTemplate.WithName(indexTemplate))
```
ruflin, line 348: Now that I played around with this API (https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-simulate-template.html) I have a much better understanding what it does. Basically it means we get the full template / current state. As you describe further down, we loose the support to use `@custom`. 

My long term preference is still having all the assets but this gets us to what we need short term. To make the @custom work, we could either do it with all the asset export we discussed yesterday or that we manually add it to the template again.

internal/benchrunner/runners/rally/runner.go
aspacca: if we export "all" ES assets, we must have a way to regenerate the track without overwriting the corpus, in order to "dump" to a different version of the package

in this case we will export everything and so we can export as well the `metrics-kubernetes.pod@custom` with the custom (well) `time_series.start_date/end_date`

I would rather suggest, at light of the discussion we had yesterday, that we install the package separately, either directly with `elastic-package` (either by the binary itself, or orchestrating `esbench`) having on option to re-play an existing rally track

ruflin: Lets go through the scenarios:

* Running a quick benchmark directly from elastic-package. The approach in this PR works. The ideal would be with @custom but this is not work in ES
* Create rally track with all assets: This template + export of all ingest pipelines which then are loaded by rally would get the job done. The rally track then is for a specific version of the package. We need to find ways to generate the corpus separately
* Corpus only generation: Is this a new command / flag we need?

What I don't like about the above is that this should also work for `stream`, meaning we adjust the date. What is the index template / component template where the time range change could be made and would work today taking the `metrics-aws.sqs` dataset?

Long term goal is that package installation happens through esbench / elastic-package and only thing the rally track needs is to reference a package and it's custom additional bits but we are too far away from this at the moment.

ruflin (approved):
The change LGTM. It will unblock the current generation of rally tracks. 

There is an ongoing discussion on how to improve this further in separating rally track from corpus generation and for rally / esbench to load the package to remove manual work. My assumption is, these can be added in addition / separately.

jsoriano (approved):

PR #1561, opened 2023-11-21
Revert "Ignore error log when a component goes to a degraded state (#1533)"

jsoriano:
This reverts commit f9a1c907d8361318238cb78345e1ea2f3c454676.

Intended for testing, **don't merge**.

I will check with integrations repo that packages using 8.11.0 fail, but the issue is fixed if these versions are increased.


jsoriano:
test integrations

PR #1563, opened 2023-11-22
Add retries to kibana client

jsoriano:
Adding it by now only to kibana client because for elasticsearch client we rely on the Go SDK, that has its own retry mechanisms.

As it is implemented, we can later consider using it in more places where an http client is used now.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/google/go-github/v32 v32.1.0
 	github.com/google/go-querystring v1.1.0
 	github.com/google/uuid v1.4.0
+	github.com/hashicorp/go-retryablehttp v0.7.5
 	github.com/jedib0t/go-pretty v4.3.0+incompatible
 	github.com/magefile/mage v1.15.0
 	github.com/mholt/archiver/v3 v3.5.1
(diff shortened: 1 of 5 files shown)


mrodm (approved):
Nice!
Did you test these changes creating a serverless project?

jsoriano:
> Nice! Did you test these changes creating a serverless project?

No... let me try...

jsoriano:
> Did you test these changes creating a serverless project?

Tested to start and stop serverless four times, and all of them were successful, except one where elastic-agent failed to start due to some authentication error, that seemed unrelated to this change. Running up again in this case started the agent.

I saw some retries querying status endpoints, and the debug logging was different. I have configured it to use the default `slog` logger that fulfils the requirements for `retryablehttp` and uses `log` under the hood as we do everywhere else.

mrodm (approved):
:rocket: 

> Tested to start and stop serverless four times, and all of them were successful, except one where elastic-agent failed to start due to some authentication error, that seemed unrelated to this change. Running up again in this case started the agent.

Thanks for checking this out ! 

> I saw some retries querying status endpoints, and the debug logging was different. I have configured it to use the default slog logger that fulfils the requirements for retryablehttp and uses log under the hood as we do everywhere else.

First usage of `slog` :muscle: At some point I would like to start using more that package

PR #1564, opened 2023-11-24
update validator to allow _ in github owner name

shahzad31:
update to allow _ in github owner name


jsoriano (approved):

jsoriano:
@shahzad31 we may need to update this in Package Spec too https://github.com/elastic/package-spec/blob/fd13526ce63051f28ef25b26208f177fec749af7/spec/integration/manifest.spec.yml#L254

PR #1566, opened 2023-11-27
Record with go-vcr instead of doing our custom recording

jsoriano:
We record requests in some tests so we can reproduce them without having running services.

We use our own framework now, that has some limitations. It can only record GET requests, and
it can only record one response for any request, so it cannot be used for tests where the state changes.

Give a try to https://pkg.go.dev/gopkg.in/dnaeon/go-vcr.v3 for this purpose.

With go-vcr, each session is stored in a file, called "cassette", that stores the sequence of interactions. By
default it matches requests per method and per URL, and each interaction is only matched once, so multiple
requests to the same endpoint can be recorded with different responses. This seems to be enough for
all the cases we have now.


internal/elasticsearch/test/httptest.go
```diff
@@ -25,64 +20,34 @@ import (
...
 	require.NoError(t, err)
 
-	t.Logf("Recording %s in %s", r.URL.Path, path)
-	var recordURL url.URL
-	recordURL.Path = r.URL.Path
-	recordURL.RawQuery = r.URL.RawQuery
-
-	req, err := http.NewRequest(r.Method, recordURL.String(), nil)
+	rec, err := recorder.NewWithOptions(&recorder.Options{
+		CassetteName:       serverDataDir,
```
mrodm, line 36: Previously, all the responses were recorded in a directory, but currently all those responses are written in a single YAML file 

I'd rename this `serverDataDir` to indicate that now it's a file (cassette file). One option:
```suggestion
		CassetteName:       serverDataFile,
```

Same for the kibana/test/http_test.go

jsoriano: Thanks, updated.

mrodm (approved):

PR #1569, opened 2023-11-27
Handle error when finalizing multipart message

jsoriano:
This stop can fail if it cannot write the trailing boundary. Unlikely, but let's handle it to avoid weird errors later.

Renamed also some files in the package to avoid underscores in file names.


kpollich (approved):
Change is simple LGTM 🚀

PR #1575, opened 2023-11-30
fix mode permission on rally track output dir

aspacca:
`os.ModeDir` instead of `os.ModePerm` was set as mode of `os.MkdirAll` with the target of the rally track output dir: this prevented the content to be written in the directory if the directory didn't exist and was created by the command.
This PR fix this.

### DIFF
--- a/internal/benchrunner/runners/rally/runner.go
+++ b/internal/benchrunner/runners/rally/runner.go
@@ -692,7 +692,7 @@ func (r *runner) runGenerator(destDir string) error {
 
 	if r.options.RallyTrackOutputDir != "" {
 		r.persistRallyTrackHandler = func() error {
-			err := os.MkdirAll(r.options.RallyTrackOutputDir, os.ModeDir)
+			err := os.MkdirAll(r.options.RallyTrackOutputDir, os.ModePerm)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


internal/benchrunner/runners/rally/runner.go
```diff
@@ -692,7 +692,7 @@ func (r *runner) runGenerator(destDir string) error {
 
 	if r.options.RallyTrackOutputDir != "" {
 		r.persistRallyTrackHandler = func() error {
-			err := os.MkdirAll(r.options.RallyTrackOutputDir, os.ModeDir)
+			err := os.MkdirAll(r.options.RallyTrackOutputDir, os.ModePerm)
```
jsoriano, line 695: `os.ModePerm` is 0777, do we need everyone to be able to make changes there? Wdyt about explicitly setting 0755?
```suggestion
			err := os.MkdirAll(r.options.RallyTrackOutputDir, 0755)
```

jsoriano (approved):
LGTM, but maybe we can limit the given mode to be less permissive.

jsoriano:
> Do we swallow some of the errors?

Yes, there were some comments about this in the original PR https://github.com/elastic/elastic-package/pull/1522#pullrequestreview-1708030141, it'd be nice to revisit error handling in benchmark runners.

PR #1576, opened 2023-11-30
Benchmark rally add flags

aspacca:
Add `package-from-registry` and `use-corpus-at-path` flags.
See updated docs for usage.

### DIFF
--- a/cmd/benchmark.go
+++ b/cmd/benchmark.go
@@ -228,6 +228,8 @@ func getRallyCommand() *cobra.Command {
 	cmd.Flags().String(cobraext.VariantFlagName, "", cobraext.VariantFlagDescription)
 	cmd.Flags().StringP(cobraext.BenchCorpusRallyTrackOutputDirFlagName, "", "", cobraext.BenchCorpusRallyTrackOutputDirFlagDescription)
 	cmd.Flags().BoolP(cobraext.BenchCorpusRallyDryRunFlagName, "", false, cobraext.BenchCorpusRallyDryRunFlagDescription)
+	cmd.Flags().StringP(cobraext.BenchCorpusRallyUseCorpusAtPathFlagName, "", "", cobraext.BenchCorpusRallyUseCorpusAtPathFlagDescription)
+	cmd.Flags().StringP(cobraext.BenchCorpusRallyPackageFromRegistryFlagName, "", "", cobraext.BenchCorpusRallyPackageFromRegistryFlagDescription)
 
 	return cmd
 }
@@ -260,12 +262,26 @@ func rallyCommandAction(cmd *cobra.Command, args []string) error {
 		return cobraext.FlagParsingError(err, cobraext.BenchCorpusRallyDryRunFlagName)
 	}
 
-	packageRootPath, found, err := packages.FindPackageRoot()
-	if !found {
-		return errors.New("package root not found")
-	}
+	corpusAtPath, err := cmd.Flags().GetString(cobraext.BenchCorpusRallyUseCorpusAtPathFlagName)
 	if err != nil {
-		return fmt.Errorf("locating package root failed: %w", err)
+		return cobraext.FlagParsingError(err, cobraext.BenchCorpusRallyUseCorpusAtPathFlagName)
+	}
+
+	packageFromRegistry, err := cmd.Flags().GetString(cobraext.BenchCorpusRallyPackageFromRegistryFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.BenchCorpusRallyPackageFromRegistryFlagName)
+	}
+
+	var packageRootPath string
+	var found bool
+	if len(packageFromRegistry) == 0 {
+		packageRootPath, found, err = packages.FindPackageRoot()
+		if !found {
+			return errors.New("package root not found")
+		}
+		if err != nil {
+			return fmt.Errorf("locating package root failed: %w", err)
+		}
 	}
 
 	profile, err := cobraext.GetProfileFlag(cmd)
@@ -299,6 +315,8 @@ func rallyCommandAction(cmd *cobra.Command, args []string) error {
 		rally.WithProfile(profile),
 		rally.WithRallyTrackOutputDir(rallyTrackOutputDir),
 		rally.WithRallyDryRun(rallyDryRun),
+		rally.WithRallyPackageFromRegistry(packageFromRegistry),
+		rally.WithRallyCorpusAtPath(corpusAtPath),
 	}
 
 	esMetricsClient, err := initializeESMetricsClient(cmd.Context())
--- a/docs/howto/rally_benchmarking.md
+++ b/docs/howto/rally_benchmarking.md
@@ -289,5 +289,5 @@ You still need to run the command from the root of the local package in order to
 
 ## Running benchmark with the same corpus against different versions of the package
 If you use both the `use-corpus-at-path` flag and the `package-from-registry` flag multiple times, you can ran benchmark with the same corpus against different versions of the same package, just updating the package version in the relevant flag between executions.  
-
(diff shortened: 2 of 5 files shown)


internal/benchrunner/runners/rally/runner.go
```diff
@@ -266,12 +267,29 @@ func (r *runner) setUp() error {
 	}
 	r.scenario = scenario
 
-	err = r.installPackage()
-	if err != nil {
-		return fmt.Errorf("error installing package: %w", err)
+	if len(r.options.PackageFromRegistry) > 0 {
```
jsoriano, line 270: Nit. We could keep the `installPackage` method, and add this new logic there.

internal/benchrunner/runners/rally/runner.go
```diff
@@ -726,6 +771,138 @@ func (r *runner) runGenerator(destDir string) error {
...
+			buffPosition += i + 1
+			count++
+		}
+
+		if err == io.EOF {
+			break
+		}
+	}
+
+	return count, nil
```
jsoriano, line 802: Nit. Wdyt about using an scanner here?
```suggestion
	var count uint64
	scanner := bufio.NewScanner(r)
	for scanner.Scan() {
		count++
	}
	if err := scanner.Err(); err != nil {
		return 0, err
	}
	return count, nil
```

internal/benchrunner/runners/rally/runner.go
```diff
@@ -726,6 +771,138 @@ func (r *runner) runGenerator(destDir string) error {
...
+		return fmt.Errorf("cannot not set permission to rally corpus file: %w", err)
+	}
+
+	existingCorpus, err := os.Open(corpusPath)
+	if err != nil {
+		if err != nil {
+			return fmt.Errorf("error while reading content for the existing rally corpus file: %w", err)
+		}
+
+	}
```
jsoriano, line 822: Redundant if?
```suggestion
	if err != nil {
		return fmt.Errorf("error while reading content for the existing rally corpus file: %w", err)
	}
```

internal/cobraext/flags.go
```diff
@@ -57,7 +57,13 @@ const (
 	BenchCorpusRallyTrackOutputDirFlagDescription = "output dir of the rally track: if present the command will save the generated rally track"
 
 	BenchCorpusRallyDryRunFlagName        = "dry-run"
-	BenchCorpusRallyDryRunFlagDescription = "Do not run rally but just generate the rally track"
+	BenchCorpusRallyDryRunFlagDescription = "do not run rally but just generate the rally track"
+
+	BenchCorpusRallyPackageFromRegistryFlagName        = "package-from-registry"
+	BenchCorpusRallyPackageFromRegistryFlagDescription = "fetch package from registry instead of local directory"
```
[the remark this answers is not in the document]
aspacca: it's in the docs.  but I'll add here! :)

internal/benchrunner/runners/rally/runner.go
aspacca: it would be a very long method that it would be better to split in two different methods anyway, imho :)

internal/benchrunner/runners/rally/runner.go
aspacca: I was looking for the most performing way to count newlines from an `io.Reader` and ended up with this: benchmarks were showing magnitude of higher speed

docs/howto/rally_benchmarking.md
```diff
@@ -279,3 +279,15 @@ esrally --target-hosts='{"defauelt":["%es_cluster_host:es_cluster_port%"]}' --tr
 
 Please refer to [esrally CLI reference](https://esrally.readthedocs.io/en/stable/command_line_reference.html) for more details.
 
+## Replaying an existing corpus
+If the `use-corpus-at-path` flag is used, the corpus will not be generated but rather loaded from the file passed as value of the flag.
+If the `rally-track-output-dir` flag is used alongside the `use-corpus-at-path` flag, the persisted rally track will reference a copy of the existing corpus saved in the directory passed as value of the `rally-track-output-dir` flag.
+
+## Loading a package from registry
+If the `package-from-registry` flag is used, the package installed in Kibana and its assets referenced in the track will be loaded from registry. The format of the flag value is `%packageName%-%packageVersion%`.
```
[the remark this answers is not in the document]
ruflin: Got it, then lets keep it consistent.

docs/howto/rally_benchmarking.md
```diff
@@ -279,3 +279,15 @@ esrally --target-hosts='{"defauelt":["%es_cluster_host:es_cluster_port%"]}' --tr
...
+You still need to run the command from the root of the local package in order to read the benchmark scenario (`benchmark` flag).
```
ruflin, line 288: Does this mean I need to run two commands? I tried the following:

```
elastic-package benchmark rally --benchmark pod-benchmark -v --package-from-registry=kubernetes-1.55.0
```

Can you share a working example that users can just copy/paste?

aspacca: your example is correct: I've reworded the docs, please let me know if it's clearer now

internal/benchrunner/runners/rally/runner.go
jsoriano: Ok, please consider adding a comment mentioning that, to avoid changing this in a future refactor :slightly_smiling_face:

internal/benchrunner/runners/rally/runner.go
aspacca: yes! that's what I did in the end :D

cmd/benchmark.go
```diff
@@ -363,6 +369,31 @@ func rallyCommandAction(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func getPackageNameAndVersion(packageFromRegistry string) (string, string, error) {
+	packageFromRegistryReverse := make([]byte, 0, len(packageFromRegistry))
```
jsoriano, line 373: Why reversing? Package names [cannot contain `-`](https://github.com/elastic/package-spec/blob/dfdfb179900285e071896f364200df23e7e80a36/spec/integration/manifest.spec.yml#L285), and versions [can](https://github.com/elastic/package-spec/blob/dfdfb179900285e071896f364200df23e7e80a36/spec/integration/manifest.spec.yml#L241).

It looks safer to just [`strings.Cut(packageFromRegistry, "-")`](https://pkg.go.dev/strings#Cut).

If we keep this I would like to have it covered by tests.

internal/cobraext/flags.go
```diff
@@ -60,7 +60,7 @@ const (
 	BenchCorpusRallyDryRunFlagDescription = "do not run rally but just generate the rally track"
 
 	BenchCorpusRallyPackageFromRegistryFlagName        = "package-from-registry"
-	BenchCorpusRallyPackageFromRegistryFlagDescription = "fetch package from registry instead of local directory"
+	BenchCorpusRallyPackageFromRegistryFlagDescription = "fetch package from registry instead of local directory, expected format: <package>-<version>"
```
jsoriano, line 63: :+1:

cmd/benchmark.go
```diff
@@ -315,7 +320,8 @@ func rallyCommandAction(cmd *cobra.Command, args []string) error {
 		rally.WithProfile(profile),
 		rally.WithRallyTrackOutputDir(rallyTrackOutputDir),
 		rally.WithRallyDryRun(rallyDryRun),
-		rally.WithRallyPackageFromRegistry(packageFromRegistry),
+		rally.WithRallyPackageName(packageName),
+		rally.WithRallyPackageVersion(packageVersion),
```
jsoriano, line 324: Nit. If they are always expected to be used together maybe we can add them as a single option.
```suggestion
		rally.WithRallyPackageFromRegistry(packageName, packageVersion),
```

jsoriano (approved):

PR #1578, opened 2023-12-01
Refactors for safer handling of strings

jsoriano:
`strings.Cut()` was introduced in Go 1.18 for the common use case of splitting a string
in two parts. In these cases it allows to directly assign each one of the parts to variables
with more meaningful names, avoiding the risk of accessing indexes that may not exist,
and improving readability.

Leverage this function in some applicable uses cases we have here.

### DIFF
--- a/internal/packages/conditions.go
+++ b/internal/packages/conditions.go
@@ -51,14 +51,14 @@ func parsePackageRequirements(keyValuePairs []string) (*packageRequirements, err
 	var pr packageRequirements
 
 	for _, keyPair := range keyValuePairs {
-		s := strings.SplitN(keyPair, "=", 2)
-		if len(s) != 2 {
+		key, value, valid := strings.Cut(keyPair, "=")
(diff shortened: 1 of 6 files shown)


internal/packages/conditions.go
```diff
@@ -73,7 +73,7 @@ func parsePackageRequirements(keyValuePairs []string) (*packageRequirements, err
 			}
 			pr.kibana.version = &withoutPrerelease
 		default:
-			return nil, fmt.Errorf("unknown package requirement: %s", s[0])
+			return nil, fmt.Errorf("unknown package requirement: %s", value)
```
mrodm, line 76: According to the previous error message, it should be set here `key` , no? 

```suggestion
			return nil, fmt.Errorf("unknown package requirement: %s", key)
```

mrodm (commented):
Thanks

mrodm (approved):

PR #1581, opened 2023-12-01
Remove documentation about how to publish packages

mrodm:
This PR removes the documentation about how to publish packages in Package Storage V2.


jsoriano (approved):

sharbuz (commented):
I would avoid deleting the documentation regarding the current approach because we still use Jenkins

Issue #1541, opened 2023-11-08
Add `benchmark stream` command


aspacca:
Similarly to `benchmark rally` command we want to generate `schema-b` documents for a given integrations.
Instead of creating a rally track out of them we will stream them, according to a configurable rate, directly to an ES cluster, using bulk requets

ruflin:
It would be nice, if you can run the command multiple times in parallel, something like:

```
# Inside nginx package
./elastic-package benchmark stream --benchmark access-logs-benchmark -v
./elastic-package benchmark stream --benchmark error-logs-benchmark -v
# Inside apache package
./elastic-package benchmark stream --benchmark access-logs-benchmark -v
```

For a developer working on the UI or testing the UI, this makes it super easy to have multiple docs mixed together.

ruflin:
Different terminal windows is good enough. I mostly want to make sure they do not get into the way each other (some global state or similar).

aspacca:
> some global state or similar

no global state: we won't even save the data to a temporary file so there should not be race condition on that

ruflin:
The discussions in https://github.com/elastic/integrations/issues/8431#issuecomment-1805355442 triggered an additional thought for the the stream command. This should not block any progress on what we have discussed so far but let me share it here. The stream command will allow devs / users to see real time data coming in. Unfortunately, the data is only from the time they started the command. What if we could backfill data quickly for the last X minutes on startup initially. Following example for system metrics:

* User starts the stream command and has set `fillup: -15min` (much better name needed)
* The stream command starts, and fills up the data for the last 15min as fast as possible
* From now on, it ships the configured events every 10 seconds
* Ideally, the two would happen in parallel so real time data is seen immediately and the fill up would happen in the background

Like this, the dev does not have to wait 15min to see data for 15min, but has historical data to build a visualisation / test a query and also has new data constantly coming in.

ruflin:
To make the stream command work without requiring a local source copy of the package, there needs a place where raw packages are available. Left a comment here: https://github.com/elastic/package-spec/issues/351#issuecomment-1833780745

ruflin:
We had recently a discussion, that the `stream` command is already useful without support for the `--package...` flag, meaning it has to be run with the source checked out.

jsoriano:
This command was already added. Closing this, we can open new issues for new features or fixes.

PR #1584, opened 2023-12-05
Benchmark stream command

aspacca:
Similarly to `benchmark rally` command we want to generate `schema-b` documents for a given integrations.
Instead of creating a rally track out of them we will stream them, according to a configurable rate, directly to an ES cluster, using bulk requets

see https://github.com/elastic/elastic-package/issues/1541 for more context

usage (from a package root):
```
elastic-package benchmark stream -v --events-per-period 10 --period-duration 1s
```
or
```
elastic-package benchmark stream -v --events-per-period 10 --period-duration 1s --backfill -15m
```
or
```
elastic-package benchmark stream -v --benchmark container-benchmark --events-per-period 10 --period-duration 1s --backfill -15m
```

flags:
* `--benchmark`: run a specific benchmark, if not present all benchmarks for a packages will be run
* `--backfill`: negative duration to backfill events ingestion for, if not present event will be ingested since now
* `--period-duration`: time between each bulk request
* `--events-per-period`: events on each bulk request
* `--timestamp-field`: field from generator config used for `@timestamp` event's field (default "timestamp": it is required for backfill and overriding `period`settings)
* `--perform-cleanup`: passing this flag will delete documents in the streaming data streams before and after the streaming, as well as uninstalling the integration at the end


ruflin:
> --benchmark: run a specific benchmark, if not present all benchmarks for a packages will be run

For now, it is really nice that it runs just all tracks. I expect eventually we need to have something like "default" tracks but we can put that for later.

> --ticker-duration: time between each bulk request

ticker seems to be very Golang specific. Ideas for alternative names?

I did a quick run of the code, so far all looks good 🎉

aspacca:
> ticker seems to be very Golang specific. Ideas for alternative names?

`--bulk-request-interval`: time between each bulk request
`--events-per-bulk-request`: events on each bulk request

?

ruflin:
Is it possible to configure the stream command to ship data to a cluster not started with elastic-package? I assume the env variables could be adjusted and it would just work 🤔

aspacca:
> I assume the env variables could be adjusted and it would just work 🤔

indeed:

```
ELASTIC_PACKAGE_ELASTICSEARCH_HOST
ELASTIC_PACKAGE_ELASTICSEARCH_PASSWORD
ELASTIC_PACKAGE_ELASTICSEARCH_USERNAME
ELASTIC_PACKAGE_KIBANA_HOST
```

internal/cobraext/flags.go
```diff
@@ -65,6 +65,18 @@ const (
 	BenchCorpusRallyUseCorpusAtPathFlagName        = "use-corpus-at-path"
 	BenchCorpusRallyUseCorpusAtPathFlagDescription = "path of the corpus to use for the benchmark: if present no new corpus will be generated"
 
+	BenchStreamBackFillFlagName        = "backfill"
+	BenchStreamBackFillFlagDescription = "amount of time to ingest events for since starting from now: expressed as a negative duration (ie: -15m)"
```
ruflin, line 69: I was thinking about making this -15m by default as the default when you go to Discover is "last 15 min".

aspacca:
> An alternative config would be `--events-per-second=0.1`. This would ship one event every 10 seconds. Is this more consumable?

it's better because we have only one flag, but it's rather tricky to express period duration where you want resolution of minutes, I would keep the two different flags, and just rename them

ruflin:
> I would keep the two different flags, and just rename them

Ok, lets go with this for now.

We have now quite a list of command and flags, as soon as things settle down a bit more there is an opportunity to look at all flags together and unify / standardise / cleanup where we can.

ruflin:
I did a quick test with the most recent AWS package by running: `elastic-package benchmark stream -v` This uses all the default. Opening Kibana with Logs Explorer shows the following result which is great!

Screenshot 2023-12-07 at 09 29 50

cmd/benchmark.go
```diff
@@ -385,6 +393,121 @@ func getPackageNameAndVersion(packageFromRegistry string) (string, string, error
...
+
+	packageRootPath, found, err := packages.FindPackageRoot()
+	if !found {
+		return errors.New("package root not found")
+	}
+	if err != nil {
+		return fmt.Errorf("locating package root failed: %w", err)
+	}
+
+	profile, err := cobraext.GetProfileFlag(cmd)
```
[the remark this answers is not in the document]
aspacca: it's documented with `elastic-package benchmark -h`

README.md
```diff
@@ -159,6 +159,16 @@ These benchmarks allow you to benchmark an integration corpus with rally.
 
 For details on how to configure rally benchmarks for a package, review the [HOWTO guide](./docs/howto/rally_benchmarking.md).
 
+#### Stream Benchmarks
+
+These benchmarks allow you to benchmark ingesting real time data.
+You can stream data to a remote ES cluster setting the following environment variables:
+
+ELASTIC_PACKAGE_ELASTICSEARCH_HOST=https://my-deployment.es.eu-central-1.aws.foundit.no
+ELASTIC_PACKAGE_ELASTICSEARCH_USERNAME=elastic
```
ruflin, line 168: Do API keys work here? Serverless has mostly API keys

internal/cobraext/flags.go
```diff
@@ -65,6 +65,21 @@ const (
 	BenchCorpusRallyUseCorpusAtPathFlagName        = "use-corpus-at-path"
 	BenchCorpusRallyUseCorpusAtPathFlagDescription = "path of the corpus to use for the benchmark: if present no new corpus will be generated"
 
+	BenchStreamBackFillFlagName        = "backfill"
+	BenchStreamBackFillFlagDescription = "amount of time to ingest events for since starting from now: expressed as a negative duration"
```
ruflin, line 69: The description is very clear, but of course I put in `1m` at first.

internal/cobraext/flags.go
```diff
@@ -65,6 +65,21 @@ const (
...
+
+	BenchStreamEventsPerPeriodFlagName        = "events-per-period"
+	BenchStreamEventsPerPeriodFlagDescription = "number of events to ingest at every ingestion cycle"
+
+	BenchStreamPeriodDurationFlagName        = "period-duration"
+	BenchStreamPeriodDurationFlagDescription = "duration of the period between each ingestion cycle: expressed as a positive duration"
+
+	BenchStreamPerformCleanupFlagName        = "perform-cleanup"
```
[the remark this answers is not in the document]
ruflin: On the fence about this one. As now `perform-cleanup` is not the default anymore, I think it is less of an issue. Lets see how it is used and come back to this but leave it for now.

internal/benchrunner/runners/stream/scenario.go
```diff
@@ -0,0 +1,114 @@
...
+
+type dataStream struct {
+	Name string `config:"name" json:"name"`
+}
+
+type corpora struct {
+	Generator *generator `config:"generator" json:"generator"`
+}
+
+type generator struct {
```
ruflin, line 35: This seems to be almost the same as https://github.com/elastic/elastic-package/blob/main/internal/benchrunner/runners/rally/scenario.go, same for other objects. Is the duplication on purpose?

ruflin (approved):
Code LGTM

As a follow up, I think there is some potential to refactor / cleanup but we can take this separate.

README.md
jsoriano: Yes, this is missing, please create an issue.

cmd/benchmark.go
```diff
@@ -385,6 +399,129 @@ func getPackageNameAndVersion(packageFromRegistry string) (string, string, error
...
+
+	variant, err := cmd.Flags().GetString(cobraext.VariantFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.VariantFlagName)
+	}
+
+	benchName, err := cmd.Flags().GetString(cobraext.BenchNameFlagName)
+	if err != nil {
+		benchName = ""
+	}
```
jsoriano, line 433: Is there any reason to ignore the error here?

internal/benchrunner/runners/stream/scenario.go
jsoriano: +1, we should probably try to reduce duplication, there is quite a lot between benchmark runners. This would also help to evaluate better how much logic to maintain we are adding with each runner.

internal/cobraext/flags.go
jsoriano: It is indeed a bit weird to have a parameter that only accepts negative numbers. Could we revert it, so it accepts only positive numbers and we negate it in code?

internal/benchrunner/runners/stream/runner.go
```diff
@@ -0,0 +1,633 @@
...
+}
+
+func (r *runner) TearDown() error {
+	r.wg.Wait()
+
+	if !r.options.PerformCleanup {
+		r.removePackageHandler = nil
+		r.wipeDataStreamHandler = nil
+		return nil
+	}
```
jsoriano, line 78: I think we don't have optional cleanup in other benchmark or test runners, and this has been indeed the topic of some discussions, see https://github.com/elastic/elastic-package/issues/582

Do we need this option for some particular reason on these benchmarks?

jsoriano (commented):
Mostly looks good to me. Added a comment about ignoring the error in one flag, and some other questions that would not be blockers before merging.

internal/benchrunner/runners/stream/scenario.go
aspacca: yes, let's plan for another PR moving repeated code to some common package under benchmark

cmd/benchmark.go
aspacca: yes, because the flag is optional and we would like that consumer of the command to have as much less barrier as possible to run it with the default settings

internal/benchrunner/runners/stream/runner.go
ruflin: I pushed for this to be optional. The reason is that the `stream` command is used slightly different from other commands. A user can ship data to an existing cluster they have and will ship some data to do some testing, stop the shipping of data and then might continue again or ship some other data. If the cleanup happens, it means the user has not chance now to look at the data after stopping. It is what happened to me when I first tested. Quickly started the shipping and then wanted to look at the data without further data coming in. I stopped the shipping and suddenly my data was gone (including the package). It is also possible that for example for system, the user has some metrics already shipping but wants to add more dimension / data with the stream command.

jsoriano (commented):
I think we need to handle the error in the flag, as we do with all other flags. For the rest it LGTM.

aspacca:
@jsoriano 

> The error here can happen if the flag is not defined or if there is some type conflict, it should not happen if the user does not provide the flag. We should report the error if it happens.

I don't want to return an error if there is some type conflict, I want the consumer of the command to be able to see the command running with the default value without any error. :)

but we can warn the user with a log entry :)

aspacca:
> @aspacca Is there a way we could have tests for these features

@ruflin we can have some end2end tests: I will add in a separate PR

cmd/benchmark.go
```diff
@@ -385,6 +399,129 @@ func getPackageNameAndVersion(packageFromRegistry string) (string, string, error
...
+	cmd.Println("Run stream benchmarks for the package")
+
+	variant, err := cmd.Flags().GetString(cobraext.VariantFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.VariantFlagName)
+	}
+
+	benchName, err := cmd.Flags().GetString(cobraext.BenchNameFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.VariantFlagName)
```
jsoriano, line 432: Oh, my suggestion was incorrect, sorry :facepalm:
```suggestion
		return cobraext.FlagParsingError(err, cobraext.BenchNameFlagName)
```

jsoriano (approved):

Issue #1306, opened 2023-06-13
Support for Docker Compose V2 + Compose V1 Nearing End of Life


nicpenning:
I have been testing the elastic-package tool inside of Windows Subsystem for Linux (WSL2) and have used the latest Docker installation that uses Compose V2 which is `docker compose`. This in turn then does not work with elastic-package as it expects `docker-compose`

```
napsta@TESTLL:~$ elastic-package stack up -d -v
2023/06/13 10:43:32 DEBUG Enable verbose logging
Boot up the Elastic stack
Using profile /home/napsta/.elastic-package/profiles/default.
Remember to load stack environment variables using 'eval "$(elastic-package stack shellinit)"'.
Elasticsearch host: https://127.0.0.1:9200
Kibana host: https://127.0.0.1:5601
Username: elastic
Password: changeme
Local package-registry will serve packages from these sources:
- Proxy to https://epr.elastic.co
2023/06/13 10:43:33 DEBUG running command: docker-compose version --short
2023/06/13 10:43:33 ERROR Unable to determine Docker Compose version: running Docker Compose version command failed: exec: "docker-compose": executable file not found in $PATH. Defaulting to 1.x
2023/06/13 10:43:33 DEBUG running command: docker-compose -f /home/napsta/.elastic-package/profiles/default/stack/snapshot.yml -p elastic-package-stack build
Error: booting up the stack failed: building docker images failed: running command failed: running Docker Compose build command failed: exec: "docker-compose": executable file not found in $PATH
```

Here are the documents to migrating to Compose V2. It seems like V1 (docker-compose) is End of Life June 2023 (this month) which means it will no longer receive updates.

https://docs.docker.com/compose/migrate/

Is this something in the works or being considered?

adam-stokes:
I think I'm running into this issue now trying to get elastic-package running on ubuntu 22.04, even with docker-compose package installed.  

```
ubuntu@ogc-layout-8zv8obfc:~$ sudo docker version
Client: Docker Engine - Community
 Version:           24.0.7
 API version:       1.43
 Go version:        go1.20.10
 Git commit:        afdd53b
 Built:             Thu Oct 26 09:07:41 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          24.0.7
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.10
  Git commit:       311b9ff
[...]

PR #1592, opened 2023-12-12
Use docker subcommand when available

jsoriano:
Recent versions of the Docker CLI include docker compose. Use it when available instead of standalone `docker-compose`.

The environment variable `ELASTIC_PACKAGE_COMPOSE_ENABLE_STANDALONE=true` can be used to force the use of standalone `docker-compose`.

Fixes https://github.com/elastic/elastic-package/issues/1306


jsoriano:
test integrations

mrodm (approved):
LGTM

That means that could it be removed docker-compose from the Buildkite pipelines of the [integrations repo](https://github.com/elastic/integrations) too ?

jsoriano:
> That means that could it be removed docker-compose from the Buildkite pipelines of the [integrations repo](https://github.com/elastic/integrations) too ?

In principle yes!

PR #1593, opened 2023-12-12
add config to enable apm-server

graphaelli:
Start and expose an APM Server under Elastic Agent when enabled in the config with:

```yml
stack.apm_server_enabled: true
```

### DIFF
--- a/internal/profile/_static/config.yml.example
+++ b/internal/profile/_static/config.yml.example
@@ -10,6 +10,10 @@
 # Region where the Serverless project is going to be created
 # stack.serverless.region: aws-us-east-1
 
+## Enable apm-server
+# Flag to enable apm-server in elastic-package stack profile config
+# stack.apm_server_enabled: true
+
 ## Enable logstash for testing
 # Flag to enable logstash in elastic-package stack profile config
 # stack.logstash_enabled: true
--- a/internal/stack/_static/docker-compose-stack.yml.tmpl
+++ b/internal/stack/_static/docker-compose-stack.yml.tmpl
@@ -112,6 +112,10 @@ services:
       - "../certs/fleet-server:/etc/ssl/elastic-agent"
     ports:
       - "127.0.0.1:8220:8220"
+      {{ $apm_server_enabled := fact "apm_server_enabled" }}
+      {{ if eq $apm_server_enabled "true" }}
+      - "127.0.0.1:8200:8200"
+      {{ end }}
 
   fleet-server_is_ready:
     image: tianon/true
--- a/scripts/test-stack-command.sh
+++ b/scripts/test-stack-command.sh
@@ -3,6 +3,7 @@
 set -euxo pipefail
 
 VERSION=${1:-default}
+APM_SERVER_ENABLED=${APM_SERVER_ENABLED:-false}
 
 cleanup() {
   r=$?
@@ -12,6 +13,11 @@ cleanup() {
 
   # Take down the stack
   elastic-package stack down -v
+
+  if [ "${APM_SERVER_ENABLED}" = true ]; then
+    # Create an apm-server profile and use it
+    elastic-package profiles delete with-apm-server
+  fi
 
   exit $r
 }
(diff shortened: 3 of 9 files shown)


internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -112,6 +112,10 @@ services:
       - "../certs/fleet-server:/etc/ssl/elastic-agent"
     ports:
       - "127.0.0.1:8220:8220"
+      {{ $apm_server_enabled := fact "apm_server_enabled" }}
+      {{ if eq $apm_server_enabled "true" }}
```
ChrsMark, line 116: This should be exposed from `elastic-agent` container not `fleet-server` right? APM should run in the same container as Agent, or I miss sth here?

internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -112,6 +112,10 @@ services:
       - "../certs/fleet-server:/etc/ssl/elastic-agent"
     ports:
```
ChrsMark, line 113: Would also need a section to expose the port from the container as well:

```yaml
expose:
- "8200"
```

internal/stack/_static/docker-compose-stack.yml.tmpl
graphaelli: this is an elastic-agent container too.  this ended up with the following containers:

```
a6d36785b0be   docker.elastic.co/elastic-agent/elastic-agent-complete:8.10.1   "/usr/bin/tini -- /u…"   20 minutes ago   Up 19 minutes (healthy)                                                          elastic-package-stack-apm-elastic-agent-1
c237ccac4ba7   docker.elastic.co/elastic-agent/elastic-agent-complete:8.10.1   "/usr/bin/tini -- /u…"   20 minutes ago   Up 20 minutes (healthy)     127.0.0.1:8200->8200/tcp, 127.0.0.1:8220->8220/tcp   elastic-package-stack-apm-fleet-server-1

```

ChrsMark: Ah ok I see, it makes sense since it will start either way.

internal/stack/_static/docker-compose-stack.yml.tmpl
ChrsMark: I tested it and it seems that it works even without this.

internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -112,6 +112,10 @@ services:
...
       - "127.0.0.1:8220:8220"
+      {{ $apm_server_enabled := fact "apm_server_enabled" }}
+      {{ if eq $apm_server_enabled "true" }}
+      - "127.0.0.1:8200:8200"
+      {{ end }}
```
jsoriano, line 118: Good idea to add this to the Fleet Server agent :+1: maybe at some point we can keep it always enabled and/or reuse it for actual performance monitoring of the testing framework.

Another option would be to create another agent, but this will probably bring unnecessary complexity.

internal/profile/_static/config.yml.example
```diff
@@ -10,6 +10,10 @@
 # Region where the Serverless project is going to be created
 # stack.serverless.region: aws-us-east-1
 
+## Enable apm-server
+# Flag to enable apm-server in elastic-package stack profile config
+# stack.apm_server_enabled: true
```
jsoriano, line 15: Could you add some test that leverages this option? It would be probably enough something like `scripts/test-stack-command.sh`, but with this setting enabled, and maybe checking that the APM server is listening.

The steps to add an integration test for this, would be:
- Add a script to `scripts`, or parameterize one of the available ones.
- Add a Makefile target that calls the script.
- Add a CI stage in `.buildkite/pipeline.trigger.integration.tests.sh`.

jsoriano (commented):
Looks good, nice approach to have an apm-server running here.

Could you also add an integration test?

internal/profile/_static/config.yml.example
graphaelli: added a test that ensures APM Server is listening, thanks for the clear instructions

scripts/test-stack-command.sh
```diff
@@ -34,7 +40,22 @@ if [ "${VERSION}" != "default" ]; then
   EXPECTED_VERSION=${VERSION}
 fi
 
+if [ "${APM_SERVER_ENABLED}" = true ]; then
+  # Create an apm-server profile and use it
+  profile=with-apm-server
+  elastic-package profiles create -v ${profile}
+  elastic-package profiles use ${profile}
```
jsoriano, line 47: Nice idea to use profiles for this :+1:

jsoriano (approved):
Great, thanks! :+1:

PR #1598, opened 2023-12-14
Enable APM in services managed by elastic-package stack

jsoriano:
Leveraging https://github.com/elastic/elastic-package/pull/1593 (thanks @graphaelli!), configure APM in the services managed by `elastic-package stack` locally with Docker Compose. By now this is only optionally enabled, when the APM server is enabled.

This allows to use APM UI for these services:
image

cc @aspacca @ruflin in case this can be useful in benchmarking scenarios.

### DIFF
--- a/internal/stack/_static/elasticsearch.yml.tmpl
+++ b/internal/stack/_static/elasticsearch.yml.tmpl
@@ -22,8 +22,8 @@ script.context.processor_conditional.cache_max_size: 2000
 script.context.template.cache_max_size: 2000
 {{ end }}
 
-{{ $apm_server_enabled := fact "apm_server_enabled" }}
-{{ if eq $apm_server_enabled "true" }}
+{{ $apm_enabled := fact "apm_enabled" }}
+{{ if eq $apm_enabled "true" }}
 tracing.apm.enabled: true
 tracing.apm.agent.server_url: "http://fleet-server:8200"
 tracing.apm.agent.environment: "dev"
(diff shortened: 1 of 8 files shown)


ruflin:
This is great!

> cc @aspacca @ruflin in case this can be useful in benchmarking scenarios.

@jsoriano Are you thinking about getting more benchmark data? It would also great to be able to generate data for the apm package but I think the team has a separate data generation tool (@axw ).

ruflin:
> I wonder if this can be useful in benchmarking scenarios, for example when benchmarking pipelines, but not sure yet about what information we have available.

I'll definitely take a look at it!

internal/stack/_static/elasticsearch.yml.tmpl
```diff
@@ -21,3 +21,10 @@ script.context.ingest.cache_max_size: 2000
 script.context.processor_conditional.cache_max_size: 2000
 script.context.template.cache_max_size: 2000
 {{ end }}
+
+{{ $apm_server_enabled := fact "apm_server_enabled" }}
+{{ if eq $apm_server_enabled "true" }}
+tracing.apm.enabled: true
```
graphaelli, line 27: I like the idea of always turning apm instrumentation on when enabling apm-server - how about we rename the config to `apm_enabled` to reflect that?

jsoriano: cc @ChrsMark who wanted to use this.

ChrsMark: My use-case is shipping OTel data to that local APM Server. So once this is covered no strong opinion on how this enablement config will be called.

graphaelli (approved):

jillguyonnet (approved):

axw:
> Are you thinking about getting more benchmark data? It would also great to be able to generate data for the apm package but I think the team has a separate data generation tool (@axw ).

Indeed we do. The tool receives APM data (intake-v2), converts it into `_bulk` requests, and generates a Rally track. We have some pre-canned APM data for this -- we could always add to it for some more diversity, but so far there hasn't been a need.

PR #1606, opened 2023-12-19
[Buildkite] Revamp buildkite scripts

mrodm:
Update buildkite scripts 

Tested:
- [x] PR pipeline
- [x] Pipeline test with integrations (https://buildkite.com/elastic/elastic-package-test-with-integrations) - [First test](https://buildkite.com/elastic/elastic-package-test-with-integrations/builds/82) - [Second test](https://buildkite.com/elastic/elastic-package-test-with-integrations/builds/84)

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -3,7 +3,8 @@
 source .buildkite/scripts/tooling.sh
 set -euo pipefail
 
-export GO_VERSION=$(cat .go-version)
+GO_VERSION=$(cat .go-version)
+export GO_VERSION
 
 GCP_SERVICE_ACCOUNT_SECRET_PATH=secret/ci/elastic-elastic-package/gcp-service-account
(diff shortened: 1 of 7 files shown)


mrodm:
test integrations

mrodm:
test integrations

mrodm:
test integrations

sharbuz (approved):
Looks good. Thank you @mrodm !
I only have one proposal: If we are going to use the common library of scripts I would start using universal (depending on the environment architecture) functions for installing some kind of tools where it's possible to avoid hardcoded "Linux", "x86_64" and so forth...

mrodm:
> I only have one proposal: If we are going to use the common library of scripts I would start using universal (depending on the environment architecture) functions for installing some kind of tools where it's possible to avoid hardcoded "Linux", "x86_64" and so forth...

I'll create a follow PR for this @sharbuz ,thanks for the heads up!

PR #1608, opened 2023-12-19
[Buildkite] Update buildkite scripts

mrodm:
Update buildkite scripts to get the platform and architecture in runtime.

Tested:
- [x] Pipeline to test with integrations - https://buildkite.com/elastic/elastic-package-test-with-integrations/builds/85

### DIFF
--- a/.buildkite/scripts/install_deps.sh
+++ b/.buildkite/scripts/install_deps.sh
@@ -4,6 +4,27 @@ source .buildkite/scripts/tooling.sh
 
 set -euo pipefail
 
+platform_type="$(uname)"
+hw_type="$(uname -m)"
+platform_type_lowercase="${platform_type,,}"
+
+check_platform_architecture() {
+  case "${hw_type}" in
+    "x86_64")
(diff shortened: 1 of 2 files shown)


mrodm:
test integrations

.buildkite/scripts/install_deps.sh
```diff
@@ -4,6 +4,27 @@ source .buildkite/scripts/tooling.sh
 
 set -euo pipefail
 
+platform_type="$(uname)"
+hw_type="$(uname -m)"
+platform_type_lowercase="${platform_type,,}"
```
jsoriano, line 9: Nit. It looks a bit unexpected to me to call platform_type to the operating system. And `platform_type` is never used, so maybe we can reserve this name to the final form that we use.
```suggestion
uname_os_type="$(uname)"
hardware_name="$(uname -m)"
os_type="${uname_os_type,,}"
```

jsoriano (approved):

sharbuz (approved):

PR #1609, opened 2023-12-20
Fix runner shutdown, improve error message

ruflin:
There are two fixes in this PR. The first one is around the runner being stuck in case of errors. Because a channel is used for the errors, in case 2 errors happen in a row before the error is consumed, the go routine gets stuck. As the effect of an error is stopping the generation anyways, now on error the progress on function is aborted with return. As the errors are mostly used for logging, it raises the question if the channels are needed in the first place but further cleanup is out of scope of this PR.

The second change is around error logging for scenarios where the template generates invalid JSON. For easier debugging, it now prints out the full invalid JSON event as a log event and gives an indication on where the problem lies.

### DIFF
--- a/internal/benchrunner/runners/stream/runner.go
+++ b/internal/benchrunner/runners/stream/runner.go
@@ -430,7 +430,9 @@ func (r *runner) collectBulkRequestBody(indexName, scenarioName string, buf *byt
 	var event map[string]any
 	err = json.Unmarshal(buf.Bytes(), &event)
 	if err != nil {
-		return bulkBodyBuilder, err
+		return bulkBodyBuilder, fmt.Errorf("problem unmarshaling json event: %s\n"+
+			"Outcome of template for scenario '%s' is invalid JSON. Check your benchmark template.",
+			buf.String(), scenarioName)
 	}
 	enriched := r.enrichEventWithBenchmarkMetadata(event)
 	src, err := json.Marshal(enriched)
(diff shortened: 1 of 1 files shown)


internal/benchrunner/runners/stream/runner.go
```diff
@@ -430,7 +430,9 @@ func (r *runner) collectBulkRequestBody(indexName, scenarioName string, buf *byt
 	var event map[string]any
 	err = json.Unmarshal(buf.Bytes(), &event)
 	if err != nil {
-		return bulkBodyBuilder, err
+		return bulkBodyBuilder, fmt.Errorf("problem unmarshaling json event: %s\n"+
+			"Outcome of template for scenario '%s' is invalid JSON. Check your benchmark template.",
+			buf.String(), scenarioName)
```
jsoriano, line 435: Linter doesn't like error strings ended with punctuation.
```
internal/benchrunner/runners/stream/runner.go:433:27: error strings should not end with punctuation or newlines (ST1005)
```
This could be changed to something like this, logging the failed document at the debug level, and including the original error too:
```suggestion
		logger.Debugf("problem found when unmarshalling document: %s", buf.String())
		return bulkBodyBuilder, fmt.Errorf("failed to unmarshal json event, check your benchmark template for scenario %s: %w", scenarioName, err)
```

ruflin:
Thanks @jsoriano for the proposal. Lets get this in as is for now. The only downside is that now a dev has to enable debug logging during development. But in the end, the problem should not be solved here but with https://github.com/elastic/elastic-package/issues/1610

jsoriano (approved):
Thanks!

Issue #1595, opened 2023-12-13
Add SonarQube generic-test-data format support for code coverage report.


sharbuz:
Currently, code coverage reports are analyzed in the [Jenkins Cobertura plugin](https://plugins.jenkins.io/cobertura/).
Jenkins will be decommissioned soon and it was decided to use Elastic SonarQube for code coverage report analysis.
Elastic SonarQube can't analyze the reports made by the elastic-package tool because the SonarQube Java Cobertura coverage reports [haven't been supported](https://community.sonarsource.com/t/cobertura-generated-coverage-xml-file-is-not-loading-so-sonar-qube-shows-0-0-coverage/42654/5) and it requires using [generic-test-data format](https://docs.sonarsource.com/sonarqube/latest/analyzing-source-code/test-coverage/generic-test-data/).
The elastic-package tool has to be updated to support the new format of the reports.
After Jenkins decommissions the support of the Java Cobertura coverage reports should be decommissioned.

Pipelines from [integrations repository](https://github.com/elastic/integrations) should be updated to enable these coverage reports too.

Example from integrations repository:
https://github.com/elastic/integrations/pull/3671#issuecomment-1179131728

Current docs about coverage: https://github.com/elastic/integrations/blob/6c808f49d404ac9f29451297a3a991e6026e7fe0/docs/testing_and_validation.md#review-test-coverage

mrodm:
In a build triggered from this PR https://github.com/elastic/integrations/pull/8464
there was an error related that the files were not found in the working copy ([build link](https://buildkite.com/elastic/integrations/builds/7289#018c59c6-d02d-42b7-91a4-54e3d32fdf3b/106-284)):

```
INFO: Sensor Cobertura Sensor for Python coverage [python]
INFO: Python test coverage
INFO: Parsing report '/buildkite/builds/bk-agent-prod-k8s-1702313828449526795/elastic/integrations/build/test-coverage/coverage-report/coverage.xml'
ERROR: Cannot resolve the file path 'elastic_package_registry/metrics' of the coverage report, the file does not exist in all 'source'.
```
Example of XML generated.

```xml
<coverage line-rate="0" branch-rate="0" version="" timestamp="1702313791649554954" lines-covered="0"
    lines-valid="0" branches-covered="0" branches-valid="0" complexity="0">
    <sources />
    <packages>
        <package name="elastic_package_registry" line-rate="0" branch-rate="0" complexity="0">
            <classes>
[...]

PR #1612, opened 2023-12-22
Update paths coverage

mrodm:
Relates https://github.com/elastic/elastic-package/issues/1595


internal/testrunner/coverageoutput.go
mrodm, line 1: This file has been renamed to be `coveragereport.go` and part of its logic related to the Cobertura format has been moved to its own file `coberturacoverage.go`

internal/packages/assets.go
```diff
@@ -153,6 +153,8 @@ func loadElasticsearchAssets(pkgRootPath string) ([]Asset, error) {
 		}
 	}
 
+	// TODO add assets for input packages
```
mrodm, line 156: Created issue for this: https://github.com/elastic/elastic-package/issues/1623

internal/testrunner/coberturacoverage.go
```diff
@@ -0,0 +1,259 @@
...
+
+// CoberturaMethod represents a method in a Cobertura XML report.
+type CoberturaMethod struct {
+	Name       string           `xml:"name,attr"`
+	Signature  string           `xml:"signature,attr"`
+	LineRate   float32          `xml:"line-rate,attr"`
+	BranchRate float32          `xml:"branch-rate,attr"`
+	Complexity float32          `xml:"complexity,attr"`
+	Lines      []*CoberturaLine `xml:"lines>line"`
+}
```
mrodm, line 66: Removed Hits field from this struct since it is not present in the DTD

jsoriano (approved):
:+1:

internal/testrunner/coveragereport.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	err = writeCoverageReportFile(report, packageName, string(testType))
+	if err != nil {
+		return fmt.Errorf("can't write test coverage report file: %w", err)
+	}
+	return nil
+}
+
+func createCoverageReport(packageRootPath, packageName, packageType string, testType TestType, results []TestResult, CoverageFormat string, timestamp int64) (CoverageReport, error) {
```
jsoriano, line 92: Nit.
```suggestion
func createCoverageReport(packageRootPath, packageName, packageType string, testType TestType, results []TestResult, coverageFormat string, timestamp int64) (CoverageReport, error) {
```

cmd/testrunner.go
```diff
@@ -127,6 +128,11 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 			return cobraext.FlagParsingError(err, cobraext.TestCoverageFlagName)
 		}
 
+		testCoverageType, err := cmd.Flags().GetString(cobraext.TestCoverageFormatFlagName)
```
jsoriano, line 131: At some point we should check if the given value is one of the allowed ones.

cmd/testrunner.go
```diff
@@ -71,6 +71,7 @@ func setupTestCommand() *cobraext.Command {
 	cmd.PersistentFlags().StringP(cobraext.ReportFormatFlagName, "", string(formats.ReportFormatHuman), cobraext.ReportFormatFlagDescription)
 	cmd.PersistentFlags().StringP(cobraext.ReportOutputFlagName, "", string(outputs.ReportOutputSTDOUT), cobraext.ReportOutputFlagDescription)
 	cmd.PersistentFlags().BoolP(cobraext.TestCoverageFlagName, "", false, cobraext.TestCoverageFlagDescription)
+	cmd.PersistentFlags().StringP(cobraext.TestCoverageFormatFlagName, "", "cobertura", cobraext.TestCoverageFormatFlagDescription)
```
jsoriano, line 74: Do we need to add the flag in the buildkite pipelines?

jsoriano: This will be added to integrations, not here.

cmd/testrunner.go
```diff
@@ -148,6 +154,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 		signal.Enable()
 
 		var testFolders []testrunner.TestFolder
+		// var dataStreamWithoutTests []testrunner.TestFolder
```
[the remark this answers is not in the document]
mrodm: Yes, I'll remove it, thanks!

internal/testrunner/coberturacoverage.go
mrodm: I just checked the Jenkins plugin code , looking at the code it does not look like it was used. And their tests do not include that field neither 

https://github.com/jenkinsci/cobertura-plugin/blob/ff0973461f60c28dbc0bf5ab4853a5aec4798210/src/test/resources/hudson/plugins/cobertura/coverage-with-data.xml

internal/testrunner/coveragereport.go
mrodm: Updated parameter and its usages

cmd/testrunner.go
mrodm: Added check about this 

Example of the error
```shell
 $ elastic-package test asset -v --test-coverage --coverage-format=gen
2024/01/18 11:18:11 DEBUG Enable verbose logging
2024/01/18 11:18:11 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
Run asset tests for the package
Error: error parsing --coverage-format flag: coverage format not available: gen

```

cmd/testrunner.go
mrodm: Exactly, this would be added in the integration.

In this PR (not ready at this moment), it has been added this new flag parameter:
https://github.com/elastic/integrations/pull/8787/files#diff-68c0021e761ebe6b8a42d32dfdf3dda2200f0f2fee50106c85c9aea530ccf3c3R19 

Probably, it would be good to update also here (elastic-package) the BK pipelines to generate the artifacts in the same coverage format as in the integrations repository. WDYT @jsoriano ?

jsoriano: Ok, as you prefer.

jsoriano (approved):

Issue #1534, opened 2023-11-03
Add support for Logstash service in system benchmarking


ebeahan:
With the Logstash service added in #1458, system benchmarking should also support using the logstash-output in the benchmarking agent policies.

Like done for [system testing](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/runner.go#L559), if `stack.logstash_enabled: true` is set in the current profile's config use the logstash output over the default.

ebeahan:
Also handing the ingest node stats will be different. Unsure what metrics/stats are exposed in Logstash compared to ES ingest node stats.

bhapas:
#1615 adds support for logstash in system benchmarking. Logstash does not support metrics similar to ingest node yet. Hence closing this issue.

bhapas:
Issue created on logstash elastic integration filter plugin to resolve this. https://github.com/elastic/logstash-filter-elastic_integration/issues/119

PR #1615, opened 2024-01-04
Support system benchmarking with the logstash-output in the benchmarking agent policies

bhapas:
Support system benchmarking with the logstash-output in the benchmarking agent policies.

If `stack.logstash_enabled: true` is set in the current profile's config use the logstash output over the default.

- Relates #1534


internal/benchrunner/runners/system/runner.go
```diff
@@ -362,6 +362,11 @@ func (r *runner) createBenchmarkPolicy(pkgManifest *packages.PackageManifest) (*
 		MonitoringEnabled: []string{"logs", "metrics"},
 	}
 
+	// Assign the data_output_id to the agent policy to configure the output to logstash. The value is inferred from stack/_static/kibana.yml.tmpl
+	if r.options.Profile.Config("stack.logstash_enabled", "false") == "true" {
+		p.DataOutputID = "fleet-logstash-output"
```
jsoriano, line 367: Could we add a test for this case?

jsoriano (approved):

internal/benchrunner/runners/system/runner.go
bhapas: @jsoriano You mean a test package?

jsoriano: It would be a test that runs benchmarks with logstash, it can be a new package or a new script and an existing package, not sure what would be the best approach.

bhapas: I can add a regression test. As currently logstash does not provide relevant stats that are comparable to ingest node stats. But I just want to make we can run system benchmarks even with logstash enabled.

bhapas: @jsoriano Added a test package

scripts/test-check-packages.sh
```diff
@@ -102,6 +102,8 @@ for d in test/packages/${PACKAGE_TEST_TYPE:-other}/${PACKAGE_UNDER_TEST:-*}/; do
       if [ "${package_to_test}" == "system_benchmark" ]; then
         elastic-package benchmark system --benchmark logs-benchmark -v --defer-cleanup 1s
       fi
+    elif [ "${PACKAGE_TEST_TYPE:-other}" == "with-logstash" ] && [ "${PACKAGE_UNDER_TEST:-*}" == "system_benchmark" ]; then
+        elastic-package benchmark system --benchmark logs-benchmark -v --defer-cleanup 1s
```
jsoriano, line 106: Can we somehow check in the results that logstash has been used?

jsoriano (approved):
Thanks for adding a test case!

PR #1617, opened 2024-01-09
[Buildkite] Remove pipeline to test package publication

mrodm:
This PR removes the pipeline "[elastic-package-package-storage-publish](https://buildkite.com/elastic/elastic-package-package-storage-publish/)". Expected to fail this pipeline since in "main" it is still configured to be triggered.

The tasks performed in this pipeline are already checked in https://buildkite.com/elastic/package-storage-infra (testing building package a publishing in dry run mode). Moreover, in the [main elastic-package pipeline](https://buildkite.com/elastic/elastic-package) `elastic-package build` and `elastic-package check` commands are also validated/tested.

As it is not used anymore, scripts to trigger new builds in Jenkins jobs are also deleted.

This PR also updates the scripts to fix some linting warnings.


jsoriano (approved):
:+1:

sharbuz (approved):
LGTM!
Only one thing.
Probably it would be nice to have one approach with using
```
    pushd "folder" > /dev/null
    do_something
    popd > /dev/null
```
instead of using:
```
  (
    cd folder
    do_something
  )
```
@mrodm WDYT?

mrodm:
> LGTM! Only one thing. Probably it would be nice to have one approach with using
> 
> ```
>     pushd "folder" > /dev/null
>     do_something
>     popd > /dev/null
> ```
> 
> instead of using:
> 
> ```
>   (
>     cd folder
>     do_something
>   )
> ```
> 
> @mrodm WDYT?

That's right, both approaches would do the same, and that's why I'd prefer not to change it for now.

test/packages/package-storage/package_storage_candidate/changelog.yml
mrodm, line 1: After removing the pipeline, this test package is just used in:
- `scripts/test-build.zip.sh`
- `scripts/test-install-zip.sh` 

For instance here
https://github.com/elastic/elastic-package/blob/85d6fcacad736e543e459a044a5e0fa48b5d43c6/scripts/test-build-zip.sh#L39-L48

I don't see any other usage of this test package in the [Makefile](https://github.com/elastic/elastic-package/blob/85d6fcacad736e543e459a044a5e0fa48b5d43c6/Makefile)

@jsoriano Do you think it could be used this test package in some other test (unit testing in some loop?)?

mrodm: I'm thinking now that maybe we could keep this test package as a package with the minimum files (changelog, manifest, readme and some imgs). This would be like a baseline to test if build/install fail or not.

jsoriano: Sounds good, yes.

PR #1625, opened 2024-01-17
[Benchmark] Add bulk request response body in case of errors

ruflin:
Currently it is tricky to debug issues in case a bulk request fails as the only item returned is the number of items failed. This adds a debug log message in case a bulk error shows up.


jsoriano (approved):

aliabbas-elastic (approved):
LGTM

PR #1628, opened 2024-01-18
Add progress flag if docker compose is used

mrodm:


internal/compose/compose.go
```diff
@@ -214,6 +216,11 @@ func NewProject(name string, paths ...string) (*Project, error) {
 		c.disablePullProgressInformation = true
 	}
 
+	v, ok = os.LookupEnv(DisableProgressOutputComposeEnv)
+	if !c.dockerComposeV1 && !c.dockerComposeStandalone && ok && strings.ToLower(v) != "false" {
```
mrodm, line 220: This option is just available using `docker compose`

.buildkite/pipeline.yml
```diff
@@ -1,6 +1,8 @@
 env:
   SETUP_GVM_VERSION: 'v0.5.1' # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
   ELASTIC_PACKAGE_COMPOSE_DISABLE_ANSI: "true"
+  # ELASTIC_PACKAGE_COMPOSE_DISABLE_PROGRESS_OUTPUT: "true"
+  ELASTIC_PACKAGE_COMPOSE_PROGRESS_OUTPUT: "quiet"
```
mrodm, line 5: @jsoriano any preference about which environment would be best to set ? 

ELASTIC_PACKAGE_COMPOSE_DISABLE_PROGRESS_OUTPUT would be just to set `true` or `false`, and we would need to decide if we want `plain` or `quiet` for the `--progress` flag.

In ELASTIC_PACKAGE_COMPOSE_PROGRESS_OUTPUT var, a few values would be allowed (auto, plain or quiet) and the user could choose which one.

README.md
```diff
@@ -625,6 +625,47 @@ The following settings are available per profile:
 * `stack.serverless.region` can be used to select the region to use when starting
   serverless projects.
 
+## Useful environment variables
```
mrodm, line 628: AFAIK there is no docs about all the environment variables that are defined in elastic-package. Just added this section to include all these references

WDYT ?

internal/compose/compose.go
```diff
@@ -36,6 +37,8 @@ var (
 	DisableANSIComposeEnv             = environment.WithElasticPackagePrefix("COMPOSE_DISABLE_ANSI")
 	DisablePullProgressInformationEnv = environment.WithElasticPackagePrefix("COMPOSE_DISABLE_PULL_PROGRESS_INFORMATION")
 	EnableComposeStandaloneEnv        = environment.WithElasticPackagePrefix("COMPOSE_ENABLE_STANDALONE")
+	DisableProgressOutputComposeEnv   = environment.WithElasticPackagePrefix("COMPOSE_DISABLE_PROGRESS_OUTPUT")
+	ProgressOutputComposeEnv          = environment.WithElasticPackagePrefix("COMPOSE_PROGRESS_OUTPUT")
```
mrodm, line 41: We should decide which one to use here.

internal/compose/compose.go
jsoriano: Why? This behaves the same in `docker-compose` (trying with standalone docker-compose `v2.23.3`).

jsoriano (commented):
> Latest Buildkite builds that are executing the docker compose for all the commands under elastic-package looks like are not honoring the parameter `--ansi never`.

I think it is not a matter of `docker-compose` vs `docker compose` but about a regression(?)  on recent versions. Trying locally with standalone `docker-compose` v2.23.3 it neither honors `--ansi never`.

Given the differences between the different versions I wonder if instead of exposing the different flags, we should have a single flag to enable or disable this verbose output, that under the hood uses any flag known to work for each version.

internal/compose/compose.go
mrodm: I was trying with `docker-compose` with version 2.17.2 (same version that was set previously in CI for it). In that version there is no `--progress` flag.

I see that in the latest version of docker-compose standalone (v2.24.1), that version is already in place.

mrodm:
buildkite test this

jsoriano (approved):
:+1:

PR #1634, opened 2024-01-22
List packages with benchmarks

ruflin:
For someone to run benchmarks or ingest sample data it is currently not easy to find the packages that have templates for benchmarks available. This is a quick POC on how the list of integrations with benchmarks can be generated. Some ideas on how this could be used:

* Add the list to the docs page around streaming data. Downside is that it needs to be manually (or automatically) updated from time to time
* Add some command to elastic-package to list the packages with templates
* Other ideas?

The current output of the script looks as following:

```
python list_benchmark_packages.py

Package: nginx
* stubstatus-benchmark
* error-benchmark

Package: mysql
* performance-benchmark

Package: aws
* sqs-benchmark
* ec2metrics-benchmark
* billing-benchmark
* ec2logs-benchmark

Package: kubernetes
* container-benchmark
* pod-benchmark

### DIFF
--- a/scripts/list_benchmark_packages.py
+++ b/scripts/list_benchmark_packages.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+
+import os
+
+integrations_repo_path = "../../integrations"
+packages_path = os.path.join(integrations_repo_path, "packages")
+
+# Iterate through all packages
+for package_path in os.listdir(packages_path):
+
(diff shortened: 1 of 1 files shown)


aspacca:
> * Add some command to elastic-package to list the packages with templates

+1 on this solution

ruflin:
@lalit-satapathy Move over to you as I think this is a nice way on how to show which packages have a benchmark.

PR #1647, opened 2024-01-26
Remove reroute processors from coverage reports

mrodm:
Relates #1595

Reroute processors are added to the ingest pipeline automatically if there is any `routing_rules.yml` defined in the data stream (added here #1372).

If these reroute processors are added to the ingest pipeline, the stats obtained from the pipelines contain also those processors that are not defined in the original pipeline (ingest pipeline defined in the package).

This PR adds the logic to remove the reroute processors from the stats to generate the coverage reports. And at the same time, to be able to use the right lines in the coverage reports, it is needed to keep the original contents (the ones read from the file) of the pipeline.

### DIFF
--- a/internal/elasticsearch/ingest/datastream.go
+++ b/internal/elasticsearch/ingest/datastream.go
@@ -98,17 +98,18 @@ func loadIngestPipelineFiles(dataStreamPath string, nonce int64) ([]Pipeline, er
 			return nil, err
 		}
 
-		c, err = addRerouteProcessors(c, dataStreamPath, path)
+		cWithRerouteProcessors, err := addRerouteProcessors(c, dataStreamPath, path)
 		if err != nil {
 			return nil, err
 		}
 
 		name := filepath.Base(path)
 		pipelines = append(pipelines, Pipeline{
-			Path:    path,
-			Name:    getPipelineNameWithNonce(name[:strings.Index(name, ".")], nonce),
-			Format:  filepath.Ext(path)[1:],
-			Content: c,
+			Path:            path,
+			Name:            getPipelineNameWithNonce(name[:strings.Index(name, ".")], nonce),
+			Format:          filepath.Ext(path)[1:],
+			Content:         cWithRerouteProcessors,
+			ContentOriginal: c,
 		})
 	}
 	return pipelines, nil
--- a/internal/elasticsearch/ingest/processors.go
+++ b/internal/elasticsearch/ingest/processors.go
@@ -36,8 +36,8 @@ func (p Pipeline) Processors() (procs []Processor, err error) {
 	return procs, nil
 }
 
-// Processors return the original list of processors in an ingest pipeline.
-func (p Pipeline) ProcessorsWithoutReroute() (procs []Processor, err error) {
+// OriginalProcessors return the original list of processors in an ingest pipeline.
+func (p Pipeline) OriginalProcessors() (procs []Processor, err error) {
 	switch p.Format {
 	case "yaml", "yml", "json":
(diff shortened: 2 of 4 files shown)


internal/elasticsearch/ingest/datastream.go
```diff
@@ -98,17 +98,18 @@ func loadIngestPipelineFiles(dataStreamPath string, nonce int64) ([]Pipeline, er
...
 		pipelines = append(pipelines, Pipeline{
-			Path:    path,
-			Name:    getPipelineNameWithNonce(name[:strings.Index(name, ".")], nonce),
-			Format:  filepath.Ext(path)[1:],
-			Content: c,
+			Path:            path,
+			Name:            getPipelineNameWithNonce(name[:strings.Index(name, ".")], nonce),
+			Format:          filepath.Ext(path)[1:],
+			Content:         cWithRerouteProcessors,
+			ContentOriginal: c,
```
mrodm, line 112: Added a new field `ContentOriginal` in this struct to be able to keep the original contents, the ones read from the ingest pipeline file defined in the package.

It is being kept `Content` to hold the same contents (containing reroute processors) to not change the current behaviour.

internal/testrunner/runners/pipeline/coverage.go
```diff
@@ -115,6 +115,17 @@ func pipelineDataForCoverage(pipeline ingest.Pipeline, stats ingest.PipelineStat
...
+	// reroute processors are added if there are any routing_rules file defined
+	var processors []ingest.ProcessorStats
+	for _, proc := range pstats.Processors {
+		if proc.Type == "reroute" {
+			continue
+		}
+		processors = append(processors, proc)
+	}
+	pstats.Processors = processors
+
```
mrodm, line 128: As the stats returned by Elasticsearch contain the `reroute` processors added by elastic-package in a previous step, here it is needed to remove those processors so it contains the same number of processors as in the file.

internal/elasticsearch/ingest/datastream.go
mrodm: Another consequence of adding those reroute processors, is that the YAML written afterwards is not the same as in the package:
- comments are not kept
- fields in the YAML do not keep the same order

internal/elasticsearch/ingest/processors.go
```diff
@@ -36,6 +36,20 @@ func (p Pipeline) Processors() (procs []Processor, err error) {
 	return procs, nil
 }
 
+// Processors return the original list of processors in an ingest pipeline.
+func (p Pipeline) ProcessorsWithoutReroute() (procs []Processor, err error) {
```
jsoriano, line 40: Nit. In case we do more manipulations in the future apart of adding reroute processors. 
```suggestion
func (p Pipeline) OriginalProcessors() (procs []Processor, err error) {
```

internal/testrunner/runners/pipeline/coverage.go
```diff
@@ -115,6 +115,17 @@ func pipelineDataForCoverage(pipeline ingest.Pipeline, stats ingest.PipelineStat
 		return "", "", nil, ingest.PipelineStats{}, fmt.Errorf("pipeline '%s' not installed in Elasticsearch", pipeline.Name)
 	}
 
+	// remove reroute processors if any so the pipeline has the same processors as in the file
+	// reroute processors are added if there are any routing_rules file defined
```
jsoriano, line 119: Nit.
```suggestion
	// Remove reroute processors if any so the pipeline has the same processors as in the file
	// reroute processors are added if there are any routing_rules file defined.
```

jsoriano (approved):

Issue #1519, opened 2023-10-19
Support JSON format in elastic-package status


jsoriano:
`elastic-package status` provides useful information about packages, but this information is currently intended for humans and displayed as a table.

It would be nice to add a `--format json` flag that can be used for easier parsing of the output, so it can be used by other tools and automations.

Originally posted by @mrodm in https://github.com/elastic/elastic-package/pull/1516#pullrequestreview-1687781704

PR #1655, opened 2024-02-02
Add JSON format to status subcommand

jsoriano:
Add a `--format json` option to `elastic-package status`, that shows the same information but in JSON format.

See sample outputs in the test files.

Fixes https://github.com/elastic/elastic-package/issues/1519

### DIFF
--- a/cmd/testdata/status-pending-changes.json
+++ b/cmd/testdata/status-pending-changes.json
@@ -1,0 +1,20 @@
+{
+  "package": "foo",
+  "owner": "team",
+  "versions": [
+    {
+      "environment": "local",
+      "version": "2.0.0-rc1",
+      "release": "release candidate",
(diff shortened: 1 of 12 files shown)


cmd/testdata/status-pending-changes.json
```diff
@@ -0,0 +1,20 @@
+{
```
mrodm, line 1: Comparing with the testdata with the table format, in this test it is also shown there a section for pending changes:
https://github.com/elastic/elastic-package/blob/b7ede8d5c91e0f19ce986112d810fdcba10f5206/cmd/testdata/status-pending-changes#L4

jsoriano: Good one, added, thanks!

mrodm (approved):
:rocket:

PR #1667, opened 2024-02-08
Avoid using coverage XML files for junit step

mrodm:
Currently, junit step last around 7 minutes:

junit step before

In this PR, it is removed from this step the XML files related to coverage reports.
After changing this the time spent in this step is around 10-15 secs:
junit step after


mrodm:
buildkite test this

jsoriano (approved):
Good one, thanks! :+1:

PR #1669, opened 2024-02-09
Enable ssl between logstash and agent

### LINKED ISSUE #1658: [logstash] Enable ssl for communication between agent and logstash in local stack
The communication between agent and logstash is currently without ssl when local stack is set up with `logstash_enabled` as a profile config option.

Similar to #1646 , change 
https://github.com/elastic/elastic-package/blob/a567a70edc1b6e74eda822f5c19ae24a91ceb470/internal/stack/_static/logstash.conf.tmpl#L4 to `true`

Relates #1645 #1432 #1458

bhapas:
Currently ssl is disabled between logstash and agent when running on local stack. This PR enables ssl communication between logstash and agent.

Closes #1658


bhapas:
There are two approaches to add the client certificates when fleet output is added through config `kibana.yml`

- Updating the fleet API with the ssl certificates with `allow_edit` . But this is blocked by https://github.com/elastic/kibana/issues/176702

- Importing the certificates generated locally into the kibana config using facters. But this is also blocked as Fleet is removing new line in the ssl certificates leading to error in elastic-agent

internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -170,9 +170,14 @@ services:
       interval: 60s
       timeout: 50s
       retries: 5
-    command: bash -c 'if [[ ! $(bin/logstash-plugin list) == *"logstash-filter-elastic_integration"* ]]; then echo "Missing plugin logstash-filter-elastic_integration, installing now" && bin/logstash-plugin install logstash-filter-elastic_integration; fi && bin/logstash -f /usr/share/logstash/pipeline/logstash.conf'
+    # logstash expects the key in pkcs8 format. Hence converting the key.pem to pkcs8 format using openssl.
+    # Also logstash-filter-elastic_integration plugin is installed by default to run ingest pipelines in logstash.
+    # elastic-package#1637 made improvements to enable logstash stats through port 9600.
+    command: bash -c 'openssl pkcs8 -inform PEM -in /usr/share/logstash/config/certs/key.pem -topk8 -nocrypt -outform PEM -out /tmp/logstash.pkcs8.key && chmod +x /tmp/logstash.pkcs8.key && if [[ ! $(bin/logstash-plugin list) ==  *"logstash-filter-elastic_integration"* ]]; then echo "Missing plugin  logstash-filter-elastic_integration, installing now" && bin/logstash-plugin install  logstash-filter-elastic_integration; fi && bin/logstash -f /usr/share/logstash/pipeline/logstash.conf'
```
bhapas, line 176: This shall be extracted to a script in https://github.com/elastic/elastic-package/pull/1668

internal/stack/certs.go
```diff
@@ -29,6 +29,7 @@ var tlsServices = []tlsService{
 	{Name: "package-registry"},
 	{Name: "fleet-server"},
 	{Name: "logstash"},
```
jsoriano, line 31: Nit. Maybe we can add another option to generate the PKCS8 files here, so we don't need to generate them later on container startup.
```suggestion
	{Name: "logstash", GeneratePKCS8: true},
```

internal/stack/resources.go
```diff
@@ -171,6 +184,45 @@ func applyResources(profile *profile.Profile, stackVersion string) error {
 	return nil
 }
 
+func addClientCertsToResources(resourceManager *resource.Manager, certResources []resource.Resource) error {
```
jsoriano, line 187: Could we use an approach like this one for serverless? So we provide the certificates when creating the output and we don't need to update it later?

internal/stack/resources.go
```diff
@@ -171,6 +184,45 @@ func applyResources(profile *profile.Profile, stackVersion string) error {
...
+				// Replace newlines with spaces to create proper indentation in the config
+				keyFile = buf.String()
+				continue
+			}
+		}
+	}
+
+	resourceManager.AddFacter(resource.StaticFacter{
+		"logstash_ssl_certificate": certFile,
+		"logstash_ssl_key":         keyFile,
```
jsoriano, line 221: Nit. Should we call it agent or client certificates? As they are not the logstash certificates.
```suggestion
		"agent_certificate": certFile,
		"agent_key":         keyFile,
```
```suggestion
		"logstash_client_certificate": certFile,
		"logstash_client_key":         keyFile,
```

internal/stack/resources.go
```diff
@@ -148,15 +154,22 @@ func applyResources(profile *profile.Profile, stackVersion string) error {
...
+	certResources, err := initTLSCertificates(CertsFolder, profile.ProfilePath, tlsServices)
 	if err != nil {
 		return fmt.Errorf("failed to create TLS files: %w", err)
 	}
-	resources = append(resources, certResources...)
 
+	// Add client certificates if logstash is enabled
+	if profile.Config("stack.logstash_enabled", "false") == "true" {
+		if err := addClientCertsToResources(resourceManager, certResources); err != nil {
+			return fmt.Errorf("Error adding client certificates: %w", err)
```
jsoriano, line 168:
```suggestion
			return fmt.Errorf("error adding client certificates: %w", err)
```

internal/stack/resources.go
bhapas: But for serverless the certificates are created at a later stage than adding the output to agent policy. And to add agent policy the output needs to be created

internal/stack/certs.go
bhapas: Can try this in a new PR

internal/stack/resources.go
jsoriano: I think we could move `applyServerlessResources` to an early stage if needed. It is only now in `startLocalServices` because so far it was only needed for local services. But we could run this earlier if needed to create the project.

PR #1670, opened 2024-02-09
Update stack versions after 8.12.1 release

jsoriano:


Makefile
```diff
@@ -59,14 +59,14 @@ test-stack-command-oldest:
 	./scripts/test-stack-command.sh 7.14.2
 
 test-stack-command-7x:
-	./scripts/test-stack-command.sh 7.17.9
+	./scripts/test-stack-command.sh 7.17.18
```
mrodm, line 62: Ouch we missed some minors here.

PR #1672, opened 2024-02-09
Disable elastic agent self monitoring by default

jsoriano:
Configure elastic-agent by default with an empty policy to avoid affecting tests with uncontrolled versions of the system package. A new user parameter `stack.self_monitor_enabled` can be set to true to revert back to the previous behaviour.

We are configuring monitoring and the system package for the default policy in Elastic Agent managed by `elastic-package stack`, but this is not required for tests or package development, and it slightly slows down startup, consumes resources unnecessarily and is a source of flakiness.

For example in https://github.com/elastic/elastic-package/pull/1670 the tests for the system package are affected by the pre-installed system package. They are failing because a missing pipeline for system package 1.53.1, but the version being tested is 1.53.0.
```
pipeline with id [metrics-system.memory-1.53.1] does not exist
```

### DIFF
--- a/internal/serverless/project.go
+++ b/internal/serverless/project.go
@@ -232,7 +232,35 @@ func (p *Project) getFleetHealth(ctx context.Context) error {
 	return nil
 }
 
-func (p *Project) CreateAgentPolicy(stackVersion string, kibanaClient *kibana.Client, outputId string) error {
+func (p *Project) CreateAgentPolicy(kibanaClient *kibana.Client, stackVersion string, outputId string, selfMonitor bool) error {
+	policy := kibana.Policy{
+		ID:                "elastic-agent-managed-ep",
+		Name:              "Elastic-Agent (elastic-package)",
+		Description:       "Policy created by elastic-package",
+		Namespace:         "default",
+		MonitoringEnabled: []string{},
+		DataOutputID:      outputId,
+	}
+	if selfMonitor {
+		policy.MonitoringEnabled = []string{"logs", "metrics"}
+	}
+
+	newPolicy, err := kibanaClient.CreatePolicy(policy)
+	if err != nil {
+		return fmt.Errorf("error while creating agent policy: %w", err)
+	}
+
+	if selfMonitor {
+		err := p.createSystemPackagePolicy(kibanaClient, stackVersion, newPolicy.ID, newPolicy.Namespace)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (p *Project) createSystemPackagePolicy(kibanaClient *kibana.Client, stackVersion, agentPolicyID, namespace string) error {
 	systemPackages, err := registry.Production.Revisions("system", registry.SearchOptions{
 		KibanaVersion: strings.TrimSuffix(stackVersion, kibana.SNAPSHOT_SUFFIX),
 	})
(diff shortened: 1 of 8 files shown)


jsoriano:
test integrations

internal/serverless/project.go
```diff
@@ -232,36 +232,49 @@ func (p *Project) getFleetHealth(ctx context.Context) error {
...
-	logger.Debugf("Found %s package - version %s", systemPackages[0].Name, systemPackages[0].Version)
-
+func (p *Project) CreateAgentPolicy(kibanaClient *kibana.Client, stackVersion string, outputId string, selfMonitor bool) error {
 	policy := kibana.Policy{
 		ID:                "elastic-agent-managed-ep",
 		Name:              "Elastic-Agent (elastic-package)",
 		Description:       "Policy created by elastic-package",
 		Namespace:         "default",
-		MonitoringEnabled: []string{"logs", "metrics"},
+		MonitoringEnabled: []string{},
```
jsoriano, line 241: Well, maybe we need logs collection to catch errors in logs :thinking:

mrodm: Checking about the system tests and the Agent policy created there does not have logs and metrics enabled.
https://github.com/elastic/elastic-package/blob/4af13a5c73432b0d7b20f595adcb1acaa8228eff/internal/testrunner/runners/system/runner.go#L560-L568

So even if the `Elastic-Agent (elastic-package)` policy has those settings enabled, when the system agent policy is assigned to that agent those logs/metrics are not collected

agent with system test policy

tools/readme/readme.md.tmpl
```diff
@@ -173,6 +173,8 @@ The following settings are available per profile:
 * `stack.logstash_enabled` can be set to true to start Logstash and configure it as the
   default output for tests using elastic-package. Supported only by the compose provider.
   Defaults to false.
+* `stack.self_monitor_enabled` enables monitoring and the system package for the managed
+  Elastic Agent. Defaults to false.
```
mrodm, line 177: Those settings are assigned to the policy, if another policy is assigned to the agent, the monitoring will depend on the values of the new policy.
Would this be needed or is it redundant ?


```suggestion
* `stack.self_monitor_enabled` enables monitoring and the system package for the default
  policy assigned to the managed Elastic Agent. Defaults to false.
```

mrodm (approved):
LGTM

Just some minor comments.

tools/readme/readme.md.tmpl
jsoriano: Well, more clear an specific this way, I will add this, thanks.

internal/stack/resources.go
```diff
@@ -130,9 +135,10 @@ func applyResources(profile *profile.Profile, stackVersion string) error {
 		"username": elasticsearchUsername,
 		"password": elasticsearchPassword,
 
-		"geoip_dir":        profile.Config("stack.geoip_dir", "./ingest-geoip"),
-		"apm_enabled":      profile.Config("stack.apm_enabled", "false"),
-		"logstash_enabled": profile.Config("stack.logstash_enabled", "false"),
+		"apm_enabled":          profile.Config(configAPMEnabled, "false"),
+		"geoip_dir":            profile.Config(configGeoIPDir, "./ingest-geoip"),
+		"logstash_enabled":     profile.Config(configLogstashEnabled, "false"),
+		"self_monitor_enabled": profile.Config(configSelfMonitorEnabled, "false"),
```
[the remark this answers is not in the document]
jsoriano: Exactly, not added because it would not be used.

PR #1675, opened 2024-02-13
Remove TODO comment about test assets in input packages

### LINKED ISSUE #1623: Coverage reports for asset tests are not generated in input packages
Relates #1595 

In #1595 , it is added support to create coverage reports for input packages. Even that, coverage report for asset tests it is not generated since it is not generated a list of assets for input packages.

This `asset` tests are based on kibana and elasticsearch assets. In case of input packages, there are no kibana assets, but there should be elasticsearch assets (e.g. index templates). Currently, the logic just takes into account generating that list based on data stream paths from the package. This is valid for integration packages but not for input packages.

https://github.com/elastic/elastic-package/blob/b3f999e5b77b91685832c08dad69d5292a52cbde/internal/packages/assets.go#L94-L107

mrodm:
This PR removes the TODO comment about test assets related to input packages.

Closes #1623

### DIFF
--- a/internal/packages/assets.go
+++ b/internal/packages/assets.go
@@ -153,8 +153,7 @@ func loadElasticsearchAssets(pkgRootPath string) ([]Asset, error) {
 		}
 	}
 
-	// TODO add assets for input packages
-
+	// No Elasticsearch asset is created when an Input package is installed through the API.
 	return assets, nil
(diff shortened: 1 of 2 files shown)


jsoriano (approved):
:+1:

PR #1676, opened 2024-02-13
Remove exception for uninstalling system package in tests

mrodm:
This PR removes the exception that was set in place for `system` package when packages were uninstalled in the tear down process (asset and system tests).

In #1672, `system` package is no longer installed in the default Agent policy, so there should not be any error when `system` package is uninstalled from now on.

- [x] Run all tests for current `system` package in integrations repository locally (Elastic stack 8.12.1)
    - system package is deleted
- [x] Run all tests for latest version supporting Elastic stack 7.17.* (version 1.11.0) - https://github.com/elastic/integrations/commit/792a55c90066756d99dbb878be2b38e6feee59fb
    - deletion of the system package is skipped

Relates:
- #1672 
- #1315

### DIFF
--- a/internal/testrunner/runners/asset/runner.go
+++ b/internal/testrunner/runners/asset/runner.go
@@ -105,9 +105,7 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 		logger.Debug("removing package...")
 		err = packageInstaller.Uninstall()
 
-		// by default system package is part of an agent policy and it cannot be uninstalled
-		// https://github.com/elastic/elastic-package/blob/5f65dc29811c57454bc7142aaf73725b6d4dc8e6/internal/stack/_static/kibana.yml.tmpl#L62
-		if err != nil && pkgManifest.Name != "system" {
+		if err != nil {
 			logger.Warnf("failed to uninstall package %q: %s", pkgManifest.Name, err.Error())
 		}
 		return nil
--- a/internal/testrunner/runners/asset/runner.go
+++ b/internal/testrunner/runners/asset/runner.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
+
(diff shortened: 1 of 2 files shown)


internal/testrunner/runners/asset/runner.go
```diff
@@ -105,9 +105,7 @@ func (r *runner) run() ([]testrunner.TestResult, error) {
 		logger.Debug("removing package...")
 		err = packageInstaller.Uninstall()
 
-		// by default system package is part of an agent policy and it cannot be uninstalled
-		// https://github.com/elastic/elastic-package/blob/5f65dc29811c57454bc7142aaf73725b6d4dc8e6/internal/stack/_static/kibana.yml.tmpl#L62
```
mrodm, line 109: Elastic stacks with version 7.17.* still have `system` package installed (tested with 7.17.18).
image

Would that be a problem ? @elastic/ecosystem 

Looking at the latest `system` package version , it requires Kibana version `^8.12.0`
https://github.com/elastic/integrations/blob/92900f6c0eb9cdc449808d8fd1dc234d05cf7f93/packages/system/manifest.yml#L11

jsoriano: Ah, you are right, actually the part of the template modified in https://github.com/elastic/elastic-package/pull/1672/files#diff-df47a0840eb947003447c6e44e2fbec3f13b1ffda4478b46ba2cced8f4ae2fc6R51 is only used starting on 8.0.0.

In 7.x probably this policy was hard-coded in Kibana, so we will need to keep the exception at least for 7.x.

mrodm: Re-added this exception but just for Elastic stack `7.*` in this commit https://github.com/elastic/elastic-package/pull/1676/commits/7d4c3f5667da3f41b03b41c321ae68e4444fa993

jsoriano (approved):

PR #1679, opened 2024-02-13
Provide contexts from the root command

jsoriano:
Apply several refactors to provide contexts inherited from the main command. This helps to have better control of interruptions and cancellation signals in a more go-native way. This can be also useful in the future when using functionality that benefits from context propagation, like tracing.

Changes applied here:
* `internal/signal` provides a single function to enable signal handling. It returns a context that is cancelled on SIGINT.
* Provide contexts where we had `context.TODO`s.
* Review all usages of `context.Background`.
* Rename `ServiceContext` to `ServiceInfo` to address ambiguities.
* New `internal/wait` package with a common implementation to replace the multiple implementations of `waitUntilTrue` that we had.
* New `CountDocsInDataStream` method to replace multiple implementation of `getTotalHits`.
* Contexts passed to cleanup handlers. They are passed without cancellation, so cleanup tasks are still executed during interruption handling.
* Reorganize code in `stream` benchmark runner to help on context cancellation.

Notes for the future:
* There are more places where we could reuse wait logic. We could also reuse https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait. I explicitly leave further refactors here for a future change.
* There are more places where we could leverage contexts, like in the clients. As they are not using any nil context now, I leave this for a future refactor.
  * Follow up for kibana started in https://github.com/elastic/elastic-package/pull/1713.

Checklist:
- [x] Ensure that test runners run the test tear downs if the test is interrupted.


cmd/benchmark.go
```diff
@@ -309,13 +309,14 @@ func rallyCommandAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	signal.Enable()
+	ctx, stop := signal.Enable(cmd.Context(), logger.Info)
+	defer stop()
```
jsoriano, line 313: @mrodm enabling by now signal handling in the same places as before, for easier control of changes.

internal/servicedeployer/custom_agent.go
```diff
@@ -99,7 +100,7 @@ func (d *CustomAgentDeployer) SetUp(inCtxt ServiceContext) (DeployedService, err
 		},
 	}
 
-	outCtxt := inCtxt
+	outCtxt := svcInfo
```
mrodm, line 103: Following the `svcInfo`naming, should `outCtxtp  be renamed to `outSvcInfo` or similar ?

internal/stack/boot.go
```diff
@@ -78,24 +80,24 @@ func BootUp(options Options) error {
...
+	err = dockerComposeUp(ctx, options)
 	if err != nil {
 		// At least starting on 8.6.0, fleet-server may be reconfigured or
 		// restarted after being healthy. If elastic-agent tries to enroll at
 		// this moment, it fails inmediately, stopping and making `docker-compose up`
 		// to fail too.
 		// As a workaround, try to give another chance to docker-compose if only
 		// elastic-agent failed.
-		if onlyElasticAgentFailed(options) {
+		if onlyElasticAgentFailed(ctx, options) && !errors.Is(err, context.Canceled) {
```
mrodm, line 96: :+1:

mrodm (approved):
It looks great! Just added some minor comments/questions, but they are non-blocking

internal/kibana/agents.go
```diff
@@ -113,8 +108,8 @@ func (c *Client) waitUntilPolicyAssigned(a Agent, p Policy) error {
 
 		logger.Debugf("Wait until the policy (ID: %s, revision: %d) is assigned to the agent (ID: %s)...", p.ID, p.Revision, a.ID)
 		select {
-		case <-timeout.C:
-			return errors.New("timeout: policy hasn't been assigned in time")
+		case <-ctx.Done():
+			return ctx.Err()
```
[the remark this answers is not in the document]
jsoriano: Error recovered in https://github.com/elastic/elastic-package/pull/1679/commits/b60f01b393df1a1d4f03d4a7b0be08f10b467818

internal/compose/compose.go
```diff
@@ -412,21 +406,25 @@ func (p *Project) WaitForHealthy(opts CommandOptions) error {
 			break
 		}
 
-		// NOTE: using sleep does not guarantee interval but it's ok for this use case
-		time.Sleep(waitForHealthyInterval)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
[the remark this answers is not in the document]
jsoriano: Error recovered in https://github.com/elastic/elastic-package/pull/1679/commits/b60f01b393df1a1d4f03d4a7b0be08f10b467818

internal/servicedeployer/custom_agent.go
jsoriano: Actually I wonder if this output variable is needed, `svcInfo` is already passed by value, I guess we don't need to copy it again.

I am removing this line in https://github.com/elastic/elastic-package/pull/1679/commits/f784e23ca85095ada3ff72953e5057479dee7cb6

mrodm (approved):

PR #1681, opened 2024-02-14
Render fields as collapsable table

jsoriano:
Add support to render collapsable field tables, so they are rendered like this:

-----

<details>
<summary><strong>Exported fields</strong></summary>

| Field | Description | Type |
|---|---|---|
| @timestamp | Event timestamp. | date |
| apache.access.ssl.cipher | SSL cipher name. | keyword |
| apache.access.ssl.protocol | SSL protocol version. | keyword |
| cloud.account.id | The cloud account or organization id used to identify different entities in a multi-tenant environment. Examples: AWS account id, Google Cloud ORG Id, or other unique identifier. | keyword |
| cloud.availability_zone | Availability zone in which this host is running. | keyword |
| cloud.image.id | Image ID for the cloud instance. | keyword |
| cloud.instance.id | Instance ID of the host machine. | keyword |
| cloud.instance.name | Instance name of the host machine. | keyword |
| cloud.machine.type | Machine type of the host machine. | keyword |
| cloud.project.id | Name of the project in Google Cloud. | keyword |
| cloud.provider | Name of the cloud provider. Example values are aws, azure, gcp, or digitalocean. | keyword |
...
</details>

----

CI in integrations repository checks that documentation files don't change after "building" them, so just changing how all packages are rendered would break CI. To avoid making this a breaking change, we cannot make this change the default. There are some possible options:

* Add a new function `fields_collapsable`.
  * Pros: Backwards compatible, package developers can chose when to change.
  * Cons: Potential inconsistent use between packages.
* Select the rendering depending on the spec version.
  * Pros: Backwards compatible, the change will eventually be applied to packages as their format is updated.
  * Cons: Won't apply till format version is updated in packages.

Complementary to this, maybe we can make Fleet to add the tags for READMEs containing the current tittle (`**Exported fields**`). This would bring consistency to existing packages.

@elastic/ecosystem @alaudazzi


jsoriano:
@alaudazzi it looks like this is not properly rendered in Fleet :confused: HTML is encoded:
imagen

Do you know if there is some way to do this with pure Markdown?

alaudazzi:
@jsoriano 

> Do you know if there is some way to do this with pure Markdown?

This is the Markdown code I found:

```
<details>
  <summary>Click me</summary>
  
  ### Heading
  1. Foo
  2. Bar
     * Baz
     * Qux

  ### Some Javascript
  ```js
  function logSomething(something) {
    console.log('Something', something);
  }
</details>
```

jsoriano:
> > Do you know if there is some way to do this with pure Markdown?
> 
> This is the Markdown code I found:
> 
> ```
> <details>
>   <summary>Click me</summary>
> ```

Yes, I mean if there would be some way to do it without HTML tags. HTML tags are not going to be rendered as expected by Kibana, it encodes the tags and they are rendered as the literal text, as seen in https://github.com/elastic/elastic-package/pull/1681#issuecomment-1943450793.

@kpollich do you have an idea of what could be done about this?

alaudazzi:
@ebeahan 
It seems that we are trying to get something that you have already tested [here](https://elastic.slack.com/archives/CED7QCL3E/p1707315721941329?thread_ts=1707119200.601369&cid=CED7QCL3E). 
Can you share the code changes that you tested to get collapsible exported fields?

kpollich:
> do you have an idea of what could be done about this?

There are plugins for `rehype` (which Fleet uses indirectly via the `react-markdown` library responsible for rendering these READMEs in the UI) that should be able to help here, e.g. https://www.npmjs.com/package/rehype-raw

I think we could update our usage of the `ReactMarkdown` component after installing `rehype-raw` and `rehype-sanitize` as follows

```jsx
<ReactMarkdown
  transformImageUri={handleImageUri}
  components={markdownRenderers(refs)}
  remarkPlugins={[remarkGfm]}
  rehypePlugins={[rehypeRaw, [rehypeSanitize, { tagNames: ['summary', 'details'] }]]}
>
```

This will allow us to "pass through" the `summary` and `details` tags as above, but keep the escaping behavior for all other tags. 

I didn't have a chance to test this directly, so this might be an incorrect assumption. However looking around GitHub and StackOverflow it seems aligned with other suggestions for similar cases.

jsoriano:
Closing by now as we are trying a different approach, modifying the markdown before rendering it.

Issue #1682, opened 2024-02-14
[serverless][local stack] Refactor ssl changes


bhapas:
As per https://github.com/elastic/elastic-package/pull/1669#discussion_r1489416945 and https://github.com/elastic/elastic-package/pull/1669#discussion_r1489425185 , check if the refactor is possible

- [ ] https://github.com/elastic/elastic-package/pull/1669#discussion_r1489425185
- [ ] https://github.com/elastic/elastic-package/pull/1669#discussion_r1489416945

kpollich:
We haven't revisited this issue in a while, so we're closing it due to inactivity. Please reopen with a comment if you think this is a mistake. Thanks!

PR #1687, opened 2024-02-15
Pre-build logstash image with plugin and configuration

jsoriano:
Polish some things in the logstash scenario:
* Pre-build logstash image:
  * It includes the plugin now so it doesn't need to be installed on every run, speeding up startup after first run.
  * It includes the configuration, so it can be modified with auto-reload, as introduced in https://github.com/elastic/elastic-package/pull/1668.
* Use private key without converting it to PKCS8.
* Use actual CA instead of elasticsearch certificate as CA in logstash output.
* Startup script removed as operations moved to other places or not needed.
* Use the same configuration template for serverless and compose providers.

Part of https://github.com/elastic/elastic-package/issues/1682.


internal/stack/_static/logstash.conf.tmpl
```diff
@@ -4,7 +4,7 @@ input {
     ssl_enabled => true
     ssl_certificate_authorities => ["/usr/share/logstash/config/certs/ca-cert.pem"]
     ssl_certificate => "/usr/share/logstash/config/certs/cert.pem"
-    ssl_key => "/tmp/logstash.pkcs8.key"
+    ssl_key => "/usr/share/logstash/config/certs/key.pem"
```
jsoriano, line 7: @bhapas was there a reason to convert to PKCS8? It seems to work with the current key.

If for some reason we need to convert, I can revert this change https://github.com/elastic/elastic-package/pull/1687/commits/b1f174b9d05d5fb4ea41af6fe454115358dd89d1 where I was doing it when generating the certificates.

internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -159,23 +159,20 @@ services:
...
-    image: ${LOGSTASH_IMAGE_REF}
     healthcheck:
       test: bin/logstash -t
       interval: 60s
       timeout: 50s
       retries: 5
-    command: bash /usr/share/logstash/startup.sh
     volumes:
       - "../certs/logstash:/usr/share/logstash/config/certs"
-      - "../certs/elasticsearch/cert.pem:/usr/share/logstash/config/certs/elasticsearch.pem:ro"
```
jsoriano, line 176: This was creating a file owned by root in the user profile, we should avoid writing there.

internal/stack/resources.go
```diff
@@ -122,10 +122,8 @@ var (
 			Content: staticSource.Template("_static/elastic-agent.env.tmpl"),
 		},
 		&resource.File{
-			Path:         "logstash_startup.sh",
-			CreateParent: true,
-			Content:      staticSource.Template("_static/logstash_startup.sh"),
-			Mode:         resource.FileMode(0755),
+			Path:    "Dockerfile.logstash",
+			Content: staticSource.File("_static/Dockerfile.logstash"),
 		},
```
mrodm, line 127: There is also another reference to this file in serverlessresources.go

Trying to creaste a serverless Elastic stack raised this error:

```
2024/02/15 18:41:48  INFO Starting local services
Error: booting up the stack failed: failed to start local services: could not initialize compose files for local services: there where errors: template: pattern matches no files: `_static/logstash_startup.sh`
```
https://github.com/elastic/elastic-package/blob/b2ee26459d1b91f07de567f0db126174d5328b34/internal/stack/serverlessresources.go#L38

internal/stack/_static/logstash.conf.tmpl
mashhurs: +1,
I have just checked that  `key.pem` is `PKCS#8` formatted.

Let's also change the `serverless-logstash.conf.tmpl` file as well.

internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -159,23 +159,20 @@ services:
 {{ $logstash_enabled := fact "logstash_enabled" }}
 {{ if eq $logstash_enabled "true" }}
   logstash:
+    build:
+      dockerfile: "./Dockerfile.logstash"
```
mashhurs, line 163: 🎉  +1 for `Dockerfile`

bhapas: +1 🎉

bhapas (approved):

internal/stack/_static/logstash.conf.tmpl
```diff
@@ -4,30 +4,32 @@ input {
...
-    password => {{ fact "password" }}
+    hosts => ["{{ $elasticsearch_host }}"]
+    user => '{{ fact "username" }}'
+    password => '{{ fact "password" }}'
     ssl_enabled => true
-    ssl_certificate_authorities => "/usr/share/logstash/config/certs/elasticsearch.pem"
+{{- if eq $elasticsearch_host "https://elasticsearch:9200" }}
+    ssl_certificate_authorities => "/usr/share/logstash/config/certs/ca-cert.pem"
+{{- end }}
+    data_stream => "true"
```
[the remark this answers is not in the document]
jsoriano: Ok, I will remove this line, thanks.

internal/stack/_static/Dockerfile.logstash
```diff
@@ -0,0 +1,10 @@
+ARG IMAGE
+FROM ${IMAGE}
+
+RUN bin/logstash-plugin install logstash-filter-elastic_integration
```
mashhurs, line 4: This will always pull new version of the plugin, the expectation to use as it is if plugin is bundled with current Logstash version. Would be problem with other plugins but this one is tightly coupled with ES.

```suggestion
RUN if [ ! $(bin/logstash-plugin list) == *logstash-filter-elastic_integration* ]; then \
        bin/logstash-plugin install logstash-filter-elastic_integration; \
    fi
```

FYI: couldn't test if this works...

internal/stack/_static/serverless-logstash.conf.tmpl
```diff
@@ -1,32 +0,0 @@
...
-}
-
-
-output {
-  elasticsearch {
-    hosts => ["{{ fact "elasticsearch_host" }}"]
-    user => '{{ fact "username" }}'
-    password => '{{ fact "password" }}'
-    ssl_enabled => true
-    data_stream => "true"
```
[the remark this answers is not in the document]
bhapas: The document_id was to avoid duplicate events to get ingested. It can be applied to serverless too.

mrodm:
I've tried to create a cluster using this branch and adding the `system` package into the Agent policy. This configuration is not able to ingest data in Elasticsearch.

This is an example of the error that I get from `logstash` container:

```
Could not index event to Elasticsearch. .... :response=>{"create"=>{"status"=>400, "error"=>{"type"=>"document_parsing_exception", "reason"=>"[1:1011] failed to parse: _id must be unset or set to [7QWrKrJpmTXFC8zWAAABjcDXW-A] but was [%{[@metadata][_ingest_document][id]}] because [.ds-metrics-system.diskio-default-2024.02.19-000001] is in time_series mode", "caused_by"=>{"type"=>"illegal_argument_exception", "reason"=>"_id must be unset or set to [7QWrKrJpmTXFC8zWAAABjcDXW-A] but was [%{[@metadata][_ingest_document][id]}] because [.ds-metrics-system.diskio-default-2024.02.19-000001] is in time_series mode"}}}}}
```

cc @jsoriano @bhapas

bhapas:
> I've tried to create a cluster using this branch and adding the `system` package into the Agent policy. This configuration is not able to ingest data in Elasticsearch.
> 
> This is an example of the error that I get from `logstash` container:
> 
> ```
> Could not index event to Elasticsearch. .... :response=>{"create"=>{"status"=>400, "error"=>{"type"=>"document_parsing_exception", "reason"=>"[1:1011] failed to parse: _id must be unset or set to [7QWrKrJpmTXFC8zWAAABjcDXW-A] but was [%{[@metadata][_ingest_document][id]}] because [.ds-metrics-system.diskio-default-2024.02.19-000001] is in time_series mode", "caused_by"=>{"type"=>"illegal_argument_exception", "reason"=>"_id must be unset or set to [7QWrKrJpmTXFC8zWAAABjcDXW-A] but was [%{[@metadata][_ingest_document][id]}] because [.ds-metrics-system.diskio-default-2024.02.19-000001] is in time_series mode"}}}}}
> ```
> 
> cc @jsoriano @bhapas

@mrodm I think for time series index mode the it cannot use the generated `_id` if I remember from @mashhurs .

One way is to remove the `[%{[@metadata][_ingest_document][id]}]` from logstash config and let it ingest duplicate entries [ As this is just for testing ] . Or have a different config for time_series mode.

jsoriano:
> One way is to remove the `[%{[@metadata][_ingest_document][id]}]` from logstash config and let it ingest duplicate entries [ As this is just for testing ] . Or have a different config for time_series mode.

I am removing it from serverless as it was before this change. In local it is not failing for me or for CI.

internal/stack/_static/Dockerfile.logstash
jsoriano: Change applied, thanks.

PR #1691, opened 2024-02-19
Add docker-compose stop before retrieving logs from service deployer

mrodm:
Add `docker compose stop` command before processing the service logs. This will ensure that all logs are written to the target file. For instance, logs from commands run as a consequence of `trap` in bash.

Terraform containers do make use of those traps to run `terraform destroy` command:

https://github.com/elastic/elastic-package/blob/26e81feb88c6203e62d1f38b6558dd76983f5e6c/internal/servicedeployer/_static/terraform_deployer_run.sh#L11-L19


mrodm:
test integrations

internal/servicedeployer/compose.go
```diff
@@ -225,6 +225,13 @@ func (s *dockerComposeDeployedService) TearDown() error {
 			s.env,
 			s.variant.Env...),
 	}
+	if err := p.Stop(compose.CommandOptions{
+		Env:       opts.Env,
+		ExtraArgs: []string{"-t", "300"}, // default shutdown timeout 10 seconds
```
jsoriano, line 230: We may want to make this configurable at some point.

jsoriano (approved):

PR #1694, opened 2024-02-20
Add linux/amd64 platform for tianon/true service

mrodm:
When trying to run `elastic-package stack up -v -d` command in Mac (M1/M2), this error was raised for all the services/containers `*_is_ready`:
- `no matching manifest for linux/arm64/v8 in the manifest list entries`

This PR updates the docker-compose version format to 2.4 to be able to set [`format`](https://docs.docker.com/compose/compose-file/compose-file-v2/#platform) and specify `linux/amd64` in those containers.

Closes #1693

### DIFF
--- a/internal/stack/_static/docker-compose-stack.yml.tmpl
+++ b/internal/stack/_static/docker-compose-stack.yml.tmpl
@@ -21,7 +21,7 @@ services:
       - "127.0.0.1:9200:9200"
 
   elasticsearch_is_ready:
-    image: tianon/true
+    image: tianon/true:multiarch
     depends_on:
       elasticsearch:
         condition: service_healthy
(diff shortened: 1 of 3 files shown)


internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -1,7 +1,7 @@
 {{ $username := fact "username" }}
 {{ $password := fact "password" }}
 {{ $apm_enabled := fact "apm_enabled" }}
-version: '2.3'
+version: '2.4'
```
jsoriano, line 4: Does the version also need to be changed in the serverless docker compose file?

jsoriano (approved):

internal/stack/_static/docker-compose-stack.yml.tmpl
mrodm: Right, It needs to be updated
Thanks, I'll change it!

jsoriano (approved):

PR #1700, opened 2024-02-22
Add daily job to cleanup stale resources

mrodm:
This PR adds a new daily pipeline to delete all stale resources created in both AWS and GCP cloud providers.

Example build: https://buildkite.com/elastic/elastic-package/builds/2495

Currently, to confirm how it works the commands executed are just plan.
If there is any resource to be deleted, a mail notification will be sent.

To be done in following PRs:
- Add support to clean AWS IAM roles
- Add support to clean AWS IAM policies
- Add support to clean AWS Redshift clusters
- Add support to clean AWS Schedulers

Based on this example:
https://github.com/elastic/elastic-agent/blob/5e17bc222dfa1475bbdf8cade30d67a44dceb436/.buildkite/scripts/steps/gce-cleanup.sh#L15

Tasks to be done before merging:
- [x] Remove step from main pipeline
- [x] Remove debug additions from `pre-command`


.buildkite/configs/cleanup.aws.yml
```diff
@@ -0,0 +1,54 @@
...
+      - type: 'node'
+        regions:
+          - us-east-1
+        filters:
+          # - type: "<"
+          #   pointer: "/created_at"
+          #   param: "${CREATION_DATE}"
+          #   converters:
+          #     param: "date"
+          #     value: "date"
```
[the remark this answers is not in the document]
mrodm: Thanks @amannocci , this solved the error ! :+1:

.buildkite/configs/cleanup.gcp.yml
```diff
@@ -0,0 +1,40 @@
...
+            param: "ci"
+          - type: "regex"
+            pointer: "/name"
+            param: "^elastic-package-(.*)"
+          - type: "!="
+            pointer: "/state"
+            param: "unknown"
+          - type: "!="
+            pointer: "/state"
+            param: "terminated"
```
jsoriano, line 39: Are states named the same in AWS and GCP?

.buildkite/configs/cleanup.aws.yml
```diff
@@ -0,0 +1,75 @@
...
+            param: "ci"
+          - type: "regex"
+            pointer: "/name"
+            param: "^elastic-package-(.*)"
+      - type: 'queue'
+        regions:
+          - us-east-1
+        filters:
+          - type: "<"
+            pointer: "/extra/tags/created_at"
```
jsoriano, line 62: Is the creation date in a tag for queues?

jsoriano (approved):
LGTM, let's see how the plan looks like.

.buildkite/configs/cleanup.gcp.yml
mrodm: Found some examples, and those values were the same for AWS and GCP (terminated and unknown).

For instance in GCP: https://github.com/elastic/elastic-agent/blob/f9310c7f664bb25f37d671cca4e2fcf3749732c8/.buildkite/misc/gce-cleanup.yml#L36

Another example found for AWS:
```yaml
      - account_name: 'CLOUD_ACCOUNT_NAME'
        resources:
          - type: 'node'
            regions:
              - 'us-east-1'
            filters:
              - type: '<'
                pointer: '/created_at'
                param: '${DELETE_RESOURCE_BEFORE_DATE}'
                converters:
                  param: 'date'
              - type: 'regex'
                pointer: '/name'
                param: 'elastic-package-test-(.*)'
              - type: '!='
                pointer: '/state'
                param: 'unknown'
              - type: '!='
                pointer: '/state'
                param: 'terminated'
```

.buildkite/configs/cleanup.aws.yml
mrodm: Yes, that is different from GCP (where it is under `labels`). In AWS, those keys are under `tags` key.

.buildkite/scripts/cloud-cleanup.sh
```diff
@@ -0,0 +1,120 @@
...
+             --context "ctx-cloud-reaper-error" \
+             --style "error"
+    fi
+    exit 1
+fi
+
+# TODO: List and delete the required resources using aws cli
+echo "--- Cleaning up other AWS resources older than ${DELETE_RESOURCES_BEFORE_DATE}"
+echo "--- Installing awscli"
+with_aws_cli
```
amannocci, line 111: What is the deadline? So we can add support for other resources in `cloud-reaper`?

mrodm: There is no hard deadline for this. We were planning to merge this PR to start checking (with `plan` command) if there is any stale resource with the current resources supported by `cloud-reaper`.

We were thinking to add in following PRs, at least listing and filtering by tags for the other resources using `aws` CLI directly starting with `redshift` clusters. But, we would prefer to do so also using `cloud-reaper` to keep all the cleanup process using the same tool.

Should we wait for the `cloud-reaper` support instead for the other resources ? Or, should we add the `redshift` listing/filtering with `aws` CLI for now and some notification ? Those `redshift` clusters were the resources that were left in some builds @jsoriano @kpollich

mrodm: > I should be able to add support for those resources.
Thanks @amannocci !

I think probably we could wait for some time, WDYT @jsoriano @kpollich ?

I've just created a PR to list if there is any AWS Redshift cluster stale due to CI builds. That would allow us to receive email notifications if there is any left cluster until that support is added into cloud-reaper:
https://github.com/elastic/elastic-package/pull/1710

Once this support is added into `cloud-reaper`, this logic could be moved to the `cloud-reaper` for sure, so the cleanup is carried out with the same tool for all resources.

In any case, this PR could be merged to start checking there is no resource like instances or queues.

PR #1703, opened 2024-02-23
Fix busy wait loop in benchmark stream

jsoriano:
This loop didn't have any blocking condition so it was looping forever consuming a whole CPU.

Introduce a ticker by now. Longer term solution in https://github.com/elastic/elastic-package/pull/1679/files#diff-a08e76793f75adeeac003dac9478256aaca336cd00f8051709949e269a081a54R204


mrodm (approved):
LGTM

PR #1710, opened 2024-03-05
Add redshift check cloud cleanup

mrodm:
Add checks into cloud cleanup script so it is sent an email notification if there is any redshift cluster stale related to CI (integrations or elastic-package).

To be replaced by `cloud-reapder` when there is support for this kind of resources.

Example of build with this cloud cleanup step:
https://buildkite.com/elastic/elastic-package/builds/2527

Added input steps when this pipeline is triggered from the UI.

Relates #1700


.buildkite/configs/cleanup.aws.yml
```diff
@@ -0,0 +1,75 @@
...
+    options:
+      key: '${ACCOUNT_KEY}'
+      secret: '${ACCOUNT_SECRET}'
+
+scanners:
+  - account_name: "${ACCOUNT_PROJECT}"
+    resources:
+      - type: 'node'
+        regions:
+          - us-east-1
```
jsoriano, line 16: Do we have the region in some environment variable that we could use here?

.buildkite/scripts/cloud-cleanup.sh
```diff
@@ -0,0 +1,188 @@
...
+if [ "${resources_to_delete}" -eq 1 ]; then
+    message="There are resources to be deleted"
+    echo "${message}"
+    if running_on_buildkite ; then
+         buildkite-agent annotate \
+             "${message}" \
+             --context "ctx-cloud-reaper-error" \
+             --style "error"
+    fi
+    exit 1
```
jsoriano, line 111: Do we want to exit here? We could continue and check other resources, right?

mrodm: Thanks, I forgot to remove this exit!
I've updated the code to allow running all steps and just exit with failure at the end.

.buildkite/configs/cleanup.aws.yml
mrodm: Currently, it is using the default region set in the terraform files, there is no value set in our scripts. For instance:
https://github.com/elastic/elastic-package/blob/e74c0219d45746fba0a6340cfefcda88509791f0/test/packages/parallel/aws/data_stream/ec2_metrics/_dev/deploy/tf/env.yml#L9

https://github.com/elastic/integrations/blob/4efd95037467d5e54270006bcd71c3b0b6982932/packages/cisco_umbrella/_dev/deploy/tf/env.yml#L9

Should we add a value there in our scripts for both repositories or rely on that default value?
Currently, all terraform tags are set in the `pre-command` hook:
https://github.com/elastic/elastic-package/blob/346081a12249f64ff4b523effb68888835d8644c/.buildkite/hooks/pre-command#L14

Maybe, we could add/set there the variable `export AWS_REGION=us-east-1` too.

jsoriano (approved):

Issue #1649, opened 2024-01-29
Migrate to ecs@mappings template


zmoog:
This issue tracks the activities needed to achieve the following goals in the Agent integrations ecosystem:

- Stop using the legacy mappings
- Start using the modern ecs@mappings from Elasticsearch

```[tasklist]
### Tasks
- [ ] https://github.com/zmoog/public-notes/issues/78
- [ ] https://github.com/elastic/package-spec/pull/726
- [ ] https://github.com/elastic/elastic-package/pull/1711
- [ ] https://github.com/elastic/integrations/pull/9268
```

zmoog:
I spent some time trying to learn what we have now and what we need to do in Step 3.

**What we have now**

I see elastic-package, when users set `dependencies.ecs.import_mappings` to `true`, can:

- Import the ECS schema to validate the fields.
- Embed a "legacy" version of the ECS dynamic mappings.


**What we need to do in the step 3**

My current understanding is we should:

- Continue using ECS schema to validate fields
- Stop embedding the "legacy" ECS dynamic mappings
- Start leveraging the `ecs@mappings` component template when it is available for integrations. 

**Questions**

We currently use `dependencies.ecs.reference` to understand which ECS version the package targets. However, once we add `ecs@mappings` to the integration component template, it will be available out-of-the-box from ES. Should we add some mechanism to the package-spec to avoid this from happening automatically, and allow developers to test their packages before opt-in to `ecs@mappings`?

jsoriano:
> Continue using ECS schema to validate fields

Yes, we need to continue doing this in elastic-package. What we can maybe change now is to somehow obtain the same mapping Elasticsearch is using, maybe copying it directly from Elasticsearch repo (from [here?](https://github.com/elastic/elasticsearch/blob/1395edf8050584a0784e7d012005a26477b30ae4/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json)).

> * Stop embedding the "legacy" ECS dynamic mappings
> * Start leveraging the `ecs@mappings` component template when it is available for integrations.

I wonder if this could be handled by Fleet. The logic could be like this:
* If the manifest contains dynamic templates prefixed by `_embedded_ecs-` (under `elasticsearch.index_template.mappings.dynamic_templates`)
  * These dynamic templates are not installed.
  * The `ecs@mappings` component template is added.

[...]

PR #1711, opened 2024-03-05
Validate fields using ECS JSON schema when all stack versions support ecs@mappings

zmoog:
### Change description

Enable `elastic-package` to import the ECS JSON schema to validate fields in two conditions:

- When package developers enabled import_mappings (existing behavior).
- When package developers did NOT enable import_mappings, but all stack versions in the Kibana constraints support the `ecs@mappings` in Elasticsearch (8.13.0+).

If the package supports one or more stack versions without ECS mappings, devs must support ECS using `import_mappings` (deprecated) or `ecs.yml` file. 

### Motivation

On 8.13, we [enabled](https://github.com/elastic/kibana/issues/174905) Fleet to include the `ecs@mappings` component template in all the integration index templates.  So, starting from 8.13, integration developers can stop using `import_mappings` and the `ecs.yml`. 

However, `elastic-package` should import the ECS JSON schema only if ALL stack versions in the Kibana constraints have ECS mappings from Elasticsearch. 

### Additional Notes

- https://github.com/elastic/elastic-package/issues/1649

### Reviewer checklist

* [x] PR address a single concern.
* [x] PR title and description are appropriately filled.
* [x] Changes will be merged in `main.`
* [x] Changes are covered by tests.
* [x] Logging is meaningful in case of troubleshooting.
* [ ] Docs are updated (at least the `README.md`, if needed).
* [x] History is clean, commit messages are meaningful (see `CONTRIBUTING.md`) and are well-formatted.


zmoog:
@jsoriano, the gist of this change is to enable `elastic-package` to import the ECS JSON schema in two conditions:

1. When package developers enabled `import_mappings` (existing behavior).
2. When package developers did NOT enable `import_mappings,` and all stack versions in the Kibana constraints support the ecs@mappings in Elasticsearch.

Kibana constraints can be as simple as `8.0.0` or more sophisticated, like  `7.14.0 | | 8.0.0`. 

The [semver](https://github.com/Masterminds/semver) package we're using does not offer to get the minimum supported version out of a constraints expression, and I want to try something simpler before custom parsing it.

So I ran a little [experiment](https://github.com/elastic/elastic-package/pull/1711/commits/96f3418f16252e6f0a6adae7c941d589a46c5df1) to infer if the Kibana constrains support a stack version without ECS support.

Let me know what you think.

internal/fields/validate.go
```diff
@@ -227,6 +257,185 @@ func initDependencyManagement(packageRoot string, specVersion semver.Version, im
...
+	//
+	// However, if the constraint is `>= 8.0.0, < 8.10.0` the check will not
+	// return the right result.
+	//
+	// To support this, we would need to check the constraint against a larger
+	// set of versions, and check if the constraint is satisfied for all
+	// of them, like in the commented out example above.
+	//
+	// lastStackVersionWithoutEcsMappings := semver.MustParse("8.12.999")
+	// return !kibanaConstraints.Check(lastStackVersionWithoutEcsMappings)
```
zmoog, line 436: This is a simpler implementation that checks the Kibana constraints agains the latest known stack version that does NOT support `ecs@mappings`.

This check works under the assumption the constraints are not limited upwards (for example, `^8.10.0` or `^7.14.0 || ^8.0.0`).

I plan to remove at the end of the review process, but I want to mention it as an option.

jsoriano: This option is actually not bad, but yeah, let's go with the other one that looks safer.

internal/fields/validate.go
```diff
@@ -215,8 +215,38 @@ func initDependencyManagement(packageRoot string, specVersion semver.Version, im
...
+		return nil, nil, fmt.Errorf("can't read package manifest: %w", err)
+	}
+
+	if len(packageManifest.Conditions.Kibana.Version) > 0 {
+		kibanaConstraints, err := semver.NewConstraint(packageManifest.Conditions.Kibana.Version)
+		if err != nil {
+			return nil, nil, fmt.Errorf("invalid constraint for Kibana: %w", err)
+		}
+		stackIncludesEcsMapping = allVersionsIncludeECS(kibanaConstraints)
+	}
```
jsoriano, line 242: Nit. Consider moving this to its own function.

internal/fields/validate.go
```diff
@@ -215,8 +215,38 @@ func initDependencyManagement(packageRoot string, specVersion semver.Version, im
...
+
+	// If the package embeds ECS mappings, or the stack version includes ECS mappings, then
+	// we should import the ECS schema to validate the package fields against it.
 	var schema []FieldDefinition
-	if buildManifest.ImportMappings() && !specVersion.LessThan(semver2_3_0) && importECSSchema {
+	if (packageEmbedsEcsMappings || stackIncludesEcsMapping) && importECSSchema {
```
jsoriano, line 247: Please, add or modify a test package so we have a test case with a package that uses ECS fields, but doesn't embeds the mappings.

internal/fields/validate.go
```diff
@@ -227,6 +257,185 @@ func initDependencyManagement(packageRoot string, specVersion semver.Version, im
...
+		"8.10.4",
+		"8.11.0",
+		"8.11.1",
+		"8.11.2",
+		"8.11.3",
+		"8.11.4",
+		"8.12.0",
+		"8.12.1",
+		"8.12.2",
+		"8.2.0",
```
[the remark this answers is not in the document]
zmoog: Thanks for the reminder; I had this on my mental to-do list. The natural order of versions hurts my mind.

flash1293 (commented):
New to the project so can't comment on small stylistic things or anything.

About the approach itself - I know I suggested this approach with listing all existing stack versions @zmoog , but looking at it now there is actually another case we don't capture here. Due to security updates, it's possible to see new patch versions for older major/minor versions (e.g. `7.17.19` could be added). As this new version won't be included, it opens a small theoretical gap in this check.

Two options:
* Just live with this - it probably won't cause issues in practice
* Come up with a process to update this list either in a runtime fashion (load the list of versions from some central place during execution) or via a code change to elastic-package

It feels like the coverage by this check is good enough and we should just move forward with this approach for now - we can always revisit and I honestly don't expect this to cause issues as I can't see why someone would retroactively add Kibana version compatibility with just a single patch release of a previous version.

zmoog:
> Two options:
> 
> * Just live with this - it probably won't cause issues in practice
> * Come up with a process to update this list either in a runtime fashion (load the list of versions from some central place during execution) or via a code change to elastic-package
> 
> It feels like the coverage by this check is good enough and we should just move forward with this approach for now - we can always revisit and I honestly don't expect this to cause issues as I can't see why someone would retroactively add Kibana version compatibility with just a single patch release of a previous version.

It's a fair point. 

Wha't the kibana constraints expression that would break the current implementation?

internal/fields/validate.go
zmoog: If we want to test the scenario where the package uses the ECS in Elasticsearch, we need a stack version 8.13.0. 

@mrodm, which version of the stack the tests run on? Can we have stack version 8.13.0 (for all tests or just one)?

flash1293:
> Wha't the kibana constraints expression that would break the current implementation?

`7.17.19 || > 8.13`

zmoog:
> 7.17.19 || > 8.13

Even if unlikely, it's still possible. 

I would address this potential risk by adding a few extra upcoming 7.17.x versions, and move the PR forward with this implementation. 

I can create a separate issue to reimplement the `allVersionsIncludeECS()` with an HTTP call to something like the GitHub releases. 

I want to enable the elastic package to deal with ECS mappings and allow the remaining step of the ecs@mappings migration to happen as soon as possible.

WDYT?

flash1293:
@zmoog sounds good to me, let's move forward with this for now and create a follow-up issue.

mrodm (commented):
Thanks @zmoog !

Just minor suggestions.

internal/fields/validate.go
```diff
@@ -215,8 +301,26 @@ func initDependencyManagement(packageRoot string, specVersion semver.Version, im
...
+	stackSupportsEcsMapping, err := supportsECSMappings(packageRoot)
+	if err != nil {
+		return nil, nil, fmt.Errorf("can't check if stack version includes ECS mappings: %w", err)
+	}
+
+	// If the package embeds ECS mappings, or the stack version includes ECS mappings, then
+	// we should import the ECS schema to validate the package fields against it.
 	var schema []FieldDefinition
-	if buildManifest.ImportMappings() && !specVersion.LessThan(semver2_3_0) && importECSSchema {
+	if (packageEmbedsEcsMappings || stackSupportsEcsMapping) && importECSSchema {
```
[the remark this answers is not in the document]
zmoog: Good point! And I  remember considering adding it, but but it slipped my mind. I'm doing it now.

internal/fields/validate.go
zmoog: > In this implementation, the ECS fields are imported before validation by the test runners (e.g. kibana version condition is `^8.13.0`), so there should not be errors related to field definitions. Is it required to start an Elastic stack ^8.13.0 too? 

Correct! 

elastic-package validates the ECS fields using the JSON schema during the static checks, so it does not need to run on Elastic stack ^8.13.0.

mrodm (approved):
Thanks @zmoog ! :+1:

jsoriano (approved):

Issue #1610, opened 2023-12-21
Validate command for benchmark templates


ruflin:
When building a template for a benchmark, currently the only way to validate the template and test it, is shipping it to Elasticsearch with the stream command, for example:

```
elastic-package benchmark stream --benchmark error-benchmark -vv
```

Instead, it would be nice to be able to do multiple things with and without Elasticsearch:

* Validate if the template can be generated
  * Validate that it is JSON
  * Validate that it works for X number of events
* Preview output
  * Show on the console how an event looks like that is sent to Elasticsearch
* Test sending event
  * X number of events are sent to Elasticsearch, on success, read out of ES again and printed to console with the result. In case of error, print out the full error from Elasticsearch

The above should make developing templates much quicker. 

Eventually, at least the validation step should also be run as part of the check command to ensure all templates in a package are valid.

ruflin:
One additional thing that would be help is being able to get a sample event on how data would be shipped by elastic-agent before. I wonder if the combination of what @marc-gr did where schema A is collect, with the sample log messages we have, if there is an easy way to generate schema B? Taking schema B, the for the template only variables have to be filled in.

ruflin:
@aspacca @aliabbas-elastic @jsoriano I would hope such a command would help with errors we found in https://github.com/elastic/integrations/pull/8800 For example, if templates exists, we can do a quick test run for the package and if it fails, we also see it directly on the PR itself.

aspacca:
> I wonder if the combination of what @marc-gr did where schema A is collect, with the sample log messages we have, if there is an easy way to generate schema B? Taking schema B, the for the template only variables have to be filled in.

a way to achieve the above is to apply a pipeline like the one from @marc-gr for system tests, and then read back the document from ES to use as base for the schema-B template.
at least for logs the solution should be generic enough.

aspacca:
There're valid reasons to split "validation" and "testing" in two different commands.
Mainly due to the fact that one requires a cluster and the other not.

I'm not sure about the "preview": it either could be enabled with a flag on both commands or another separated command, I have no strong opinion on that.

I suggest to start with the "validation" command.

@ruflin are we on the same page?

ruflin:
> @ruflin are we on the same page?

Yep. In general I see this more as a meta issue and we will get to the end goal in small steps.

kpollich:
We haven't revisited this issue in a while, so we're closing it due to inactivity. Please reopen with a comment if you think this is a mistake. Thanks!

PR #1718, opened 2024-03-08
Add benchmark validate command for static validation

flash1293:
Context: I'm mostly working on this to familiarize myself with the elastic-package project (and golang in the first place).

Related to https://github.com/elastic/elastic-package/issues/1610

This PR adds a `validate` command to benchmarks to statically validate the benchmark generator is configured properly:

In this example there is a missing `{` for the data_stream object in the template of one out of multiple templates:
```
> elastic-package benchmark validate
Run benchmark asset validation
Error: validation failed: [access-benchmark] failed to unmarshal json event: invalid character ':' after object key:value pair, generated output:
{
    "@timestamp": "2024-03-08T16:55:55.248+01:00",
    "agent": {
        "ephemeral_id": "mistycarver",
        "id": "ef5e274d-4b53-45e6-943a-a5bcf1a6f523",
        "name": "cottonhugger",
        "type": "filebeat",
        "version": "8.5.1"
    },
    "data_stream":
        "dataset": "nginx.access",
        "namespace": "ep",
        "type": "logs"
    },
...
```

It generates 10 events using each generator and checks whether the result is valid JSON.

This does not require a running elasticsearch instance, it's purely checking the benchmark configuration. This implementation validates the following things:
* All config files are in their expected location
* The template is valid
* The generated output is valid JSON

## Implementation details

As this is very close to the existing stream logic, I made it part of that module. I pondered to split the necessary bits out into a common and validation-specific bit, but it seemed overkill.

The existing `setUp` routine is talking to Elasticsearch already, so I split out the static parts into an `initialize` function which can be called for the validation before actually querying the generators.

@ruflin feel free to add the right people as reviewers


internal/benchrunner/runners/stream/runner.go
```diff
@@ -120,15 +124,58 @@ func (r *runner) setUp() error {
...
-
 	err = r.collectGenerators()
 	if err != nil {
 		return fmt.Errorf("can't initialize generator: %w", err)
 	}
 
+	return nil
+}
+
+const NumberOfEvents = 10
```
ruflin, line 135: Nit: It is common to have const on top of the doc. And as it is not used outside the scope (yet), it can start with lowercase.

ruflin:
This is a great first step for validation. Did you test what kind of error is returned if you made an error in the template itself that the template is invalid instead of the JSON?

flash1293:
> Did you test what kind of error is returned if you made an error in the template itself that the template is invalid instead of the JSON?

@ruflin In this case the error looks like this:

```
Run benchmark asset validation
Error: validation failed: can't initialize generator: [access-benchmark]: template: generator:35: undefined variable "$remoteUser"
```

ruflin:
@flash1293 Neat, that already gives a lot more details on what goes wrong and makes it easy to debug.

internal/benchrunner/runners/stream/runner.go
```diff
@@ -295,25 +341,25 @@ func (r *runner) collectGenerators() error {
 	for scenarioName, scenario := range r.scenarios {
 		config, err := r.getGeneratorConfig(scenario)
 		if err != nil {
-			return err
+			return fmt.Errorf("[%s]: %w", scenarioName, err)
```
jsoriano, line 344: Thanks for adding more context here, I would also give context about the action that has failed here, something like this:
```suggestion
			return fmt.Errorf("failed to obtain generator configuration for scenario %q: %w", scenarioName, err)
```

cmd/benchmark.go
```diff
@@ -85,6 +85,9 @@ func setupBenchmarkCommand() *cobraext.Command {
 	systemCmd := getSystemCommand()
 	cmd.AddCommand(systemCmd)
 
+	validateCmd := getValidateCommand()
+	cmd.AddCommand(validateCmd)
```
jsoriano, line 89: Instead of adding this as a new subcommand, I think we could add this as part of one of the subcommands we have now for validation. This would have the benefit of being executed by CI.

I think a good place could be in `elastic-package test static`, there we are only checking the sample events, but checking the events generated by `stream` could also fit there.

Other place could be `elastic-package check`, but this is more concerned about the structure of the package more than its behavior.

internal/benchrunner/runners/stream/runner.go
```diff
@@ -120,15 +126,55 @@ func (r *runner) setUp() error {
...
+			err := generator.Emit(buf)
+			if err != nil {
+				return fmt.Errorf("[%s] error while generating event: %w", scenarioName, err)
+			}
+			// check whether the generated event is valid json
+			var event map[string]any
+			err = json.Unmarshal(buf.Bytes(), &event)
+			if err != nil {
+				return fmt.Errorf("[%s] failed to unmarshal json event: %w, generated output: %s", scenarioName, err, buf.String())
+			}
```
jsoriano, line 150: It would be also nice to validate the fields, as we do in tests and when validating the sample documents, for example here: https://github.com/elastic/elastic-package/blob/a3e3a802b2d2c3f933c4e151f32cf12d5945005f/internal/testrunner/runners/static/runner.go#L117

jsoriano (commented):
Thanks for adding this! This is going to be really useful when preparing benchmarks.

I wonder if we could add this as part of `elastic-package test static`, this is used now to validate sample documents, and the stream subcommand is kind-of an automation for the same thing. We could also validate fields there.

aspacca (approved):

internal/benchrunner/runners/stream/runner.go
```diff
@@ -37,6 +37,8 @@ import (
 	"github.com/elastic/elastic-package/internal/signal"
 )
 
+const numberOfEvents = 10
```
aspacca, line 40: beware that 10 events  might not be enough to check an error in a template

for example this template: https://github.com/elastic/integrations/blob/main/packages/aws/_dev/benchmark/rally/ec2metrics-benchmark/template.ndjson#L117-L210

here we render different content according the the value of `$dimensionType`: https://github.com/elastic/integrations/blob/main/packages/aws/_dev/benchmark/rally/ec2metrics-benchmark/config.yml#L4-L12

there is 1/40 possibility that a specific content will be rendered

internal/benchrunner/runners/stream/runner.go
flash1293: I looked into this, but the fields validator partially depends on test configurations like `numeric_keyword_fields` which might not be available / can't be matched properly to the benchmark config.

I would like to split that part out of this PR if you don't mind.

cmd/benchmark.go
flash1293: It makes sense to me to test as much as possible via a central command - I added this to `test static`, but IMHO there is value in having a dedicated command as well to be able to iterate on only the benchmark config.

What do you think about the most recent version that wires up the same logic in both places?

internal/benchrunner/runners/stream/runner.go
flash1293: I bumped it to 100, the generation is fairly quick, it shouldn't matter. Is this enough or should we go even higher?

flash1293:
Thanks @aspacca and @jsoriano for review, could you have another look?

flash1293:
I'm seeing the same test failure on other PRs a well, I'm not sure it is related

jsoriano (commented):
Looking to this change makes me think that we should move benchmarks from the root of the package to each data stream, as they are so far specific to data streams. Having benchmarks disconnected from data streams is going to cause problems, as the ones I mention of running repeated tests if we integrate this validation into test subcommands.

An option to progress with this change before these refactors could be to add this `StaticValidation` as a fail fast mechanism to `benchmark rally` and `benchmark stream`. Developers could use `benchmark stream` to quickly validate the generated data.

internal/testrunner/runners/static/runner.go
```diff
@@ -72,7 +74,33 @@ func (r runner) run(ctx context.Context) ([]testrunner.TestResult, error) {
 		return result.WithError(fmt.Errorf("failed to read manifest: %w", err))
 	}
 
-	return r.verifySampleEvent(pkgManifest), nil
+	// join together results from verifyStreamConfig and verifySampleEvent
+	return append(r.verifyStreamConfig(ctx, r.options.PackageRootPath), r.verifySampleEvent(pkgManifest)...), nil
```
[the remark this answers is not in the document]
ruflin: We had the discussion in the past about where to start. If I remember correctly, we went for global first as it was easier to implement and kept the option open, that a single benchmark could be run for more then just one data stream. Like you generate a rally track for all k8s data in parallel, you don't test just one data stream at the time. 

Overall I share your sentiment, that the benchmark definition should be part of the dataset, but there could be global scenarios (likely even across integrations). The corpus tool had it fully disconnected which was not ideal either from my perspective.

ruflin:
@jsoriano It would be nice to still have the option for a quick win here without requiring refactoring to data streams to have validation.

jsoriano:
> @jsoriano It would be nice to still have the option for a quick win here without requiring refactoring to data streams to have validation.

Yep, agree, this is why I propose to do the validation as a pre-check in the commands that use these files, it is pretty fast.

> An option to progress with this change before these refactors could be to add this `StaticValidation` as a fail fast mechanism to `benchmark rally` and `benchmark stream`. Developers could use `benchmark stream` to quickly validate the generated data.

internal/testrunner/runners/static/runner.go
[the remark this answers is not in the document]
jsoriano: > a single benchmark could be run for more then just one data stream

Looking to the templates we have now, I think it would be pretty convoluted to create one to generate data suitable for multiple data streams, so even when they are at the root of the package, and run at the same time, they are specific to each data stream. I think we could have the definitions per data-stream, and still be able to run multiple of them at the same time.

internal/testrunner/runners/static/runner.go
```diff
@@ -72,7 +74,33 @@ func (r runner) run(ctx context.Context) ([]testrunner.TestResult, error) {
...
+}
+
+func (r runner) verifyStreamConfig(ctx context.Context, packageRootPath string) []testrunner.TestResult {
+	resultComposer := testrunner.NewResultComposer(testrunner.TestResult{
+		Name:       "Verify benchmark config (if available)",
```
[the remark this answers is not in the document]
jsoriano: At the moment all benchmarks are named `<data stream>-benchmark`. We could maybe follow this convention here to check if the config is available and to run the checks per data-stream.

flash1293:
Thanks for the discussion @jsoriano and @ruflin 

Agreed that having the benchmark config defined on the stream level looks like it makes sense, but of course it's a bigger refactoring.

> An option to progress with this change before these refactors could be to add this StaticValidation as a fail fast mechanism to benchmark rally and benchmark stream. Developers could use benchmark stream to quickly validate the generated data.

We can start with that, but I really like the point you brought up about this being checked in CI - having it as part of rally and stream would only marginally speed up these commands failing.

Two other thoughts:

What do you think about making the `static` runner work on both packages and data streams? This is not a category right now, but we could add it by a similar flag like `CanRunPerDataStream` defined on each runner: `CanRunForPackageRoot` - the static testrunner would return true for both and its run method would be called for both the package and each individual data stream (and the test runner would decide based on that what static checks to run). This seems like it could be useful more universally as well.

What do you think about adding this check to a new test command (`test benchmark`) which can then be run in CI along with the static test command?

jsoriano:
> What do you think about making the `static` runner work on both packages and data streams?

> What do you think about adding this check to a new test command (`test benchmark`) which can then be run in CI along with the static test command?

I would be cautious about adding these features, we would be circumventing current tech debt, by adding more tech debt.

Current benchmark scenarios are already bound to data streams, they require a data stream name in their configs, we don't really need to support them at the package level. `test static` looks like a good place to add this, because it is already being used to test sample documents, and runs per data stream.

What do you think about my comment [here](https://github.com/elastic/elastic-package/pull/1718#discussion_r1540816959)?

We could read the scenarios, and from them know if there is any scenario for a given data stream, and if there are, validate only them. This would allow to include this in the current `test static`.

flash1293:
> What do you think about my comment https://github.com/elastic/elastic-package/pull/1718#discussion_r1540816959?

> We could read the scenarios, and from them know if there is any scenario for a given data stream, and if there are, validate only them. This would allow to include this in the current test static.

As they are already bound to the data stream this way I agree, let's do it like that. I will update the PR.

flash1293:
@jsoriano could you take another look?

jsoriano (commented):
Looks great now, thanks!

flash1293:
Thanks @jsoriano , updated

jsoriano (approved):
Thanks!

Issue #787, opened 2022-04-11
Add ability to deploy custom elastic-agents


marc-gr:
There are some integrations that might require elastic-agents with custom configurations for them or their container ie: winlogbeat requires a windows container, auditbeat requires special container capabilities, etc.

I initially created https://github.com/elastic/elastic-package/pull/786 that adds ability to deploy custom agents as test services, there is still missing code specific to deal with the windows scenario.

I open this thread to discuss other approaches that might avoid adding the complexity to the test runner if possible.

mtojek:
pinging @jsoriano for his thoughts around this and https://github.com/elastic/elastic-package/pull/786.

jsoriano:
Another option could be to move agent initialization to the system test runner, and remove it from the default stack definition. Having it in the runner would allow it to have full control of the started agents, allowing to start them with different options, and handling platform-specific needings, as could be the case of Windows.
It could also allow to start different tests with different agent configuration, for example to test different auditbeat configurations with different capabilities.

We already have a custom agent for the Kubernetes service deployer, we could follow a similar strategy on any other deployer.
There could be general options, such as options to select the version to use, or things like capabilities to add. And there could be platform specific options, that would also allow things like selecting daemonset vs deployment in Kubernetes (https://github.com/elastic/elastic-package/issues/465). On a second iteration these options could be easily overridden with flags or environment variables.

If we remove agent from the default stack definition, we could still have a stack subcommand to start agents for manual tests. Something like this could also cover https://github.com/elastic/elastic-package/issues/548.

I think this could be a more future-proof option, but it can require an important effort.

---

And another option for the use case of starting an agent but no service, can be to add a [new "system" deployer](https://github.com/elastic/elastic-package/blob/bbdb1f3238ac78b17fdeedc72a75c2e4fef52997/internal/testrunner/runners/system/servicedeployer/factory.go#L40), that just starts an agent with a given configuration, intended for system-level monitoring. This could help with packages for auditbeat or for the system module itself. This could be extended in the future to start completely different OSs using VMs.

This would be more in line of https://github.com/elastic/elastic-package/pull/786, but without needing to hack over the current test runner and compose deployer.

mtojek:
> Another option could be to move agent initialization to the system test runner, and remove it from the default stack definition. Having it in the runner would allow it to have full control of the started agents, allowing to start them with different options, and handling platform-specific needings, as could be the case of Windows.
[...]

Issue #1586, opened 2023-12-06
[System tests] Run system tests for integrations with a non root Elastic Agent


jlind23:
With the addition of the two features below, integrations will now be able to run on non root Elastic Agent. If not specified otherwise, an integration will be runnable on a non root Agent. In order to ensure that integrations are properly running without problems we need to update Elastic Package system test in order to run a non root Elastic Agent.

## Related issues
- https://github.com/elastic/package-spec/issues/537
- https://github.com/elastic/elastic-agent/pull/3598

Blocked on https://github.com/elastic/elastic-package/issues/787

cc @nimarezainia @pierrehilbert

jlind23:
@blakerouse I guess we will need https://github.com/elastic/elastic-agent/pull/3799 to be reverted before being able to run these tests right?

@mrodm @jsoriano I guess we will have to add this flag to the Elastic Agent docker image we use right?

pierrehilbert:
@jlind23 the flag is only hidden so I think there is no problem to run these tests

jlind23:
@jsoriano @kpollich After chatting with @blakerouse it seems that we should already be able to do it by changing the user that the docker image relies on.
@jen-huang this is then ready to be picked up whenever someone has time.

jsoriano:
> it seems that we should already be able to do it by changing the user that the docker image relies on.

One challenge we will face is that `elastic-package test` doesn't manage the Elastic Agent used for tests, it is managed by `elastic-package stack` and is independent of the package to be tested. So as things are now, we cannot just change the docker user depending on the package.

We have though a number of issues that would be solved if we move the Elastic Agent management from the `stack` subcommand to the `test` one. This is specially discussed in https://github.com/elastic/elastic-package/issues/787, that could be a generalization of this issue.

The kubernetes and custom agent service deployers, used by some integrations, circumvent this limitation by managing their own agents, but this exceeds their responsibilities, and cannot be generalized to all integrations.

I think we should take the opportunity to dedicate some time to solve https://github.com/elastic/elastic-package/issues/787. It would help on this issue and other similar ones, and it would allow to pay off some technical debt.

If we cannot dedicate time to this now, we could add a temporary profile setting, that makes `elastic-package stack up` aware of the root requirement of the package in the current directory. This will likely be enough for CI, as we start one stack per package, but can lead to unexpected situations when working locally on multiple packages with the same stack.

jlind23:
[...]

PR #1724, opened 2024-03-13
Create different agent per each test execution

mrodm:
This PR adds support to create a new agent per each test run.

Created a new module `agentdeployer` that is in charge of creating agents separately from the `servicedeployer` module.

Added a new environment variable ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT to enable creating independent agents for tests (system tests). In order to enable this feature, this environment variable must be set to `true`:
```shell
ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT=true elastic-package test system -v

ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT=true elastic-package test system --config-file <path> --setup --test-independent-agent
ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT=true elastic-package test system --test-independent-agent --no-provision
ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT=true elastic-package test system --test-independent-agent --tear-down
```

Pending:

- [x] Test with Terraform services (gcp, aws and aws_logs).
- [x] Issues using stages in system tests
    - Check enrolled agents does not find any agent
    - ~Use of tags in elastic agents instead of checking enrolling time ? https://www.elastic.co/guide/en/fleet/current/filter-agent-list-by-tags.html#add-tags-at-enrollment~
    - Create Test Policy in advance, so agents can be enrolled in that policy and filtered with it.
- [x] Issues with tests using `elastic-agent` as a hostname to run requests
    - [x] Choose between environment variable or hard-coded variables
    - Added environment variables but also ensure that it can be used `elastic-agent` alias as previously.
- [x] Test with previous versions
    - [x] Elastic stack 7.17.x (e.g. 7.17.18) - Policy name is not the same
    - [x] Elastic stack >=8.0.0,<8.3.0 - No tags in this version.
        - Use agent hostname adding as suffix the Test RunID (integer).
        - ~Kubernetes requires to use tags to try to distinguish from other agents.~ Filtered agents with Policy ID (created in advance)
- [x] Ensure logs are reviewed after each test: elastic-agent does not run in the same docker-compose project
    - Working for docker-compose scenarios
- [x] Update if needed `internal/service/boot.go`
    - Added a new options field to keep the previous behavior (DeployIndependentAgent).
- [x] Update benchrunner system to use also agentdeployer module
    - Added a new options field to keep the previous behavior (DeployIndependentAgent).
- [x] Issues with enrollment token after some executions
    - More info at https://github.com/elastic/elastic-package/pull/1724#issuecomment-1998093975
- [ ] custom agents (`_dev/deploy/agent`) should be moved to `agentdeployer` module?
    - These compose scenarios contain both agent and service definitions
- [ ] Add system configuration name into the agent hostname?
    - Not sure if needed, would that allow to run more tests in parallel ?
- [ ] Include variants in custom agents (servicedeployer) ?
- [ ] Allow more than one test run in parallel for Kubernetes? It would require allowing more than one Elastic agent in Kubernetes.

Relates #787


mrodm:
Containers are using the alias `elastic-agent` to run requests to the agent. That happens for instance in these test packages: `hits_count_assertion` and `ti_anomali`. Example:

- https://github.com/elastic/elastic-package/blob/12a8d0df74150ad70d606f3db27cf005d0a5e347/test/packages/parallel/ti_anomali/_dev/deploy/docker/docker-compose.yml#L21

Two different solutions here:
- Update each package to use the corresponding new hostname in these commands
    - Used for now this approach.
- Try to add a new environment variable that contains this information.

mrodm:
After some time the Elastic stack is running, this error is raised when trying to enroll a new elastic-agent 

```
docker-custom-agent-1  | Kibana fetch policy failed: http GET request to https://kibana:5601/api/fleet/agent_policies fails: fail to execute the HTTP GET request: Get "https://kibana:5601/api/fleet/agent_policies": lookup kibana on 127.0.0.11:53: server misbehaving. Response: 
docker-custom-agent-1  | Error: unable to find enrollment token named "Default" in policy "Elastic-Agent (elastic-package)"
```

.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -73,6 +76,10 @@ done
 pushd test/packages/parallel > /dev/null
 for package in $(find . -maxdepth 1 -mindepth 1 -type d) ; do
     package_name=$(basename "${package}")
+    if [[ "$package_name" == "aws" || "$package_name" == "aws_logs" || "$package_name" == "gcp" ]] ; then
+        echoerr "Skip temporarily ${package_name}"
+        continue
+    fi
```
mrodm, line 82: To be removed, first try with the test packages that are not using terraform.

internal/agentdeployer/_static/elastic-agent-managed.yaml.tmpl
mrodm, line 1: These changes allowed me to create different elastic-agents inside the kubernetes cluster created that it has just one node. Not sure how this would work with different setups.

Related to this comment in the code: 
https://github.com/elastic/elastic-package/blob/a3e3a802b2d2c3f933c4e151f32cf12d5945005f/internal/kubectl/kubectl_apply.go#L132-L136

test/packages/parallel/ti_anomali/_dev/deploy/docker/files/test-intel.ndjson
```diff
@@ -1,3 +1,3 @@
-{"message_type":"start","integrator_metadata":{"user_metadata":"{\"url\":\"http://elastic-agent:9080/\",\"secret\":\"TheSecret\",\"request_timeout\":10}","ioc_field_names":["domain","itype","classification","lat","update_id","source_feed_id","id","confidence","severity","trusted_circle_ids","lon","date_first","source","state","import_session_id","value_type","srcip","org","date_last","country","detail2","resource_uri","added_at"]}}
```
mrodm, line 1: Not able to use here environment variables.

internal/servicedeployer/compose.go
```diff
@@ -220,14 +227,18 @@ func (s *dockerComposeDeployedService) TearDown() error {
 		return fmt.Errorf("could not create Docker Compose project for service: %w", err)
 	}
 
+	extraArgs := []string{}
+	if s.shutdownTimeout != "" {
+		extraArgs = append(extraArgs, "-t", s.shutdownTimeout)
+	}
```
mrodm, line 233: Created a separate Pull Request for this change:
https://github.com/elastic/elastic-package/pull/1727

internal/servicedeployer/kubernetes.go
```diff
@@ -100,8 +93,8 @@ func NewKubernetesServiceDeployer(opts KubernetesServiceDeployerOptions) (*Kuber
 	}, nil
 }
 
-// SetUp function links the kind container with elastic-package-stack network, installs Elastic-Agent and optionally
-// custom YAML definitions.
+// SetUp function links the kind container with elastic-package-stack network, installs
+// custom YAML definitions if any.
```
mrodm, line 97: Elastic agent is now created inside the `agentdeployer`

internal/testrunner/runners/system/runner.go
```diff
@@ -683,14 +725,57 @@ func (r *runner) prepareScenario(config *testConfig, ctxt servicedeployer.Servic
...
+	logger.Debug("setting up agent...")
+	agentOptions := r.createAgentOptions(scenario)
+	agentInfo, err := r.createAgentInfo(scenario)
+	if err != nil {
+		return nil, err
+	}
+
+	// FIXME: this should be stored in service state or not used in
+	// --no-provision and --tear-down stages
+	enrollingTime := time.Now()
```
mrodm, line 738: To be tested if it can be used Elastic Agent tags to be able to select the right agent without depending on the enrolling time or the order in the response. Currently, it selects the first agent (item) in the response.

internal/kubectl/kubectl_delete.go
```diff
@@ -4,8 +4,17 @@
...
 	_, err := modifyKubernetesResources("delete", definitionsPath)
 	return err
 }
+
+// DeleteStdin function removes resources from the Kubernetes cluster based on provided definitions.
+func DeleteStdin(out []byte) error {
+	logger.Debugf("Delete Kubernetes stdin")
+	_, err := deleteKubernetesResourcesStdin(out)
+	return err
+}
```
mrodm, line 20: Previously, in Service Deployer, the elastic-agent in the kubernetes cluster was not deleted.
In the context of the PR / change, I think the elastic-agents should be deleted once the test is finished.

To be checked wether or not it would be possible to trigger as it is more than one test. Probably, it would be needed to create different names for the DaemonSet resource https://github.com/elastic/elastic-package/blob/a3e3a802b2d2c3f933c4e151f32cf12d5945005f/internal/servicedeployer/elastic-agent-managed.yaml.tmpl#L12

jsoriano:
> WDYT @jsoriano ? Should we go for the environment variables ?

Yes, the solution with the environment variable looks better.

Or could it be an option to rename the elastic agent started by `elastic-package stack up` and use `elastic-agent` alias for the container managed by the agent deployer?

jsoriano:
> Do these enrollment token expire ?

Actually I don't know if we are using enrollment tokens at all. I think we use the policy ID and the default credentials `elastic`/`changeme`, which happens to be enough.

But I think it would be good to migrate to use enrollment tokens, so for each test we create a new policy, get the enrollment key, and use it to create and enroll the new agent. This way we don't need to assume the presence of policies or credentials, and we test in a more similar way to a real deployment.

mrodm:
> Or could it be an option to rename the elastic agent started by `elastic-package stack up` and use `elastic-agent` alias for the container managed by the agent deployer?

The concern that I have with setting that alias is that there could not be two test runs in parallel, since two or more containers would have the same alias / hostname.

internal/servicedeployer/_static/elastic-agent-managed.yaml.tmpl
mrodm, line 1: Just moved to `_static` folder.

mrodm:
Changed network connections between agent and services.

Previously, service container just connected to the Elastic stack network (docker-compose project `elastic-package-stack`) where the agent was running:

```mermaid

flowchart LR
service[Service]
stack[Elastic stack]

service --> stack
```


With this PR, agents can be run independently on its own network (another docker-compose project). This allows the service to just connect to the agent network instead of the Elastic stack network and therefore keep using the same alias `elastic-agent` for all test suites:


```mermaid
flowchart LR
service[Service]
stack[Elastic stack]
agent[Agent]

service --> agent
agent --> stack
```

internal/agentdeployer/_static/elastic-agent-managed.yaml.tmpl
```diff
@@ -0,0 +1,295 @@
...
+metadata:
+  name: elastic-package-certs
+  namespace: kube-system
+data:
+  ca-cert.pem: {{ .caCertPem }}
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: elastic-agent
```
mrodm, line 12: This probably should be changed to allow different runs at the same time

internal/agentdeployer/info.go
```diff
@@ -0,0 +1,80 @@
...
+// templates in system test configuration files, for example: {{ Hostname }}.
+type AgentInfo struct {
+	// Name is the name of the service.
+	Name string
+
+	// Hostname is the host name of the service, as addressable from
+	// the Agent container.
+	Hostname string
+
+	NetworkName string
```
jsoriano, line 27: Nit. Document also this attribute?

internal/agentdeployer/deployed_agent.go
```diff
@@ -0,0 +1,34 @@
...
+	TearDown(ctx context.Context) error
+
+	// Signal sends a signal to the service.
+	Signal(ctx context.Context, signal string) error
+
+	// Info returns the current context from the service.
+	Info() AgentInfo
+
+	// SetInfo sets the current context for the service.
+	SetInfo(str AgentInfo) error
```
jsoriano, line 27: On what cases this would return an error?

internal/testrunner/runners/system/runner.go
```diff
@@ -1603,20 +1786,93 @@ func deleteDataStreamDocs(ctx context.Context, api *elasticsearch.API, dataStrea
...
+			logger.Debugf("filtered agent (policy id) %q", agent.ID) // TODO: remove
 			continue
 		}
+
+		if agent.Status != "online" {
+			logger.Debugf("filtered agent (not online) %q", agent.ID) // TODO: remove
+			continue
+		}
+
+		if runIndependentElasticAgent && agent.EnrolledAt.Before(threshold) {
```
jsoriano, line 1826: There are many ifs based on `runIndependentElasticAgent`. Should we have two different `filterAgent` methods?

internal/testrunner/runners/system/runner.go
```diff
@@ -1603,20 +1786,93 @@ func deleteDataStreamDocs(ctx context.Context, api *elasticsearch.API, dataStrea
 	return nil
 }
 
-func filterAgents(allAgents []kibana.Agent, svcInfo servicedeployer.ServiceInfo) []kibana.Agent {
+func filterAgents(allAgents []kibana.Agent, agentInfo agentdeployer.AgentInfo, threshold time.Time, svcInfo servicedeployer.ServiceInfo, runIndependentElasticAgent bool) []kibana.Agent {
```
jsoriano, line 1789: Quite a lot of complexity is added here. Could we maybe base the search on the policy name, and create different policy names on each execution with independent elastic agents?

jsoriano (commented):
Looks good overall, specially given that is hidden behind a feature flag. But I wonder if we are adding even more complexity to an already complex code. At some point we have to make a big refactor in the system test runner.

I wonder if there would be some way to abstract the code that we are branching on conditions based on `runIndependentElasticAgent`.

It would be also nice to add a test script for this.

jlind23:
@mrodm NIT - will this is close or only relates to https://github.com/elastic/elastic-package/issues/787

mrodm:
> @mrodm NIT - will this is close or only relates to #787

@jlind23 It only relates to #787.
This PR sets the foundations allowing us to start working on #1586 too. But there is still some work to do in following PRs (and maybe changes in package-spec) to allow all the scenarios of custom agents. For instance, probably it would be needed to change how custom agents are defined in packages. Currently, those packages define the agent and the service in the same docker-compose scenario, cc @jsoriano

internal/agentdeployer/deployed_agent.go
mrodm: I cannot think of any scenario where this could return an error.
I'll update this method to return nothing and update the corresponding implementations of `DeployedAgent`.

This was based on this interface (`DeployedService`):

https://github.com/elastic/elastic-package/blob/63624b4f4ccf07391099ef5c37b45fa4e331c42e/internal/servicedeployer/deployed_service.go#L26

Should we update that other interface too ?

internal/agentdeployer/deployed_agent.go
```diff
@@ -0,0 +1,34 @@
...
+
+var ErrNotSupported error = errors.New("not supported")
+
+// DeployedAgent defines the interface for interacting with a service that has been deployed.
+type DeployedAgent interface {
+	// TearDown implements the logic for tearing down a service.
+	TearDown(ctx context.Context) error
+
+	// Signal sends a signal to the service.
+	Signal(ctx context.Context, signal string) error
```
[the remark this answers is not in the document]
mrodm: Re-thinking about this one, it should not be needed to send a signal to the agent. I'll remove it.

internal/testrunner/runners/system/runner.go
mrodm: Sure, restored the previous `filterAgents` method and created a new one for the new independent agents.

internal/agentdeployer/agent.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	opts := compose.CommandOptions{
+		Env: append(
+			s.env,
+			s.variant.Env...,
+		),
+	}
+
+	return p.ServiceExitCode(ctx, agent, opts)
```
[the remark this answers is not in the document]
mrodm: Updated, this change allows to remove one parameter of the `ExitCode` function :+1:

internal/servicedeployer/terraform.go
```diff
@@ -113,15 +117,14 @@ func (tsd TerraformServiceDeployer) SetUp(ctx context.Context, svcInfo ServiceIn
 		env:             tfEnvironment,
 		shutdownTimeout: 300 * time.Second,
 	}
-	outCtxt := svcInfo
```
mrodm, line 116: Avoid creating a temporal variable

internal/testrunner/runners/system/runner.go
```diff
@@ -261,33 +264,82 @@ func (r *runner) Run(ctx context.Context, options testrunner.TestOptions) ([]tes
...
-	if err != nil {
-		return servicedeployer.ServiceInfo{}, fmt.Errorf("could not create output dir for terraform deployer %w", err)
+	if r.options.RunTearDown || r.options.RunTestsOnly {
+		logger.Debug("Skip creating output directory")
+	} else {
+		outputDir, err := servicedeployer.CreateOutputDir(r.locationManager, svcInfo.Test.RunID)
+		if err != nil {
+			return servicedeployer.ServiceInfo{}, fmt.Errorf("could not create output dir for terraform deployer %w", err)
+		}
+		svcInfo.OutputDir = outputDir
```
mrodm, line 341: Ensure that this folder is just created with --setup or regular testing commands.

The OutputDir and RunID values will be retrieved from ServiceState in case testing is run in stages (e.g. --no-provision)

internal/testrunner/runners/system/runner.go
```diff
@@ -692,32 +760,58 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, servic
 		return nil, fmt.Errorf("failed to find the selected policy_template: %w", err)
 	}
 
-	// Setup service.
-	logger.Debug("setting up service...")
-	serviceDeployer, err := servicedeployer.Factory(serviceOptions)
-	if err != nil {
-		return nil, fmt.Errorf("could not create service runner: %w", err)
+	// Configure package (single data stream) via Fleet APIs.
```
mrodm, line 763: Move creation of test Agent Policy as one of the first steps, so it can be used to enroll the new agents.

internal/testrunner/runners/system/runner.go
```diff
@@ -1005,6 +1091,94 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, servic
...
+
+	// Set the right folder for logs execpt for custom agents that are still deployed using "servicedeployer"
+	if r.options.RunIndependentElasticAgent && agentDeployed != nil {
+		svcInfo.Logs.Folder.Local = agentInfo.Logs.Folder.Local
+	}
+
+	// In case of custom agent (servicedeployer) enabling independent agents, update serviceOptions to include test policy too
+	if r.options.RunIndependentElasticAgent {
+		serviceOptions.PolicyName = agentInfo.Policy.Name
+	}
```
mrodm, line 1115: Currently, it is not set the Agent Policy name in all the scenarios, to avoid changing the current behaviour. Currently, in servicedeployer it is used the default policy `Elastic-Agent (elastic-package)`

jsoriano (approved):

jsoriano (approved):

jsoriano (approved):

internal/agentdeployer/_static/elastic-agent-managed.yaml.tmpl
mrodm: Follow-up PR #1759

PR #1727, opened 2024-03-14
Use custom shutdown time for service deployer compose

mrodm:
Set the shutdown timeout in docker-compose stop and docker-compose down customizable depending on the service deployer used.

Difference in times for `ti_anomali` test package:
- Before approx. 18 min:  https://buildkite.com/elastic/elastic-package/builds/2596#018e3955-cafd-45f0-ac31-e2faec7ded0d
- In this PR 9 min:  https://buildkite.com/elastic/elastic-package/builds/2606#018e3d1c-ebdc-4e50-9140-e940dba84cce

For packages using terraform, it still uses the timeout of 300 seconds:

https://buildkite.com/elastic/elastic-package/builds/2606#018e3d1c-ebdf-4d9c-bbfe-2ce7864a2bb5/115-1105

### DIFF
--- a/internal/servicedeployer/compose.go
+++ b/internal/servicedeployer/compose.go
@@ -42,6 +42,8 @@ type DockerComposeServiceDeployerOptions struct {
 type dockerComposeDeployedService struct {
 	ctxt ServiceContext
 
+	shutdownTimeout string // default shutdown timeout 10 seconds
+
 	ymlPaths []string
 	project  string
 	variant  ServiceVariant
@@ -225,9 +227,14 @@ func (s *dockerComposeDeployedService) TearDown() error {
 			s.env,
 			s.variant.Env...),
 	}
+
+	extraArgs := []string{}
+	if s.shutdownTimeout != "" {
+		extraArgs = append(extraArgs, "-t", s.shutdownTimeout)
+	}
 	if err := p.Stop(compose.CommandOptions{
 		Env:       opts.Env,
-		ExtraArgs: []string{"-t", "300"}, // default shutdown timeout 10 seconds
+		ExtraArgs: extraArgs,
 	}); err != nil {
 		return fmt.Errorf("could not stop service using Docker Compose: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


internal/servicedeployer/compose.go
```diff
@@ -42,6 +42,8 @@ type DockerComposeServiceDeployerOptions struct {
 type dockerComposeDeployedService struct {
 	ctxt ServiceContext
 
+	shutdownTimeout string // default shutdown timeout 10 seconds
```
jsoriano, line 45: Nit. Use `time.Duration` for this variable?

jsoriano, line 45: Is this intended to be internal only? Maybe we can expose it to the user through some test setting.

jsoriano (commented):
Looks good, added a couple of questions/suggestions.

internal/servicedeployer/compose.go
mrodm: Updated to use `time.Duration` :+1: 
I'll consider default value zero, as the timeout has not been set in that case. So it will use the default value from docker-compose 
https://docs.docker.com/compose/faq/#why-do-my-services-take-10-seconds-to-recreate-or-stop

mrodm: Updated here: https://github.com/elastic/elastic-package/pull/1727/commits/fd8321f920536ccb725eb727921578394505146b

internal/servicedeployer/compose.go
```diff
@@ -229,8 +230,10 @@ func (s *dockerComposeDeployedService) TearDown() error {
 	}
 
 	extraArgs := []string{}
-	if s.shutdownTimeout != "" {
-		extraArgs = append(extraArgs, "-t", s.shutdownTimeout)
+	// if not set "-t" , default shutdown timeout is 10 seconds
+	// https://docs.docker.com/compose/faq/#why-do-my-services-take-10-seconds-to-recreate-or-stop
+	if s.shutdownTimeout.Seconds() > 0 {
+		extraArgs = append(extraArgs, "-t", fmt.Sprintf("%d", int(math.Round(s.shutdownTimeout.Seconds()))))
```
jsoriano, line 236: Nit.
```suggestion
	if seconds := s.shutdownTimeout.Seconds(); seconds > 0 {
		extraArgs = append(extraArgs, "-t", fmt.Sprintf("%d", int(math.Round(seconds))))
```

jsoriano (approved):

internal/servicedeployer/compose.go
mrodm: I'd prefer to keep it internal for now, to ensure that in the terraform service type it is set a high timeout. Currently, for that case is 300 secs, and all the other scenarios that flag (-t) would not be set, so it is 10 seconds according to the documentation.

PR #1728, opened 2024-03-14
Add platform package-registry container

mrodm:
Add platform `linux/amd64` as platform in Elastic Package Registry container, to avoid errors like this:
> The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8

This is following the same approach as it was done for `tianon/true` docker images #1694 

It's likely that this change could be removed if this PR is merged: https://github.com/elastic/package-registry/pull/1162


gsantoro (approved):

mrodm:
Closing this PR , installing `rosetta` emulation looks like it solves this error.

PR #1733, opened 2024-03-19
Add dynamic mappings for non-indexed ECS fields

jsoriano:
Proposal to add dynamic mappings for non-indexed ECS fields.

This is needed for some fields like `event.original` that are documented and defined in ECS as non-indexed, with `index: false` and `doc_values: false`.


mrodm:
test integrations

jsoriano:
test integrations

internal/builder/_static/ecs_mappings.yaml
```diff
@@ -198,6 +204,12 @@ mappings:
             mapping:
                 type: match_only_text
             path_match: message
+        - event_original_non_indexed_keyword:
+            mapping:
+                type: keyword
+                index: false
+                doc_values: false
+            path_match: 'event.original'
```
jsoriano, line 212: If we just apply these changes here, rebuilding current packages will give different results. So I think we may need to include them only after something is updated in the package. I see two options for this:
* Use the ECS version set in `_dev/build/build.yml`. This would allow us to introduce the change in the same version as in Elasticsearch (https://github.com/elastic/elasticsearch/pull/106714).
* Use a Package Spec version, as we do with other features.

Or we can just go on and apply it to existing packages if rebuilt, as this can be considered a fix, even if it breaks a bit of reproducibility.

@ruflin @flash1293 @mrodm thoughts?

flash1293: I don't think users will insist on this level of reproducibility - it's unlikely anyone actually relied on this behavior and by not requiring the integration author to jump through hoops, we make it easier for this fix to get adopted.

mrodm: Looking into this it would be great that it would be easy for owners to get this fix in their packages without any change (e.g. updating package-spec version) to not force to more changes (e.g. new validation rules).

So, it looks ok to me if this applied as it is too.

mrodm (approved):

jsoriano:
Unrelated failures in benchmarks.

Issue #732, opened 2022-03-16
Add unit testing for pipeline test runner to cover json decoding and encoding


jsoriano:
In https://github.com/elastic/elastic-package/pull/717 a json decoder was introduced to parse numbers as json numbers instead of `float64`. This is needed to avoid truncating big integers, what can contaminate test files and can hide actual problems.

Add unit testing in a way that allows to check that the final written result is exactly what is expected, covering all the uses of json decoding and encoding.

Some ideas can be found in this comment https://github.com/elastic/elastic-package/pull/717#issuecomment-1062788770 and in https://github.com/elastic/elastic-package/pull/721.

PR #1734, opened 2024-03-19
Add false positive package to detect regressions in JSON marshalling

jsoriano:
Add a test package that reproduces the issue fixed in https://github.com/elastic/elastic-package/pull/717. This will help catching regressions caused by not using JSON numbers while decoding documents.

Fixes https://github.com/elastic/elastic-package/issues/732.


mrodm (approved):
:+1:

PR #1735, opened 2024-03-19
Refactors in pipeline test runner

jsoriano:
A couple of refactors in the pipeline test runner, implemented while investigating #732.
* Reuse JSON unmarshaler.
* Extract test case execution to its own function.


internal/common/json.go
```diff
@@ -0,0 +1,35 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package common
```
mrodm, line 5: Would it make sense to move this function defined here to the internal `formatter` package ?

mrodm (approved):

Issue #1276, opened 2023-05-23
Fail tests if documents contained _ignored fields


andrewkroh:
Fleet turned on `ignore_malformed` for logs data streams in https://github.com/elastic/kibana/pull/157184. This causes Elasticsearch to silently ignore malformed fields (for certain types). This is great for users, but silent failures are the bane of developers and unit tests.

`elastic-package` should be updated to check if any documents returned in tests contain [`_ignored`](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-malformed.html#_dealing_with_malformed_fields). This would indicate the document contains malformed data and the test should fail.

PR #1738, opened 2024-03-22
Extend system test to validate absence of _ignored

flash1293:
Closes https://github.com/elastic/elastic-package/issues/1276

This PR adds a check whether any indexed docs have an `_ignored` field - this would indicate a mapping problem:

Example output:
> Error: error running package system tests: could not complete test run: found _ignored fields in data stream logs-cel.cel-ep

## How to test

Cause a mapping error by configuring a field in a wrong way - I used the `cel` package:
```diff
--- a/packages/cel/fields/input.yml
+++ b/packages/cel/fields/input.yml
@@ -3,7 +3,8 @@
 - name: message
-  external: ecs
+  type: keyword
+  ignore_above: 5
 - name: input.name
   type: constant_keyword
 - name: input.type
```

As message is longer than 5 characters, this will cause `message` to not be mapping correctly. Running the system tests for `cel` will fail with the error shown above.

## Notes

The check is done via aggregations instead of checking through each documents to be able to validate all docs, not just the ones that are fetched (which are capped at 500). Once https://github.com/elastic/elasticsearch/pull/101373 is merged, we can use a terms query here for better debug information (but it's also possible to check the index manually to find the root cause).

This cannot be meaningfully applied to the pipeline tests currently, as they don't take mappings into account (https://github.com/elastic/elasticsearch/pull/99920).


flash1293:
It actually looks like this check finds an error right away:

> Error: error running package system tests: could not complete test run: found _ignored fields in data stream logs-auditd_manager.auditd-ep
 
Who would be a good person to look into this?

ruflin:
++, nice addition. The more we add of these checks, the higher the quality of integrations will be.

jsoriano:
test integrations

jsoriano:
> I ran this test locally and the issue is the `event.original` field, which is mapped as a `keyword` with `ignore_above` of 1024 - I think it will be resolved by #1733 , can you confirm @jsoriano ?

It is possible, but we would need to test it when ready.

jsoriano (commented):
LGTM, thanks!

But let's wait to see if it would affect many existing integrations. If it does, we could introduce this only for packages using the next version of the spec.

flash1293:
test integrations

flash1293:
test integrations

flash1293:
Still the same error in the test package, I will look into this

internal/testrunner/runners/system/runner.go
```diff
@@ -42,8 +42,8 @@ const (
 	testRunMaxID = 99999
 	testRunMinID = 10000
 
-	allFieldsBody = `{"fields": ["*"]}`
-	DevDeployDir  = "_dev/deploy"
+	checkFieldsBody = `{"fields": ["*"], "aggs": {"degraded": {"filter": {"exists": {"field": "_ignored"}}}}}`
```
andrewkroh, line 45: I think the returned error should at a minimum include the names of ignored fields. You could aggregate those field names. (It needs a runtime field to access the names as per https://github.com/elastic/elasticsearch/issues/59946#issuecomment-947796041).

```
POST logs-*/_search
{
  "size": 10,
  "runtime_mappings": {
    "my_ignored": {
      "type": "keyword",
      "script": {
        "source": "for (def v : params['_fields']._ignored.values) { emit(v); }"
      }
    }
  },
  "aggs": {
    "all_ignored": {
      "filter": {
        "exists": {
          "field": "_ignored"
        }
      },
      "aggs": {
        "ignore_fields": {
          "terms": {
            "size": 100,
            "field": "my_ignored"
          }
        }
      }
    }
  }
}
```

flash1293: As mentioned I wanted to wait with this until the change to make _ignored aggregatable is merged, but you are right that we don't need to block on it - I changed the code to use the runtime field workaround for now.

flash1293:
test integrations

flash1293:
The test `auditd_manager` package still fails as it relies on the `ecs@mappings` component template which won't be fixed until https://github.com/elastic/elasticsearch/pull/106714 is released - we can silence the test by explicitly mapping `event.original` correctly, but let's check how many of the real integrations are affected in a similar way - if there are a lot, we either should:
* wait for 8.14.0 until we merge this
* or make an exception for `event.original` for now to at least detect other mapping errors

flash1293:
test integrations

jsoriano:
test integrations

flash1293:
Test failures:
* ti_threatconnect: `threat_connect.indicator.tags.data.description `
* tenable_io: `tenable_io.vulnerability.plugin.description vulnerability.description`
* tanium: `tanium.threat_response.other_parameters.log_details.payload`, `tanium.threat_response.other_parameters.original`
* spring_boot: `host.ip`
* panw: `panw.panos.after_change_detail`
* oracle: `oracle.database_audit.action`
* mysql: `mysql.slowlog.query`
* m365_defender: `m365_defender.incident.alert.recommended_actions`
* fortinet_fortimanager: `fortimanager.log.changes`
* crowdstrike: `crowdstrike.ConfigStateData`, `crowdstrike.FeatureVector`, `crowdstrike.OSVersionFileData`, `process.command_line`
* arista_ngfw: unrelated failure
* activemq: `error.stack_trace`

These are actual issues in the integrations that need to be fixed 1 by 1. cc @narph and @lalit-satapathy - this is hinting at real problems in these integrations.

internal/testrunner/runners/system/runner.go
```diff
@@ -1141,6 +1184,18 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
 		return result.WithError(err)
 	}
 
+	// TODO: remove this once Elasticsearch will map event.original correctly (8.14)
```
ruflin, line 1187: This is https://github.com/elastic/elasticsearch/pull/106714 I assume? Maybe worth to add the link here.

andrewkroh (approved):

internal/testrunner/runners/system/runner.go
```diff
@@ -1333,6 +1376,18 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
...
+	ignoredFields := make([]string, 0, len(scenario.ignoredFields))
+	for _, field := range scenario.ignoredFields {
+		if field != "event.original" {
+			ignoredFields = append(ignoredFields, field)
+		}
+	}
+
+	if len(ignoredFields) > 0 {
+		return result.WithError(fmt.Errorf("found ignored fields in data stream %s: %v", scenario.dataStream, ignoredFields))
+	}
```
jsoriano, line 1389: If looks like this change is going to be released in 8.14.0, and I guess we need to keep this code for older versions, so we can add an if to decide if we need to run this code, and remove the TODO:
```suggestion
	stackVersion, err := semver.NewVersion(serviceOptions.StackVersion)
	if err != nil {
		....
	}
	if stackVersion.LessThan(semver.MustParse("8.14.0")) {
		ignoredFields := make([]string, 0, len(scenario.ignoredFields))
		for _, field := range scenario.ignoredFields {
			if field != "event.original" {
				ignoredFields = append(ignoredFields, field)
			}
		}
		if len(ignoredFields) > 0 {
			return result.WithError(fmt.Errorf("found ignored fields in data stream %s: %v", scenario.dataStream, ignoredFields))
		}
	}
```

internal/testrunner/runners/system/runner.go
```diff
@@ -1333,6 +1376,18 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
 		return result.WithError(err)
 	}
 
+	// TODO: remove this once Elasticsearch will map event.original correctly (https://github.com/elastic/elasticsearch/pull/106714)
```
jsoriano, line 1379: Btw, if we remove this, we won't be checking ignored fields, right? :thinking: Or we will have to add a different way to check ignored fields?

jsoriano (commented):
Code LGTM, but I am a bit confused about the plan to remove the code after the expected changes are merged in Elasticsearch.

Also, before merging, we need a plan for the failing integrations, otherwise we won't be able to update elastic-package in integrations repo without breaking CI.

internal/testrunner/runners/system/runner.go
flash1293: I just meant removing the exception for event.original and failing on any ignored field. But I agree with checking the Elasticsearch version is the better approach.

flash1293:
> Also, before merging, we need a plan for the failing integrations, otherwise we won't be able to update elastic-package in integrations repo without breaking CI.

Agreed, I'm in contact with @narph and @lalit-satapathy to clear up this part.

internal/testrunner/runners/system/runner.go
jsoriano: Oh ok, I see. Then yes, we can conditionally omit `event.original` depending on the stack version.

internal/testrunner/runners/system/runner.go
```diff
@@ -1333,6 +1376,18 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
...
+	ignoredFields := make([]string, 0, len(scenario.ignoredFields))
+	for _, field := range scenario.ignoredFields {
+		if field != "event.original" {
+			ignoredFields = append(ignoredFields, field)
+		}
+	}
+
+	if len(ignoredFields) > 0 {
+		return result.WithError(fmt.Errorf("found ignored fields in data stream %s: %v", scenario.dataStream, ignoredFields))
```
andrewkroh, line 1388: As I was thinking about getting someone to investigate the errors that were surfaced by this enhancement, I am realizing that this error needs more information in it. As it stands now, there isn't enough information in the error to attempt solving it.

The first thing a package developer needs to do is replicate the test setup and then run the same query to obtain information that elastic-package already has in order to see raw values from the `_source` or `ignored_field_values`. Can we create a structured error that includes this information or log the raw data in some way to improve DX?

flash1293: That's a fair point @andrewkroh - Getting the top documents with `_ignored` and making them part of the error message seems like the best way to help the user fix the problem. I will also add a note to `docs/howto/system_testing.md` to explain what's this error is about and how to debug it.

efd6: It would be nice if this were significantly less verbose; the vast majority of the output from this is unhelpful in debugging the issue and we already have a very high noise to signal ratio here.

If only the relevant fields are emitted this becomes much more manageable. For example with this change, running system tests on m365_defender
```diff
diff --git a/internal/testrunner/runners/system/runner.go b/internal/testrunner/runners/system/runner.go
index 332925a..838da5b 100644
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -21,6 +21,7 @@ import (
        "github.com/Masterminds/semver/v3"
        "gopkg.in/yaml.v3"
 
+       "github.com/elastic/elastic-agent-libs/mapstr"
        "github.com/elastic/elastic-package/internal/agentdeployer"
        "github.com/elastic/elastic-package/internal/common"
        "github.com/elastic/elastic-package/internal/configuration/locations"
@@ -1403,7 +1404,22 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
        }
 
        if len(ignoredFields) > 0 {
-               degradedDocsJSON, err := json.MarshalIndent(scenario.degradedDocs, "", "  ")
+               issues := make([]struct {
+                       ID            any `json:"_id"`
+                       Timestamp     any `json:"@timestamp,omitempty"`
+                       IgnoredFields any `json:"ignored_field_values"`
+               }, len(scenario.degradedDocs))
+               for i, d := range scenario.degradedDocs {
+                       issues[i].ID = d["_id"]
+                       switch s := d["_source"].(type) {
+                       case map[string]any:
+                               issues[i].Timestamp = s["@timestamp"]
+                       case mapstr.M:
+                               issues[i].Timestamp = s["@timestamp"]
+                       }
+                       issues[i].IgnoredFields = d["ignored_field_values"]
+               }
+               degradedDocsJSON, err := json.MarshalIndent(issues, "", "  ")
                if err != nil {
                        return nil, fmt.Errorf("failed to marshal degraded docs to JSON: %w", err)
                }
```
we get something like this
```
Error: error running package system tests: could not complete test run: found ignored fields in data stream logs-m365_defender.incident-ep: [m365_defender.incident.alert.recommended_actions]. Affected documents: [
  {
    "_id": "RJDPoPySgbgCrqqObuC4ic2x9QE=",
    "@timestamp": "2021-09-30T09:35:45.113Z",
    "ignored_field_values": {
      "m365_defender.incident.alert.recommended_actions": [
        "Collect artifacts and determine scope\n�\tReview the machine timeline for suspicious activities that may have occurred before and after the time of the alert, and record additional related artifacts (files, IPs/URLs) \n�\tLook for the presence of relevant artifacts on other systems. Identify commonalities and differences between potentially compromised systems.\n�\tSubmit relevant files for deep analysis and review resulting detailed behavioral information.\n�\tSubmit undetected files to the MMPC malware portal\n\nInitiate containment \u0026 mitigation \n�\tContact the user to verify intent and initiate local remediation actions as needed.\n�\tUpdate AV signatures and run a full scan. The scan might reveal and remove previously-undetected malware components.\n�\tEnsure that the machine has the latest security updates. In particular, ensure that you have installed the latest software, web browser, and Operating System versions.\n�\tIf credential theft is suspected, reset all relevant users passwords.\n�\tBlock communication with relevant URLs or IPs at the organization�s perimeter."
      ]
    }
  }
]
```

lalit-satapathy:
> Test failures:
> 
> * ti_threatconnect: `threat_connect.indicator.tags.data.description `
> * tenable_io: `tenable_io.vulnerability.plugin.description vulnerability.description`
> * tanium: `tanium.threat_response.other_parameters.log_details.payload`, `tanium.threat_response.other_parameters.original`
> * spring_boot: `host.ip`
> * panw: `panw.panos.after_change_detail`
> * oracle: `oracle.database_audit.action`
> * mysql: `mysql.slowlog.query`
> * m365_defender: `m365_defender.incident.alert.recommended_actions`
> * fortinet_fortimanager: `fortimanager.log.changes`
> * crowdstrike: `crowdstrike.ConfigStateData`, `crowdstrike.FeatureVector`, `crowdstrike.OSVersionFileData`, `process.command_line`
> * arista_ngfw: unrelated failure
> * activemq: `error.stack_trace`
> 
> These are actual issues in the integrations that need to be fixed 1 by 1. cc @narph and @lalit-satapathy - this is hinting at real problems in these integrations.

CC: @ishleenk17 to watch for any package level changes.

efd6:
I've looked at m365_defender, tenable_io and ti_threatconnect, and in all cases, AFAICS, the issue is the length of text in the field. ISTM that this is not necessarily something that can be resolved. In all the cases that I looked at there is no sensible way to reduce the length of the field in the general case (in some other cases maybe doing a split would be an appropriate action, but that is no the case here, both because there is no obvious split, or the field definition does not recommend an array, or both).

Unfortunately, the documents that are ignored by kibana when `ignore_malformed: true` do not have the reason for an assessment of malformation added to the document (probably it should). Given this and the fact that text length can be the cause, the test failure assessment based on the presence of fields in the `ignored_field_values` fields appears to be harsh, unless there is an option to mark fields as acceptably ignorable (again, having a reason to validate against would be a good, idea, but we are where we are).

flash1293:
Thanks for your comments @efd6 !

> Unfortunately, the documents that are ignored by kibana when ignore_malformed: true do not have the reason for an assessment of malformation added to the document (probably it should)

Agreed, that's the next thing we want to look into adding to Elasticsearch, but there are no committed plans yet.

> Given this and the fact that text length can be the cause, the test failure assessment based on the presence of fields in the ignored_field_values fields appears to be harsh, unless there is an option to mark fields as acceptably ignorable

Yeah, I started out without this because I hoped we could get by without it, but it might not be feasible in practice. This might be a silly question, but could you elaborate a bit more on why we can't change the mapping? `m365_defender.incident.alert.recommended_actions` for example doesn't look like something you would typically query as a keyword, right? What about turning it into a text field? Not necessarily pushing for it, just trying to understand the cases that happen in practice.

efd6:
@flash1293 Yeah, it could be a text field. Changing this in existing installations may be problematic though.

flash1293:
@efd6 

> Changing this in existing installations may be problematic though.

Why's that? Because the user could have something set up (a dashboard or similar) that depends on it being a keyword field?

Two thoughts about it:
* The fact that sometimes (often?) the values for this field get ignored and are not actually searchable also seems problematic, right? It means the user won't see the values they are looking for
* I'm not sure how this is normally handled, but there is probably a process for breaking changes in integrations, right?

efd6:
> Why's that? Because the user could have something set up (a dashboard or similar) that depends on it being a keyword field?

Maybe, though more just that to pick it up in their installation they would need to re-index to get the new mapping (unless I am misunderstanding how this would work).

efd6:
> I'm not sure how this is normally handled, but there is probably a process for breaking changes in integrations, right?

Yes, there is, but we'd like to avoid them if at all possible.

flash1293:
> Maybe, though more just that to pick it up in their installation they would need to re-index to get the new mapping (unless I am misunderstanding how this would work).

IMHO it would be OK to just do this for new indices. The mapping is updated in the template and once the index rolls over, it gets mapped as a `text` field. What do you think about this approach? It also wouldn't require a breaking change IMHO, but you now more about the notion of "breaking change" here.

> > I'm not sure how this is normally handled, but there is probably a process for breaking changes in integrations, right?

> Yes, there is, but we'd like to avoid them if at all possible.

Makes sense

efd6:
> IMHO it would be OK to just do this for new indices. The mapping is updated in the template and once the index rolls over, it gets mapped as a text field. What do you think about this approach?

This is probably reasonable, though how do we express that? (this is getting outside my area of expertise)

flash1293:
Another thought - if we are OK with the field not being indexed if it's longer than x characters (as that's the status quo), what about making that explicit and adding an ingest pipeline processor that does this?

There might be a case for an option to "ignore _ignored fields", but it feels a bit like muting linter warnings in code - it's probably a symptom of a bigger problem if you have to do it all the time.

flash1293:
> This is probably reasonable, though how do we express that? (this is getting outside my area of expertise)

Not sure what you mean by expressing that - I thought it's just about changing the mapping in `fields.yml` - am I missing something here @jsoriano ?

efd6:
> There might be a case for an option to "ignore _ignored fields", but it feels a bit like muting linter warnings in code - it's probably a symptom of a bigger problem if you have to do it all the time.

Indeed, which is why I'd liked to have had the reason; if I can say ignore ignored field X when the reason is Y, then if it gets ignored for another reason, that's still a failure.

> Not sure what you mean by expressing that 

"(this is getting outside my area of expertise)" is the relevant caveat, but I suspect you're correct.

ishleenk17:
> * spring_boot: `host.ip`

@flash1293 : I checked this and other Observvability Integrations you have tagged. I'm trying to understand why these would be considered errors.

Eg: 
Springboot: Suspected Issue in host.ip field 
```
  - set:
      field: host.ip
      value: ["{{{host.ip}}}"]
      ignore_empty_value: true
      ignore_failure: true

```
Here the ignore_ looks appropriate only. 

The Mapping of the field happens through ecs.yml

```
-  external: ecs
   name: host.ip
```

andrewkroh:
test integrations

efd6 (approved):

flash1293:
@jsoriano  As the need came up in the existing failures, I opened a PR on the package-spec to add a flag to add exceptions to this test: https://github.com/elastic/package-spec/pull/743

Once this is merged, I'm going to update this PR to check this definition and skip the check if it makes sense. Once this is in place, it should be possible to mute existing errors in the integrations repo and merge this PR.

ShourieG:
@efd6 @flash1293, I've often seen encoded payload field values exceed 6-7k characters easily. This can easily go up even more in real world scenarios and the issue is these are always clumped into **'keyword'** types. We really need a way to mark 'keyword' types to have an attribute like **'ignore_length: true'** or some sort or introduce a new type for storing response payloads, otherwise in a real scenario these fields will not get indexed after the recent changes, just addressing the test scenarios won't work.

flash1293:
> otherwise in a real scenario these fields will not get indexed after the recent changes and just addressing the test scenarios won't work

The changes around `_ignored` didn't change when values got indexed or not, it's just about exposing the problem.

ShourieG:
> > otherwise in a real scenario these fields will not get indexed after the recent changes and just addressing the test scenarios won't work
> 
> The changes around `_ignored` didn't change when values got indexed or not, it's just about exposing the problem.

@flash1293, Ah alright, I guess I misunderstood initially the issue around indexing. Then maybe adding the  **'ignore_length: true'** to the package spec so we can specify that during the tests would be pretty awesome. Something that we can store in the validation.yml ?

flash1293:
@ShourieG Currently working on it here: https://github.com/elastic/package-spec/pull/743

internal/testrunner/runners/system/runner.go
```diff
@@ -1417,16 +1431,56 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
...
+	atLeast814, err := semver.NewConstraint(">=8.14")
+	if err != nil {
+		// Pre 8.14 Elasticsearch commonly has event.original not mapped correctly, exclude from check: https://github.com/elastic/elasticsearch/pull/106714
+		return nil, fmt.Errorf("failed to parse version constraint: %w", err)
+	}
+	skipIgnoredFields := append([]string(nil), config.SkipIgnoredFields...)
+	isAtLeast814 := atLeast814.Check(semver.MustParse(r.stackVersion.Number))
+	if !isAtLeast814 {
+		skipIgnoredFields = append(skipIgnoredFields, "event.original")
+	}
```
jsoriano, line 1443: I think we can directly use `LessThan` here, and we should handle the error when parsing the stack version, as this one can come from user input.
```suggestion
	skipIgnoredFields := append([]string(nil), config.SkipIgnoredFields...)
	stackVersion, err := semver.New(r.stackVersion.Number)
	if err != nil {
		return nil, fmt.Errorf("failed to parse stack version: %w", err)
	}
	if stackVersion.LessThan(semver.MustParse("8.14.0")) {
		// Pre 8.14 Elasticsearch commonly has event.original not mapped correctly, exclude from check: https://github.com/elastic/elasticsearch/pull/106714
		skipIgnoredFields = append(skipIgnoredFields, "event.original")
	}
```

internal/testrunner/runners/system/runner.go
```diff
@@ -1374,6 +1431,58 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
...
+			}
+			issues[i].IgnoredFields = d["ignored_field_values"]
+		}
+		degradedDocsJSON, err := json.MarshalIndent(issues, "", "  ")
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal degraded docs to JSON: %w", err)
+		}
+
+		return result.WithError(fmt.Errorf("found ignored fields in data stream %s: %v. Affected documents: %s", scenario.dataStream, ignoredFields, degradedDocsJSON))
+	}
```
jsoriano, line 1484: Nit. Could we move the logic to check for ignored fields to its own function?

internal/testrunner/runners/system/runner.go
```diff
@@ -1374,6 +1431,58 @@ func (r *runner) validateTestScenario(ctx context.Context, result *testrunner.Re
...
+		for _, skipField := range skipIgnoredFields {
+			if field == skipField {
+				shouldSkip = true
+				break
+			}
+		}
+
+		if !shouldSkip {
+			ignoredFields = append(ignoredFields, field)
+		}
```
jsoriano, line 1460:
```suggestion
		if !slices.Contains(skipIgnoredFields, field) {
			ignoredFields = append(ignoredFields, field)
		}
```

flash1293:
Muting existing failures here: https://github.com/elastic/integrations/pull/9823

jsoriano (approved):
LGTM, thanks for the work here! Feel free to merge the change in the spec too.

PR #1742, opened 2024-03-26
[cleanup] Fix creation date tag for buckets

mrodm:
Set the right field to check for creation date in Storage Bucket resources.

### DIFF
--- a/.buildkite/configs/cleanup.aws.yml
+++ b/.buildkite/configs/cleanup.aws.yml
@@ -40,17 +40,18 @@ scanners:
           - us-east-1
         filters:
           - type: "<"
-            pointer: "/created_at"
+            pointer: "/extra/creation_date"
             param: "${CREATION_DATE}"
             converters:
               param: "date"
               value: "date"
-          - type: "regex"
-            pointer: "/extra/tags/repo"
-            param: "^(elastic-package|integrations)"
-          - type: "="
-            pointer: "/extra/tags/environment"
-            param: "ci"
+          # It looks like tags are not reported by cloud-reaper for buckets
+          # - type: "regex"
+          #   pointer: "/extra/tags/repo"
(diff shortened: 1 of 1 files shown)


.buildkite/configs/cleanup.aws.yml
```diff
@@ -40,17 +40,18 @@ scanners:
...
-          - type: "="
-            pointer: "/extra/tags/environment"
-            param: "ci"
+          # It looks like tags are not reported by cloud-reaper for buckets
+          # - type: "regex"
+          #   pointer: "/extra/tags/repo"
+          #   param: "^(elastic-package|integrations)"
+          # - type: "="
+          #   pointer: "/extra/tags/environment"
+          #   param: "ci"
```
mrodm, line 54: I have some doubts for these tags in Storage Bucket resources.
@amannocci , are tags also reported/used by cloud-reaper to filter resources ? 

At least it looks like that in the verbose output other tags a part from creation_date are not shown:

```
ObjectStorageBucket(
  name=xxx
  extra={'creation_date': '2024-03-26T06:57:28.000Z'}
)
```

jsoriano (approved):

.buildkite/configs/cleanup.aws.yml
mrodm: Tested adding this filter to force showing some queues:
```yaml
          - type: "regex"
            pointer: "/extra/tags/repo"
            param: "^(elastic-package|integrations|beats)"            
```
but, `cloud-reaper` does not report anything

mrodm: Talked offline to @amannocci , and currently this is the expected behavior. `cloud-reaper` needs to make an additional API call to get all the tags set in Buckets.

Leaving the filters commented until this support is added into `cloud-reaper`.

PR #1747, opened 2024-04-04
Use rerefence to update terraform outputs in serviceInfo struct

mrodm:
This PR updates `addTerraformOutputs` to use reference instead of value parameter.

It also removes an unnecessary temporal variable (`outCtxt`).

### DIFF
--- a/internal/servicedeployer/terraform_test.go
+++ b/internal/servicedeployer/terraform_test.go
@@ -65,6 +65,29 @@ func TestAddTerraformOutputs(t *testing.T) {
 			},
 		},
 		{
+			testName: "add_single_value_output",
+			runId:    "99999",
+			svcInfo: ServiceInfo{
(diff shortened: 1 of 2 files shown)


internal/servicedeployer/terraform_test.go
```diff
@@ -138,15 +161,14 @@ func TestAddTerraformOutputs(t *testing.T) {
 	for _, tc := range testCases {
 
 		t.Run(tc.testName, func(t *testing.T) {
-			tc.svcInfo.CustomProperties = make(map[string]interface{})
```
mrodm, line 141: As now it is used as reference, there is no need to create in advance the map.

jsoriano: :+1: this was indeed cheating the test.

jsoriano (approved):
Thanks!

PR #1752, opened 2024-04-08
internal/fields: fix handling of boolean keyword fields

efd6:
In some packages boolean fields are treated as keywords. Changing the mapping of those fields would be a breaking change so allow these fields to be considered u1 integers.

Please take a look.


jsoriano:
test integrations

internal/fields/validate.go
```diff
@@ -695,15 +695,17 @@ func createDocExpandingObjects(doc common.MapStr) (common.MapStr, error) {
 func isNumericKeyword(definition FieldDefinition, val interface{}) bool {
 	var isNumber bool
 	switch val := val.(type) {
-	case float64, []float64:
+	case bool, []bool, float64, []float64:
```
jsoriano, line 698: Could you please add an example of this in a test package?

internal/fields/validate.go
```diff
@@ -695,15 +695,17 @@ func createDocExpandingObjects(doc common.MapStr) (common.MapStr, error) {
...
-			_, ok := v.(float64)
-			if !ok {
+			switch v.(type) {
+			case bool, float64:
+			default:
 				isNumber = false
-				break
+				break loop
 			}
 		}
```
jsoriano, line 710: Nit. I wonder if we can avoid the label.
```
		for _, v := range val {
			switch v.(type) {
			case bool, float64:
			default:
				isNumber = false
			}
			if !isNumber {
				break
			}
		}
```

efd6: The generated code is worse for arguably less clear source. https://godbolt.org/z/fqrvEfrcW

jsoriano: In general I consider the use of labels as less clear if there are alternatives. Here itself I had to think twice if this `loop:` was a special keyword for the `switch` that I didn't know about. But I guess this is a matter of opinions, so as you prefer :slightly_smiling_face:

jsoriano (approved):
LGTM, Thanks!

PR #1756, opened 2024-04-09
Use declarative definitions for asset test runner expected state

jsoriano:
Use go-resource to define the state expected by the asset test runner.

The idea would be to apply this model also to the system test runner, to try to reduce its complexity.

With this architecture, each runner would have a `resources` method, that generate the expected resources based on some given parameters. On this example the only parameter would be whether the package must be installed or not. Then test runners would apply these resources before running the tests or on tear down stages.


internal/testrunner/runners/asset/runner.go
```diff
@@ -94,60 +104,33 @@ func (r *runner) run(ctx context.Context) ([]testrunner.TestResult, error) {
...
-	r.removePackageHandler = removePackageHandler
+	manifest, err := packages.ReadPackageManifestFromPackageRoot(r.packageRootPath)
+	if err != nil {
+		return result.WithError(fmt.Errorf("cannot read the package manifest from %s: %w", r.packageRootPath, err))
+	}
+	installedPackage, err := r.kibanaClient.GetPackage(ctx, manifest.Name)
+	if err != nil {
+		return result.WithError(fmt.Errorf("cannot get installed package %q", manifest.Name, err))
+	}
+	installedAssets := installedPackage.Assets()
```
jsoriano, line 133: On apply we were already getting the assets. We wouldn't need this if `go-resource` had support to get info from defined resources.

internal/testrunner/runners/asset/runner.go
```diff
@@ -94,60 +104,21 @@ func (r *runner) run(ctx context.Context) ([]testrunner.TestResult, error) {
...
-
-		logger.Debug("removing package...")
-		err = packageInstaller.Uninstall(ctx)
-		if err != nil {
-			// logging the error as a warning and not returning it since there could be other reasons that could make fail this process
-			// for instance being defined a test agent policy where this package is used for debugging purposes
-			logger.Warnf("failed to uninstall package %q: %s", pkgManifest.Name, err.Error())
-		}
-		return nil
+		// FIXME: We won't detect here if this has been a cancelation, see https://github.com/elastic/go-resource/pull/2.
```
jsoriano, line 109: Requires https://github.com/elastic/go-resource/pull/2.

jsoriano: Fixed in https://github.com/elastic/elastic-package/pull/1761

jsoriano:
Opening for review as it seems to work mostly as before. @mrodm wdyt of this approach?

I like that tear down gets the state from the system itself and doesn't need to keep track of what has been set up or not. But maybe this is complex in the system runner scenario. I also like that we can test each resource and its lifecycle more easily.

mrodm:
> Opening for review as it seems to work mostly as before. @mrodm wdyt of this approach?
> 
> I like that tear down gets the state from the system itself and doesn't need to keep track of what has been set up or not. But maybe this is complex in the system runner scenario. I also like that we can test each resource and its lifecycle more easily.

It looks great this approach!
Easier to manage the conditions when to install/uninstall each resource (in this case Packages)  and being able to add tests easily! Nice !!

Let's see if this can be applied easily also in system tests :crossed_fingers:

internal/testrunner/runners/system/runner.go
```diff
@@ -389,11 +403,13 @@ func (r *runner) tearDownTest(ctx context.Context) error {
...
 	}
 
-	if r.deletePackageHandler != nil {
-		if err := r.deletePackageHandler(cleanupCtx); err != nil {
-			return err
-		}
-		r.deletePackageHandler = nil
+	resourcesOptions := resourcesOptions{
+		// Keep it installed only if we were running setup, or tests only.
+		installedPackage: r.options.RunSetup || r.options.RunTestsOnly,
```
mrodm, line 408: :+1:

internal/testrunner/runners/system/runner.go
```diff
@@ -817,58 +837,15 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
...
-		Kibana:         r.options.KibanaClient,
-		RootPath:       r.options.PackageRootPath,
-		SkipValidation: true,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize package installer: %v", err)
+	logger.Debug("Installing package...")
+	resourcesOptions := resourcesOptions{
+		// Install it unless we are running the tear down only.
+		installedPackage: !r.options.RunTearDown,
```
mrodm, line 843: :+1:

internal/resources/fleetpackage_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+func TestPackageLifecycle(t *testing.T) {
+	cases := []struct {
+		title string
+		name  string
+	}{
+		{title: "nginx", name: "nginx"},
+		{title: "package not found", name: "sql_input"},
```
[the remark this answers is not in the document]
jsoriano: This is a special case happening with some of our test packages, it looks like they don't exist in the registry, so when trying to get information about them through the registry when they are not installed we receive a 404.

mrodm (approved):
This looks really promising, thanks! :rocket:

PR #1757, opened 2024-04-09
Update Go Runtime to 1.22.2

jsoriano:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/elastic/elastic-package
 
-go 1.21.0
+go 1.22.0
```
pkoutsovasilis, line 3: shouldn't this be `1.22.2`?

jsoriano: This is like the lower version supported by the package, not really the version used for compilation.

pkoutsovasilis (approved):

PR #1762, opened 2024-04-11
Remove variant data from agent deployer package

mrodm:
Relates #787

Variant information was added into the agent deployer package in #1724, however this data is not relevant to spin up a new Elastic Agent. This PR removes `AgentVariant` and its usages under `agentdeployer` package.


jsoriano (approved):
:+1:

PR #1765, opened 2024-04-12
Allow to define custom agents in system test configuration files

mrodm:
Relates #787 

In this PR, it is added into the system test configuration files first fields in order to run custom agents with different parameters (capabilities, pid mode, user...). This is just taken into account when independent agents are enabled.

A new test package has been added to test this `auditd_manager_independent_agent`.


internal/agentdeployer/kubernetes.go
```diff
@@ -39,11 +38,10 @@ type KubernetesAgentDeployer struct {
...
 type KubernetesAgentDeployerOptions struct {
-	Profile        *profile.Profile
-	DefinitionsDir string
-	StackVersion   string
-	PolicyName     string
-	DataStream     string
+	Profile      *profile.Profile
+	StackVersion string
+	PolicyName   string
+	DataStream   string
```
mrodm, line 44: If it is preferred, I can create a separated PR to remove `definitionsDir` from this code.

internal/agentdeployer/factory.go
```diff
@@ -80,19 +76,16 @@ func Factory(options FactoryOptions) (AgentDeployer, error) {
 		// FIXME: this docker-compose scenario contains both agent and service
 		return nil, nil
 	case "k8s":
-		if _, err := os.Stat(agentDeployerPath); err == nil {
```
mrodm, line 83: It should not be needed to check that this folder exists, WDYT ? @jsoriano 
`findAgentDeployer` method returns already the folders found under `_dev/deploy`

internal/service/boot.go
```diff
@@ -48,6 +48,11 @@ func BootUp(ctx context.Context, options Options) error {
 		return fmt.Errorf("can't create the service deployer instance: %w", err)
 	}
 
+	if serviceDeployer == nil {
+		fmt.Println("No service defined")
+		return nil
+	}
+
```
mrodm, line 55: Running this command with the `auditd_manager` package in this branch:

```shell
 $ elastic-package service up 
Boot up the service stack
No service defined
```

internal/agentdeployer/factory.go
jsoriano: Yeah, looks correct, and indeed `agentDeployerPath` is not used anymore here after this change.

internal/agentdeployer/kubernetes.go
mrodm: Moved these changes to a new PR #1769

test/packages/with-custom-agent/auditd_manager_independent_agent/data_stream/auditd/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,333 @@
...
+      field: auditd.warnings
+      separator: "\n"
+      target_field: error.message
+      if: ctx?.auditd?.warnings != null && ctx.auditd.warnings instanceof List
+  # if it is not a list, rename it back to error.message
+  - set:
+      field: error.message
+      copy_from: auditd.warnings
+      ignore_empty_value: true
+      if: ctx?.auditd?.warnings != null && !(ctx.auditd.warnings instanceof List)
```
mrodm, line 30: Added this new step in the pipeline to detect if there is just one warning (string)

warnings as a string

Adding this would make fail this test package when ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT is false, since this will use the Elastic Agent from the stack that does not have the required capabilities.

Should we skip this test package in the CI scripts ? @jsoriano

jsoriano: > Should we skip this test package in the CI scripts ? @jsoriano

Yeah, we will need to find a way to test this package only when independent agents are enabled.

test/packages/with-custom-agent/auditd_manager_independent_agent/data_stream/auditd/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,12 @@
...
+    audit_rules:
+      - "-a always,exit -F arch=b64 -S execve,execveat -k exec"
+    preserve_original_event: true
+agent:
+  user: "root"
+  # shared_pid_address_space: true ?
+  pid_mode: "host"
+  capabilities:
+    - AUDIT_CONTROL
+    - AUDIT_READ
```
mrodm, line 12: Currently, this file allows additional properties. However, it probably would be better to at least add `agent` field as an object in this file. WDYT ?

jsoriano: Yes please.

mrodm: Created PR for this : https://github.com/elastic/package-spec/pull/739

test/packages/with-custom-agent/auditd_manager_independent_agent/data_stream/auditd/_dev/test/system/test-default-config.yml
```diff
@@ -0,0 +1,13 @@
+data_stream:
+  vars:
+    audit_rules:
+      - "-a always,exit -F arch=b64 -S execve,execveat -k exec"
+    preserve_original_event: true
+agent:
+  user: "root"
+  runtime: docker
+  # shared_pid_address_space: true ?
+  pid_mode: "host"
```
mrodm, line 10: This setting looks like that it is more related to docker runtime.

It is needed to specify somehow this setting for `auditd_manager` package to keep the same behavior. That setting is present in the current configuration:
https://github.com/elastic/elastic-package/blob/052cfc8d3d0bfba8355cdea12289ee1a40fa4949/test/packages/with-custom-agent/auditd_manager/data_stream/auditd/_dev/deploy/agent/custom-agent.yml#L4-L8

Should it be added here even if it is more docker related ? or maybe do we need to add a map for docker specific settings ? 
```
agent:
  user: root
  runtime: docker
  capabilities:
    - AUDIT_CONTROL
    - AUDIT_READ
  docker:
    pid_mode: host
```

Maybe it can be changed the name of this field to `use_host_pid_namespace: true`

WDYT @jsoriano ?

jsoriano: PID namespacing is not specific of docker, so I would not use an specific docker variable. Not sure about the variable name to use but I wouldn't use a boolean because there can be more than one option.

internal/agentdeployer/info.go
```diff
@@ -71,6 +71,17 @@ type AgentInfo struct {
...
+		// User user to run Elastic Agent process
+		User string
+		// PidMode selects the host PID mode
+		// (From docker-compose docs) Turns on sharing between container and the host
+		// operating system the PID address space
+		PidMode string
+		// Runtime is the selected runtime to run the Elastic Agent process
+		Runtime string
+		// Capabilities is a list of the capabilities needed to run the Elastic Agent process
+		Capabilities []string
```
jsoriano, line 84: Nit. Maybe we can call this `LinuxCapabilities` to avoid confusion with other capabilities.
```suggestion
		// LinuxCapabilities is a list of the capabilities needed to run the Elastic Agent process
		LinuxCapabilities []string
```

jsoriano (approved):
:+1:

internal/agentdeployer/info.go
mrodm: Agreed, do you think it should be updated that naming in the configuration file for system tests ? @jsoriano 

Currently, those settings in the configuration file looks like:
```yaml
agent:
  runtime: docker
  user: "root"
  pid_mode: "host"
  capabilities:
    - AUDIT_CONTROL
    - AUDIT_READ
```
```diff
   runtime: docker
   user: "root"
   pid_mode: "host"
-  capabilities:
+  linux_capabilities:
     - AUDIT_CONTROL
     - AUDIT_READ
```

jsoriano: Yes, call them `linux_capabilities` in the config.

jsoriano (approved):

PR #1768, opened 2024-04-12
Allow packages without service deployer to have system tests

jsoriano:
This will allow to leverage the new custom definitions added in https://github.com/elastic/elastic-package/pull/1765, without forcing the packages to define a service deployer.

This is useful when the only requirement to test the package is a running agent, for example when collecting information from the system where it is running.

Part of https://github.com/elastic/elastic-package/issues/787


internal/agentdeployer/factory.go
```diff
@@ -80,23 +67,51 @@ func Factory(options FactoryOptions) (AgentDeployer, error) {
...
+	if agentDeployerName == "docker" || agentDeployerName == "tf" {
+		return "default", "", nil
+	}
+
+	agentDeployerPath := filepath.Join(devDeployPath, agentDeployerName)
+	if info, err := os.Stat(agentDeployerPath); err != nil {
+		return "", "", fmt.Errorf("could not find deployer path for agent deployer %q: %w", agentDeployerName, err)
+	} else if errors.Is(err, os.ErrNotExist) || !info.IsDir() {
+		return "", "", fmt.Errorf("expected to find a directory in %s", agentDeployerPath)
+	}
```
mrodm, line 110: In case, this file path does not exist I think it is not going to execute the `else` block since `err != nil`.

Should this condition be added inthe first `if` ?

```suggestion
	if info, err := os.Stat(agentDeployerPath); err != nil && !errors.Is(err, os.ErrNotExist) {
		return "", "", fmt.Errorf("could not find deployer path for agent deployer %q: %w", agentDeployerName, err)
	} else if errors.Is(err, os.ErrNotExist) || !info.IsDir() {
		return "", "", fmt.Errorf("expected to find a directory in %s", agentDeployerPath)
	}
```

internal/agentdeployer/factory.go
```diff
@@ -80,23 +67,51 @@ func Factory(options FactoryOptions) (AgentDeployer, error) {
...
+		return "default", "", nil
+	}
+	// if package defines `_dev/deploy/docker` folder to start their services, it should be
+	// using the default agent deployer`
+	if agentDeployerName == "docker" || agentDeployerName == "tf" {
+		return "default", "", nil
+	}
+
+	agentDeployerPath := filepath.Join(devDeployPath, agentDeployerName)
+	if info, err := os.Stat(agentDeployerPath); err != nil {
```
mrodm, line 106: Question: should this path be already ensure that exists ? 

`devDeployPath` should be a valid path returned by `FindDevDeployPath` but looking at the code it could be a file instead of a directory.

What about moving this check if that path is a folder into `FindDevDeployPath` method? 


The string returned by `findAgentDeployer` should be already a directory (maybe there is an edge case that I missed):

https://github.com/elastic/elastic-package/blob/f8153ab029d82c2107f3c6d0f6de5a9a478550d9/internal/agentdeployer/factory.go#L127

internal/agentdeployer/factory.go
```diff
@@ -80,23 +67,51 @@ func Factory(options FactoryOptions) (AgentDeployer, error) {
...
+	}
+	if err != nil {
+		return "", "", fmt.Errorf("can't find \"%s\" directory: %w", options.DevDeployDir, err)
+	}
+
+	agentDeployerName, err := findAgentDeployer(devDeployPath)
+	if err != nil {
+		logger.Debugf("Not found any agent deployer, using default one")
+		return "default", "", nil
+	}
```
mrodm, line 98: The error returned by `findAgentDeployer` could be that there are more than one deployer folders defined.

It needs to be taken into account that too. One option could be adding an specific error an check if that error is returned.

internal/servicedeployer/factory.go
mrodm, line 1: This factory (the methods in it) should be updated too, shouldn't it ?

If there is no deployer folder, should not this be failing ?
https://github.com/elastic/elastic-package/blob/f8153ab029d82c2107f3c6d0f6de5a9a478550d9/internal/servicedeployer/factory.go#L47-L57

internal/testrunner/runners/system/runner.go
```diff
@@ -1118,7 +1120,10 @@ func (r *runner) setupService(ctx context.Context, config *testConfig, serviceOp
 	}
 
 	serviceDeployer, err := servicedeployer.Factory(serviceOptions)
-	if err != nil {
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, svcInfo, nil
+	}
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
```
mrodm, line 1126: This change should also be applied into `internal/service/boot.go` 
https://github.com/elastic/elastic-package/blob/f8153ab029d82c2107f3c6d0f6de5a9a478550d9/internal/service/boot.go#L38

and `internal/benchrunner/runners/system/runner.go`
https://github.com/elastic/elastic-package/blob/f8153ab029d82c2107f3c6d0f6de5a9a478550d9/internal/benchrunner/runners/system/runner.go#L251

test/packages/other/test_without_service_deployer/manifest.yml
```diff
@@ -0,0 +1,36 @@
+format_version: 3.1.3
+name: test_without_service_deployer
+title: "Package with system tests and no service deployer."
```
mrodm, line 3: Looking at the system test config , this package looks like the `log` input package from the integrations repository.

internal/testrunner/runners/system/runner.go
```diff
@@ -1118,7 +1120,10 @@ func (r *runner) setupService(ctx context.Context, config *testConfig, serviceOp
 	}
 
 	serviceDeployer, err := servicedeployer.Factory(serviceOptions)
-	if err != nil {
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, svcInfo, nil
```
mrodm, line 1124: It could be added a logger Debug message here to note that no service is going to be setup.

```suggestion
	if errors.Is(err, os.ErrNotExist) {
		logger.Debug("skip setting service, no deployer defined")	
		return nil, svcInfo, nil
```

internal/servicedeployer/factory.go
jsoriano: The situation is slightly different between the service and the agent deployer. We always want an agent, but we don't always want a service, so the service deployer is easier, if there is no one, we just return an `ErrNotExist`.
But you were right that `findServiceDeployer` needed to be modified to return `ErrNotExist` when no service deployer is found in the dev deploy path.

test/packages/other/test_without_service_deployer/manifest.yml
jsoriano: This is basically the vanilla package created with `elastic-package create package`, with the system test and some additional fields.

internal/testrunner/runners/system/runner.go
jsoriano: Done in both places, but not really tested in benchmarks, I will address issues there if any in a follow up.

jsoriano:
Thanks @mrodm for the review!

mrodm (approved):
Thanks!

PR #1746, opened 2024-04-03
internal/fields: fix handling of numeric keyword fields in arrays

efd6:
Currently, the keyword validation check does not allow for the existence of arrays of numbers. Add that support.

Please take a look.

PR #1773, opened 2024-04-16
Fix validation of empty numeric keywords arrays with normalization

jsoriano:
Fixes issues happening when validating fields defined as numeric keywords with array normalization, and documents can contain empty arrays.

Arrays of numeric keywords are converted to arrays of strings, instead of a string of the array.

This fixes validation issues like this one where the empty array should be valid:
```
field "tags" is not normalized as expected: expected array, found "[]" (string)
```

Related to https://github.com/elastic/elastic-package/pull/1746.

### DIFF
--- a/internal/fields/testdata/fields/fields.yml
+++ b/internal/fields/testdata/fields/fields.yml
@@ -3,6 +3,10 @@
   fields:
     - name: code
       type: keyword
+    - name: pid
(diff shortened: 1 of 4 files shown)


jsoriano:
test integrations

mrodm (approved):
Thanks!

PR #1774, opened 2024-04-16
Reduce verbosity when checking if package embeds ECS

jsoriano:
We are logging now `Package does not embed ECS mappings` every time we check if we should embed ECS mappings because they are referenced in the package.
This is not so relevant because the final reason to import ECS schema during validation depends on two other variables at the moment, but it sometimes floods the logs. For example when running the pipeline tests of the Azure package, this line is printed [33 times in a row](https://buildkite.com/elastic/integrations/builds/10450#018ee48b-d9f0-4ff5-ae38-260bdaaeedd4/857-945), and is not deterministic on its own for the actual result.

This change removes this debug log, and when the mappings are effectively considered for fields validation, the possible reasons for importing are printed too.

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -320,7 +320,7 @@ func initDependencyManagement(packageRoot string, specVersion semver.Version, im
 		if err != nil {
 			return nil, nil, err
 		}
-		logger.Debug("Imported ECS fields definition from external schema for validation (package: %v, stack: %v)", packageEmbedsEcsMappings, stackSupportsEcsMapping)
+		logger.Debugf("Imported ECS fields definition from external schema for validation (package: %v, stack: %v)", packageEmbedsEcsMappings, stackSupportsEcsMapping)
 		schema = ecsSchema
 	}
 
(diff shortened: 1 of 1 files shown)


internal/fields/validate.go
```diff
@@ -323,7 +320,7 @@ func initDependencyManagement(packageRoot string, specVersion semver.Version, im
 		if err != nil {
 			return nil, nil, err
 		}
-		logger.Debug("Imported ECS fields definition from external schema for validation")
+		logger.Debug("Imported ECS fields definition from external schema for validation (package: %v, stack: %v)", packageEmbedsEcsMappings, stackSupportsEcsMapping)
```
mrodm, line 323: Nit. Requires `Debugf` and `packageEmbedsECSMappings` is a bool

This is the current message shown ([link](https://buildkite.com/elastic/elastic-package/builds/2824#018ee696-716e-4961-9127-5cd9a74690e6/111-938)):
```
Imported ECS fields definition from external schema for validation (package: %v, stack: %v)false true
```

Maybe too descriptive like this?

```suggestion
		logger.Debugf("Imported ECS fields definition from external schema for validation (package embeds ECS: %v, stack supports ECS: %v)", packageEmbedsEcsMappings, stackSupportsEcsMapping)
```

mrodm (approved):

PR #1771, opened 2024-04-15
Remove created agents in service deployer

mrodm:
Part of #787

Currently, service deployers like `custom_agent` or `kubernetes` create new Elastic Agent instances and those agents are never removed (or unenrolled) from Fleet once the tests are finished (system tests).

This PR removes these Elastic Agents created during tests from Fleet, so the stack state after running the tests is always without those Elastic Agents.

PR #1776, opened 2024-04-16
Replace agent names in condition to remove agent handler

mrodm:
Follows #1771 

This PR removes the usage of Elastic Agent names as part of the condition used in the remove Agent handler.

A new field is added in the `Info` struct to set whether or not that service uses an independent Elastic Agent. Currently, this happens for packages using custom and k8s agents.

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -340,6 +340,8 @@ func (r *runner) createServiceInfo() (servicedeployer.ServiceInfo, error) {
 		svcInfo.OutputDir = outputDir
 	}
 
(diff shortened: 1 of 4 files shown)


mrodm:
test integrations

internal/testrunner/runners/system/runner.go
```diff
@@ -340,6 +340,8 @@ func (r *runner) createServiceInfo() (servicedeployer.ServiceInfo, error) {
 		svcInfo.OutputDir = outputDir
 	}
 
+	svcInfo.Agent.Independent = false
```
mrodm, line 343: By default services are not using independent agents

PR #1782, opened 2024-04-18
Restore append before checking error

mrodm:
Restore appending results even if there is an error.

For instance, system tests could exit using `result.WithError()` method and that method could return both results and error:
https://github.com/elastic/elastic-package/blob/6eda8fe74f7d014cfbbc49a382f4c3096f1cb98b/internal/testrunner/testrunner.go#L145


jsoriano (approved):
Thanks!

PR #1785, opened 2024-04-18
Remove support for Docker Compose v1

jsoriano:
elastic-package is likely not working with Docker Compose v1 at this point, and it is discontinued since last year. Completely remove support for it.

### DIFF
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -195,16 +195,13 @@ func NewProject(name string, paths ...string) (*Project, error) {
 		c.dockerComposeStandalone = c.dockerComposeStandaloneRequired()
 	}
 
-	if logger.IsDebugMode() {
-		// Passing a nil context here because we are on initialization.
-		ver, err := c.dockerComposeVersion(context.Background())
-		if err != nil {
-			logger.Errorf("Unable to determine Docker Compose version: %v", err)
-			return &c, nil
-		}
-		versionMessage := fmt.Sprintf("Determined Docker Compose version: %v", ver)
-		logger.Debug(versionMessage)
+	// Passing a nil context here because we are on initialization.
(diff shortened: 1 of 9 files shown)


internal/compose/compose.go
```diff
@@ -199,20 +198,13 @@ func NewProject(name string, paths ...string) (*Project, error) {
...
 	ver, err := c.dockerComposeVersion(context.Background())
 	if err != nil {
-		logger.Errorf("Unable to determine Docker Compose version: %v. Defaulting to 1.x", err)
-		c.dockerComposeV1 = true
+		logger.Errorf("Unable to determine Docker Compose version: %v", err)
 		return &c, nil
 	}
-
-	versionMessage := fmt.Sprintf("Determined Docker Compose version: %v", ver)
-	if ver.Major() == 1 {
```
mrodm, line 208: Should we keep this check and raise an error if versions 1.* are used ?

internal/compose/compose.go
```diff
@@ -199,20 +198,15 @@ func NewProject(name string, paths ...string) (*Project, error) {
 	// Passing a nil context here because we are on initialization.
 	ver, err := c.dockerComposeVersion(context.Background())
 	if err != nil {
-		logger.Errorf("Unable to determine Docker Compose version: %v. Defaulting to 1.x", err)
-		c.dockerComposeV1 = true
-		return &c, nil
+		return nil, fmt.Errorf("unable to determine Docker Compose version: %w", err)
```
jsoriano, line 201: Making this an error now, the main reason for this to fail is that docker compose command is not found.

internal/compose/compose.go
jsoriano: Kept, good point.

mrodm (approved):
:rocket:

PR #1791, opened 2024-04-18
Store start time just before installing pipelines

mrodm:
Ensure that the start time for pipeline is stored just before installing the pipelines in case there are warnings in Elasticsearch during the pipeline installation itself.

To avoid errors like in this build: https://buildkite.com/elastic/elastic-package/builds/2892#018ef14c-bb78-4d02-9080-bc1daabc9503


mrodm:
test integrations

jsoriano (approved):

Issue #1355, opened 2023-07-13
[Feature Request] Allow Opening Ports on Elastic Agent Container


MakoWish:
As it is now, the Elasticsearch container opens port 127.0.0.1:9200, the Kibana contain opens port 127.0.0.1:5601, and the Fleet Agent container opens port 127.0.0.1:8220. Unfortunately, the Elastic Agent container does not open any ports at all, and there does not appear to be a way to open one.

It would be nice to have a feature that allows opening custom ports, preferably allowing externally-accessible like 0.0.0.0:9010, so package development can be done with live data. This would exactly match a production environment, from source, to final indexed documents, instead of relying on a fixed set of sample logs using [Stream](https://github.com/elastic/stream/). This would give the best chance of catching anomalous events during development, as well as a complete A-to-B test of the integration.

jsoriano:
Related: https://github.com/elastic/elastic-package/issues/1262

Issue #1262, opened 2023-05-11
add ability to set publish ports to `config.yml`


leehinman:
Previously we could edit `snapshot.yml` and change the publish ports to do things like listen on a different network interface from localhost.  With #1140 we can't do that anymore because `snapshot.yml` gets overwritten every time "stack up" is called.

Use Case:  When testing different OSes, for example you want the stack running on Linux, but an elastic-agent running on Windows.

dot-mike:
Any update to this? An idea could be to support multiple compose files: https://docs.docker.com/compose/how-tos/multiple-compose-files/extends/#understanding-multiple-compose-files
and use the `!override` feature to replace value https://docs.docker.com/reference/compose-file/merge/#replace-value

Command to boot the stack would then be: `docker compose -f docker-compose.yml -f docker-compose.override.yml up -d`


A base `docker-compose.yml` file:

```
services:
  kibana:
    image: "${KIBANA_IMAGE_REF}"
    ports:
      - "127.0.0.1:5601:5601"
```

Override port mapping for kibana with file `docker-compose.override.yml`

```
services:
  kibana:
    ports: !override
      - "0.0.0.0:5601:5601"
```

This results in:

```
services:
  kibana:
    image: "${KIBANA_IMAGE_REF}"
    ports:
      - "0.0.0.0:5601:5601"
```


-----

I had the need to access Kibana on port 5601 in Docker using WSL, but from Windows., I can't access 127.0.0.1 directly in Windows.   So I would like to change listening host for 5601. But this is not possible.

A workaround for now is using socat to access Kibana with WSL for port-forwarding to 5600 so I can access on port 5600 in Windows which Socat listen on for the WSL IP-address.

```bash
elastic-package stack up -d -s elasticsearch,kibana
HOST=$(ip addr show eth0 | grep 'inet\b' | awk '{print $2}' | cut -d/ -f1)
# note that ssl is enabled with stack so we need to listen on SSL with port 5600
socat OPENSSL-LISTEN:5600,cert=$HOME/.elastic-package/profiles/default/certs/kibana/cert.pem,key=$HOME/.elastic-package/profiles/default/certs/kibana/key.pem,verify=0,fork OPENSSL:127.0.0.1:5601,verify=0

curl -vv -k https://172.25.208.198:5600/
> GET / HTTP/1.1
> Host: 172.25.208.198:5600
>
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
[...]

PR #1793, opened 2024-04-18
feat(compose/agent): add stack.agent.ports option in config.yml to publish ports for agent

pkoutsovasilis:
This PR introduces support for defining inside `config.yml` of the profile ports for agent service to be published

- Relates to https://github.com/elastic/elastic-package/issues/1262
- Relates to https://github.com/elastic/elastic-package/issues/1355

### DIFF
--- a/internal/profile/config.go
+++ b/internal/profile/config.go
@@ -5,7 +5,6 @@
 package profile
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -13,6 +12,8 @@ import (
 	"github.com/elastic/go-ucfg/yaml"
 
 	"github.com/elastic/elastic-package/internal/common"
+
+	"github.com/mitchellh/mapstructure"
 )
 
 type config struct {
@@ -50,7 +51,7 @@ func (c *config) get(name string) (string, bool) {
 	}
 }
 
-func (c *config) Unmarshal(name string, out any) error {
+func (c *config) Decode(name string, out any) error {
 	v, err := c.settings.GetValue(name)
 	if err != nil {
 		if errors.Is(err, common.ErrKeyNotFound) {
@@ -58,13 +59,7 @@ func (c *config) Unmarshal(name string, out any) error {
 		}
 		return err
 	}
-
-	data, err := json.Marshal(v)
-	if err != nil {
-		return err
-	}
-
-	if err := json.Unmarshal(data, out); err != nil {
+	if err := mapstructure.Decode(v, out); err != nil {
 		return err
 	}
 
--- a/internal/profile/config.go
+++ b/internal/profile/config.go
@@ -10,10 +10,9 @@ import (
 	"os"
 
 	"github.com/elastic/go-ucfg/yaml"
+	"github.com/mitchellh/mapstructure"
 
 	"github.com/elastic/elastic-package/internal/common"
-
-	"github.com/mitchellh/mapstructure"
 )
 
 type config struct {
--- a/internal/profile/config.go
+++ b/internal/profile/config.go
@@ -5,6 +5,7 @@
 package profile
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -47,4 +48,25 @@ func (c *config) get(name string) (string, bool) {
 	default:
 		return fmt.Sprintf("%v", v), true
 	}
+}
+
+func (c *config) Unmarshal(name string, out any) error {
+	v, err := c.settings.GetValue(name)
+	if err != nil {
+		if errors.Is(err, common.ErrKeyNotFound) {
+			return nil
+		}
+		return err
+	}
+
+	data, err := json.Marshal(v)
+	if err != nil {
+		return err
+	}
+
+	if err := json.Unmarshal(data, out); err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/internal/profile/profile.go
+++ b/internal/profile/profile.go
@@ -141,13 +141,13 @@ type Profile struct {
 }
 
 // Path returns an absolute path to the given file
-func (profile Profile) Path(names ...string) string {
+func (profile *Profile) Path(names ...string) string {
 	elems := append([]string{profile.ProfilePath}, names...)
 	return filepath.Join(elems...)
 }
 
 // Config returns a configuration setting, or its default if setting not found
-func (profile Profile) Config(name string, def string) string {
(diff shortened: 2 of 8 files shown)


jsoriano:
I wonder if https://github.com/elastic/elastic-package/pull/1795 supersedes this one.

@pkoutsovasilis is there any use case that would not be covered by defining the ports in the system test cases?

pkoutsovasilis:
> I wonder if https://github.com/elastic/elastic-package/pull/1795 supersedes this one.
> 
> @pkoutsovasilis is there any use case that would not be covered by defining the ports in the system test cases?

hmmm interesting 🤔 the answer is yes when you are developing a new integration or testing something about an existing one. Sure this should end up in a system test but before you have a test you have to do certain manual steps to get there. Also, since this PR is now open for 6 days, if you need to extend this one to custom agents deployments as this https://github.com/elastic/elastic-package/pull/1795 (2 days ago) does, wouldn't make sense to ask for it here than open a complete different PR?

jsoriano:
> the answer is yes when you are developing a new integration or testing something about an existing one.

Makes sense. This is probably also useful in cases where the stack is used for manual testing.

> since this PR is now open for 6 days, if you needed support for custom agents deployments as this #1795 (2 days ago) does, wouldn't make sense to ask for it here than open a complete different PR?

Well, they likely come from two different needings, we probably need both approaches. The one here for manual tests, and the other one for testing and CI in the new custom agents.

pkoutsovasilis:
> Makes sense. This is probably also useful in cases where the stack is used for manual testing.
> 
> >

and for manual testing/development sometimes we need also custom volume mounts 😄 e.g. for this [one](https://github.com/elastic/integrations/blob/main/packages/fim/manifest.yml) we need this `/sys/kernel/debug:/sys/kernel/debug`

internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -139,6 +139,9 @@ services:
       interval: 5s
     hostname: docker-fleet-agent
     env_file: "./elastic-agent.env"
+    {{- if eq (fact "agent_has_publish_ports") "true" }}
+    ports: {{ fact "agent_publish_ports" }}
+    {{- end }}
```
jsoriano, line 144: Nit. Maybe we can pass here the comma separated list of ports, without the brackets, and check here if the list is empty? This way we would not need the `has_publish_ports` fact.
```suggestion
    {{- if ne (fact "agent_publish_ports") "" }}
    ports: {{ fact "agent_publish_ports" }}
    {{- end }}
```

internal/stack/resources.go
```diff
@@ -152,9 +152,25 @@ func applyResources(profile *profile.Profile, stackVersion string) error {
...
+		facter["agent_publish_ports"] = fmt.Sprintf("[%s]", strings.Join(agentPorts, ","))
+	} else {
+		facter["agent_has_publish_ports"] = "false"
+	}
+
+	resourceManager.AddFacter(facter)
+
+	if err := os.MkdirAll(stackDir, 0755); err != nil {
+		return fmt.Errorf("failed to create stack directory: %w", err)
+	}
```
jsoriano, line 173: This should not be needed :thinking:

internal/profile/profile.go
```diff
@@ -141,13 +141,13 @@ type Profile struct {
 }
 
 // Path returns an absolute path to the given file
-func (profile Profile) Path(names ...string) string {
+func (profile *Profile) Path(names ...string) string {
```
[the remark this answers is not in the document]
pkoutsovasilis: so here it depends, it is considered an unusual pattern to have a struct definition that mixes both value and pointer receivers, and from a quick look at the code, profile is widely used as a pointer. Thus I homogenise the struct by making these pointer receivers but if there is a reason behind this divergence happy revert this change

internal/stack/resources.go
pkoutsovasilis: I just added the err, `os.MkdirAll...` was already there but without being checked. If we want the error to be ignored there then I would prefer have it as `_ = os.MkdirAll...` so it is evident that we explicitly ignore this 🙂

pkoutsovasilis:
> Well, they likely come from two different needings, we probably need both approaches. The one here for manual tests, and the other one for testing and CI in the new custom agents.

the more I look the agentdeployer the more it makes sense to me expose this through cli and someone be able to deploy "custom" agents in the stack and focus any extendability effort there, @jsoriano what do you think?

internal/profile/config.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/elastic/go-ucfg/yaml"
 
 	"github.com/elastic/elastic-package/internal/common"
+
+	"github.com/mitchellh/mapstructure"
```
jsoriano, line 16: Nit. Imports order.
```suggestion
	"github.com/elastic/go-ucfg/yaml"
	"github.com/mitchellh/mapstructure"

	"github.com/elastic/elastic-package/internal/common"
```

internal/profile/profile.go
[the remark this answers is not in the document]
jsoriano: Good point. I think I would rather use the profile by value, but it is true that it is mostly used as pointer now, so as you prefer.

internal/profile/config.go
```diff
@@ -48,3 +50,18 @@ func (c *config) get(name string) (string, bool) {
 		return fmt.Sprintf("%v", v), true
 	}
 }
+
+func (c *config) Decode(name string, out any) error {
```
jsoriano, line 54: Please consider adding a test for this.

internal/stack/resources.go
jsoriano: It is fine.

jsoriano (approved):
Thanks! I think that with this change and with https://github.com/elastic/elastic-package/pull/1795, we can close https://github.com/elastic/elastic-package/issues/1355.

jsoriano:
> the more I look the agentdeployer the more it makes sense to me expose this through cli and someone be able to deploy "custom" agents in the stack and focus any extendability effort there, @jsoriano what do you think?

Yeah, I see the agent deployer as the future to deploy any kind of agents. There is still some work to do in this area though, @mrodm is now working on this.

In any case we have to keep in mind that the purpose of `elastic-package` is to be the main tool for package developers. We should be cautious about adding features that go beyond package development.

jsoriano (approved):
:+1:

PR #1794, opened 2024-04-22
Drop all capabilities by default in Elastic Agent containers

mrodm:
Part of #787 

In docker/docker-compose there are some capabilities that are added by default to every container (https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities).
This PR ensures that agents are started with the minimum required capabilities, dropping by default all of them.

If `ALL` it is defined in `cap_drop` field, then it just adds the Linux capabilities for the container defined in `cap_add`:
https://github.com/moby/moby/blob/82d8f8d6e6dbc88ed437b8bf6c38399b46ba7d93/oci/caps/utils.go#L112-L114

### DIFF
--- a/internal/agentdeployer/_static/docker-agent-base.yml.tmpl
+++ b/internal/agentdeployer/_static/docker-agent-base.yml.tmpl
@@ -18,6 +18,8 @@ services:
       - {{ . }}
     {{- end }}
     {{ end }}
+    cap_drop:
+      - ALL
     environment:
       - FLEET_ENROLL=1
       - FLEET_URL=https://fleet-server:8220
--- a/internal/servicedeployer/_static/docker-custom-agent-base.yml
+++ b/internal/servicedeployer/_static/docker-custom-agent-base.yml
@@ -6,6 +6,8 @@ services:
       retries: 180
       interval: 1s
     hostname: docker-custom-agent
+    cap_drop:
+      - ALL
     environment:
       - FLEET_ENROLL=1
       - FLEET_URL=https://fleet-server:8220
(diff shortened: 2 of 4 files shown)


internal/agentdeployer/_static/docker-agent-base.yml.tmpl
```diff
@@ -18,6 +18,8 @@ services:
       - {{ . }}
     {{- end }}
     {{ end }}
+    cap_drop:
+      - ALL
```
jsoriano, line 22: What do you think about trying to drop all capabilities also in the main docker compose, and test with integrations?

mrodm: Ok, I'll update those scenarios and run a test with integrations.

Wondering what to do in the template used for custom agents (servicedeployer). For that case, packages could also define some `cap_drop`. It's also true that until now some capabilities are just added with `cap_add`. But looking at the code of `moby`, if ALL is present in cap_drop , all capabilities are drop...

mrodm:
test integrations

internal/agentdeployer/_static/docker-agent-base.yml.tmpl
mrodm: It looks like there are no issues when cap_drop ALL is set:
https://github.com/elastic/integrations/pull/9694

mrodm: Should we update also custom agent template ? As there are no packages using `cap_drop` in the integrations repository, probably it's a good idea, WDYT ? 

https://github.com/elastic/elastic-package/blob/5e322d7be61a739b855689aa0c4342abd5fa9d68/internal/servicedeployer/_static/docker-custom-agent-base.yml

jsoriano: Yes, good idea.

jsoriano (approved):
Nice :+1:

mrodm:
test integrations

jsoriano (approved):

PR #1795, opened 2024-04-22
Add support to define expose ports for independent agents

mrodm:
Part of #787
Relates #1262 #1355

This PR allows users to define which ports should be exposed from the Elastic Agent container.ç
This definition should be added to the system configuration file. Example:

```yaml
agent:
  ports:
    - "127.0.0.1:3000:3000"
```

This example is based on agents using docker as a runtime https://docs.docker.com/compose/compose-file/compose-file-v2/#ports

### DIFF
--- a/internal/testrunner/runners/system/test_config.go
+++ b/internal/testrunner/runners/system/test_config.go
@@ -58,6 +58,7 @@ type testConfig struct {
 		PidMode           string   `config:"pid_mode"`
 		LinuxCapabilities []string `config:"linux_capabilities"`
 		Runtime           string   `config:"runtime"`
+		ExposePorts       []string `config:"expose_ports"`
 	} `config:"agent"`
 }
 
--- a/internal/testrunner/runners/system/test_config.go
+++ b/internal/testrunner/runners/system/test_config.go
@@ -58,7 +58,7 @@ type testConfig struct {
 		PidMode           string   `config:"pid_mode"`
 		LinuxCapabilities []string `config:"linux_capabilities"`
 		Runtime           string   `config:"runtime"`
-		ExposePorts       []string `config:"expose_ports"`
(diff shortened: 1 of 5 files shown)


internal/testrunner/runners/system/test_config.go
```diff
@@ -58,6 +58,7 @@ type testConfig struct {
 		PidMode           string   `config:"pid_mode"`
 		LinuxCapabilities []string `config:"linux_capabilities"`
 		Runtime           string   `config:"runtime"`
+		ExposePorts       []string `config:"expose_ports"`
```
jsoriano, line 61: Nit. Have you thought about calling this just `ports`?
```suggestion
		ExposePorts       []string `config:"ports"`
```

jsoriano (approved):
:+1:

internal/testrunner/runners/system/test_config.go
mrodm: I wanted to emphasize that these ports are the ones to be exposed. But re-thinking that idea, if it used `ports` as the name in those settings, it looks like is clear that it refers to the ports to be exposed (or forwarded from the host like in VMs).

For instance, docker-compose uses ports: https://docs.docker.com/compose/compose-file/compose-file-v3/#ports
But in Vargrantfile, it is used forwarded_ports: https://developer.hashicorp.com/vagrant/docs/networking/forwarded_ports#defining-a-forwarded-port

I think it could be just "ports".

I'll update the PR

mrodm: Updated in https://github.com/elastic/elastic-package/pull/1795/commits/2af3facfe61da8aa6fc621cc249b252033713e64

jsoriano (approved):

PR #1796, opened 2024-04-22
Update volumes to use bind type

mrodm:
Use the same definitions for volumes as in the Elastic Agent defined in the stack:
https://github.com/elastic/elastic-package/blob/c2ff98b32e5cca207cb5858a973a91d1a3ffaf3f/internal/stack/_static/docker-compose-stack.yml.tmpl#L142-L150

### DIFF
--- a/internal/agentdeployer/_static/docker-agent-base.yml.tmpl
+++ b/internal/agentdeployer/_static/docker-agent-base.yml.tmpl
@@ -31,9 +31,9 @@ services:
       - type: bind
         source: ${SERVICE_LOGS_DIR}
         target: /tmp/service_logs/
-        read_write: true
+        read_only: false
       # Mount service_logs under /run too as a testing workaround for the journald input (see elastic-package#1235).
       - type: bind
         source: ${SERVICE_LOGS_DIR}
         target: /run/service_logs/
-        read_write: true
+        read_only: false
--- a/internal/agentdeployer/_static/docker-agent-base.yml.tmpl
+++ b/internal/agentdeployer/_static/docker-agent-base.yml.tmpl
@@ -24,11 +24,16 @@ services:
       - KIBANA_HOST=https://kibana:5601
       - FLEET_TOKEN_POLICY_NAME=${FLEET_TOKEN_POLICY_NAME}
     volumes:
-      - ${LOCAL_CA_CERT}:/etc/ssl/certs/elastic-package.pem
+      - type: bind
+        source: ${LOCAL_CA_CERT}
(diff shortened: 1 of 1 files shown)


internal/agentdeployer/_static/docker-agent-base.yml.tmpl
```diff
@@ -24,5 +24,11 @@ services:
       - KIBANA_HOST=https://kibana:5601
       - FLEET_TOKEN_POLICY_NAME=${FLEET_TOKEN_POLICY_NAME}
     volumes:
-      - ${SERVICE_LOGS_DIR}:/tmp/service_logs/
       - ${LOCAL_CA_CERT}:/etc/ssl/certs/elastic-package.pem
```
jsoriano, line 27: Do you know what is the default type when using this syntax?

jsoriano (approved):

internal/agentdeployer/_static/docker-agent-base.yml.tmpl
mrodm: Just checked in a local run, and the type is also "bind", but the Mode changes from "rw" to empty string. I haven't found what that means exactly... 

```json
        "Mounts": [
            {
                "Type": "bind",
                "Source": "/home/mariorodriguez/.elastic-package/profiles/default/certs/ca-cert.pem",
                "Destination": "/etc/ssl/certs/elastic-package.pem",
                "Mode": "rw",
                "RW": true,
                "Propagation": "rprivate"
            },
            {
                "Type": "bind",
                "Source": "/home/mariorodriguez/.elastic-package/tmp/service_logs/agent-sophos-utm",
                "Destination": "/tmp/service_logs",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
            },
            {
                "Type": "bind",
                "Source": "/home/mariorodriguez/.elastic-package/tmp/service_logs/agent-sophos-utm",
                "Destination": "/run/service_logs",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
            }
        ],
```

mrodm: Maybe the service that was failing was because it was expecting the logs in `/run/service_logs` ? 

At the end, it looks similar the configuration shown with a `docker inspect`

mrodm: Updated settings to set explicitly read only or read write for those volumes, and keeping the same syntax in all mounts.

jsoriano (approved):

PR #1798, opened 2024-04-23
Shorten docker container name

mrodm:
Relates #787
Relates https://github.com/elastic/integrations/pull/9660

Until now the elastic agent names created (`hostname`) are set as:
`elastic-agent-<package>-<data-stream>-<run_id>`

Using that pattern, it caused that for some packages the resulting name is longer than 64 characters causing an error in docker-compose.
```
 $ echo -n "elastic-agent-carbon_black_cloud-asset_vulnerability_summary-62403" |wc -c 
66
```

This PR changes the prefix to be just `agent`. For the example above it results in 58 characters:
```
 $ echo -n "agent-carbon_black_cloud-asset_vulnerability_summary-62403" |wc -c 
58
```

### DIFF
--- a/internal/agentdeployer/agent.go
+++ b/internal/agentdeployer/agent.go
@@ -24,7 +24,8 @@ import (
 )
 
 const (
-	dockerTestAgentNamePrefix    = "elastic-agent"
+	dockerTestAgentNamePrefix    = "agent"
+	dockerTestAgentServiceName   = "elastic-agent"
 	dockerTestAgentDockerCompose = "docker-agent-base.yml"
 	defaultAgentPolicyName       = "Elastic-Agent (elastic-package)"
 )
@@ -163,7 +164,7 @@ func (d *DockerComposeAgentDeployer) SetUp(ctx context.Context, agentInfo AgentI
 	}
 
 	// Service name defined in the docker-compose file
-	agentInfo.Name = dockerTestAgentNamePrefix
+	agentInfo.Name = dockerTestAgentServiceName
 	agentName := agentInfo.Name
 
 	opts := compose.CommandOptions{
--- a/internal/agentdeployer/agent.go
+++ b/internal/agentdeployer/agent.go
@@ -24,7 +24,6 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


internal/agentdeployer/agent.go
```diff
@@ -163,7 +164,7 @@ func (d *DockerComposeAgentDeployer) SetUp(ctx context.Context, agentInfo AgentI
 	}
 
 	// Service name defined in the docker-compose file
-	agentInfo.Name = dockerTestAgentNamePrefix
+	agentInfo.Name = dockerTestAgentServiceName
```
jsoriano, line 167: I wonder if with this solution we are still at risk of having the same issue if some package or data stream name is longer.

Could we truncate the package/data stream part?

So for `elastic-agent-carbon_black_cloud-asset_vulnerability_summary-62403` we'd have the truncated version `elastic-agent-carbon_black_cloud-asset_vulnerability_s-62403`.

I would be ok though with applying both changes, the change in the prefix and the change to truncate.

jsoriano: Or maybe we could just call it `elastic-agent` plus the run id, that should be unique: `elastic-agent-62403`.

jsoriano (commented):
Looks good, but I wonder if we can have a more future-proof approach.

internal/agentdeployer/agent.go
mrodm: The goal to add the package and datastream was to help the user to find the agent in the UI. It's true that as it is right now, there could be issues again if some package or data stream name is longer, so this solution would not work if that happen.

Until now the agent was set as `docker-fleet-agent`. Maybe it's not needed to set in the agent name all that information, since the agents could be found filtering by the Agent Policy. Do you thin that would be good @jsoriano ? 

So I think it could be set just adding as a suffix the RunID: `elastic-agent-12345`

mrodm: Changed here: https://github.com/elastic/elastic-package/pull/1798/commits/a161483576a3c6c42c7eba6947b9ac07e41982fc

jsoriano (approved):

PR #1801, opened 2024-04-25
Return stderr as part of the error when compose fails

jsoriano:
Collect the output from stderr when an error running docker compose happens. This should give more information about the cause of the failure, even when debug mode is not used.


mrodm (approved):
LGTM ! 
This would help us to find easier the errors raised by docker. For instance errors like: 
```
validating /home/user/Coding/work/integrations/packages/oracle_weblogic/_dev/deploy/docker/docker-compose.yml: version must be a string
```
that were just shown in debug mode.

Thanks!!

PR #1803, opened 2024-04-26
Add support for working copies created with git worktree

mrodm:
Add support for working copies created with [`git worktree` command](https://git-scm.com/docs/git-worktree/2.30.0).

Working copies created with this command do not have a `.git` folder, instead those working copies have a `.git` file where it is indicated where the reference repository is. For isntance:

```
 $ cat .git
gitdir: /home/mariorodriguez/Coding/work/elastic-package/.git/worktrees/elastic-package-add-support-git-worktree
```

Currently, `elastic-package` fails since `.git` is not a folder with this error:
- In this example, `build` folder does not exist and `elastic-package` needs to create it
```
Error: error running package system tests: could not complete test run: can't install the package: there was an apply error: failed to build package: can't locate build directory: can't locate build packages root directory: can't create new build directory: package can be only built inside of a Git repository (.git folder is used as reference point)
```

### DIFF
--- a/internal/files/repository.go
+++ b/internal/files/repository.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 
 	"gopkg.in/yaml.v3"
+
+	"github.com/elastic/elastic-package/internal/logger"
 )
 
 func FindRepositoryRootDirectory() (string, error) {
@@ -18,7 +20,10 @@ func FindRepositoryRootDirectory() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("locating working directory failed: %w", err)
 	}
+	return findRepositoryRootDirectory(workDir)
+}
 
+func findRepositoryRootDirectory(workDir string) (string, error) {
 	// VolumeName() will return something like "C:" in Windows, and "" in other OSs
 	// rootDir will be something like "C:\" in Windows, and "/" everywhere else.
 	rootDir := filepath.VolumeName(workDir) + string(filepath.Separator)
@@ -30,6 +35,7 @@ func FindRepositoryRootDirectory() (string, error) {
 			return "", err
 		}
 		if gitRepo {
+			logger.Debugf("Found git repository directory: %q", dir)
 			return dir, nil
 		}
 		if dir == rootDir {
@@ -44,7 +50,7 @@ func FindRepositoryRootDirectory() (string, error) {
 func isGitRootDir(dir string) (bool, error) {
 	path := filepath.Join(dir, ".git")
 	fileInfo, err := os.Stat(path)
-	if err != nil && errors.Is(err, os.ErrNotExist) {
+	if errors.Is(err, os.ErrNotExist) {
 		return false, nil
 	}
 	if err != nil {
--- a/internal/files/repository.go
+++ b/internal/files/repository.go
@@ -5,32 +5,13 @@
 package files
 
 import (
-	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 
 	"gopkg.in/yaml.v3"
 )
-
-func isGitWorktree(path string) error {
-	content, err := os.ReadFile(path)
-	if err != nil {
-		return fmt.Errorf("failed to read %s: %w", path, err)
-	}
-	type gitWorktree struct {
-		GitDir string `yaml:"gitdir"`
-	}
-	var worktree gitWorktree
-	dec := yaml.NewDecoder(bytes.NewBuffer(content))
-	dec.KnownFields(true)
-
-	if err := dec.Decode(&worktree); err != nil {
-		return fmt.Errorf("failed to decode %s: %w", path, err)
-	}
-
-	return nil
-}
 
 func FindRepositoryRootDirectory() (string, error) {
 	workDir, err := os.Getwd()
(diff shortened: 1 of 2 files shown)


internal/files/repository.go
```diff
@@ -5,11 +5,33 @@
...
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %w", path, err)
+	}
+	type gitWorktree struct {
+		GitDir string `yaml:"gitdir"`
+	}
+	var worktree gitWorktree
+	dec := yaml.NewDecoder(bytes.NewBuffer(content))
+	dec.KnownFields(true)
```
jsoriano, line 26: Nit. Do we need this? This can be an issue if some other field is added to this file, right?

internal/files/repository.go
```diff
@@ -5,11 +5,33 @@
...
+
+	if err := dec.Decode(&worktree); err != nil {
+		return fmt.Errorf("failed to decode %s: %w", path, err)
+	}
+
```
jsoriano, line 31: Maybe we should also check if `worktree.GitDir` is not empty?

internal/files/repository.go
```diff
@@ -5,11 +5,33 @@
...
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"gopkg.in/yaml.v3"
 )
 
+func isGitWorktree(path string) error {
```
jsoriano, line 16: Nit. We could maybe add some test cases for this function.

internal/files/repository.go
```diff
@@ -24,6 +46,14 @@ func FindRepositoryRootDirectory() (string, error) {
 	for dir != "." {
 		path := filepath.Join(dir, ".git")
 		fileInfo, err := os.Stat(path)
+		if err == nil && !fileInfo.IsDir() {
+			errWorktree := isGitWorktree(path)
+			if errWorktree != nil {
```
jsoriano, line 51: Nit. This can be `err`.
```suggestion
			err := isGitWorktree(path)
			if err != nil {
```

internal/files/repository.go
jsoriano, line 16: Nit. It would make sense for this function to return a boolean telling if this is a git worktree.
```suggestion
func isGitWorktree(path string) (bool, error) {
```

internal/files/repository.go
```diff
@@ -24,6 +46,14 @@ func FindRepositoryRootDirectory() (string, error) {
...
+				return "", errWorktree
+			}
+
+			return dir, nil
+		}
```
jsoriano, line 56: Nit. Maybe we can move all this logic to a `isGitRootDir(dir string) (bool, err)`  function, that checks both known conditions.

jsoriano (approved):
Nice! Thanks for adding this. Added some comments, but mostly nitpicking.

internal/files/repository.go
mrodm: TBH I don't know if that file could have other fields, so probably it's better to just remove it and ensure that at least the file contains "gitdir"...

internal/files/repository.go
mrodm: In this case, would the boolean returned be the same as the error? Currently, if the path is not a git worktree folder it returns an error.

internal/files/repository.go
mrodm: Sure, I'll add this

internal/files/repository.go
```diff
@@ -36,3 +40,41 @@ func FindRepositoryRootDirectory() (string, error) {
 
 	return "", os.ErrNotExist
 }
+
+func isGitRootDir(dir string) (bool, error) {
+	path := filepath.Join(dir, ".git")
+	fileInfo, err := os.Stat(path)
+	if err != nil && errors.Is(err, os.ErrNotExist) {
```
jsoriano, line 47: Nit.
```suggestion
	if errors.Is(err, os.ErrNotExist) {
```

internal/files/repository_test.go
```diff
@@ -0,0 +1,141 @@
...
+				err := os.MkdirAll(gitWorktreeFile, 0o755)
+				require.NoError(t, err)
+			}
+			err := os.MkdirAll(otherDir, 0o755)
+			require.NoError(t, err)
+
+			err = os.Chdir(otherDir)
+			require.NoError(t, err)
+
+			dir, err := FindRepositoryRootDirectory()
```
jsoriano, line 55: Changing cwd is a bit dangerous, it changes cwd in the whole process and might affect other tests. It is also probably the reason why this fails in Windows, when the test finishes, the temporary directory that should be removed is the current working directory.
Maybe we can define an internal function that looks for the root directory under a given path, and test this function.
It would be something like this:
```
func FindRepositoryRootDirectory() (string, error) {
	workDir, err := os.Getwd()
	if err != nil {
		return "", fmt.Errorf("locating working directory failed: %w", err)
	}
	return findRepositoryRootDirectory(workDir)
}

func findRepositoryRootDirectory(workDir string) (string, error) {
	...
```
And then here:
```suggestion
			dir, err := findRepositoryRootDirectory(otherDir)
```
We wouldn't be testing the part about getting the local dir, but I think this is fine enough.

mrodm: > Maybe we can define an internal function that looks for the root directory under a given path, and test this function.

I'll try with this, thanks!

jsoriano (approved):
:+1:

PR #1807, opened 2024-04-30
[Buildkite] Add pipeline to test with Elastic serverless daily

### LINKED ISSUE #1656: [Buildkite] Add tests to check elastic-package with serverless projects
Currently, there is no step in CI to check about `elastic-package` creating Serverless projects or running system/pipeline/asset/static tests with it.

There is a daily job in the [integrations repository](https://github.com/elastic/integrations) where `elastic-package` test the packages using Serverless projects: 
- [Pipeline definition](https://github.com/elastic/integrations/blob/24edad8f8a726f3e88ed50e6dee1578f4cc23037/.buildkite/pipeline.serverless.yml)
- [Buildkite link](https://buildkite.com/elastic/integrations-schedule-daily/)

However, this pipeline tests the latest elastic-package release merged into integrations. It does not test the latest changes in elastic-package before being released.

It would be interesting to have dedicated steps to check that `elastic-package` CLI can create a Serverless project and run the given tests with some package.

To be considered:
- Run tests with Observability and Security or just one project type?
- Create a new daily job or add a new step in the main pipeline? 
    - Adding an step in the main pipeline would do this test in every Pull Request and every merge to main.
- Test using also different configurations (enabled Logstash? #1645)

Related issues:
- https://github.com/elastic/elastic-package/issues/1451
- https://github.com/elastic/elastic-package/issues/1645

mrodm:
Closes #1656

Add a new pipeline to test `elastic-package` using the Elastic tack from a Serverless project (by default observability project). For that:
- A new pipeline that is going to be executed daily for Observability project
- Currently, this will be tested with all the packages under `test/packages/parallel`
    - No system tests (as in the integrations repository)
    - No coverage for pipeline tests, there are errors like:
      ```
      Error: error running package pipeline tests: could not complete test run: error calculating pipeline coverage: error fetching pipeline stats for code coverage calculations: need exactly one ES node in stats response (got 4)
      ```
    - Set an option to trigger this new pipeline using a comment in the Pull Request: `test serverless`: to be validated once this PR is merged

Example of build running Serverless step:
- https://buildkite.com/elastic/elastic-package/builds/2988
- https://buildkite.com/elastic/elastic-package/builds/2999

### DIFF
--- a/.buildkite/pipeline.serverless.yml
+++ b/.buildkite/pipeline.serverless.yml
@@ -55,7 +55,7 @@ steps:
   - wait: ~
     continue_on_failure: true
 
-  - label: ":pipeline: Test Serverless Integration tests"
+  - label: ":elastic: Serverless Integration tests"
     key: test-serverless
     command: ".buildkite/scripts/test_packages_with_serverless.sh"
     env:
--- a/.buildkite/pipeline.serverless.yml
+++ b/.buildkite/pipeline.serverless.yml
@@ -43,17 +43,6 @@ steps:
       cpu: "8"
       memory: "4G"
 
-  - label: ":go: :linux: Run unit tests"
-    key: unit-tests-linux
-    command: "make test-go-ci"
-    artifact_paths:
-      - "build/test-results/*.xml"
-      - "build/test-coverage/*.xml"
-    agents:
-      image: "${LINUX_AGENT_IMAGE}"
-      cpu: "8"
-      memory: "4G"
-
   - wait: ~
     continue_on_failure: true
 
@@ -68,8 +57,6 @@ steps:
       - build/test-coverage/coverage-*.xml
     depends_on:
       - step: check-static
-        allow_failure: false
-      - step: unit-tests-linux
         allow_failure: false
 
   - wait: ~
--- a/.buildkite/pipeline.serverless.yml
+++ b/.buildkite/pipeline.serverless.yml
@@ -1,0 +1,73 @@
+env:
+  SETUP_GVM_VERSION: 'v0.5.2' # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
+  ELASTIC_PACKAGE_COMPOSE_DISABLE_VERBOSE_OUTPUT: "true"
+  DOCKER_COMPOSE_VERSION: "v2.24.1"
+  DOCKER_VERSION: "26.1.0"
+  KIND_VERSION: 'v0.20.0'
+  K8S_VERSION: 'v1.29.0'
+  LINUX_AGENT_IMAGE: "golang:${GO_VERSION}"
+
+steps:
+  - input: "Input values for the variables"
+    key: "input-variables"
+    fields:
+    - select: "SERVERLESS_PROJECT"
+      key: "SERVERLESS_PROJECT"
+      options:
+        - label: "observability"
+          value: "observability"
+        - label: "security"
+          value: "security"
+      default: "observability"
+    if: "build.source == 'ui'"
+
+  - wait: ~
+    if: "build.source == 'ui'"
+    allow_dependency_failure: false
+
+  - label: ":go: Run check-static"
+    key: check-static
+    command: "make check-static"
+    agents:
+      image: "${LINUX_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
+
+  - label: ":go: :linux: Run unit tests"
+    key: unit-tests-linux
+    command: "make test-go-ci"
+    artifact_paths:
+      - "build/test-results/*.xml"
+      - "build/test-coverage/*.xml"
+    agents:
+      image: "${LINUX_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
+
+  - wait: ~
+    continue_on_failure: true
+
+  - label: ":pipeline: Test Serverless Integration tests"
+    key: test-serverless
+    command: ".buildkite/scripts/test_packages_with_serverless.sh"
+    env:
+      SERVERLESS_PROJECT: "${SERVERLESS_PROJECT:-observability}"
+      UPLOAD_SAFE_LOGS: 1
+    artifact_paths:
+      - build/test-results/*.xml
+      - build/test-coverage/coverage-*.xml
+    depends_on:
+      - step: check-static
+        allow_failure: false
+      - step: unit-tests-linux
+        allow_failure: false
+
+  - wait: ~
+    continue_on_failure: true
+
+  - label: ":junit: Junit annotate"
+    plugins:
+      - junit-annotate#v2.4.1:
+          artifacts: "build/test-results/*.xml"
+    agents:
+      provider: "gcp"  # junit plugin requires docker
--- a/.buildkite/pipeline.serverless.yml
+++ b/.buildkite/pipeline.serverless.yml
@@ -35,17 +35,6 @@ steps:
     if: "build.source == 'ui'"
     allow_dependency_failure: false
 
-  - label: ":go: Run check-static"
-    key: check-static
-    command: "make check-static"
-    agents:
-      image: "${LINUX_AGENT_IMAGE}"
-      cpu: "8"
-      memory: "4G"
-
-  - wait: ~
-    continue_on_failure: true
-
   - label: ":elastic: Serverless Integration tests"
     key: test-serverless
     command: ".buildkite/scripts/test_packages_with_serverless.sh"
@@ -55,9 +44,6 @@ steps:
     artifact_paths:
       - build/test-results/*.xml
       - build/test-coverage/coverage-*.xml
-    depends_on:
-      - step: check-static
-        allow_failure: false
 
   - wait: ~
     continue_on_failure: true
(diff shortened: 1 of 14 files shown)


scripts/test-check-packages.sh
```diff
@@ -68,13 +71,15 @@ if [ "${PACKAGE_TEST_TYPE:-other}" == "with-logstash" ]; then
   echo "stack.logstash_enabled: true" >> ~/.elastic-package/profiles/logstash/config.yml
 fi
 
-# Update the stack
-elastic-package stack update -v
+if [[ "${SERVERLESS}" != "true" ]]; then
```
mrodm, line 74: When testing with Serverless, Elastic stack is just started once at the beginning of the pipeline step. That's why here it is skipped.

catalog-info.yaml
```diff
@@ -155,3 +155,54 @@ spec:
...
+
+spec:
+  type: buildkite-pipeline
+  owner: group:ingest-fp
+  system: buildkite
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: elastic-package-test-serverless
```
mrodm, line 177: Is this pipeline name ok? do you prefer to add something else into this name ?

.buildkite/scripts/test_packages_with_serverless.sh
```diff
@@ -0,0 +1,87 @@
...
+with_docker
+echo "--- Install docker-compose"
+with_docker_compose_plugin
+
+if [[ "${BUILDKITE_PULL_REQUEST}" != "false" ]]; then
+    echo "--- install gh cli"
+    with_github_cli
+
+    add_pr_comment "${BUILDKITE_PULL_REQUEST}" "${BUILDKITE_BUILD_URL}"
+fi
```
mrodm, line 48: Add a comment in case it is triggered with a Github comment.
Example:
https://github.com/elastic/elastic-package/pull/1807#issuecomment-2089998917

.buildkite/scripts/integration_tests.sh
```diff
@@ -78,57 +73,37 @@ if [[ "${TARGET}" == "" ]]; then
     exit 1
 fi
 
-google_cloud_auth_safe_logs() {
```
mrodm, line 81: Moved these functions to `.buildkite/scripts/tooling.sh`

.buildkite/pipeline.serverless.yml
```diff
@@ -0,0 +1,82 @@
...
+
+  - label: ":go: Run check-static"
+    key: check-static
+    command: "make check-static"
+    agents:
+      image: "${LINUX_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
+
+  - label: ":go: :linux: Run unit tests"
```
mrodm, line 45: Skipped in this pipeline windows tests. They should already be run by the PR pipeline or by the push to the main branch.

.buildkite/pull-requests.json
```diff
@@ -31,6 +31,22 @@
...
+        "enabled": true,
+        "pipelineSlug": "elastic-package-test-serverless",
+        "allow_org_users": true,
+        "allowed_repo_permissions": ["admin", "write"],
+        "allowed_list": [ ],
+        "set_commit_status": false,
+        "build_on_commit": false,
+        "build_on_comment": true,
+        "trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:test)\\W+(?:serverless))$",
+        "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:test)\\W+(?:serverless))$",
```
mrodm, line 45: Allow to trigger this pipeline on demand by adding a comment in the Pull Request.

.buildkite/pipeline.serverless.yml
jsoriano: Umm, and do we need any unit tests here? :thinking: We could run only the integration tests for serverless.

.buildkite/pipeline.serverless.yml
```diff
@@ -0,0 +1,87 @@
...
+    artifact_paths:
+      - "build/test-results/*.xml"
+      - "build/test-coverage/*.xml"
+    agents:
+      image: "${LINUX_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
+
+  - wait: ~
+    continue_on_failure: true
```
jsoriano, line 58: Or if we keep the static and unit tests, maybe we can use them to gate if we continue with integration tests? So if static or unit tests fail, we fail fast and don't continue with the integration tests.
```suggestion
    continue_on_failure: false
```

.buildkite/scripts/test_packages_with_serverless.sh
```diff
@@ -0,0 +1,87 @@
...
+
+echo "--- Install elastic-package"
+# Required to start the Serverless Elastic stack
+make install
+
+prepare_serverless_stack
+
+echo "Waiting time to avoid getaddrinfo ENOTFOUND errors..."
+sleep 120
+echo "Done."
```
jsoriano, line 58: Could we do some active wait here?

.buildkite/scripts/tooling.sh
```diff
@@ -74,3 +80,112 @@ running_on_buildkite() {
...
+    echo "--- Prepare serverless stack"
+    local stack_version=${STACK_VERSION:-""}
+
+    local args="-v"
+    if [ -n "${stack_version}" ]; then
+        args="${args} --version ${stack_version}"
+    fi
+
+    # Currently, if STACK_VERSION is not defined, for serverless it will be
+    # used as Elastic stack version (for agents) the default version in elastic-package
```
jsoriano, line 100: Nit. Move this comment closer to `elastic-package stack up`?

catalog-info.yaml
jsoriano: It is fine.

scripts/test-check-packages.sh
jsoriano: Could we also add a comment here mentioning this?

jsoriano: Or maybe we can extract the common logic to check packages, and follow the same approach as in serverless in all targets? For a future refactor in any case.

jsoriano (approved):

.buildkite/pipeline.serverless.yml
mrodm: Ok, I'll remove this too :+1: 
These unit tests are going to be run in each PR , so it's not needed.

.buildkite/pipeline.serverless.yml
mrodm: It is required to set `continue_on_failure` as true, in order to process the XML files from JUnit in the last step.

.buildkite/pipeline.serverless.yml
mrodm: Do you mean to remove also the step related to `make check-static`? @jsoriano

.buildkite/scripts/test_packages_with_serverless.sh
mrodm: This is based on the script run in the integrations repository. At least a few months ago, there were some errors even if elastic-package does an active wait to report as initialized the project.

Probably, we could try here in this pipeline to remove this `sleep` and check if it works. WDYT @jsoriano ?

.buildkite/pipeline.serverless.yml
jsoriano: Would it be possible to process the junit files for unit tests without executing the step with integration tests?

.buildkite/scripts/test_packages_with_serverless.sh
jsoriano: As you prefer, we can also leave this for a future change.

.buildkite/pipeline.serverless.yml
jsoriano: Yep, we could also remove this step if we are not going to use them to block the execution of integration tests.

.buildkite/scripts/tooling.sh
mrodm: Done.

scripts/test-check-packages.sh
mrodm: Added a comment for now in this script, here and also in the `cleanup` function.

.buildkite/pipeline.serverless.yml
mrodm: I would keep `check-static` step, that will ensure that `elastic-package` can be compiled without errors.

mrodm: > Yep, we could also remove this step if we are not going to use them to block the execution of integration tests.

The script that starts the Elastic serverless project already performs a `make install` command https://github.com/elastic/elastic-package/pull/1807/files#diff-6b482aae5cb3ce434c95fe0f9238d6c8ed14a602031d6473ae03373f78ea937fR52

If there is any error there, that script should fail. So there should not be needed to keep `check-static` step.

I'll remove that step too.

.buildkite/pipeline.serverless.yml
mrodm: Not sure if there would be any issue with that.
It could be added a new step to just process the junit files from those two steps.

But we would need to keep the latest step to process the junit files from the integration steps, and that would re-process those junit files.

In any case, as discussed in the other thread, both static and unit tests/steps have been removed from the pipeline.

.buildkite/scripts/test_packages_with_serverless.sh
mrodm: Ok. so then I'll keep it for now, and we can test to remove it later.

PR #1809, opened 2024-04-30
Generalize dump stack function and fix issue when checking elasticsearch logs

jsoriano:
Merge the multiple functionality to dump logs into a single `Dump` method, and add different implementations for each stack provider.

`Dump` can be used now to dump logs to file or to variable, and to collect logs only since a given time. This replaces `GetServiceLogs`.

`Dump` implementation in the `serverless` provider returns an specific error telling that this functionality is not implemented for any service except `elastic-package`. This is used to fix a regression that was producing failures when trying to get logs from `elasticsearch`.

Errors that were previously being ignored are returned now.


jsoriano:
test integrations

mrodm (approved):
Nice refactor here!

PR #1815, opened 2024-05-03
Enable independent Elastic Agent if root privileges is set to true

### LINKED ISSUE #1586: [System tests] Run system tests for integrations with a non root Elastic Agent

mrodm:
Relates #787 
Closes #1586

This PR allows to spin up new Elastic agents independents from the Elastic Agent run in the stack (from `elastic-package stack up` command) for testing, if the packages defined in their manifest `agent.privileges.root` key as true.

That action can be overwritten if it is needed by setting the ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT environment variable. This variable has preference over the manifest file.

**NOTE**: if packages contain a folder to deploy a previous "custom agent" (`_dev/deploy/agent`), elastic-package will not create an independent Elastic Agent with User root. In that scenario, elastic-package will still create the previous custom agents.


scripts/test-check-packages.sh
```diff
@@ -121,11 +121,6 @@ for d in test/packages/${PACKAGE_TEST_TYPE:-other}/${PACKAGE_UNDER_TEST:-*}/; do
     elif [ "${PACKAGE_TEST_TYPE:-other}" == "with-logstash" ] && [ "${package_to_test}" == "system_benchmark" ]; then
         elastic-package benchmark system --benchmark logs-benchmark -v --defer-cleanup 1s
     else
-      if [[ "${ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT}" == "false" && "${package_to_test}" == "auditd_manager_independent_agent" ]]; then
-          echo "Package \"${package_to_test}\" skipped: not supported with Elastic Agent running in the stack (missing capabilities)."
-          exit # as it is run in a subshell, it cannot be used "continue"
-      fi
```
mrodm, line 127: Now elastic-package should run automatically independent Elastic Agents.

cmd/testrunner.go
```diff
@@ -189,10 +189,14 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 			return fmt.Errorf("cannot determine if package has data streams: %w", err)
 		}
 
-		runIndependentElasticAgent := false
+		// Enable independent agents if package defines that requires root privileges
```
jsoriano, line 192: Maybe we can mention here that this is temporary till independent agents are enabled by default.

test/packages/with-custom-agent/auditd_manager/manifest.yml
```diff
@@ -1,15 +1,17 @@
-format_version: 1.0.0
+format_version: 2.12.0
```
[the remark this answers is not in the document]
jsoriano: Ok, looking at this, we can probably remove the custom agent, and the `auditd_manager_independent_agent` test package.

jsoriano (commented):
Nice!

scripts/test-check-packages.sh
mrodm: It is needed to not set ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_FALSE as false, since env. vars have preference. Just added variable if it is "true", done here https://github.com/elastic/elastic-package/pull/1815/commits/2dfbedd4248275a3543c6f7a46df8228dbf18bbd

test/packages/with-custom-agent/auditd_manager/manifest.yml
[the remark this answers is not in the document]
mrodm: I was wondering if we should keep that package to test the scenario of the custom agents. Although, `oracle` package is still there in this test folder.

If we remove the custom-agent folder of `auditd_manager`, we should add the settings into the system configuration file at least to add the capabilities required for the process.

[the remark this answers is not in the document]
jsoriano: Ok, I guess we can do that. But as you prefer, having both packages also makes sense.

mrodm:
It needs to take into account that if any package contains a folder to deploy a previous "custom agent" (`_dev/deploy/agent`), elastic-package will not create an independent Elastic Agent with User root. In that scenario, elastic-package will still create the previous custom agents. cc @jsoriano 

Updated the description also with this note.

scripts/test-check-packages.sh
```diff
@@ -121,11 +119,6 @@ for d in test/packages/${PACKAGE_TEST_TYPE:-other}/${PACKAGE_UNDER_TEST:-*}/; do
     elif [ "${PACKAGE_TEST_TYPE:-other}" == "with-logstash" ] && [ "${package_to_test}" == "system_benchmark" ]; then
         elastic-package benchmark system --benchmark logs-benchmark -v --defer-cleanup 1s
     else
-      if [[ "${ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT}" == "false" && "${package_to_test}" == "auditd_manager_independent_agent" ]]; then
```
jsoriano, line 124: Umm, this `auditd_manager_independent_agent` package could be moved out of the `with-custom-agent` directory, right? And then maybe renamed back to `auditd_manager`.

jsoriano (approved):
:tada:

jsoriano:
> It needs to take into account that if any package contains a folder to deploy a previous "custom agent" (`_dev/deploy/agent`), elastic-package will not create an independent Elastic Agent with User root. In that scenario, elastic-package will still create the previous custom agents. cc @jsoriano

The only custom agent that requires root is [the one for `auditd_manager`](https://github.com/elastic/integrations/blob/main/packages/auditd_manager/data_stream/auditd/_dev/deploy/agent/custom-agent.yml), and it correctly defines the user and capabilities it requires, this will continue to be honored, right?

scripts/test-check-packages.sh
mrodm: Ok, moving this package out of this folder :+1:

mrodm:
> The only custom agent that requires root is [the one for `auditd_manager`](https://github.com/elastic/integrations/blob/main/packages/auditd_manager/data_stream/auditd/_dev/deploy/agent/custom-agent.yml), and it correctly defines the user and capabilities it requires, this will continue to be honored, right?

Yes, that will be honored. `elastic-package` will create a custom agent (same as before) for that package with the capabilities and user defined in that YAML file.

mrodm:
test integrations

jsoriano (approved):

PR #1816, opened 2024-05-03
Update system flags test with independent Elastic Agents

mrodm:
Part of #787 

Update tests related to system flags to take into account if independent Elastic Agents are created.


scripts/test-system-test-flags.sh
```diff
@@ -271,13 +285,20 @@ elastic-package stack status
 
 FOLDER_PATH="${HOME}/.elastic-package/profiles/default/stack/state"
 
+# Check also if independent Elastic Agents are running too
+# depending on the environment variable
+service_deployer_type="docker"
+if [[ "${ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT:-"false"}" == "true" ]]; then
```
jsoriano, line 291: Out of curiosity, why setting a default here?
```suggestion
if [[ "${ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT}" == "true" ]]; then
```

jsoriano (approved):

scripts/test-system-test-flags.sh
mrodm: If the variable is not defined, as it happens in the CI to test non-independent Elastic Agents (Elastic Agent from stack), this could cause an error of `unbound variable` due to the flags set for this script `set -euo pipefail`.

As the environment variable has preference over `agent.privileges.root`, it was required to remove that environment variable in case of "false" in order to be considered the value of that `agent.privileges.root` key in our test packages (the last test package added `test/packages/parallel/auditd_manager`). That package is always tested with independent Elastic Agent since it requires root privileges.

That means that currently it is not tested the scenario with these conditions in the CI:
- Setting `ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT` as false
- Defining `agent.privileges.root` as true
This scenario should use the Elastic Agent from the stack. It was tested locally.

PR #1820, opened 2024-05-07
Improve healthcheck for Fleet Server

jsoriano:
Fleet Server restarts once after being healthy. Agents enrolling during this restart fail and need to be restarted. Improve the healthcheck so it expects to have three healthy healthchecks before considering Fleet Server ready to enroll. This should reduce the number of restarts required by Elastic Agent on first boot.

Reduce also the time to wait when the agent is restarted in case it fails on a first boot.

And rename some files to avoid legacy names.


internal/stack/resources.go
```diff
@@ -24,8 +24,8 @@ import (
 var static embed.FS
 
 const (
-	// SnapshotFile is the docker-compose snapshot.yml file name.
-	SnapshotFile = "snapshot.yml"
+	// ComposeFile is the docker compose file.
+	ComposeFile = "docker-compose.yml"
```
mrodm, line 28: There is an issue here but not sure if it would be too complicated to add support for this scenario(s).

In one of my tests I faced one error:
1. Elastic stack running with a previous elastic-package version (using snapshot.yaml). 
2. Install elastic-package with the contents of this branch.
3. Try to run `elastic-package stack down -v` command failed:
    ```bash
     $ elastic-package stack down -v 
    2024/05/07 15:40:54 DEBUG Enable verbose logging
    2024/05/07 15:40:54 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
    Take down the Elastic stack
    Error: tearing down the stack failed: stopping docker containers failed: could not create docker compose project: could not find Docker Compose configuration file: /home/mariorodriguez/.elastic-package/profiles/default/stack/docker-compose.yml: stat /home/mariorodriguez/.elastic-package/profiles/default/stack/docker-compose.yml: no such file or directory
    ``` 


Same error happens with these sequence of commands:
1. Tear down Elastic stack with previous version `elastic-package stack down`
2. Install elastic-package with the contents of this branch.
3. Run again `elastic-package stack down -v`
    - `elastic-package` does not find the new `docker-compose.yml` because `elastic-package stack up` has not run yet.

mrodm: Just checked that`snapshot.yml` file is kept in `~/.elastic-package/profiles/default/stack/snapshot.yml` even if the new file is created and used by elastic-package.

I guess there should be any issue, depending on the elastic-package version one or other file would be used.

internal/stack/boot.go
```diff
@@ -94,10 +94,14 @@ func BootUp(ctx context.Context, options Options) error {
...
+			sleepTime := 2 * time.Second
 			fmt.Printf("Elastic Agent failed to start, trying again in %s.\n", sleepTime)
-			time.Sleep(sleepTime)
-			err = dockerComposeUp(ctx, options)
+			select {
+			case <-time.After(sleepTime):
+				err = dockerComposeUp(ctx, options)
+			case <-ctx.Done():
+				err = ctx.Err()
+			}
```
mrodm, line 104: :+1:

internal/stack/logs.go
```diff
@@ -22,7 +22,7 @@ func dockerComposeLogsSince(ctx context.Context, serviceName string, profile *pr
 		return nil, fmt.Errorf("can't read application configuration: %w", err)
 	}
 
-	snapshotFile := profile.Path(ProfileStackPath, SnapshotFile)
+	snapshotFile := profile.Path(ProfileStackPath, ComposeFile)
```
mrodm, line 25: Nit. I'd rename this variable too, as `composeFile` for instance:

```suggestion
	composeFile := profile.Path(ProfileStackPath, ComposeFile)
```

internal/stack/_static/fleet-server-healthcheck.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
```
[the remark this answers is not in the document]
jsoriano: Yeah, from my tests it is working as expected, but I can add this option just in case.

internal/stack/resources.go
jsoriano: Solved by adding "migrations" in https://github.com/elastic/elastic-package/pull/1820/commits/5bfb5dfe032dfc6c15cb5bde89cad78cfcfe14a1.
I will split the PR in two because it is starting to have too many things.

jsoriano: Refactors reverted in new PR https://github.com/elastic/elastic-package/pull/1824.

jsoriano:
File renames moved to https://github.com/elastic/elastic-package/pull/1835, closing this.

PR #1822, opened 2024-05-07
Allow to define custom scripts in independent Elastic Agents

mrodm:
Part of  #787 

This PR allows to define custom scripts for independent Elastic Agents. This custom scripts can be used to install required software to run the required tests successfully. For instance some libraries as it happens with `oracle` package.

Example of the new keys added into the configuration file:

```yaml
agent:
  custom_script: |
    mkdir -p /my/path
    wget ...
  pre_start_script: |
    export PATH=${PATH}:/my/path
```

Now the files required to spin a new independent Elastic Agent using containers has been migrated to `go-resources`. Files generated can be checked within the profile folder:
```shell
 $ ls -l ~/.elastic-package/profiles/default/deployer/docker-agent-oracle-memory-61511/
total 12
-rw-r--r-- 1 user user 1046 may  8 16:15 docker-agent-base.yml
-rw-r--r-- 1 user user  203 may  8 16:15 Dockerfile
-rwxr-xr-x 1 user user  764 may  8 16:15 script.sh
```

If custom scripts are required, `elastic-package` builds a new Docker image adding the suffix `-custom` to the tag. Example:
```shell
 $ docker images |grep custom
docker.elastic.co/elastic-agent/elastic-agent-complete   8.13.2-custom     ee55b4ece29b   22 hours ago    3.95GB
```

## Testing

Added a new test package `test/packages/parallel/oracle` based on the one from `test/packages/with-custom-agent/oracle` but with some changes:
- Removed custom "agent" deploy (deleted folder).
- Added just the `oracle` service in `_dev/deploy/docker`.
- Added the required agent settings in the configuration file for system tests

This package does not work when tested with the Elastic stack running from the stack since it is missing some required libraries/dependencies the container.

- Process to test this locally:
```shell
elastic-package stack up -v -d
export ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT=true

# test new package with independent Elastic Agents
pushd test/packages/parallel/oracle
elastic-package test system -v
popd

# test old package - it should keep using previous Custom Agents with servicedeployer
pushd test/packages/with-custom-agent/oracle
elastic-package test system -v
popd

elastic-package stack down -v
```

PR #1828, opened 2024-05-10
Add support for system tests using independent Elastic agents in Elastic Serverless

mrodm:
Part of #787 

This PR adds support to run tests with Elastic Serverless.

## Author's Checklist
- [x] Update branch after #1822 is merged

## How to test

```shell
export EC_API_KEY="...."
export EC_HOST="..."

elastic-package profiles create serverless
elastic-package profiles use serverless

elastic-package stack up -v -d --provider serverless
cd test/packages/parallel/custom_entrypoint

ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT=true elastic-package test system -v

elastic-package stack down -v

elastic-package profiles use default
```


jlind23 (approved):

internal/agentdeployer/agent.go
```diff
@@ -261,18 +261,39 @@ func (d *DockerComposeAgentDeployer) installDockerCompose(agentInfo AgentInfo) (
...
+	}
+
+	fleetURL := "https://fleet-server:8220"
+	kibanaHost := "https://kibana:5601"
+	stackVersion := d.stackVersion
+	if config.Provider == stack.ProviderServerless {
+		fleetURL = config.Parameters[stack.ParamServerlessFleetURL]
+		kibanaHost = config.KibanaHost
+		stackVersion = config.Parameters[stack.ParamServerlessLocalStackVersion]
+	}
```
jsoriano, line 276: Nit. Couldn't we get this info from the config in all cases?
```suggestion
	fleetURL = config.Parameters[stack.ParamServerlessFleetURL]
	kibanaHost = config.KibanaHost
	stackVersion = config.Parameters[stack.ParamServerlessLocalStackVersion]

```

jsoriano (approved):
:+1:

internal/agentdeployer/agent.go
mrodm: Currently, it cannot be done. When compose provider is used, kibanaHost is `https://127.0.0.1:5601`, so it cannot be used to set the environment variable for the Elastic Agent. It requires the container name `kibana`.

In case of fleet-server, that value is not available as a configuration parameter now in the Compose provider. Probably, it could be added to the compose provider, but it we do so, it should be renamed (`stack_fleet_url`?) or create a new parameter for the compose provider, eg. `compose_fleet_url`.

Should I try to use just one parameter for both providers? @jsoriano 

This is the current definition of this parameter for serverless: https://github.com/elastic/elastic-package/blob/c2a0f6df3eeffee964c8c2b1ee4c5e96d2df464c/internal/stack/serverless.go#L27

jsoriano: Well, lets leave it like this by now, maybe we can implement this logic in the provider later.

PR #1834, opened 2024-05-15
Remove unused extra docker-compose files in default agent deployer

mrodm:
Part of #787 

This PR removes the option to add extra docker-compose files to the default agent deployer.

This was added in case it would be needed to add a new docker-compse file to override the settings of the Elastic Agent. However, this has been changed in favor of a more generic approach of using provisioning (or pre-start) scripts. This approach could be used also in different environments too.

Related PRs:
- https://github.com/elastic/elastic-package/pull/1822


jsoriano (approved):
:+1: :+1:

PR #1835, opened 2024-05-15
Introduce profile versioning

jsoriano:
Introduce profiles versioning, so we can make controlled changes in profile files without having to rely on specific versions of the elastic-package binary, that are sometimes unknown or unpredictable.

Leverage this to migrate some stack files for consistency:
* `docker-compose.yml` is used instead of the legacy `snapshot.yml`.
* All script names are separated with `-` instead of a mix of `-` and `_`.

Also rename some internal variables to avoid confusion between data home path and profiles path.


internal/profile/profile_json.go
```diff
@@ -10,46 +10,39 @@ import (
...
-	"github.com/elastic/elastic-package/internal/version"
 )
 
 // Metadata stores the data associated with a given profile
 type Metadata struct {
 	Name        string    `json:"name"`
 	DateCreated time.Time `json:"date_created"`
-	User        string    `json:"user"`
 	Version     string    `json:"version"`
-	Path        string    `json:"path"`
```
jsoriano, line 27: User and Path removed because they are not used, and the home directory could be moved and the data here would be incoherent.

I wonder if we should do the same with the profile name and the creation date, not sure if we need them.
The creation date can be useful though to give an idea of when a profile was created when listing profiles.
The name is also useful, but could be problematic to have it in the metadata in case it is moved. Though we don't support manually handling files inside the data dir.

mrodm: As you mentioned, I would also remove User and Path.

Creation date it could be interesting to be kept to show in the profiles list view for the users in case they need compare which profile should be using (or deleting).

> The name is also useful, but could be problematic to have it in the metadata in case it is moved.

Currently, we don't support moving. Just copying a profile based on another profile. So it should not be a problem now that, based on users are not updating those files manually :sweat_smile:

Maybe raise an error if there are two profiles with the same name ? But that can be done just when running `elastic-package profiles list` , no?

mrodm (approved):
Nice!

PR #1838, opened 2024-05-16
Update system testrunner to allow parallel executions

mrodm:
Relates https://github.com/elastic/elastic-package/issues/787

This PR prepares the environment in such a way that two executions of `elastic-package test system -v` could be run in parallel.

For that it is required:
- [x] Update compose project name used in service deployer adding a random string
- [x] Update compose project name used in agent deployer adding a random string
- [x] Update integration policy name adding a random string
- [x] ~Do not raise error if uninstall package fails in the tear down~
      ```
      2024/05/17 16:16:49 ERROR Failed to uninstall package is there any other execution in parallel?: there was an apply error: can't remove the package: could not remove package; API status code = 400; response body = {"statusCode":400,"error":"Bad Request","message":"Unable to remove package with existing package policy(s) in use by agent(s)"}
      ```
    - Reverted this change in favor of installing/uninstalling package as a first/last step of all the test executions.
- [x] Update service hostnames used in test config, some options:
    - ~Allow to inject variables in system test configuration files, at least service hostname~.
    - [x] Modify compose scenarios
    - [x] Use placeholders for hostname in test package oracle `{{Hostname}}`
        - Require to update test packages (oracle)
- [x] Fix tests system flags

For a follow up PRs:
- [ ] Build directory used for all routines needs to be revisited
```
2024/05/17 11:58:05 ERROR error running package system tests: could not complete test run: can't install the package: there was an apply error: failed to build package: clearing package contents failed: removing directory failed (path: /home/user/Coding/work/elastic-package/build/packages/nginx/999.999.999): unlinkat /home/user/Coding/work/elastic-package/build/packages/nginx/999.999.999: directory not empty
```
- [ ] Show in each log as a context at least the data stream tested

These steps would allow also to be able to trigger parallel tests directly from `elastic-package`.


mrodm:
test integrations

internal/servicedeployer/compose.go
```diff
@@ -75,7 +75,7 @@ func (d *DockerComposeServiceDeployer) SetUp(ctx context.Context, svcInfo Servic
 	logger.Debug("setting up service using Docker Compose service deployer")
 	service := dockerComposeDeployedService{
 		ymlPaths: d.ymlPaths,
-		project:  "elastic-package-service",
+		project:  fmt.Sprintf("elastic-package-service-%s", svcInfo.Test.RunID),
```
mrodm, line 78: This change changes also the compose project name used without independent Elastic Agents.
This compose project name should not be taken into account in any package.

Should we keep the same compose project name in case independent Elastic Agents are not used (`d.deployIndependentAgent`)?

internal/testrunner/runners/system/runner.go
```diff
@@ -1530,10 +1531,11 @@ func createPackageDatastream(
 	ds packages.DataStreamManifest,
 	config testConfig,
 ) kibana.PackageDataStream {
+	suffix := createTestRunID()
```
mrodm, line 1534: Would it be better to just create this suffix (runID) in each create (input or integration) datastream function? Currently, there are no tests of these functions, but having this as a parameter, I think it will be easier to add tests eventually.

internal/servicedeployer/compose.go
jsoriano: It looks good like this.

internal/testrunner/runners/system/runner.go
jsoriano: Or maybe we can pass it from the runner, in case we need to use the runID in more places.

jsoriano (approved):

internal/testrunner/runners/system/runner.go
```diff
@@ -452,7 +452,8 @@ func (r *runner) tearDownTest(ctx context.Context) error {
 	}
 	_, err := r.resourcesManager.ApplyCtx(cleanupCtx, r.resources(resourcesOptions))
 	if err != nil {
-		return err
+		logger.Errorf("Failed to uninstall package is there any other execution in parallel?: %s", err)
+		// return err
```
mrodm, line 456: Should it be reported as a Warn level ?

internal/testrunner/runners/system/runner.go
mrodm: :+1: 
In runner it can be set as this suffix `svcInfo.Test.RunID`. That struct is always created.

internal/testrunner/runners/system/runner.go
mrodm: > It would be interesting to install package just once before running all tests. And once all tests have finished, uninstall that package.

Related to this, a new PR has been created: https://github.com/elastic/elastic-package/pull/1845

PR #1840, opened 2024-05-16
Enable independent agents if root privileges are required at both package or data stream level

mrodm:
Follows #1724 
Relates #1586

This PR also enables independent Elastic Agents in case root privileges are defined in the Data Stream manifest.

If any of the two files (package manifest or Data Stream manifest) defines that root privileges are required, it is enabled independent Elastic Agents.

As previously, if the environment variable `ELASTIC_PACKAGE_TEST_ENABLE_INDEPENENT_AGENT` is set, it has preference over the values defined in those manifests.

### DIFF
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -190,9 +190,6 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 		}
 
 		runIndependentElasticAgent := false
-
-		// If the environment variable is present, it always has preference over the root
-		// privileges value (if any) defined in the manifest file
 		v, ok := os.LookupEnv(enableIndependentAgents)
 		if ok {
 			runIndependentElasticAgent = strings.ToLower(v) == "true"
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -189,9 +189,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 			return fmt.Errorf("cannot determine if package has data streams: %w", err)
 		}
 
-		// Temporarily until independent Elastic Agents are enabled by default,
-		// enable independent Elastic Agents if package defines that requires root privileges
-		runIndependentElasticAgent := manifest.Agent.Privileges.Root
+		runIndependentElasticAgent := false
 
 		// If the environment variable is present, it always has preference over the root
 		// privileges value (if any) defined in the manifest file
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -18,7 +18,6 @@ import (
 	"github.com/elastic/elastic-package/internal/cobraext"
 	"github.com/elastic/elastic-package/internal/common"
 	"github.com/elastic/elastic-package/internal/elasticsearch"
-	"github.com/elastic/elastic-package/internal/environment"
 	"github.com/elastic/elastic-package/internal/install"
 	"github.com/elastic/elastic-package/internal/kibana"
 	"github.com/elastic/elastic-package/internal/logger"
@@ -52,8 +51,6 @@ For details on how to run static tests for a package, see the [HOWTO guide](http
 These tests allow you to test a package's ability to ingest data end-to-end.
 
 For details on how to configure amd run system tests, review the [HOWTO guide](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md).`
-
-var enableIndependentAgents = environment.WithElasticPackagePrefix("TEST_ENABLE_INDEPENDENT_AGENT")
 
 func setupTestCommand() *cobraext.Command {
 	var testTypeCmdActions []cobraext.CommandAction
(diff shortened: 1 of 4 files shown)


cmd/testrunner.go
```diff
@@ -189,9 +189,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 			return fmt.Errorf("cannot determine if package has data streams: %w", err)
 		}
 
-		// Temporarily until independent Elastic Agents are enabled by default,
-		// enable independent Elastic Agents if package defines that requires root privileges
-		runIndependentElasticAgent := manifest.Agent.Privileges.Root
+		runIndependentElasticAgent := false
```
mrodm, line 192: Here, in `cmd` just check about the environment variable.
Checks about `agent.privileges.root` in package or data-stream manifest have been moved to `prepareScenario` method in system test runner.

internal/testrunner/runners/system/runner.go
```diff
@@ -847,6 +851,19 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
...
+	if scenario.pkgManifest.Agent.Privileges.Root || scenario.dataStreamManifest.Agent.Privileges.Root {
+		r.options.RunIndependentElasticAgent = true
+	}
+
+	// If the environment variable is present, it always has preference over the root
+	// privileges value (if any) defined in the manifest file
+	v, ok := os.LookupEnv(enableIndependentAgents)
+	if ok {
+		r.options.RunIndependentElasticAgent = strings.ToLower(v) == "true"
+	}
```
mrodm, line 865: It needs to be checked again the environment variable, to ensure it has preference over the value in the manifest files.

internal/testrunner/runners/system/runner.go
```diff
@@ -847,6 +851,19 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
 		return nil, fmt.Errorf("reading data stream manifest failed: %w", err)
 	}
 
+	// Temporarily until independent Elastic Agents are enabled by default,
+	// enable independent Elastic Agents if package defines that requires root privileges
+	if scenario.pkgManifest.Agent.Privileges.Root || scenario.dataStreamManifest.Agent.Privileges.Root {
```
jsoriano, line 856: Maybe we need to check if the data stream manifest is nil, for input packages.
```suggestion
	if pkg, ds := scenario.pkgManifest, scenario.dataStreamManifest; pkg.Agent.Privileges.Root || (ds != nil && ds.Agent.Privileges.Root) {
```

cmd/testrunner.go
```diff
@@ -189,12 +189,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
...
 		}
 
-		// Temporarily until independent Elastic Agents are enabled by default,
-		// enable independent Elastic Agents if package defines that requires root privileges
-		runIndependentElasticAgent := manifest.Agent.Privileges.Root
-
-		// If the environment variable is present, it always has preference over the root
-		// privileges value (if any) defined in the manifest file
+		runIndependentElasticAgent := false
 		v, ok := os.LookupEnv(enableIndependentAgents)
```
jsoriano, line 193: Do we need to check the environment variable here? We are also checking it in the runner now.

mrodm: Yes, it could be removed.

internal/testrunner/runners/system/runner.go
mrodm: Sure, better to be sure that it will not raise any error/panic :+1:

cmd/testrunner.go
```diff
@@ -368,7 +354,7 @@ func testTypeCommandActionFactory(runner testrunner.TestRunner) cobraext.Command
 				RunSetup:                   runSetup,
 				RunTearDown:                runTearDown,
 				RunTestsOnly:               runTestsOnly,
-				RunIndependentElasticAgent: runIndependentElasticAgent,
+				RunIndependentElasticAgent: false,
```
mrodm, line 357: Leave it as default value.

internal/testrunner/runners/system/runner.go
```diff
@@ -257,7 +260,7 @@ func (r *runner) Run(ctx context.Context, options testrunner.TestOptions) ([]tes
 	}
 	result = r.newResult(fmt.Sprintf("%s - %s", resultName, testConfig.Name()))
 
-	scenario, err := r.prepareScenario(ctx, testConfig, svcInfo, serviceOptions)
+	scenario, err := r.prepareScenario(ctx, testConfig, svcInfo)
```
mrodm, line 263: Take the change to move `createServiceOptions` within `prepareScenario` method.

From this object, it was just used `DataStreamRootPath` but this value is also available from the `runner` struct once `r.initRun()` has been called.

internal/testrunner/runners/system/runner.go
```diff
@@ -275,7 +278,7 @@ func (r *runner) Run(ctx context.Context, options testrunner.TestOptions) ([]tes
 		if err != nil {
 			return result.WithError(fmt.Errorf("failed to prepare scenario: %w", err))
 		}
-		return r.validateTestScenario(ctx, result, scenario, testConfig, serviceOptions)
+		return r.validateTestScenario(ctx, result, scenario, testConfig)
```
mrodm, line 281: This parameter can be removed after some refactors in two steps.
These are the commits in two steps:
- https://github.com/elastic/elastic-package/pull/1840/commits/d20c1827635e0619f5431ff3664861bd214535b5
- https://github.com/elastic/elastic-package/pull/1840/commits/bcdaf8a401b2b4c8348bd67771399dbd2f1b2494

mrodm:
test integrations

jsoriano (approved):

mrodm:
Failures are unrelated, merging.

PR #1843, opened 2024-05-17
PoC: Run system tests in parallel (per data stream)

mrodm:
Follows #1838 

- [x] Add mutex to install package process (all 3 packages use the same directory)
- [x] Check terraform outputs folder
    - Looks good, it adds the runID per each output dir folder https://github.com/elastic/elastic-package/blob/49d021e7437b83fab0356aab7eacec9dc71ba3ae/internal/servicedeployer/terraform.go#L220 
- [x] Check terraform deployer dir folder
    - Requires to be changed per test https://github.com/elastic/elastic-package/blob/49d021e7437b83fab0356aab7eacec9dc71ba3ae/internal/servicedeployer/terraform.go#L180C56-L180C76
- [ ] Check kubernetes service deployer
    - What happens if two different data streams deploy the same resources for testing ?
- [ ] Check how it should be executed these parallel tests using system flags.
- [ ] Add some flag or configuration to disable parallelization


mrodm:
Closing in favor of #1909

PR #1845, opened 2024-05-20
Update installation/uninstallation package process in system tests

mrodm:
Relates #787 

Package installation process is repeated for every data stream and configuration file defined in every package.

This process is just needed once:
- No need to install multiple times the same package.
- Every data stream and configuration file would be defining the required Package Policy.

This PR updates the test runner to include two new methods in the interface (`SetupRunner` and `TearDownRunner`). These two methods are in charge of creating (or deleting) the global resources required to run all the tests for every data streams and configuration file defined in the package:
- For system tests, installation (and uninstallation) of the package would be performed in these two new methods.

Depends on #1866


mrodm:
test integrations

mrodm:
Closing PR in favor of #1892

PR #1846, opened 2024-05-20
Add suffix to compose project names

mrodm:
Follows #1838

Add suffix (run ID string) to compose project created by terraform and custom agent service deployers.

### DIFF
--- a/internal/servicedeployer/custom_agent.go
+++ b/internal/servicedeployer/custom_agent.go
@@ -106,7 +106,7 @@ func (d *CustomAgentDeployer) SetUp(ctx context.Context, svcInfo ServiceInfo) (D
 
 	service := dockerComposeDeployedService{
 		ymlPaths: ymlPaths,
-		project:  "elastic-package-service",
+		project:  fmt.Sprintf("elastic-package-service-%s", svcInfo.Test.RunID),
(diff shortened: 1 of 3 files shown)


internal/servicedeployer/custom_agent.go
```diff
@@ -106,7 +106,7 @@ func (d *CustomAgentDeployer) SetUp(ctx context.Context, svcInfo ServiceInfo) (D
 
 	service := dockerComposeDeployedService{
 		ymlPaths: ymlPaths,
-		project:  "elastic-package-service",
+		project:  fmt.Sprintf("elastic-package-service-%s", svcInfo.Test.RunID),
```
mrodm, line 109: I've updated this service deployer to keep consistency.
But it would not be needed, since this deployer is not used with independent Elastic Agents. 

Would it be better to revert the change in this file?

This would make mandatory to change test configuration files for `oracle` package using the placeholders `{{Hostname}}` and `{{Port}}`.

jsoriano: Well, it doesn't do any harm, right? Keep the change in case this continues to be used.

jsoriano (approved):

PR #1848, opened 2024-05-21
[Buildkite] Fix cloud cleanup pipeline

mrodm:
Ensure that the Buildkite pipeline in charge of checking for stale cloud resources fails if there is any resource to delete, so the team is notified.

Example of build with the expected failure:
https://buildkite.com/elastic/elastic-package-cloud-cleanup/builds/84

### DIFF
--- a/.buildkite/scripts/cloud-cleanup.sh
+++ b/.buildkite/scripts/cloud-cleanup.sh
@@ -101,6 +101,7 @@ if any_resources_to_delete "${AWS_RESOURCES_FILE}" ; then
     resources_to_delete=1
 fi
 
+echo "Resources to delete: ${resources_to_delete}"
 if [ "${resources_to_delete}" -eq 1 ]; then
     message="There are resources to be deleted"
     echo "${message}"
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/cloud-cleanup.sh
```diff
@@ -168,8 +169,9 @@ jq -c '.Clusters[]' redshift_clusters.json | while read i ; do
         #   --skip-final-cluster-snapshot
         echo "Done."
     fi
-done
+done <<< "$(jq -c '.Clusters[]' redshift_clusters.json)"
```
mrodm, line 172: This change is to ensure that the while command is not run in a subshell as it happens running after a pipe.

If `while` runs in a subshell, none of the variables modified in those iterations would be remain their values after finishing the loop.

jsoriano: Would it work with a for loop like `for i in $(jq -c '.Clusters[]' redshift_clusters.json)`?

jsoriano (approved):

.buildkite/scripts/cloud-cleanup.sh
mrodm: Tested here: https://github.com/elastic/elastic-package/pull/1848/commits/0e34e0bc4b7d56cb3e0d963d04253f9b50492aaf
but it fails: https://buildkite.com/elastic/elastic-package-cloud-cleanup/builds/85#018f9aad-05c8-4703-b1d2-9c3c777cccf0/112-114

I'll keep the while loop

mrodm:
Tested pipeline triggering it from the UI:
https://buildkite.com/elastic/elastic-package-cloud-cleanup/builds/86

PR #1849, opened 2024-05-21
[Buildkite] Enable delete redshift clusters with dry_run false

mrodm:
Enable in cloud cleanup pipeline the action to delete Redshift Clusters when DRY_RUN is set to false.

Currently, keep DRY_RUN=true by default when it is triggered the pipeline automatically, since it has not been tested for the other resources.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -14,7 +14,7 @@ steps:
     trigger: elastic-package-cloud-cleanup
     build:
       env:
-        DRY_RUN: false 
+        DRY_RUN: false
         BUILDKITE_REFSPEC: "refs/pull/${BUILDKITE_PULL_REQUEST}/merge"
         BUILDKITE_COMMIT: "${BUILDKITE_COMMIT}"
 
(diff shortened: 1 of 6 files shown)


.buildkite/scripts/cloud-cleanup.sh
```diff
@@ -161,18 +180,37 @@ while read -r i ; do
...
+        redshift_clusters_to_delete=1
+        continue
+    fi
+
+    echo "Deleting: $identifier. It was created > ${RESOURCE_RETENTION_PERIOD} ago"
+    if ! aws redshift delete-cluster \
+      --cluster-identifier "${identifier}" \
+      --skip-final-cluster-snapshot \
+      --output json \
+      --query "Clusters[*].{ClusterStatus:ClusterStatus,ClusterIdentifier:ClusterIdentifier}" ; then
```
mrodm, line 193: Just show those values from the response.

.buildkite/pipeline.yml
```diff
@@ -10,76 +10,84 @@ env:
   IMAGE_UBUNTU_X86_64: "family/core-ubuntu-2204"
 
 steps:
-  - label: ":go: Run check-static"
```
mrodm, line 13: I would restore this file before merging, in case it is needed to do more tests.

PR #1853, opened 2024-05-23
Add github action to bump stack versions

jsoriano:
Add github action to automate bump of stack versions.

It bumps:
- Latest snapshot build for 8.x.
- Latest snapshot build for 7.17.x.
- Stack default version.
- Package Registry version.

### DIFF
--- a/.github/workflows/bump-elastic-stack-version.yml
+++ b/.github/workflows/bump-elastic-stack-version.yml
@@ -15,7 +15,6 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
-    needs: [filter]
     strategy:
       fail-fast: false
     steps:
--- a/.github/workflows/bump-elastic-stack-version.yml
+++ b/.github/workflows/bump-elastic-stack-version.yml
@@ -1,0 +1,30 @@
+---
+name: bump-elastic-stack-version
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 15 * * 1-5'
+
+permissions:
+  contents: read
+
+jobs:
+  bump-elastic-stack:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    needs: [filter]
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Updatecli in the runner
+        uses: updatecli/updatecli-action@9a37c7e35598d7b37d8e7568b40ed9538112be01 # v0.76.1
+
+      - name: Run Updatecli in Apply mode
+        run: updatecli --experimental apply --config .github/workflows/updatecli.d/bump-elastic-stack-version.yml --values .github/workflows/updatecli.d/scm.yml
+        env:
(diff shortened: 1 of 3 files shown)


jsoriano:
@mrodm do you know if there is any way to test these actions before merging?

mrodm:
> @mrodm do you know if there is any way to test these actions before merging?

I think of two options
- Create first a PR with a minimal Github workflow, and then triggering from UI the Github Workflow from your branch
- Try to use `gh`. Not tested, but it would be worth to try it: https://cli.github.com/manual/gh_workflow_run

  Within the elastic-pacakge repository:
  ```shell
  gh workflow run bump-elastic-stack-version -r `your branch`
  ```

Not sure if it requires to create this PR in the upstream repository because of this:
https://github.com/elastic/elastic-package/pull/1853/files#diff-cf638fce84d5cc76017b884e67009ea1e65b57e9dd6f2ec2f3de859c1f0bcbdaR30

.github/workflows/bump-elastic-stack-version.yml
```diff
@@ -0,0 +1,30 @@
...
+permissions:
+  contents: read
+
+jobs:
+  bump-elastic-stack:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    needs: [filter]
```
mrodm, line 18: There is no jobs `filter` defined. Probably, this job fails or could not start. Is it required another job here ? 

I guess it is probably based on this workflow: https://github.com/elastic/beats/blob/d23b4d33622137f7245b628d77d15f8ecfef3a0d/.github/workflows/bump-elastic-stack-snapshot.yml#L16

Currently, the `filter` job looks like would not be needed. If so, this could be removed.

jsoriano: Yep, this is a leftover from copying from beats.

mrodm:
> This doesn't seem to work, I have tried different options but it doesn't seem to find the workflow.

So at the end, it looks like it should be defined in main... I cannot think of any other ways to test this without merging.

> I have modified the workflow in this PR to show only the diff and don't apply the changes, let me know if this is ok for you. We could merge this and change to apply after testing in a different PR.

Perfect, let's do it in that way :+1:

.github/workflows/updatecli.d/bump-elastic-stack-version.yml
```diff
@@ -0,0 +1,91 @@
...
+      replacepattern: '$1 "{{ source "latest8xVersion" }}"'
+  update-package-registry-base-image:
+    name: "Update Package Registry base image"
+    kind: file
+    sourceid: latestRegistryVersion
+    scmid: default
+    spec:
+      file: internal/stack/resources.go
+      matchpattern: '"(docker.elastic.co/package-registry/package-registry):v[0-9\.]+"'
+      replacepattern: '"$1:{{ source "latestRegistryVersion" }}"'
```
mrodm, line 91: If this works, we could remove the step to create this PR from other Buildkite job.

mrodm (approved):
These workflows looks good, let's merge it and try it :+1:

jsoriano:
I can test changes in the workflow now that it is in main :+1:

PR #1866, opened 2024-05-24
Create Agent policies per each test execution

mrodm:
Relates #787

This PR adds a new Agent Policy that will be created and configured per each system test execution.

This new Agent Policy is going to help us:
- Each test will ingest documents in its own Data Stream.
- It should not be needed to manage deletion of docs, since all executions will create a new Data Stream for it
    - Previously, using the same Agent Policy, the data streams could contain documents form previous runs and sometimes it is not possible to check if all documents have been removed or not, as it happens in this PR when running per stages system tests:
        - https://github.com/elastic/elastic-package/pull/1845
        - https://buildkite.com/elastic/elastic-package/builds/3162

After each test run defined that new Agent Policy is deleted as well as the Data Stream created for it.


internal/testrunner/runners/system/runner.go
```diff
@@ -821,6 +842,36 @@ type scenarioTest struct {
...
 
+func (r *runner) shouldCreateNewAgentPolicyForTest() bool {
+	if r.options.RunTestsOnly {
+		// always that --no-provision is set, it should create new Agent Policies.
+		return true
+	}
+	if !r.options.RunIndependentElasticAgent {
+		// keep same behaviour as previously when Elastic Agent of the stack is used.
+		return false
+	}
```
mrodm, line 853: I thought to keep the same behaviour when using Elastic Agent of the stack. WDYT ?

internal/testrunner/runners/system/runner.go
```diff
@@ -1378,14 +1462,6 @@ func (r *runner) deleteOldDocumentsDataStreamAndWait(ctx context.Context, dataSt
...
 		return fmt.Errorf("error deleting old data in data stream: %s: %w", dataStream, err)
 	}
-	startHits, err := r.getDocs(ctx, dataStream)
-	if err != nil {
-		return err
-	}
-	// First call already reports zero documents
-	if startHits.size() == 0 {
-		return nil
-	}
```
mrodm, line 1388: No need to keep this logic (the same as line 1398), as every test will use a fresh data stream without any documents written on it.

internal/testrunner/runners/system/runner.go
```diff
@@ -278,7 +279,13 @@ func (r *runner) Run(ctx context.Context, options testrunner.TestOptions) ([]tes
 		if err != nil {
 			return result.WithError(fmt.Errorf("failed to prepare scenario: %w", err))
 		}
-		return r.validateTestScenario(ctx, result, scenario, testConfig)
+		results, err := r.validateTestScenario(ctx, result, scenario, testConfig)
+		tdErr := r.tearDownTest(ctx)
+		if tdErr != nil {
+			logger.Errorf("failed to tear down runner: %s", tdErr.Error())
+		}
```
mrodm, line 286: Required to delete the policy created just for testing and also the related Data Streams.

mrodm:
test integrations

internal/testrunner/runners/system/runner.go
```diff
@@ -917,6 +967,9 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
 		}
 	}
 	r.deleteTestPolicyHandler = func(ctx context.Context) error {
+		if r.options.RunTestsOnly {
+			return nil
+		}
```
mrodm, line 972: Handlers have been updated depending on whether or not they should be running with `--no-provision` flag.

mrodm:
test integrations

internal/testrunner/runners/system/runner.go
jsoriano: And do we need the `deleteOldDocumentsDataStreamAndWait` method at all then?

jsoriano (approved):
I like the concept, but I wonder if we need as many policies as we have. We are creating three kinds of policies now.

internal/testrunner/runners/system/runner.go
mrodm: I'm not 100% sure.

This logic was already present before adding the flags to run system tests in different stages:
https://github.com/elastic/elastic-package/blob/ff60d256311c85d9b140abd395ab4cb4f0fa9550/internal/testrunner/runners/system/runner.go#L621-L641

Considering the three Agent Policies are kept, one scenario that I could think of where it would be better to keep it is when the Elastic Agent from the stack is used, and it is tested twice the same integration.
There is a handler to wipe the data from the data stream in the `tearDownTest`, but this tear down process could fail and leave the data stream with old documents in the following run.

One way that maybe it could be removed, it is also changing the behavior of the system testing when the Elastic Agent from the stack is used. It would mean to update the method `shouldCreateNewAgentPolicyForTest`.

One doubt about this, would the developers/users be using the current data streams created (e.g. logs-nginx.access-ep)? @jsoriano ?

internal/testrunner/runners/system/runner.go
```diff
@@ -948,10 +1001,24 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
...
 	logger.Debug("adding package data stream to test policy...")
-	ds := createPackageDatastream(*policyToTest, *scenario.pkgManifest, policyTemplate, *scenario.dataStreamManifest, *config, svcInfo.Test.RunID)
-	if r.options.RunTearDown || r.options.RunTestsOnly {
+	policyTesting := kibana.Policy{
+		Name:        fmt.Sprintf("ep-one-test-system-%s-%s-%s", r.options.TestFolder.Package, r.options.TestFolder.DataStream, scenario.startTestTime.Format("20060102T15:04:05Z")),
+		Description: fmt.Sprintf("test policy created by elastic-package test system for data stream %s/%s", r.options.TestFolder.Package, r.options.TestFolder.DataStream),
+		Namespace:   common.CreateTestRunID(),
+	}
+	policyToAssignDatastreamTests := policyToTest
+	if r.shouldCreateNewAgentPolicyForTest() {
```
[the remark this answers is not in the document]
mrodm: If I just update the `policyToTest` to be like this one, that means that the namespace is random, it would work just while the system tests are not run per stages (--setup, --no-provision, --tear-down).

When these flags are used it would be used the same data stream created in the first run, retrieved from the service state:
https://github.com/elastic/elastic-package/blob/858808bfec7b3dee5de84cdbbbf71ab09b87c09a/internal/testrunner/runners/system/runner.go#L859

That would mean that the method to delete old documents from the data stream could not be delete since that data stream would contain documents for sure.

Not tested, maybe it could be created always the `policyToTest` and just used the one from the state file to assign back the policy to the agent in the scenarios where the tests are performed in stages.

And for those test runs without these flags (common usage), there would be data streams created for each run:
data streams created

It could be added the logic to remove those data streams in one of the handlers.

In this PR these 3 policies are used as following:
- one for enrolling without any package policy
    - there were having issues when the stack is running for more than X minutes that agents were not able to enroll using the default policy.
- one to assign the package policy to the agent
    - this one would be used to assign back when running per stages after each --no-provision run 
- one for the actual testing
    - this policy is the one that would be using the Elastic Agent to run the system tests.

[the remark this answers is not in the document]
mrodm: I'll give it a try to remove one agent policy. I'll need to test with system test running per stages too.

internal/testrunner/runners/system/runner.go
jsoriano: > One way that maybe it could be removed, it is also changing the behavior of the system testing when the Elastic Agent from the stack is used. It would mean to update the method

Yes, it would be nice if when using the Agent from the stack, we also use a clean data stream while testing. This way we would have the same logic independently of the agent used:
* Create the policy for testing.
* Assign the policy to the agent.
* Run the tests.
* Assign the previously assigned policy to the agent.
* Remove the data stream created by the test policy.

I guess that in this case we need at most two policies:
* When `--setup` is used, the package policy can be assigned to the enroll policy. Then the test policy when using `--no-provision` would be basically a copy of the same policy, but with different namespace.
* When the stack agent is used, it already has a previous policy.

> One doubt about this, would the developers/users be using the current data streams created (e.g. logs-nginx.access-ep)? @jsoriano ?

Not sure of understanding the question. When `--setup` is used, the deployment should be functional, so I guess that in this case yes, users would be using these data streams. In other cases I don't think so, as the policies will be configured for specific packages only during tests.

internal/testrunner/runners/system/runner.go
```diff
@@ -920,7 +919,7 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
 		policyEnroll := kibana.Policy{
 			Name:        fmt.Sprintf("ep-test-system-enroll-%s-%s-%s", r.options.TestFolder.Package, r.options.TestFolder.DataStream, testTime),
 			Description: fmt.Sprintf("test policy created by elastic-package to enroll agent for data stream %s/%s", r.options.TestFolder.Package, r.options.TestFolder.DataStream),
-			Namespace:   "ep",
+			Namespace:   "enrollep",
```
jsoriano, line 922: This could be also randomized, right?

internal/testrunner/runners/system/runner.go
```diff
@@ -1150,6 +1151,9 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
 		}
 	}
 	r.resetAgentLogLevelHandler = func(ctx context.Context) error {
+		if r.options.RunTestsOnly {
+			return nil
+		}
```
jsoriano, line 1156: Nit. In these cases, could we directly avoid adding the handler?
```suggestion
	if r.options.RunTestsOnly {
		r.resetAgentLogLevelHandler = func(ctx context.Context) error {
```

jsoriano (approved):
LGTM. I think we can go further in the direction of using an isolated policy and data stream for testing, so we don't need to clean documents at any moment. But we can incrementally go in this direction.

internal/testrunner/runners/system/runner.go
mrodm: > Yes, it would be nice if when using the Agent from the stack, we also use a clean data stream while testing. This way we would have the same logic independently of the agent used:
> 
> * Create the policy for testing.
> * Assign the policy to the agent.
> * Run the tests.
> * Assign the previously assigned policy to the agent.
> * Remove the data stream created by the test policy.

:+1: 


> * When `--setup` is used, the package policy can be assigned to the enroll policy. Then the test policy when using `--no-provision` would be basically a copy of the same policy, but with different namespace.

The enroll policy does not contain any package policy, is just to be able to enroll the agent. In `--setup` stage, it is needed to create another policy that will be assigned back to the agent every time `--no-provision` finishes.

The two policies created (in --setup and --no-provision) are created from the contents of the package. It could be changed by the developer locally.

> Not sure of understanding the question. When `--setup` is used, the deployment should be functional, so I guess that in this case yes, users would be using these data streams. In other cases I don't think so, as the policies will be configured for specific packages only during tests.

Those data streams continue being functional, but from now on they will have another namespace (random) in the data stream.

internal/testrunner/runners/system/runner.go
mrodm: It could be done, but there are handlers that have more conditions to not run. For instance 

```
	r.removeAgentHandler = func(ctx context.Context) error {
		if r.runTestsOnly {
			return nil
		}
		// When not using independent agents, service deployers like kubernetes or custom agents create new Elastic Agent
		if !r.runIndependentElasticAgent && !svcInfo.Agent.Independent {
			return nil
		}
```

I think I'll keep all the handlers as they are to keep the same pattern in all of them.

internal/testrunner/runners/system/runner.go
mrodm: Sure, I'll update it.
This policy has no package policy assigned so it will not create any data stream in any case, but better to have it randomized to avoid future issues.

mrodm:
test integrations

internal/testrunner/runners/system/runner.go
```diff
@@ -911,47 +937,62 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
...
+	} else {
+		// Create a specific Agent Policy just for testing this test.
+		// This allows us to ensure that the Agent Policy used for testing is
+		// assigned to the agent with all the required changes (e.g. Package DataStream)
+		logger.Debug("creating test policy...")
+		policyToAssignDatastreamTests := kibana.Policy{
 			Name:        fmt.Sprintf("ep-test-system-%s-%s-%s", r.testFolder.Package, r.testFolder.DataStream, testTime),
 			Description: fmt.Sprintf("test policy created by elastic-package test system for data stream %s/%s", r.testFolder.Package, r.testFolder.DataStream),
-			Namespace:   "ep",
+			Namespace:   common.CreateTestRunID(),
```
mrodm, line 976: Now, every execution will create a new Agent Policy for testing, except when running with `--tear-down`. This tear down stage does not run any tests, so it is not needed.

jsoriano (approved):
Great change :+1:

internal/testrunner/runners/system/runner.go
```diff
@@ -456,13 +462,33 @@ func (r *runner) tearDownTest(ctx context.Context) error {
...
 	if r.resetAgentPolicyHandler != nil {
 		if err := r.resetAgentPolicyHandler(cleanupCtx); err != nil {
 			return err
 		}
 		r.resetAgentPolicyHandler = nil
 	}
 
+	// Shutting down the service should be run one of the first actions
+	// to ensure that resources created by terraform are deleted even if other
+	// errors fail.
```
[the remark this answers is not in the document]
mrodm: That's a good point.

There could be handlers that depend on others, but if so there will be another error. Or it could be detected the error and no execute the handler.

As you mentioned, for a future change.

internal/testrunner/runners/system/runner.go
```diff
@@ -911,47 +937,64 @@ func (r *runner) prepareScenario(ctx context.Context, config *testConfig, svcInf
...
+		// required to assign the policy stored in the service state file
+		// so data stream related to this Agent Policy can be obtained (and deleted)
+		// in the cleanTestScenarioHandler handler
+		policyToTest = policyCurrent
+	} else {
+		// Create a specific Agent Policy just for testing this test.
+		// This allows us to ensure that the Agent Policy used for testing is
+		// assigned to the agent with all the required changes (e.g. Package DataStream)
+		logger.Debug("creating test policy...")
+		policyToAssignDatastreamTests := kibana.Policy{
```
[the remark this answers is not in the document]
mrodm: I just wrote that name while developing the feature and I didn't re-check it :sweat_smile: 
I'll rename it with just `policy` :+1:

PR #1850, opened 2024-05-21
Review service container aliases and hostname

mrodm:

PR #1871, opened 2024-05-29
Add missing Test Run ID for service up command

mrodm:
Follows #1850 

Add missing test run ID when creating the service using `elastic-package service up` command.

Take the chance to remove some unused fields from `servicedeployer.FactoryOptions`.

### DIFF
--- a/internal/service/boot.go
+++ b/internal/service/boot.go
@@ -8,11 +8,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/rand"
 	"os"
 	"os/signal"
 	"syscall"
 
+	"github.com/elastic/elastic-package/internal/common"
 	"github.com/elastic/elastic-package/internal/logger"
 	"github.com/elastic/elastic-package/internal/profile"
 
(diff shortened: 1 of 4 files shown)


internal/service/boot.go
```diff
@@ -20,6 +21,15 @@ import (
...
 
+const (
+	testRunMaxID = 99999
+	testRunMinID = 10000
+)
+
+func createTestRunID() string {
+	return fmt.Sprintf("%d", rand.Intn(testRunMaxID-testRunMinID)+testRunMinID)
+}
+
```
mrodm, line 32: Copied from testrunner/system. 

WDYT about dding this to `internal/common/helpers.go` ?

jsoriano: Yes, if used in multiple places we could move it to some common package.

jsoriano (approved):

internal/service/boot.go
mrodm: Ok, I'll move it to that location then.

PR #1876, opened 2024-05-31
[Spacetime] Use structured logging (log/slog)

mrodm:
Currently, `elastic-package` uses a global logger based directly on `log.Print` calls: https://github.com/elastic/elastic-package/blob/486698806a279e44f7c6e03d896b9fa8e1cb1408/internal/logger/logger.go

This PR updates the code base to start using the new `log/slog` package introduced in Go 1.21 (https://go.dev/blog/slog):

- Revisit commands to use always a new instance of logger `log/slog`.
- Added a new Handler to show the message as plain text and all the Log Attributes that have been added as JSON. Example:
  ```
  2024/05/31 00:50:00 DEBUG: Compress using archiver.Zip 
  {
    "destination": "/home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.2.0.zip",
    "elastic-package.command": "test",
    "package": "elastic_package_registry",
    "source": "/home/mariorodriguez/Coding/work/integrations/packages/elastic_package_registry",
    "testrunner": "asset"
  }

  2024/05/31 11:25:04 DEBUG: adding service container internal ports to context 
  {
    "config": "default",
    "elastic-package.command": "test",
    "package.datastream": "metrics",
    "package.name": "elastic_package_registry",
    "package.type": "integration",
    "service": "elastic-package-service-44240-elastic_package_registry-1",
    "service.deployer": "compose",
    "service.name": "elastic_package_registry",
    "stack.version": "8.13.4",
    "test.config": "default",
    "testrunner": "system",
    "variant": ""
  }
  ```
- Leverage the usage of log Attributes, to show in each log message a better context.
   - Required some refactors to be able to use the same instance around the code base.
   - Update testrunner to include the logger instance in every runner
       - Created a factory to create runners on demand
- Previous logger calls, based on `log`, have been updated to run under the hood with an instance of `log/Slog` Logger.
- A new log level TRACE has been added, to show even more information than DEBUG level.
    - Examples: show more information about container health status, show the body from the responses, etc.
- Revisited CLI:
    - Updated `-v` flag, and now it can be set multiple times:
        - `-v`: debug level
        - `-vv`: trace level
        - `-vvv`: adds source information (file, function and line of that log)
          ```
          2024/05/31 01:00:07 DEBUG: Kibana request 
          {
            "elastic-package.command": "test",
            "method": "POST",
            "source": {
              "file": "/home/mariorodriguez/Coding/work/elastic-package-use-global-slog/internal/kibana/client.go",
              "function": "github.com/elastic/elastic-package/internal/kibana.(*Client).newRequest",
              "line": 189
            },
            "testrunner": "asset",
            "url": "https://127.0.0.1:5601/api/fleet/epm/packages"
          }
          ```        
    - Added a new `--log-format` flag to set the format to be used in our logging:
        - `elastic-package test system -v --log-format [json|text|default]`

## Space, Time
This project started during an ON Week, a time we give each other in Elastic to explore ideas or learn new things, in alignment with our [Source Code](https://www.elastic.co/about/our-source-code).


internal/testrunner/testrunner.go
```diff
@@ -284,13 +287,49 @@ func RegisterRunner(runner TestRunner) {
...
+	return &r
+}
+
+type RunnerLauncherOption func(r *runnerLauncher)
+
+func WithLogger(logger *slog.Logger) RunnerLauncherOption {
+	return func(r *runnerLauncher) {
+		r.logger = logger
+	}
+}
```
mrodm, line 309: Another option would be to set a new method in the `TestRunner` interface (e.g. `SetLogger()`), so it can be updated the logger of each runner dynamically.

mrodm: Updated with this option in https://github.com/elastic/elastic-package/pull/1876/commits/6cb19a0dfff335a04e46ab3d387f2ecfc781754b

PR #1879, opened 2024-06-03
Enable http2 in kibana

jsoriano:
Kibana 8.15.0 will have support for HTTP2 (after https://github.com/elastic/kibana/pull/183465). Let's use it in local stacks managed by `elastic-package`.

Or should we wait till this is closer to GA?

### DIFF
--- a/README.md
+++ b/README.md
@@ -648,6 +648,8 @@ The following settings are available per profile:
 * `stack.geoip_dir` defines a directory with GeoIP databases that can be used by
   Elasticsearch in stacks managed by elastic-package. It is recommended to use
   an absolute path, out of the `.elastic-package` directory.
+* `stack.kibana_http2_enabled` can be used to control if HTTP/2 should be used in versions of
+  kibana that support it. Defaults to true.
 * `stack.logstash_enabled` can be set to true to start Logstash and configure it as the
   default output for tests using elastic-package. Supported only by the compose provider.
   Defaults to false.
--- a/internal/stack/_static/kibana.yml.tmpl
+++ b/internal/stack/_static/kibana.yml.tmpl
@@ -1,6 +1,10 @@
 {{ $version := fact "kibana_version" }}
 server.name: kibana
 server.host: "0.0.0.0"
+{{ if not (semverLessThan $version "8.15.0") }}
+server.protocol: http2
+{{ end }}
+
 server.ssl.enabled: true
 server.ssl.certificate: "/usr/share/kibana/config/certs/cert.pem"
 server.ssl.key: "/usr/share/kibana/config/certs/key.pem"
--- a/internal/stack/_static/kibana.yml.tmpl
+++ b/internal/stack/_static/kibana.yml.tmpl
@@ -1,10 +1,9 @@
-{{ $version := fact "kibana_version" }}
+{{- $version := fact "kibana_version" }}
 server.name: kibana
 server.host: "0.0.0.0"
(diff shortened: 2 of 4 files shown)


jsoriano:
Moving to draft till this is available in snapshots.

ruflin:
This is great! It will help us to test http2 much more.

jsoriano:
Snapshots already have the flag, opening for review.

ruflin (approved):
I checked out the branch and did spin up the cluster quickly, clicking around. Looks good to me. How do I confirm that http2 is enabled? Is there some indicator or similar?

jsoriano:
> How do I confirm that http2 is enabled? Is there some indicator or similar?

In Kibana itself no idea, but you can check with curl.

Before this change:
```
$ curl -v -k https://localhost:5601/api
*   Trying 127.0.0.1:5601...
* Connected to localhost (127.0.0.1) port 5601 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
...
* ALPN, server accepted to use http/1.1
...
> GET /api HTTP/1.1
```

After this change:
```
$ curl -v -k https://localhost:5601/api
*   Trying 127.0.0.1:5601...
* Connected to localhost (127.0.0.1) port 5601 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
...
* ALPN, server accepted to use h2
...
> GET /api HTTP/2
```

In the browser developer tools it is also possible to see the `HTTP/2` version for any of the requests done.

ruflin:
Checking with curl confirms things are working as expected 👍

jsoriano:
@kpollich @mrodm wdyt about merging this?

mrodm:
> @kpollich @mrodm wdyt about merging this?

This will enable HTTP/2 always starting with 8.15.0-SNAPSHOT, would it be good to have some setting in the stack configuration to disable this (going back to HTTP/1.1)? Similar to the `self_monitor_enabled` setting ? 

https://github.com/elastic/elastic-package/blob/1765efb3e842ad959c6ee69c2ebfc27928c406d2/internal/stack/resources.go#L63

Thinking in case it is needed to test both scenarios. Would that be needed?

kpollich:
> This will enable HTTP/2 always starting with 8.15.0-SNAPSHOT, would it be good to have some setting in the stack configuration to disable this (going back to HTTP/1.1)? Similar to the self_monitor_enabled setting ?

+1 on having a setting to disable this as a potential troubleshooting mechanism. Otherwise I'm good with merging this.

jsoriano:
> This will enable HTTP/2 always starting with 8.15.0-SNAPSHOT

Correct.

> would it be good to have some setting in the stack configuration to disable this (going back to HTTP/1.1)? Similar to the `self_monitor_enabled` setting ?

We could, but I don't think it would be needed. If we add a setting, I would add it as escape hatch to disable http2 if found problematic, but I would keep using http2 when possible by default. But if we find problems with http2 we might also consider reverting this change. I would leave this for a future change, if we find we need it for some reason.

> Thinking in case it is needed to test both scenarios. Would that be needed?

I don't think so. This only changes the transport, nothing functional should change.

jsoriano:
> +1 on having a setting to disable this as a potential troubleshooting mechanism.

Ok, added flag as potential troubleshooting mechanism.

mrodm (approved):
:rocket:

kpollich (approved):

Issue #1884, opened 2024-06-04
Bug: Elastic-Package v100 not working on Windows.


Alphayeeeet:
Due to the most recent changes and the usage of a pseudo terminal library (pty), which is not supported under windows, the following error occurs when trying to boot up the elastic stack on windows:

Error: booting up the stack failed: building docker images failed: could not create docker compose project: unable to determine Docker Compose version: running Docker Compose version command failed: failed to open pseudo-tty to capture stderr: unsupported

Alphayeeeet:
Breaking change was in this PR: #1801

jsoriano:
@Alphayeeeet thanks for reporting, let me take a look.

PR #1885, opened 2024-06-04
Revert use of PTY for docker compose invocation in Windows

jsoriano:
In https://github.com/elastic/elastic-package/pull/1801 we introduced a virtual tty to capture docker compose output as it is generated in normal consoles. The library we use for that doesn't work on Windows, effectively breaking stack management and tests depending on int.

While we find an alternative, remove the use of PTY in Windows, effectively reverting #1801 on this platform.

Add also a test to check that `elastic-package stack` keeps working on Windows.

Fixes https://github.com/elastic/elastic-package/issues/1884.


.buildkite/pipeline.yml
```diff
@@ -48,8 +64,6 @@ steps:
         allow_failure: false
       - step: unit-tests-linux
         allow_failure: false
-      - step: unit-tests-windows
-        allow_failure: false
```
mrodm: Ok, no need to wait for windows unit tests step if linux unit tests are successful. :+1: 

What I would suggest is to move the `wait` step before the windows unit test step

```yaml
  - label: ":go: :linux: Run unit tests"
    key: unit-tests-linux
    command: "make test-go-ci"
    artifact_paths:
      - "build/test-results/*.xml"
      - "build/test-coverage/*.xml"
    agents:
      image: "${LINUX_AGENT_IMAGE}"
      cpu: "8"
      memory: "4G"

  - wait: ~
    continue_on_failure: true

  - label: ":go: :windows: Run unit tests"
    key: unit-tests-windows
    command: ".buildkite/scripts/unit_tests_windows.ps1"
    agents:
      provider: "gcp"
      image: "${WINDOWS_AGENT_IMAGE}"
    artifact_paths:
      - "TEST-unit-windows.xml"

  - label: ":pipeline: Trigger Integration tests"
    command: ".buildkite/pipeline.trigger.integration.tests.sh | buildkite-agent pipeline upload"
    depends_on:
      - step: check-static
        allow_failure: false
      - step: unit-tests-linux
        allow_failure: false
```

Mainly for the UI view because the trigger integration tests will run even windows steps are still running

buildkite steps

mrodm: Ok, I see it has been moved the steps, and the step to run stack in windows is running at the same time as the other integration test.

Ignore this comment.

.buildkite/pipeline.yml
```diff
@@ -48,8 +49,20 @@ steps:
...
+    agents:
+      provider: "gcp"
+      image: "${WINDOWS_AGENT_IMAGE}"
+    depends_on:
+      - step: check-static
+        allow_failure: false
       - step: unit-tests-windows
         allow_failure: false
+    artifact_paths:
+      - 'C:\ProgramData\chocolatey\logs\chocolatey.log'
```
mrodm, line 65: Is this needed? Could it be removed?

.buildkite/scripts/stack_tests_windows.ps1
```diff
@@ -0,0 +1,68 @@
...
+Write-Host "--- Docker Info"
+docker info
+
+echo "--- Downloading Go modules"
+go version
+go mod download -x
+
+echo "--- Running stack tests"
+go version
+$ErrorActionPreference = "Continue" # set +e
```
mrodm, line 59: Here it will be shown the same go version as in line 54, no ?

If so, this could be removed.

```suggestion
echo "--- Running stack tests"
$ErrorActionPreference = "Continue" # set +e
```

mrodm: This is also in the other windows script, I think that command could be removed in both scripts.

jsoriano: Yeah, I copied it around, didn't think a lot about this, I will review it, thanks!

.buildkite/pipeline.yml
jsoriano: I found it useful to see what chocolatey is doing, so I left it. But I can remove it and we can add it back when debugging issues installing things, as you prefer.

mrodm: Ok, if it has been helpful, let's keep it.

mrodm (approved):

PR #1886, opened 2024-06-04
Update elastic-package test command to define each command separately

mrodm:
This PR defines each subcommand under `elastic-package test` in its own function.

This allows us to remove functions from the TestRunner interface: TestFolderRequired, CanRunPerDataStream, CanRunSetupTeardownIndependent

### DIFF
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -155,12 +155,13 @@ func testRunnerAssetCommandAction(cmd *cobra.Command, args []string) error {
 	_, pkg := filepath.Split(packageRootPath)
 
 	results, err := testrunner.Run(ctx, testType, testrunner.TestOptions{
-		Profile:         profile,
-		TestFolder:      testrunner.TestFolder{Package: pkg},
-		PackageRootPath: packageRootPath,
-		WithCoverage:    testCoverage,
-		CoverageType:    testCoverageFormat,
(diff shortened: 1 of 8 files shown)


cmd/testrunner.go
```diff
@@ -58,345 +56,842 @@ These tests allow you to test different configuration options and the policies t
...
+			return fmt.Errorf("wrong number of test folders (expected 1): %d", len(testFolders))
 		}
 
-		deferCleanup, err := cmd.Flags().GetDuration(cobraext.DeferCleanupFlagName)
+		cmd.Printf("Running tests per stages (technical preview)\n")
+	}
+
+	var results []testrunner.TestResult
+	for _, folder := range testFolders {
+		r, err := testrunner.Run(ctx, testType, testrunner.TestOptions{
```
jsoriano, line 694: Now that we have specific commands for each test runner, maybe we can call here the specific runner, with an specific set of options? This would also allow to remove the registry of tests.

Actually now we wouldn't even need the test runner interface, right? :thinking:

jsoriano (approved):

cmd/testrunner.go
mrodm: It looks like this could be possible, I'll check this out in a follow-up PR.

PR #1888, opened 2024-06-05
Instantiate runners directly on test commands

mrodm:
Relates #787
Follows #1886 

This PR instantiates in each command the corresponding runner, for that new constructors have been added into each runner. This allows us to create independent Options structs for each runner with just the required options for each runner.

These changes allow to remove the registry of the runners.

For now, it is kept the TestRunner interface since that will force to define Run and TearDown methods so they can be used in `testrunner.Run()`

### DIFF
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -28,6 +28,7 @@ import (
 	_ "github.com/elastic/elastic-package/internal/testrunner/runners" // register all test runners
 	"github.com/elastic/elastic-package/internal/testrunner/runners/asset"
 	"github.com/elastic/elastic-package/internal/testrunner/runners/pipeline"
+	"github.com/elastic/elastic-package/internal/testrunner/runners/policy"
 	"github.com/elastic/elastic-package/internal/testrunner/runners/static"
 	"github.com/elastic/elastic-package/internal/testrunner/runners/system"
 )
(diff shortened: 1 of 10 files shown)


internal/testrunner/runners/policy/runner.go
```diff
@@ -98,14 +116,10 @@ func (r *runner) runTest(ctx context.Context, manager *resources.Manager, option
...
-		// issue with the final policy.
-		expectedRevision := 1 + len(policy.PackagePolicies)
-		if options.GenerateTestResult {
-			testErr = dumpExpectedAgentPolicy(ctx, options, testPath, policy.ID, expectedRevision)
+		if r.generateTestResult {
+			testErr = dumpExpectedAgentPolicy(ctx, r.kibanaClient, testPath, policy.ID)
 		} else {
-			testErr = assertExpectedAgentPolicy(ctx, options, testPath, policy.ID, expectedRevision)
+			testErr = assertExpectedAgentPolicy(ctx, r.kibanaClient, testPath, policy.ID)
 		}
```
mrodm, line 123: Removed `expectedRevision` as parameter since in none of those functions was used.

Should it be used that variable in those functions ? @jsoriano

cmd/testrunner.go
```diff
@@ -164,16 +168,13 @@ func testRunnerAssetCommandAction(cmd *cobra.Command, args []string) error {
...
-		TestFolder:                 testrunner.TestFolder{Package: pkg},
-		PackageRootPath:            packageRootPath,
-		WithCoverage:               testCoverage,
-		CoverageType:               testCoverageFormat,
-		KibanaClient:               kibanaClient,
-		RunIndependentElasticAgent: false,
+	runner := asset.NewAssetRunner(asset.AssetRunnerOptions{
+		TestFolder:      testrunner.TestFolder{Package: pkg},
+		PackageRootPath: packageRootPath,
+		KibanaClient:    kibanaClient,
```
jsoriano, line 174: Good to have specific option types :+1:

internal/testrunner/runners/pipeline/runner.go
```diff
@@ -321,8 +354,13 @@ func (r *runner) runTestCase(ctx context.Context, testCaseFile string, dsPath st
 		return tr, nil
 	}
 
-	if r.options.WithCoverage {
-		tr.Coverage, err = GetPipelineCoverage(r.options, r.pipelines)
+	if r.withCoverage {
+		tr.Coverage, err = GetPipelineCoverage(PipelineRunnerOptions{
```
jsoriano, line 358: Nit. Is this method used in other packages?
```suggestion
		tr.Coverage, err = getPipelineCoverage(PipelineRunnerOptions{
```

internal/testrunner/runners/policy/runner.go
jsoriano: Ah no, I think the code using this variable was removed.

jsoriano (approved):
Nice. This change goes in the good direction :+1:

internal/testrunner/runners/pipeline/runner.go
mrodm: I'll check it, and if it is not used in other packages I'll update it  :+1:

mrodm: Done in https://github.com/elastic/elastic-package/pull/1888/commits/51b825de892fcc46fde39ef0c1bc5d09e8297c3b

PR #1892, opened 2024-06-06
Move installation package in policy and system tests

mrodm:
Part of #787 
Supersedes #1845 

Package installation process is repeated for every data stream and configuration file defined in every package (system and policy tests).

This process is just needed once:
- No need to install multiple times the same package.
- Every data stream and configuration file defines the required Package Policy.

This PR updates the runner to include two new methods in the interface (`SetupRunner` and `TearDownRunner`). These two methods are in charge of creating (or deleting) the global resources required to run all the tests for every data streams and configuration file defined in the package:
- For system and policy tests, installation (and uninstallation) of the package would be performed in these two new methods.

Depends on #1866

### DIFF
--- a/internal/testrunner/runners/system/runner.go
+++ b/internal/testrunner/runners/system/runner.go
@@ -168,7 +168,7 @@ type runner struct {
 	shutdownAgentHandler      func(context.Context) error
 }
 
-type SystemRunnerOptions struct {
+type SystemTesterOptions struct {
 	PackageRootPath string
 	KibanaClient    *kibana.Client
 
@@ -198,7 +198,7 @@ type SystemTestRunnerOptions struct {
 	RunTestsOnly   bool
 }
 
-func NewSystemRunner(options SystemRunnerOptions) *runner {
+func NewSystemTester(options SystemTesterOptions) *runner {
 	r := runner{
 		packageRootPath: options.PackageRootPath,
 		kibanaClient:    options.KibanaClient,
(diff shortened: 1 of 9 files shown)


mrodm:
test integrations

internal/testrunner/runners/system/runner.go
```diff
@@ -207,10 +231,53 @@ func NewSystemRunner(options SystemRunnerOptions) *runner {
...
+	r.serviceStateFilePath = filepath.Join(testrunner.StateFolderPath(r.profile.ProfilePath), testrunner.ServiceStateFileName)
 	// TODO: check if logic in initRun could be moved to this constructor
 	return &r
 }
 
+// SetupRunner prepares global resources required by the test runner.
+func (r *runner) SetupRunner(ctx context.Context) error {
+	if r.runTearDown {
+		logger.Debug("Skip installing package")
+	} else {
```
jsoriano, line 246: Nit.
```suggestion
		return nil
	}
```

jsoriano (approved):
:+1:

internal/testrunner/runners/system/runner.go
efd6: This could have the condition inverted and then the else block removed.

mrodm: Hi @efd6 , that change to remove the else block was done in https://github.com/elastic/elastic-package/pull/1892/commits/68dfcb24e5e525d10f18e1aeb29f7b64e77a45c0 keeping the same condition.

PR #1895, opened 2024-06-07
Move getting all tests to run inside each runner

mrodm:
Part of #787

This PR moves the logic to obtain tests to each runner.

For that it has been separated into two different structs each runner:
- TestRunner: holds the logic to perform global setup or tear down process and retrieve the list of tests (testers)
- Tester: holds the logic to run the one test.

With this PR each TestRunner would return a list of Testers. Each one of these Tester instances must have all the information to run its own test.

### DIFF
--- a/internal/testrunner/testrunner.go
+++ b/internal/testrunner/testrunner.go
@@ -6,7 +6,6 @@ package testrunner
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -19,12 +18,6 @@ import (
 	"github.com/elastic/elastic-package/internal/kibana"
 	"github.com/elastic/elastic-package/internal/packages"
 	"github.com/elastic/elastic-package/internal/profile"
-	"github.com/elastic/elastic-package/internal/stack"
-)
-
-const (
-	ServiceStateFileName = "service.json"
-	StateFolderName      = "state"
 )
 
 // TestType represents the various supported test types
@@ -368,11 +361,6 @@ func findPackageTestFolderPaths(packageRootPath, testTypeGlob string) ([]string,
 	return paths, err
 }
 
-// StateFolderPath returns the folder where the state data is stored
-func StateFolderPath(profilePath string) string {
-	return filepath.Join(profilePath, stack.ProfileStackPath, StateFolderName)
-}
-
 func PackageHasDataStreams(manifest *packages.PackageManifest) (bool, error) {
 	switch manifest.Type {
 	case "integration":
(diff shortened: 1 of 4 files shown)


cmd/testrunner.go
```diff
@@ -774,53 +707,20 @@ func testRunnerPolicyCommandAction(cmd *cobra.Command, args []string) error {
...
-		testFolders, err = testrunner.FindTestFolders(packageRootPath, nil, testType)
+		err = validateDataStreamsFlag(packageRootPath, dataStreams)
 		if err != nil {
-			return fmt.Errorf("unable to determine test folder paths: %w", err)
-		}
-		if failOnMissing && len(testFolders) == 0 {
-			return fmt.Errorf("no %s tests found", testType)
+			return cobraext.FlagParsingError(err, cobraext.DataStreamsFlagName)
 		}
 	}
```
jsoriano, line 725: We could move this to a function that we could reuse.
```
dataStreams, err := getDataStreamsFlag(cmd)
if err != nil {
... 
```

cmd/testrunner.go
```diff
@@ -774,53 +707,20 @@ func testRunnerPolicyCommandAction(cmd *cobra.Command, args []string) error {
...
-				return cobraext.FlagParsingError(err, cobraext.DataStreamsFlagName)
-			}
-		}
-
-		testFolders, err = testrunner.FindTestFolders(packageRootPath, dataStreams, testType)
+	var dataStreams []string
+	if cmd.Flags().Lookup(cobraext.DataStreamsFlagName) != nil {
+		// We check for the existence of the data streams flag before trying to
+		// parse it because if the root test command is run instead of one of the
+		// subcommands of test, the data streams flag will not be defined.
```
jsoriano, line 714: If we want the root command to support this flag, maybe we can just define it in the root command too?

internal/testrunner/testrunner.go
```diff
@@ -66,6 +68,23 @@ type TestRunner interface {
...
+
+// TestRunner is the interface test runners that require a global initialization must implement.
+type TestRunner interface {
+	// Type returns the test runner's type.
+	Type() TestType
+
+	// SetupRunner prepares global resources required by the test runner.
+	SetupRunner(context.Context) error
+
+	GetTests(context.Context) ([]TestFolder, error)
```
jsoriano, line 81: Umm, I would expect this to directly return test[er]s.
```suggestion
	GetTests(context.Context) ([]Tester, error)
```

jsoriano (commented):
Nice, this is making test commands much more straightforward.

cmd/testrunner.go
mrodm: Probably, this "if" could be removed now that test commands are defined each one in its own function.

`elastic-package test` command does not define `--data-streams` flag so it fails if it is defined:
```bash
 $ elastic-package test  -v --data-streams access
Error: unknown flag: --data-streams
```

And if it is executed the `elastic-package test system` subcommand, as the required sub-commands define the data stream flag if they needed, it does not fail neither.

cmd/testrunner.go
mrodm: Done in https://github.com/elastic/elastic-package/pull/1895/commits/cbe756f624dc6271d744c916d5ae90f9e45be940

internal/testrunner/testrunner.go
mrodm: Update method to return a Tester instead of TestFolder.
Each tester returned must have all the configuration / parameters set.

internal/testrunner/testrunner.go
```diff
@@ -76,6 +71,8 @@ type TestRunner interface {
 	// SetupRunner prepares global resources required by the test runner.
 	SetupRunner(context.Context) error
 
+	GetTests(context.Context) ([]Tester, error)
```
jsoriano, line 74: :+1:

jsoriano (approved):
Nice! Haven't reviewed a lot the moved code, I guess this is mostly the same?

mrodm:
> Nice! Haven't reviewed a lot the moved code, I guess this is mostly the same?

Exactly, I've separated the code into different files according to what struct they refer (runner, tester or state), but the code remains the same.

The main changes are in runners , where the function GetTests returns in this PR a tester per each TestFolder 

https://github.com/elastic/elastic-package/pull/1895/files#diff-6f892c7c658aca0069c0d5c93c0bff77f0a52f608687126380ace89df6c10fc3R120-R130


In a follow PR , this function would return a tester per each configuration file for instance in system tests.

Issue #1743, opened 2024-04-01
Test pipelines with pipelines provided by fleet


jsoriano:
Current pipeline tests install the pipelines included in packages and test them standalone using the simulate API.

In real scenarios Fleet also adds some processing, this is tested on system tests, but not on pipeline tests.

It would be good to somehow test pipelines with the processing added by Fleet, for scenarios that are difficult to test with system tests, such as the Windows integration.

This could be done by installing the whole package and still use the simulate API, instead of installing the pipelines only. Validation of received documents would need to be different in this mode because we wouldn't have the same level of control on the resulting documents.

kpollich:
We haven't revisited this issue in a while, so we're closing it due to inactivity. Please reopen with a comment if you think this is a mistake. Thanks!

jsoriano:
Let's keep this open as this would perform more realistic pipeline tests.

PR #1898, opened 2024-06-10
Update Tester instances to trigger just one test

mrodm:
Part of #787 
Depends on #1895

Changed how Tester instances are created in each runner. In this PR, now it is created a Tester per each configuration file and variant (if required), so each Tester instance should be in charged of just one test.

This PR affects the following testers:
- pipeline: one Tester per each configuration file.
- policy: one Tester per each configuration file.
- system: one Tester per each configuration file and variant.

Static tests keep creating one Tester per data stream folder, and Asset tests just create one test per package.

Note:
- Runner instance is now in charge of performing the safeguards related to running system tests per stages.


test/packages/false_positives/cisco_asa.expected_errors
```diff
@@ -2,3 +2,14 @@
...
+<failure>detected ingest pipeline warnings: 2: character class has &#39;-&#39; without escape&#xA;regular expression has redundant nested repeat operator.*</failure>
+<failure>detected ingest pipeline warnings: 2: character class has &#39;-&#39; without escape&#xA;regular expression has redundant nested repeat operator.*</failure>
+<failure>detected ingest pipeline warnings: 2: character class has &#39;-&#39; without escape&#xA;regular expression has redundant nested repeat operator.*</failure>
+<failure>detected ingest pipeline warnings: 2: character class has &#39;-&#39; without escape&#xA;regular expression has redundant nested repeat operator.*</failure>
+<failure>detected ingest pipeline warnings: 2: character class has &#39;-&#39; without escape&#xA;regular expression has redundant nested repeat operator.*</failure>
+<failure>detected ingest pipeline warnings: 2: character class has &#39;-&#39; without escape&#xA;regular expression has redundant nested repeat operator.*</failure>
+<failure>detected ingest pipeline warnings: 2: character class has &#39;-&#39; without escape&#xA;regular expression has redundant nested repeat operator.*</failure>
```
mrodm, line 15: There is one failure per pipeline test.

This error is the result of installing the pipelines of `cisco_asa`. Previously, pipelines were installed just onace, now those pipelines are installed once per test (Run()).

jsoriano: Will these failures be predictable if we parallelize?

At some point we may consider installing the pipelines by installing the packages, then we could install the package only once. The pipelines we test now are not exactly the same as what is executed at the end, see https://github.com/elastic/elastic-package/issues/1743.

jsoriano: Maybe another mid-ground option is to install all pipelines included in the package in the `Setup` step, and remove them on `TearDown`, but then we would be installing all pipelines, even the ones that are not tested. Not sure I like this approach.

mrodm: > Will these failures be predictable if we parallelize?

I guess this would be a source of flakiness. A pipeline test running in parallel with another one could report the errors from the other pipeline test and vice versa. So I guess the number of errors could vary.

It's also true that for now there would be support to run in parallel just the system tests. The other type of tests would be running sequentially.

> Maybe another mid-ground option is to install all pipelines included in the package in the Setup step, and remove them on TearDown, but then we would be installing all pipelines, even the ones that are not tested.

Do you mean installing and removing them in the `SetupRunner` and `TearDownRunner` methods? so those are just performed once

Currently, those methods just return `error` in the interface, if we want to do that probably they should also return `[]testrunner.Result` , to return errors if any in those steps.

jsoriano (approved):

PR #1906, opened 2024-06-13
Fix race condition reading from PTTY

jsoriano:
Execution of fast docker compose commands are getting stuck sometimes. One possible theory is that we are closing the pseudo-tty before the copy has finished, and it hangs there.

Close the pseudo-TTY after waiting for the finalization of the goroutine, so we avoid problems caused by closing it while it is being read.

Thanks @mrodm for reporting!


mrodm (approved):
Perfect, thanks!!
Let's try running several times the CI :crossed_fingers:

jsoriano:
Three builds green, I am merging this.

PR #1913, opened 2024-06-18
Added support for data stream overrides for specific integrations based on a defined list

ShourieG:
Some integrations like **amazon security lake** use routing rules to route results to a different data stream from a central data stream. Having such a mechanism the standard system tests fail since it always searches for hits in the central data stream. We need a way to dynamically instruct elastic-package to look for hits in a specific data stream during system tests. 
A similar functionally was recently added with a recent PR [here](https://github.com/elastic/elastic-package/pull/1851). This is just a small iteration on top of that change to allow elastic-package to honour dynamically defined data streams for select integration packages. 

NOTE: I have tested the functionally locally.

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -122,6 +122,9 @@ var (
 			},
 		},
 	}
+	dataStreamOverrides = map[string]bool{
+		"amazon_security_lake": true,
+	}
 	enableIndependentAgents = environment.WithElasticPackagePrefix("TEST_ENABLE_INDEPENDENT_AGENT")
 )
 
@@ -937,7 +940,7 @@ func (r *tester) prepareScenario(ctx context.Context, config *testConfig, svcInf
 
 	// Input packages can set `data_stream.dataset` by convention to customize the dataset.
 	dataStreamDataset := ds.Inputs[0].Streams[0].DataStream.Dataset
-	if scenario.pkgManifest.Type == "input" {
+	if scenario.pkgManifest.Type == "input" || dataStreamOverrides[scenario.pkgManifest.Name] {
 		v, _ := config.Vars.GetValue("data_stream.dataset")
 		if dataset, ok := v.(string); ok && dataset != "" {
 			dataStreamDataset = dataset
@@ -1290,7 +1293,7 @@ func (r *tester) validateTestScenario(ctx context.Context, result *testrunner.Re
 		}
 		expectedDatasets = []string{expectedDataset}
 	}
-	if scenario.pkgManifest.Type == "input" {
+	if scenario.pkgManifest.Type == "input" || dataStreamOverrides[scenario.pkgManifest.Name] {
 		v, _ := config.Vars.GetValue("data_stream.dataset")
 		if dataset, ok := v.(string); ok && dataset != "" {
 			expectedDatasets = append(expectedDatasets, dataset)


internal/testrunner/runners/system/tester.go
```diff
@@ -122,6 +122,9 @@ var (
 			},
 		},
 	}
+	dataStreamOverrides = map[string]bool{
+		"amazon_security_lake": true,
```
jsoriano, line 126: We don't add exceptions to specific packages in elastic-package. We should have generic mechanism so this is available for any package that supports it.

On this specific case, we have code to support routing rules, as well as custom datasets. If this is not enough for this package we should check why.

internal/testrunner/runners/system/tester.go
```diff
@@ -937,7 +940,7 @@ func (r *tester) prepareScenario(ctx context.Context, config *testConfig, svcInf
 
 	// Input packages can set `data_stream.dataset` by convention to customize the dataset.
 	dataStreamDataset := ds.Inputs[0].Streams[0].DataStream.Dataset
-	if scenario.pkgManifest.Type == "input" {
+	if scenario.pkgManifest.Type == "input" || dataStreamOverrides[scenario.pkgManifest.Name] {
```
jsoriano, line 943: I wonder if we should do this in all cases and not only for input packages :thinking: Though integration packages should not need to configure datasets.

Or maybe this `amazon_security_lake` package should be an input package if it is expected to be used with custom datasets?

jsoriano (changes requested):
We need to investigate this more, it is unexpected to need to add a exception for a package.

internal/testrunner/runners/system/tester.go
jsoriano: Support for reroute was added in https://github.com/elastic/elastic-package/pull/1372 and https://github.com/elastic/elastic-package/pull/1391.

internal/testrunner/runners/system/tester.go
ShourieG: Adding this for all integration packages seems unnecessary and might unexpected results in tests. The thing is with this specific integration, there's a central data stream that runs the tests, but then re routes the results to a separate data stream. This re route mechanism causes the system test to fail. In future other integration packages might also do this, hence why I created this override list.

This also cannot be an input package cause it has multiple data streams and ingest pipelines had have a lot of mapping logic. Generally input packages just consume raw inputs in a single generic data stream and dump them into elastic search.

internal/testrunner/runners/system/tester.go
jsoriano: The examples and the test package in https://github.com/elastic/elastic-package/pull/1372 are integration packages.

internal/testrunner/runners/system/tester.go
jsoriano: > The thing is with this specific integration, there's a central data stream that runs the tests, but then reroutes the results to a separate data stream.

I see we have some packages with routing rules, but don't have system tests, it would be good to identify what is missing and fix it so all these packages can have system tests.

On the other hand I see that the `entityanalytics_entra_id` package has routing rules and system tests. Is it doing some trick to workaround the current limitations? Or is this package doing something we could apply in other cases?

> Maybe we can add support for a new package spec variable, whose presence could trigger this ? Then we can remove the list.

Yes, I would prefer this before adding a list with exceptions, but lets try to identify first what is missing, or what other packages are doing to have system tests with routing rules.

I think it would be good to open an issue identifying the problem, and then we could discuss about possible solutions. @ShourieG would this work for you?

ShourieG: @jsoriano I saw the **entityanalytics_entra_id** integration, and I don't think the systems tests over there are working properly, if you see the sample_event.json, it only contains agent data and does not contain any fields under the **entityanalytics_entra_id** namespace as per the field mappings. It also seems that the entityanalytics input has some internal mechanism to publish logs to all datasets which is not supported by the aws input, hence why a sample_event.json is even being generated in the first place even though the contents are incorrect.

ShourieG:
created a new issue as suggested here : https://github.com/elastic/elastic-package/issues/1917

PR #1914, opened 2024-06-18
Add flag to change directory before running the command

jsoriano:
Add `-C`/`--change-directory` flag to change the directory before running any command. This allows to build, install or test packages without needing to move to the directory with the source of the package, in a similar fashion to similar flags other commands have.


mrodm (approved):
:rocket:

Issue #1624, opened 2024-01-17
Coverage for pipeline tests are not generated for all data streams in some scenarios


mrodm:
Relates #1595 

Currently, there are three scenarios for reporting coverage in pipeline tests considering
just one data stream:

1. No ingest pipeline defined (and no tests)
   - No report is generated, empty coverage file.
     ```xml
     <?xml version="1.0" encoding="UTF-8"?>
     
     <coverage version="1">
       <!--Coverage for pipeline test--></coverage>
     ```
2. There is an ingest pipeline but no tests
   - It creates a coverage report with a Missing method*
     ```xml
     <?xml version="1.0" encoding="UTF-8"?>
     
     <coverage version="1">
       <file path="packages/elastic_package_registry/data_stream/metrics/manifest.yml">
         <lineToCover lineNumber="2" covered="false"></lineToCover>
       </file>
       <!--Coverage for pipeline test-->
     </coverage>
     ```
3. There is an ingest pipeline and tests for that pipeline
   - It creates a coverage report with all the processors used as methods
     ```xml
     <?xml version="1.0" encoding="UTF-8"?>
     
     <coverage version="0">
       <file path="packages/elastic_package_registry/data_stream/metrics/elasticsearch/ingest_pipeline/default.yml">
         <lineToCover lineNumber="4" covered="true"></lineToCover>
         <lineToCover lineNumber="5" covered="true"></lineToCover>
         <lineToCover lineNumber="6" covered="true"></lineToCover>
         <lineToCover lineNumber="10" covered="true"></lineToCover>
     ....
         <lineToCover lineNumber="85" covered="true"></lineToCover>
         <lineToCover lineNumber="86" covered="true"></lineToCover>
         <lineToCover lineNumber="87" covered="true"></lineToCover>
         <lineToCover lineNumber="88" covered="true"></lineToCover>
       </file>
       <!--Cobertura for pipeline test-->
     </coverage>
     ```
   
This is working correctly while there is just one data stream with ingest pipelines.
If there are multiple data streams and not all of them with ingest pipelines, there is an scenario not covered. As an example:

- Package containing 2 data streams
- Both data streams have an ingest pipeline defined
- Just one data stream has pipeline tests defined for the ingest pipelines.

In this case, the coverage report will just show the coverage report for the data stream
with tests, but it will not add any reference for the second data stream that has no tests. It should also add a coverage reference adding a Missing (not covered) method/line.

[...]

PR #1915, opened 2024-06-18
Add best-effort coverage reports per test type

jsoriano:
Ensure that each executed test type marks some content as covered, so we can have an idea of the files covered by tests, and the test types executed.

Currently we report if a test type has been executed by marking a line in a manifest as covered, but this fails if the manifest doesn't have enough lines. This change attempts to solve this kind of issues.

At the moment the following files are fully or partially marked as covered if an specific type of test is executed:
* Pipeline tests:
  - [x] No changes in this PR, we have coverage based on ES stats.
* Asset tests:
  - [x] All files under the kibana directory.
* Policy tests:
  - [x] All manifests.
  - [x] All template files in the executed data stream.
* Static tests
  - [x] Sample events.
  - [x] ~~Benchmark configs.~~ No need to cover dev files.
* System tests
  - [x] All manifests.
  - [x] Fields files.

The rest of non-development files are marked as not covered.

Fixes https://github.com/elastic/elastic-package/issues/1624.


jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

cmd/check.go
```diff
@@ -24,7 +24,6 @@ func setupCheckCommand() *cobraext.Command {
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			err := cobraext.ComposeCommands(args,
-				setupFormatCommand(),
```
jsoriano, line 27: We can make this change in a different PR: https://github.com/elastic/elastic-package/pull/1936

jsoriano:
Ok, it is working now. I will open for review. There is a working report in https://sonar.elastic.dev/component_measures?metric=new_conditions_to_cover&view=list&pullRequest=10273&id=elastic_integrations_AYu5LCaKQZlFqhqWIrk_

Perhaps we can review the files added to the coverage and we will need to review the list of files ignored by sonarqube in the integrations repository, the latter will be done in a different PR.

internal/testrunner/coveragereport.go
```diff
@@ -33,20 +31,6 @@ func CoverageFormatsList() []string {
...
-		"system":   3,
-		"static":   4,
-	}
-	lineNumber, ok := lineNumberPerTestType[testType]
-	if !ok {
-		lineNumber = 5
-	}
-	return lineNumber
-}
-
```
mrodm, line 49: :fire:

internal/testrunner/coveragereport_test.go
```diff
@@ -212,49 +79,14 @@ func TestCreateCoverageReport(t *testing.T) {
...
-				TestType:  "Coverage for asset test",
-				Timestamp: 10,
+				TestType: "Coverage for system test",
 			},
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			report, err := createCoverageReport(tt.rootPath, tt.packageName, tt.packageType, tt.testType, tt.results, tt.coverageFormat, tt.timestamp)
```
mrodm, line 257: Is timestamp not mandatory for the coverage reports using Cobertura?

internal/testrunner/coverage.go
```diff
@@ -0,0 +1,229 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/elastic/elastic-package/internal/files"
+)
+
+// GenerateBasePackageCoverageReport generates a coverage report where all files under the root path are
+// marked as not covered. It ignores files under _dev directories.
+func GenerateBasePackageCoverageReport(pkgName, rootPath, format string) (CoverageReport, error) {
```
mrodm, line 23: Could this function solved also this issue for pipeline tests ? 
https://github.com/elastic/elastic-package/issues/1624

If all files present in the package are marked as uncovered, there will be coverage (zero) for those files without tests.

Just one concern regarding also about pipelines, if there is a coverage report created by pipeline in some package. With this change there would be lines marked as uncovered (that previously there was no information), that could cause a decrease in the coverage (%). Would that be an issue?

internal/testrunner/coveragereport_test.go
jsoriano: Maybe, but there is no coverage report generated by this method now, it uses the coverage in the results, and then the timestamp is not used anymore, this is why I removed the parameter.

internal/testrunner/testrunner.go
```diff
@@ -145,6 +145,23 @@ func NewResultComposer(tr TestResult) *ResultComposer {
...
+
+// CoveragePackageName returns a package name that can be used in coverage reports, based on information
+// in the composer.
+func (rc *ResultComposer) CoveragePackageName() string {
+	if rc.DataStream != "" {
+		return rc.Package + "." + rc.DataStream
+	}
+
+	return rc.Package
+}
```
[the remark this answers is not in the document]
jsoriano: Done in 51a85a356b7804f41ebdd6168e9f4422c14c21be.

internal/testrunner/coverage.go
jsoriano: > Could this function solved also this issue for pipeline tests ?
> #1624

Probably yes. Tests will generate coverage for all files now, even if it is empty.

> If all files present in the package are marked as uncovered, there will be coverage (zero) for those files without tests.

Correct, if there are no tests, there is no coverage.

> Just one concern regarding also about pipelines, if there is a coverage report created by pipeline in some package. With this change there would be lines marked as uncovered (that previously there was no information), that could cause a decrease in the coverage (%). Would that be an issue?

Yes, there can be a decrease in coverage in several cases. I wouldn't consider this as an issue. Before there was more coverage % just because we were not reporting coverage for many files. We will probably revisit what files we mark as covered or uncovered, and how.

internal/testrunner/runners/pipeline/tester.go
```diff
@@ -300,74 +300,67 @@ func (r *tester) checkElasticsearchLogs(ctx context.Context, startTesting time.T
...
-	if e, ok := err.(testrunner.ErrTestCaseFailed); ok {
-		tr.FailureMsg = e.Error()
-		tr.FailureDetails = e.Details
-		return tr, nil
-	}
 	if err != nil {
-		err := fmt.Errorf("verifying test result failed: %w", err)
-		tr.ErrorMsg = err.Error()
-		return tr, nil
+		results, _ := rc.WithErrorf("verifying test result failed: %w", err)
```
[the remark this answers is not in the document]
jsoriano: This is interesting, because what we are doing now is quite similar, `WithErrorf` includes a similar logic, that should be setting the failure fields:
```
        var tcf *ErrTestCaseFailed
        if errors.As(err, &tcf) {
                rc.FailureMsg += tcf.Reason
                rc.FailureDetails += tcf.Details
                return []TestResult{rc.TestResult}, nil
        }
```
Taking a look...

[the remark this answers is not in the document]
jsoriano: Condition was incorrect, the variable doesn't have to be a pointer. Apart of that I adjusted when the prefix `test case failed:` is expected.

PR #1909, opened 2024-06-13
Run system tests in parallel

mrodm:
Depends on https://github.com/elastic/package-spec/pull/759
Part of #787 

Run system tests in parallel:

- Created new methods to trigger in parallel (using routines)
- Added new global test configuration files to define whether or not a package can be run in parallel, or all tests of a given test type should be skipped.
    - By default, **tests are run sequentially**
- Added environment variables to define independent agents and how many routines can be triggered in parallel
    - Updated `nginx` and `apache` packages to run system tests in parallel.
- ~Reordered when the `shutdownAgentHandler` is set, just after setting up the Agent~
    - Required to keep the same order while agents could be created in `servicedeployer` package (custom agents or k8s agents).
- ~Changed how shutdown service handler is set, even if there is an error~  Reverted

### Checklist
- [x] New version of package-spec containing https://github.com/elastic/package-spec/pull/759 - PR for release https://github.com/elastic/package-spec/pull/764

### How to test

1. Add in a package a global test configuration file
```yaml
system:
  parallel: true
```
2. Run system tests 
```shell
elastic-package stack up -v -d

cd test/packages/parallel/nginx/

ELASTIC_PACKAGE_MAXIMUM_NUMBER_PARALLEL_TESTS=3 ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT=true elastic-package test system -v

cd test/packages/parallel/apache/

ELASTIC_PACKAGE_MAXIMUM_NUMBER_PARALLEL_TESTS=3 ELASTIC_PACKAGE_TEST_ENABLE_INDEPENDENT_AGENT=true elastic-package test system -v


elastic-package stack down -v
```

PR #1919, opened 2024-06-19
Create different deployer folders per each test

mrodm:
Follows #1909 
Part of #787 

This PR creates a different deployer folder for each test. This will ensure that if at some point those kind of tests could be parallelized, they would not have issues creating all the resources in the same folder.

Currently, all resources are created in the same folder and they are not cleaned up after finishing the tests:
- For packages using Terraform: `~/.elastic-package/deployer/terraform`
- For packages using custom agents: `~/.elastic-package/deployer/docker_custom_agent`

For instance:
```
/home/user/.elastic-package/deployer/
├── docker_custom_agent
│   └── docker-custom-agent-base.yml

```
This PR creates a new subfolder under those folders to create the resources needed for each test in different directories. For instance:
```
/home/user/.elastic-package/deployer/
├── docker_custom_agent
│   └── oracle-34417
│       └── docker-custom-agent-base.yml
```


internal/servicedeployer/custom_agent.go
```diff
@@ -94,7 +94,7 @@ func (d *CustomAgentDeployer) SetUp(ctx context.Context, svcInfo ServiceInfo) (D
 		fmt.Sprintf("%s=%s", fleetPolicyEnv, d.policyName),
 	)
 
-	configDir, err := d.installDockerfile()
+	configDir, err := d.installDockerfile(deployerFolderName(svcInfo))
```
jsoriano, line 97: I wonder if we could use go-resource to manage these files.

internal/servicedeployer/custom_agent.go
```diff
@@ -197,13 +198,13 @@ func (d *CustomAgentDeployer) SetUp(ctx context.Context, svcInfo ServiceInfo) (D
...
 // the directory with these files.
-func (d *CustomAgentDeployer) installDockerfile() (string, error) {
+func (d *CustomAgentDeployer) installDockerfile(folder string) (string, error) {
 	locationManager, err := locations.NewLocationManager()
 	if err != nil {
 		return "", fmt.Errorf("failed to find the configuration directory: %w", err)
 	}
 
-	customAgentDir := filepath.Join(locationManager.DeployerDir(), dockerCustomAgentDir)
+	customAgentDir := filepath.Join(locationManager.DeployerDir(), dockerCustomAgentDir, folder)
```
jsoriano, line 207: Would it make sense to use tmp dirs for these files? Though probably we want them in more persistent places for `test system --setup`?

jsoriano (approved):

internal/servicedeployer/custom_agent.go
mrodm: Yes, I would prefer to keep them in these known/persistent locations.

internal/common/helpers.go
```diff
@@ -58,3 +60,13 @@ func ToStringSlice(val interface{}) ([]string, error) {
...
+
+func ProcessResourceApplyResults(results resource.ApplyResults) string {
+	var errors []string
+	for _, result := range results {
+		if err := result.Err(); err != nil {
+			errors = append(errors, err.Error())
+		}
+	}
+	return strings.Join(errors, ", ")
+}
```
mrodm, line 72: Moved to the helpers since it is going to be used in both agentdeployer and servicedeployer packages.

internal/servicedeployer/custom_agent.go
mrodm: Sure, updated `custom_agent` to create the required files using go-resource.

Terraform (servicedeployer) was already using `go-resource` to create the required configuration files or scripts.

internal/agentdeployer/agent.go
```diff
@@ -74,9 +75,10 @@ var _ AgentDeployer = new(DockerComposeAgentDeployer)
 type dockerComposeDeployedAgent struct {
 	agentInfo AgentInfo
 
-	ymlPaths []string
-	project  string
-	env      []string
+	ymlPaths  []string
+	project   string
+	env       []string
+	configDir string
```
mrodm, line 81: Moved from agentInfo struct to be just defined here in the deployed agent.

Currently, this variable is not used outside this struct `dockerComposeDeployedAgent`, so it is not needed to keep it afterwards

jsoriano (approved):

PR #1924, opened 2024-06-20
Refactor PTTY handling in compose commands

jsoriano:
Use the helper [`pty.Start()`](https://pkg.go.dev/github.com/creack/pty/v2#Start) instead of handling ourselves the tty and associated ptty. Wait for the finalization of the command just with the copy, to avoid a separate goroutine in an attempt to avoid the hangs we are seeing in CI.

Downgrade to pty 1.1.19, that seems to solve some kind of race condition https://github.com/creack/pty/issues/196

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/boumenot/gocover-cobertura v1.2.0
 	github.com/cbroglie/mustache v1.4.0
 	github.com/cespare/xxhash/v2 v2.3.0
-	github.com/creack/pty v1.1.21
+	github.com/creack/pty v1.1.19
 	github.com/dustin/go-humanize v1.0.1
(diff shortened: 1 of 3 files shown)


jsoriano:
test integrations

mrodm:
test integrations

mrodm (approved):
Nice !! 

Just one question , will dependabot try to update this version again to the version that has those issues?

jsoriano:
> will dependabot try to update this version again to the version that has those issues?

Yes, I think so. We will have to close the PR so the version is ignored, and in future versions review if the issue has been fixed.

jsoriano:
This should be enough: https://github.com/elastic/elastic-package/pull/1931#issuecomment-2191239223

PR #1936, opened 2024-06-28
Don't run format on check

jsoriano:
Running format on check modifies the working copy in an unexpected moment. This is problematic for example for coverage, as tests run check indirectly, so the resulting coverage may not fit the actual files.


jsoriano:
test integrations

mrodm (approved):
:+1:

PR #1937, opened 2024-06-28
feat: update capabilities to the minimum that agent requires

pkoutsovasilis:
When [this PR](https://github.com/elastic/elastic-agent/pull/4925) gets merged the minimum capabilities required by elastic-agent, when running inside a container, will change. This PR prepares `elastic-package` for such a change

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -421,6 +421,9 @@ func (r *tester) createAgentInfo(policy *kibana.Policy, config *testConfig, runI
 		info.Agent.User = "root"
 	}
 
+	// Set linux capabilities to the minimum set required by the agent
+	info.Agent.LinuxCapabilities = []string{"CHOWN", "SETFCAP", "SETPCAP"}
+
 	return info, nil
 }
 
--- a/internal/profile/testdata/migration-home-legacy/profiles/default/stack/snapshot.yml
+++ b/internal/profile/testdata/migration-home-legacy/profiles/default/stack/snapshot.yml
@@ -136,6 +136,10 @@ services:
     env_file: "./elastic-agent.env"
     cap_drop:
     - ALL
+    cap_add:
+    - CHOWN
+    - SETPCAP
+    - SETFCAP
     ports: []
     volumes:
     - "../certs/ca-cert.pem:/etc/ssl/certs/elastic-package.pem"
--- a/internal/servicedeployer/_static/docker-custom-agent-base.yml
+++ b/internal/servicedeployer/_static/docker-custom-agent-base.yml
@@ -8,6 +8,10 @@ services:
     hostname: docker-custom-agent
     cap_drop:
       - ALL
+    cap_add:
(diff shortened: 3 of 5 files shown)


mrodm:
Checking that the PR mentioned in the description is already merged, should this PR be open for review @pkoutsovasilis ?

internal/testrunner/runners/system/tester.go
```diff
@@ -421,6 +421,9 @@ func (r *tester) createAgentInfo(policy *kibana.Policy, config *testConfig, runI
 		info.Agent.User = "root"
 	}
 
+	// Set linux capabilities to the minimum set required by the agent
+	info.Agent.LinuxCapabilities = []string{"CHOWN", "SETFCAP", "SETPCAP"}
```
mrodm, line 425: This would overwrite the capabilities defined in the configuration file if any.

The settings from the config file are copied into the `info` variable here:
https://github.com/elastic/elastic-package/pull/1937/files#diff-0bdd696fb9abd5e4aa12a3bab634cbac31ec1a1349b3a61e8a784a6bdeb6e7d8R416

It would needed to merge both arrays/slices to not lose the values from the configuration file.

pkoutsovasilis:
Hi @mrodm 👋 thanks for catching this, this PR is no longer relevant as elastic-agent is now handling raising capabilities for non-root users in a non-mandating way. thx

mrodm:
> Hi @mrodm 👋 thanks for catching this, this PR is no longer relevant as elastic-agent is now handling raising capabilities for non-root users in a non-mandating way. thx

Ok @pkoutsovasilis !

Just one question, I was looking into this PR since in the elastic-agent logs for 8.16.0-SNAPSHOT stack appear these messages:
```
elastic-agent-1  | agent container initialisation - effective capabilities
elastic-agent-1  | agent container initialisation - ambient capabilities
elastic-agent-1  | Warning: cannot chown agent paths without CAP_CHOWN or CAP_DAC_OVERRIDE capabilities
elastic-agent-1  | agent container initialisation - chown paths
```

Would it be needed to add some capability because of that warning ?

pkoutsovasilis:
> Just one question, I was looking into this PR since in the elastic-agent logs for 8.16.0-SNAPSHOT stack appear these messages:

So from 8.16.0 elastic-agent is trying to chown all agent-related data to the uid:gid that the process is running under. 

This is done to guarantee access to it's state when this is a host path mount (useful in k8s daemonset deployment scenarios) and to make sure that any beat config files are matching the uid of the process resolving issues like [this one](https://github.com/elastic/elastic-agent/issues/1540#issuecomment-2217737916). To this end, this warning essentially says that since CAP_CHOWN capability is not granted elastic-agent won't perform this step.

Now all that said, since elastic-agent, in the context of elastic-package, doesn't require any host folder mounted as agent state and it doesn't run into config ownership issues I would say that you can ignore this warning message. Does this help @mrodm ?

mrodm:
Thanks for this explanation! That really helps :+1:

Issue #1842, opened 2024-05-17
Test LogsDB early with integrations


ruflin:
The Elasticsearch team is working on the `logsdb` mode for data streams which will provide an optimised way to store logs. To provide the Elasticsearch team with early feedback if this causes any issues with integrations, we should find a way to enable this on integrations at least for test runs. More details on logsdb here: https://github.com/elastic/elasticsearch/issues/106489

Like TSDS, logsdb is an index mode setting on the data stream. How can we inject this setting for a test run on all integrations without having to change each integration? At first, enabling logsdb might reject documents so this goes well together with the testing with failure store enabled: https://github.com/elastic/elastic-package/issues/1832

An alternative way of approaching this is using Fleet / EPM to enable the flag for all logs integrations @kpollich . For example a feature flag in Kibana. Last, it could also be a feature flag in Elasticsearch that enables it. The best place to enable it would be `logs@settings`. This would not only enable it for all integrations but also for all data going into logs-*-* which seems ideal for broader testing.

Below a screenshot on all the templates that an integration imports.

![Screenshot 2024-05-17 at 08 39 40](https://github.com/elastic/elastic-package/assets/244900/8f74bfbf-2c9d-4886-9472-15520be99c42)

@salvatore-campagna Having written down the above option, to me the easiest way to test LogsDB seems to be having a feature flag / setting in Elasticsearch that changes `logs@settings` and adds the logsdb index mode. It is already possible today to start up elastic-package with specific settings for Elasticsearch so no modifications to elastic-package would be needed. In addition, it means we have a simple way to also test it in our test clusters.

## Links

* Setting template in code: https://github.com/elastic/elasticsearch/blob/088f051a893dd9d9742e6023d1f62a3e28293dad/x-pack/plugin/core/template-resources/src/main/resources/logs%40settings.json#L4

ruflin:
https://github.com/elastic/elasticsearch/issues/108762 was just filed which should bring a feature flag to Elasticsearch to change the setting. This means, to test integrations, the only thing needed would be to pass in a config option on startup to Elasticsearch. @jsoriano Am I missing something?

jsoriano:
> [elastic/elasticsearch#108762](https://github.com/elastic/elasticsearch/issues/108762) was just filed which should bring a feature flag to Elasticsearch to change the setting. This means, to test integrations, the only thing needed would be to pass in a config option on startup to Elasticsearch. @jsoriano Am I missing something?

[...]

PR #1939, opened 2024-07-02
Add flag to enable logsdb mode in logs data streams

jsoriano:
Add a new profile parameter, `stack.logsdb_enabled`, to enable the feature flag that enables the logs index mode in all logs data streams (https://github.com/elastic/elasticsearch/issues/108762), in stacks that support it.

This can be tested now with the 8.15 snapshots, for example:
```
elastic-package stack up -v -d -U stack.logsdb_enabled=true --version 8.15.0-SNAPSHOT
```

Closes https://github.com/elastic/elastic-package/issues/1842.

### DIFF
--- a/internal/stack/_static/elasticsearch.yml.tmpl
+++ b/internal/stack/_static/elasticsearch.yml.tmpl
@@ -19,7 +19,6 @@ ingest.geoip.downloader.enabled: false
 cluster.logsdb.enabled: true
 {{- end -}}
 
-{{ $version := fact "elasticsearch_version" }}
 {{ if semverLessThan $version "8.0.0" }}
 script.max_compilations_rate: "use-context"
 script.context.template.max_compilations_rate: "unlimited"
--- a/internal/stack/_static/elasticsearch.yml.tmpl
+++ b/internal/stack/_static/elasticsearch.yml.tmpl
@@ -13,6 +13,12 @@ xpack.security.http.ssl.certificate: "certs/cert.pem"
 
 ingest.geoip.downloader.enabled: false
 
+{{- $version := fact "elasticsearch_version" -}}
+{{- $logsdb_enabled := fact "logsdb_enabled" -}}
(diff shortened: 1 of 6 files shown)


ruflin (approved):
I did a manual test and the flag works as expected.

internal/stack/_static/elasticsearch.yml.tmpl
```diff
@@ -13,6 +13,12 @@ xpack.security.http.ssl.certificate: "certs/cert.pem"
 
 ingest.geoip.downloader.enabled: false
 
+{{- $version := fact "elasticsearch_version" -}}
```
ruflin, line 16: Nit: elasticsearch version is also fetched on line 22. I was also wondering about the diff with `-` vs without.

jsoriano: Ah, thanks, I will remove the line below. The `-` is to avoid adding newlines when rendering the template.

mrodm (approved):
:+1:

jsoriano:
As discussed internally, we need to disable validation for normalized fields when logsdb is used, because it is going to use synthetic mode.

In https://github.com/elastic/elastic-package/pull/1939/commits/fd68bfcb5d9f7321d4290f70a17ce9287e66b087 I have refactored our detection of synthetics mode to be based on the simulated index template for the data stream, instead of using the component template provided by the package.

mrodm (approved):
:+1: 
Checked that it reports what it is expected in two different packages:
- (synthetic test package) synthetic enabled: https://buildkite.com/elastic/elastic-package/builds/3541#0190779f-4c8b-4063-8999-f4b5c3ba512d/193-2385
- (nginx test package) synthetic disabled: https://buildkite.com/elastic/elastic-package/builds/3541#0190779f-4cae-43dd-aa64-cdb512a97132/195-855

PR #1946, opened 2024-07-03
Fix detection of synthetic source mode

jsoriano:
When some index modes are used, synthetic source mode is also enabled, even when not explicitly mentioned in the template. Assume that synthetic mode is enabled when these index modes are used.

Fixes issue introduced in https://github.com/elastic/elastic-package/pull/1939.


jsoriano:
test integrations

jsoriano:
test integrations

PR #1948, opened 2024-07-04
Fix error handling when deleting data stream in system tests

jsoriano:
Error returning was being ignored, and error handled was a different one.


jsoriano:
test integrations

mrodm (approved):
Thanks!

PR #1949, opened 2024-07-04
Use different data stream name when simulating index template

jsoriano:
Use a different data stream name when simulating to check if synthetic is enabled. There seems to be some conflict if the same name is reused. Maybe related to https://github.com/elastic/elasticsearch/issues/84256 or some similar issue.

For the matter of testing, this PR also includes these changes:
- https://github.com/elastic/elastic-package/pull/1948
- https://github.com/elastic/elastic-package/pull/1946


internal/testrunner/runners/system/tester.go
```diff
@@ -949,7 +964,7 @@ func (r *tester) prepareScenario(ctx context.Context, config *testConfig, svcInf
 
 	r.cleanTestScenarioHandler = func(ctx context.Context) error {
 		logger.Debugf("Deleting data stream for testing %s", scenario.dataStream)
-		r.deleteDataStream(ctx, scenario.dataStream)
+		err := r.deleteDataStream(ctx, scenario.dataStream)
```
mrodm, line 967: good catch!

mrodm (approved):
Nice! I see that issues in the integrations repo were solved with these changes :+1:

mrodm (approved):

PR #1957, opened 2024-07-05
[CI] Replace exit by continue in loop to check packages - serverless

mrodm:
Previously, that loop was based on sub-shells and it was required to use `exit` to continue to the next loop iterations.

With the latest changes, subshells are not used anymore and it can be used directly `continue`.


mrodm:
buildkite test serverless

jlind23 (approved):

PR #1960, opened 2024-07-09
Add security-rule kibana assets to be checked by tests

mrodm:
Follows #1915 

Include `security-rule` assets into the assets checked by `elastic-package test asset`.

Those kibana assets are reported by the fleet API:
```
GET kbn:/api/fleet/epm/packages/security_detection_engine
```

Example of response (truncated):
```json
{
  "response": {
    ...
   "savedObject": {
      "id": "security_detection_engine",
      "type": "epm-packages",
      "namespaces": [],
      "updated_at": "2024-07-09T14:33:46.715Z",
      "created_at": "2024-07-09T14:33:45.355Z",
      "version": "WzcxMTEsMV0=",
      "attributes": {
        "installed_kibana": [
          {
            "id": "000047bb-b27a-47ec-8b62-ef1a5d2c9e19_106",
            "type": "security-rule"
          },
          ...
        ],
      },
    },
  },
}
```


internal/packages/assets.go
```diff
@@ -159,7 +162,7 @@ func loadElasticsearchAssets(pkgRootPath string) ([]Asset, error) {
 }
 
 func loadFileBasedAssets(kibanaAssetsFolderPath string, assetType AssetType) ([]Asset, error) {
-	assetsFolderPath := filepath.Join(kibanaAssetsFolderPath, string(assetType))
+	assetsFolderPath := filepath.Join(kibanaAssetsFolderPath, strings.ReplaceAll(string(assetType), "-", "_"))
```
mrodm, line 165: Folder in packages is names `security_rule`, but the type name is `security-rule`.

mrodm:
test integrations

internal/packages/assets.go
jsoriano: Nit. Maybe add this as a comment to understand why we do the replace? Or maybe we could make `AssetType` more explicit, with a type that contains the asset type name and the folder name.

jsoriano (approved):

internal/packages/assets.go
```diff
@@ -17,16 +17,35 @@ import (
 // AssetType represents the type of package asset.
 type AssetType string
 
+type assetTypeFolder struct {
+	Type       AssetType
+	folderName string
```
jsoriano, line 22: Nit. Why is one public and the other one private?

jsoriano (approved):

internal/packages/assets.go
mrodm: That is a leftover, I was trying first with another type of struct. But it was needed to updated more code.

Thanks for catching this! I'll rename it to `typeName` (I cannot set `type` since it is a reserved keyword).

Issue #1872, opened 2024-05-17
Fleet: Make subobjects: false the default for all data streams


flash1293:
**Describe the feature:** It's currently possible for data streams to opt into `subobjects: false` for data streams. As this is meant to be the default for logs streams in the future and also a sensible default for metrics, it should be the default for newly introduced data streams in integrations.

Also after releasing Package Spec 3.2.0, `subobjects: false` should be added to the archetypes of `elastic-package create` for data streams, so new packages created by this tool include it.

flash1293:
Thinking this through, we should also set `subobjects: false` on metrics streams in the same way - it's even more common there and all the considerations still apply.

Should I create a separate issue for this or can it be handled as part of this one @jsoriano ?

jsoriano:
> Thinking this through, we should also set `subobjects: false` on metrics streams in the same way - it's even more common there and all the considerations still apply.
> 
> Should I create a separate issue for this or can it be handled as part of this one @jsoriano ?

Lets keep it in the same issue so we define a common plan for both.

flash1293:
Adjusted the description

PR #1965, opened 2024-07-10
Disable subobjects by default on new packages and data streams

jsoriano:
Modify the archetypes used by `elastic-package create` to disable subobjects by default on new packages and data streams.

A new question is added to the survey so users can chose to opt out for this behavior.

Closes https://github.com/elastic/elastic-package/issues/1872.


internal/packages/archetype/_static/package-manifest.yml.tmpl
```diff
@@ -72,6 +72,22 @@ policy_templates:
...
 {{ end }}
 {{ end -}}
+{{ if .Manifest.Elasticsearch }}
+elasticsearch:
+{{ if .Manifest.Elasticsearch.SourceMode }}
+  source_mode: {{ .Manifest.Elasticsearch.SourceMode }}
+{{- end}}
+{{ if .Manifest.Elasticsearch.IndexMode }}
+  index_mode: {{ .Manifest.Elasticsearch.IndexMode }}
+{{- end }}
```
mrodm, line 82: Currently, there are no questions for these values when creating the package. So, right now they would not be written here in this manifest.

Should they be added also as part of the questions?

jsoriano: Not sure, I could add them in a follow up if necessary. Or would you prefer the questions to be added here?

mrodm: Sure, they can be added in a follow up, it's not needed to do it now.

mrodm (approved):
:+1:

Issue #1832, opened 2024-05-15
Fail tests if failure store contains any documents


flash1293:
The failure store is already available as an experimental feature in the 8.15 snapshot build of Elasticsearch.

While it's not ready yet to be enabled as a default feature for integrations, it can be used as an additiona layer of security, making sure no documents are dropped during system tests which would indicate an issue with the integration.

I propose the following steps:
* In stack version >=8.15, always enable the failure store on all data streams
* Similar to https://github.com/elastic/elastic-package/pull/1738, validate that a datastream under test didn't drop any documents in the failure store (fail the test in case it happens)
* Add a configuration option to the test config that allows skipping this check
* Document what the failure store means and how to skip the check

What do you think @ruflin @jsoriano ? It seems like this can be done even before the failure store is ready to be used on customers clusters.

jsoriano:
Yes, I think we could try it in elastic-package even if it is not ready for GA.

ruflin:
++ on enabling it for the tests and check if any documents end up in it.

jsoriano:
@flash1293 @ruflin do you know where can I find documentation about how to use the failure store?

ruflin:
Here is a gist to get you started: https://gist.github.com/jbaiera/cecf6b973c0a15e72e2b189120caba51 Note, some of the APIs still might change.

@jbaiera Are there already some more official docs somewhere?

jsoriano:
@flash1293 @ruflin I have been playing a bit with the failure store and I think that the way it is enabled is a bit problematic for its use while testing packages.

It is enabled by setting `failure_store` to `true` in the index template for the data stream. This setting has to be set inside `data_stream`, like this:
```
PUT _index_template/my-index-template
{
  "index_patterns": ["my-index-template-*"],
  "data_stream": {
    "failure_store": true
  }
}
```

But when using packages, customization of index templates is based on component templates, and component templates don't allow to set `data_stream` (or I haven't managed to do it), so we cannot selectively enable it on specific data streams or packages.

I see some paths forward for this:
* Adding support in Elasticsearch to enable the failure store in existing data streams. For example with something like `POST my-data-stream/_rollover?enable_failure_store=true`. I think this would be quite valuable in testing or debugging scenarios, where the user may not have enabled the failure store beforehand, but wants to look for failures in future ingestions. cc @jbaiera
[...]

PR #1973, opened 2024-07-12
Check failure store when available during system tests

jsoriano:
Enable failure store and check if it has any documents after executing system tests.

Failure store is used if the version of the stack is 8.15.0 or newer. In lower versions there are unexpected failures on tests.

Fixes https://github.com/elastic/elastic-package/issues/1832.


jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

cmd/testrunner.go
```diff
@@ -534,6 +535,15 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
...
 	}
+	versionInfo, err := kibanaClient.Version()
+	if err != nil {
+		return fmt.Errorf("can't get version info from Kibana client: %w", err)
+	}
+	stackVersion, err := semver.NewVersion(versionInfo.Number)
+	if err != nil {
+		return fmt.Errorf("can't parse Kibana version %q: %w", versionInfo.Number, err)
+	}
+	checkFailureStore := !stackVersion.LessThan(semver.MustParse("8.14.0"))
```
jsoriano, line 546: We should not rely only on the version, but actually check that ES supports the failure store.

jsoriano:
test integrations

jsoriano:
Ok, it looks like even when the failure store is available in older versions, it has some issues with time series before 8.15. Reopening for review after enabling failure store starting on 8.15, and pinning the stack version for the related test package.

jsoriano:
test integrations

jlind23 (approved):
LGTM but waiting for @flash1293  review and test anyway

flash1293:
The code looks good to me, however I'm not sure how to test locally anymore - I used to 

`elastic-package test system -v --setup --config-file ./my/config.yml`, then `elastic-package test system -v --no-provision`.

The first command still works and finds the docs in the failure store, however I don't know how to see the test failing locally (as the setup routine is not running the validation bits). Seems like this code was recently restructured.

Not a blocker to merge though, I read through the code and it seems to work exactly like the ignored docs stuff.

jsoriano:
@flash1293 do you mean that it doesn't fail when running with `--no-provision`? It should fail the same as when running directly with `elastic-package test system -v`.

flash1293:
It never finds any hits when running with --no-provision and doesn't terminate

jsoriano:
> It never finds any hits when running with --no-provision and doesn't terminate

Have you tried with `elastic-package test system -v`? Could you share the package you are testing with?

flash1293:
@jsoriano When I'm running just with `test system` it doesn't get to that point because it fails during teardown:
> Error: failed to tear down system runner: there was an apply error: can't remove the package: could not remove package; API status code = 400; response body = {"statusCode":400,"error":"Bad Request","message":"Unable to remove package with existing package policy(s) in use by agent(s)"}

I'm testing with the `test/packages/false_positives/failure_store` package from this PR.

I don't think it's something wrong with this PR, seems like a general thing.

flash1293 (approved):
nvm, I think it was just some stale state in my local stack. Tested again with a fresh stack and it works as expected 👍 

One nit: Maybe we should show more details of the failure than just the error message (e.g. which processor was involved and so on) to make it easier to debug what's happening.

jsoriano:
> Maybe we should show more details of the failure than just the error message (e.g. which processor was involved and so on) to make it easier to debug what's happening.

Thanks @flash1293 for the feedback. I have pushed another commit including more information in the error messages, they include the error type, the processor and the pipelines that caused the issue:
```
--- Test results for package: failure_store - START ---
FAILURE DETAILS:
failure_store/test fail:
[0] illegal_argument_exception: unable to convert [eight] to integer (processor: convert, pipelines: logs-failure_store.test-0.0.1)
[1] illegal_argument_exception: unable to convert [five] to integer (processor: convert, pipelines: logs-failure_store.test-0.0.1)
[2] illegal_argument_exception: unable to convert [seven] to integer (processor: convert, pipelines: logs-failure_store.test-0.0.1)
[3] illegal_argument_exception: unable to convert [six] to integer (processor: convert, pipelines: logs-failure_store.test-0.0.1)
```

Additionally, the stack trace is logged at the debug level when collecting results.

Wdyt? Is it too much?

flash1293:
Love it! Thanks for adding

flash1293 (approved):

jsoriano:
test integrations

PR #1990, opened 2024-07-26
Add services name to containers logged while waiting for healthy

jsoriano:
While investigating issues in the sql_input package I missed being able to see
what was the service that was failing to start, in the previous version the logs
only included the ID, what cannot be correlated with the service when running
in CI.

Take also the opportunity to refactor labels to use a struct instead of a map
for known labels.


mrodm (approved):
:+1:

PR #2009, opened 2024-08-08
updatecli: pass the username to fix the issue with auth

v1v:
### What

Pin version `v2.64.0`
Set `username` to solve the issues with the authentication needed.
Simplify GitHub action to use a folder with all the pipelines
Restructure the folder layout for the updatecli, updatecli.d for pipelines and values.d for configuration files.

### Why

Use the latest version and fix the issue when using a version over 2.62.0

### Tests

See https://github.com/elastic/elastic-package/actions/runs/10299711920/job/28507533818


.github/workflows/bump-elastic-stack-version.yml
```diff
@@ -35,28 +35,10 @@ jobs:
...
-      - name: Update latest testing stack version
-        # --experimental needed for commitusingapi option.
-        run: updatecli --experimental ${{ env.UPDATECLI_ACTION }} --config .github/workflows/updatecli.d/bump-latest-snapshot-version.yml --values .github/workflows/updatecli.d/scm.yml
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Update Package Registry version
-        # --experimental needed for commitusingapi option.
-        run: updatecli --experimental ${{ env.UPDATECLI_ACTION }} --config .github/workflows/updatecli.d/bump-package-registry-version.yml --values .github/workflows/updatecli.d/scm.yml
+        run: updatecli --experimental ${{ env.UPDATECLI_ACTION }} --config .github/workflows/updatecli --values .github/workflows/updatecli/values.d/scm.yml
```
jsoriano, line 42: Nice, does it work the same?

.github/workflows/bump-elastic-stack-version.yml
```diff
@@ -35,28 +35,10 @@ jobs:
           echo "UPDATECLI_ACTION=apply" >> $GITHUB_ENV
 
       - name: Install Updatecli in the runner
-        uses: updatecli/updatecli-action@v2.62.0
+        uses: updatecli/updatecli-action@3a8785d88ec4fa03d86521a181f37c0e74627463 #v2.64.0
```
jsoriano, line 38: I prefer the use of tags, they are easier to find in the releases. Is there any advantage or reason to use the commit hash instead?
```suggestion
        uses: updatecli/updatecli-action@v2.64.0
```

Or dependabot also adds the comment when updating the action?

jsoriano (commented):
Nice, thanks! Could we also add `github/workflows/values.d` to the list of `pull_request` trigger paths in https://github.com/elastic/elastic-package/blob/2d64d0b649ba96c2c18f89d471677eb624a4ac3a/.github/workflows/bump-elastic-stack-version.yml#L8?

.github/workflows/bump-elastic-stack-version.yml
v1v: >  Is there any advantage or reason to use the commit hash instead?

Tags are mutable and full sha commits are considered the way to do it as explained [here](https://docs.github.com/en/enterprise-cloud@latest/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

>> Pin actions to a full length commit SHA
Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

> Or dependabot also adds the comment when updating the action?

Dependabot does it , see https://github.com/elastic/oblt-actions/pull/105/files

.github/workflows/bump-elastic-stack-version.yml
v1v: it does run all the defined pipelines in the folder; see:

- https://github.com/elastic/elastic-package/actions/runs/10299711920/job/28507533818#step:6:142

.github/workflows/bump-elastic-stack-version.yml
jsoriano: Perfect, thanks!

jsoriano (approved):

jsoriano:
Applying these changes to the integrations repo in https://github.com/elastic/integrations/pull/10739

Issue #1811, opened 2024-05-02
Add negative tests for detection of issues in elastic agent logs


jsoriano:
`elastic-agent` supports finding issues in Elastic Agent logs. We have had some regressions and fixes around there, that could have been detected earlier by negative tests.

We should find a test case that can be used for negative tests. One challenge for this is that these issues use to be related to specific versions of the stack, and we always test now with the default version of the stack.

For example we know that prometheus package had issues in 8.12.0, but these issues were caused by a regression in the stack.

mrodm:
This should be reproduced with the package `prometheus` from the integrations repository using Elastic stack 8.12.0 with this changeset : a2c55c4cbf752e0490

```
cd /path/integrations/repo
cd packages/prometheus
git co a2c55c4cbf752e0490

elastic-package stack up -v -d --version 8.12.0
elastic-package test system -v
elastic-package stack down -v
```

PR #2017, opened 2024-08-13
Add expected errors for benchmark test package [false positive test] 

mrodm:
Follows #1718 

Force to fail `rally_benchmark_malformed_template` test package by updating the `template.ndjson` file as suggested in the README.md:
https://github.com/elastic/elastic-package/blob/f0272ca7cb429d6d10c7ac40577598a1d6ba0eb7/test/packages/false_positives/rally_benchmark_malformed_template/docs/README.md?plain=1#L3

This PR also adds the corresponding expected errors.


scripts/test-check-false-positives.sh
```diff
@@ -23,8 +23,9 @@ function check_expected_errors() {
   local package_root=$1
   local package_name=""
   package_name=$(basename "$1")
+  package_name_manifest=$(cat "$package_root/manifest.yml" | yq -r '.name')
   local expected_errors_file="${package_root%/}.expected_errors"
-  local result_tests="build/test-results/${package_name}-*.xml"
+  local result_tests="build/test-results/${package_name_manifest}-*.xml"
```
mrodm, line 28: Test results file names are using the package name defined in the manifest. Example:
`build/test-results/rally_benchmarks-static-1723556393923280393.xml`

test/packages/false_positives/rally_benchmark_malformed_template/_dev/benchmark/rally/logs-benchmark/template.ndjson
```diff
@@ -1,7 +1,7 @@
 {{- $timestamp := generate "timestamp" -}}
 {
 "@timestamp": "{{ $timestamp.Format "2006-01-02T15:04:05.999999Z07:00" }}",
-"data_stream.type": "logs",
+data_stream.type": "logs",
```
mrodm, line 4: As it was this file, `elastic-package test static` did not detect any failure.

With this change, I tried to force a failure here (as mentioned in the README.md of the package).

test/packages/false_positives/failure_store.stack_version
mrodm, line 1: Since #2012 , default version in elastic-package is 8.15.0. I think this file can be removed now.

I would keep the logic to add the custom Elastic stack version, in case it is needed for another test package. Probably here it can be used: https://github.com/elastic/elastic-package/issues/1811

jsoriano (approved):

jsoriano (approved):

Issue #539, opened 2021-10-12
[META] Support ARM in developer pipeline


jlind23:
Purpose of this Meta issue is to track all issues related to ARM support for all of our developer pipelines.

Linked issues:
- https://github.com/elastic/package-registry/issues/723
- https://github.com/elastic/elastic-package/issues/497
- https://github.com/elastic/integrations/issues/4456

mrodm:
After being merged the PRs mentioned above, now there is some testing in Macos ARM:
- elastic-package and package-registry run unit tests on Macos ARM.
- elastic-package tests also a specific command `elastic-package build`.
    - Not able to run docker/docker-compose in CI in Macos ARM VMs for now.

Checking both issues related in the description, they look related to not be available a docker distribution for architecture ARM. This support was added in https://github.com/elastic/package-registry/pull/1162

As a next steps for this issue:
- I would say both issues could be closed since now all images run by `elastic-package` support multi-architecture.
- It could be kept this issue open until there is support to run/install docker and run more tests in `elastic-package` CI pipelines.

WDYT @jlind23 @jsoriano ?

jsoriano:
Agree on next steps, thanks!

mrodm:
Closed both issues https://github.com/elastic/elastic-package/issues/497 https://github.com/elastic/package-registry/issues/723 since Package Registry docker image already supports multi-platform images (linux/amd64 and linux/arm64).

mrodm:
As mentioned in https://github.com/elastic/elastic-package/issues/539#issuecomment-2326551469 , this issue is kept open until there is support to run/install docker in our CI pipelines for Macos ARM cc @kpollich

PR #2030, opened 2024-08-21
[CI] Add some more steps to test on Macos ARM

mrodm:
Relates https://github.com/elastic/elastic-package/issues/539

This PR adds a new step in the CI pipeline to test to build packages using `elastic-package build` command in MacOS ARM VMs. Doing that allows us to check that the at least can be executed this command in that platform and architecture.

For the time being, it is not available docker/docker-compose commands in those MacOS ARM VMs.

### DIFF
--- a/.buildkite/scripts/install_deps.sh
+++ b/.buildkite/scripts/install_deps.sh
@@ -35,10 +35,6 @@ add_bin_path(){
 }
 
 with_docker() {
-    if [[ "${platform_type_lowercase}" == "darwin" ]]; then
-        echo "Skip. Not supported in Darwin."
-        return
-    fi
     if ! command -v docker &> /dev/null ; then
         echo "Skip. Docker is not installed by default"
         return
@@ -79,15 +75,6 @@ with_docker() {
 }
 
 with_docker_compose_plugin() {
-    if [[ "${platform_type_lowercase}" == "darwin" ]]; then
-        echo "Skip. Not supported in Darwin."
-        return
-    fi
-    if ! command -v docker &> /dev/null ; then
-        echo "Skip. Docker is not installed by default"
-        return
-    fi
-
     if [[ "${DOCKER_COMPOSE_VERSION:-"false"}" == "false" ]]; then
         echo "Skip docker compose installation (plugin)"
         return
--- a/.buildkite/scripts/integration_tests.sh
+++ b/.buildkite/scripts/integration_tests.sh
@@ -31,6 +31,8 @@ PARALLEL_TARGET="test-check-packages-parallel"
 FALSE_POSITIVES_TARGET="test-check-packages-false-positives"
 KIND_TARGET="test-check-packages-with-kind"
 SYSTEM_TEST_FLAGS_TARGET="test-system-test-flags"
+TEST_JUST_BUILD_ZIP_TARGET="test-just-build-zip"
+
 GOOGLE_CREDENTIALS_FILENAME="google-cloud-credentials.json"
 
 REPO_NAME=$(repo_name "${BUILDKITE_REPO}")
@@ -80,11 +82,14 @@ if [[ "$SERVERLESS" == "false" ]]; then
     echo "--- install go"
     with_go
 
-    echo "--- install docker"
-    with_docker
+    if [[ "${TARGET}" != "${TEST_JUST_BUILD_ZIP_TARGET}" ]]; then
+        # Not supported in Macos ARM
+        echo "--- install docker"
+        with_docker
 
-    echo "--- install docker-compose plugin"
-    with_docker_compose_plugin
+        echo "--- install docker-compose plugin"
+        with_docker_compose_plugin
+    fi
 fi
 
 if [[ "${TARGET}" == "${FALSE_POSITIVES_TARGET}" ]]; then


.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -139,6 +137,17 @@ echo "        artifact_paths:"
 echo "          - build/elastic-stack-dump/build-zip/logs/*.log"
 echo "          - build/packages/*.sig"
 
+# TODO: Missing docker & docker-compose in MACOS ARM agent image, skip installation of packages in the meantime.
+# If docker and docker-compose are available for this platform/architecture, it could be added a step to test the stack commands (or even replace this one).
+echo "      - label: \":macos: :go: Integration test: build-zip\""
+echo "        command: ./.buildkite/scripts/integration_tests.sh -t test-build-zip-skip-install"
```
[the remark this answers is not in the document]
jsoriano: Sounds good, yes.

scripts/test-build-zip.sh
```diff
@@ -46,6 +77,10 @@ done
 # Remove unzipped built packages, leave .zip files
 rm -r build/packages/*/
 
+if [[ "${SKIP_INSTALL}" == 1 ]]; then
+    exit 0
+fi
```
jsoriano, line 82: Nit. Instead of having this if and the ones in `with_docker` and `with_docker_compose`, maybe we can add a new script that just builds the zip and doesn't call to `with_docker` and `with_docker_compose`. Even if we have some duplication between the scripts, it would be more explicit. WDYT?

jsoriano (approved):
LGTM, but added suggestion about having a test script that just builds the zip.

scripts/test-build-zip.sh
mrodm: Sure, I'll do that. It is going to be better to have that script with less "ifs"

.buildkite/scripts/install_deps.sh
```diff
@@ -35,6 +35,10 @@ add_bin_path(){
 }
 
 with_docker() {
+    if [[ "${platform_type_lowercase}" == "darwin" ]]; then
+        echo "Skip. Not supported in Darwin."
+        return
+    fi
```
jsoriano, line 41: This is not needed now, right?

.buildkite/scripts/install_deps.sh
```diff
@@ -75,6 +79,15 @@ with_docker() {
...
 
 with_docker_compose_plugin() {
+    if [[ "${platform_type_lowercase}" == "darwin" ]]; then
+        echo "Skip. Not supported in Darwin."
+        return
+    fi
+    if ! command -v docker &> /dev/null ; then
+        echo "Skip. Docker is not installed by default"
+        return
+    fi
```
jsoriano, line 89: Or this.

.buildkite/scripts/install_deps.sh
mrodm: It is still needed. These function (with_docker and with_docker_compose) are called from the script `.buildkite/scripts/integrations_tests.sh`. This script is used for all the steps that could run with Linux containers or Linux VMs, so it is needed to keep these safeguards/checks.

https://github.com/elastic/elastic-package/blob/10509f6bcb3a22a2c57a353e97ea1bfa46a9608e/.buildkite/scripts/integration_tests.sh#L83-L88

.buildkite/scripts/install_deps.sh
mrodm: As mentioned here, I would keep this condition too.
In this case, if there is no `docker` there is no need to install docker-compose.

.buildkite/scripts/install_deps.sh
jsoriano: Maybe I am missing something, but do we want these scripts to continue if docker installation cannot be performed?

> This script is used for all the steps that could run with Linux containers or Linux VMs, so it is needed to keep these safeguards/checks.

Is it run in darwin?

jsoriano: > > This script is used for all the steps that could run with Linux containers or Linux VMs, so it is needed to keep these safeguards/checks.
> 
> Is it run in darwin?

Ok, I see it is run with `-t test-just-build-zip`, could we avoid calling `with_docker` and `with_docker_compose_plugin` when the target is `test-just-build-zip`? The same way as we are avoiding it when running for serverless.

mrodm: Exactly, for now just Macos ARM is used for that specific make target.
It could be added as you mentioned that condition to not call those functions if that target is the one to be executed.

I'll give it a try :+1:

mrodm: Applied the suggestion here https://github.com/elastic/elastic-package/pull/2030/commits/2e7220d9bdb73d6b8f403875ceed6549e09a1efd

Makefile
```diff
@@ -109,6 +109,9 @@ test-check-packages-with-custom-agent:
 test-build-zip:
 	./scripts/test-build-zip.sh
```
jsoriano, line 110: What does this apart of just building the zip? :sweat_smile:

jsoriano (approved):

Makefile
mrodm: This script builds the packages and then installs them (without the zip file). It would be briefly like:
```shell

# for each package
elastic-package build -v --zip --sign 
elastic-package install -v

elastic-package stack down -v
```

Then, there is another script that builds the packages and installs them using the zip file (`scripts/test-install-zip.sh`):
```shell

# for each package
elastic-package build -v --zip --sign 
elastic-package install -v --zip <zip_file_path>
# check via curl that it is installed

elastic-package stack down -v
```

And there is another script file, that just builds the zip file. Used specially for testing in Macos ARM:
```shell

# for each package
elastic-package build -v --zip --sign 

elastic-package stack down -v
```

Probably, they could be renamed so it is clear what they do. What about rename them as this? @jsoriano

- `scripts/test-build.zip.sh`: `scripts/test-build-install.sh`
    - Package build and installed from the working copy 
- `scripts/test-install.zip.sh`: `scripts/test-build-install-zip-file.sh`
    - Package built and installed using the zip file
- `scripts/test-just-build-zip.sh`: `scripts/test-build.zip.sh`
    - Package built

If so, I could create a follow-up PR to rename those files as well as Makefile and buildkite scripts accordingly.

mrodm: Created PR to rename test scripts: https://github.com/elastic/elastic-package/pull/2072

PR #2033, opened 2024-08-22
Update cleanup to remove service logs from independent Elastic Agents

mrodm:
Update `elastic-package cleanup` to remove also service logs folders created for independent Elastic Agents.

Tested with `nginx` (from test package) and `sql_input` (from integrations repository).

## How to test locally

```shell

cd test/packages/parallel/nginx

# run tests and press Ctrl-C once the process is waiting for elastic-agent containers to be healthy

# check that some folders have been created in the profile default folder:
ls -l ~/.elastic-package/profiles/default/service_logs

# for instance
# drwxr-xr-x 2 user user 4096 ago 28 12:40 agent-nginx-access-85847
# drwxr-xr-x 2 user user 4096 ago 28 12:40 agent-nginx-error-59191
# drwxr-xr-x 2 user user 4096 ago 28 12:40 agent-nginx-stubstatus-91770

# run cleanup command
elastic-package clean -v

# the previous folders in the service logs dir should be deleted
ls -l ~/.elastic-package/profiles/default/service_logs
```

### DIFF
--- a/internal/cleanup/service_logs.go
+++ b/internal/cleanup/service_logs.go
@@ -44,11 +44,9 @@ func ServiceLogsIndependentAgents(profile *profile.Profile) (string, error) {
 		return "", fmt.Errorf("locating package root failed: %w", err)
 	}
 
-	packageFolder := filepath.Base(packageRootPath)
-	serviceLogsDir := agentdeployer.ServiceLogsDir(profile)
+	serviceLogDirGlob := agentdeployer.ServiceLogsDirGlobPackage(profile, packageRootPath)
 
-	folderGlob := fmt.Sprintf("%s/agent-%s-*", serviceLogsDir, packageFolder)
-	folders, err := filepath.Glob(folderGlob)
+	folders, err := filepath.Glob(serviceLogDirGlob)
 	if err != nil {
 		return "", fmt.Errorf("pattern malformed: %w", err)
 	}
@@ -59,5 +57,5 @@ func ServiceLogsIndependentAgents(profile *profile.Profile) (string, error) {
 		}
 	}
 
-	return folderGlob, nil
+	return serviceLogDirGlob, nil
 }
--- a/internal/cleanup/service_logs.go
+++ b/internal/cleanup/service_logs.go
@@ -6,6 +6,7 @@ package cleanup
 
 import (
 	"fmt"
+	"os"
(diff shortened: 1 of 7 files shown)


internal/cleanup/service_logs.go
```diff
@@ -26,5 +31,33 @@ func ServiceLogs() (string, error) {
...
+
+	packageRootPath, err := packages.MustFindPackageRoot()
+	if err != nil {
+		return "", fmt.Errorf("locating package root failed: %w", err)
+	}
+
+	packageFolder := filepath.Base(packageRootPath)
+	serviceLogsDir := agentdeployer.ServiceLogsDir(profile)
+
+	folderGlob := fmt.Sprintf("%s/agent-%s-*", serviceLogsDir, packageFolder)
```
mrodm, line 50: As this package removes the built packages of the given package (folder), I tried to remove just the service logs folders for the given package.

internal/cleanup/service_logs.go
```diff
@@ -26,5 +31,33 @@ func ServiceLogs() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("can't remove content (path: %s): %w", locationManager.ServiceLogDir(), err)
 	}
+
 	return locationManager.ServiceLogDir(), nil
 }
+
+// ServiceLogs function removes service logs from temporary directory in the `~/.elastic-package`.
```
jsoriano, line 38:
```suggestion
// ServiceLogsIndependentAgent function removes service logs from temporary directory for independent agents in `~/.elastic-package`.
```

internal/cleanup/service_logs.go
jsoriano: Do we use the name of the directory for these logs directories or the name of the package in the manifest? (There are packages with names different to their directories).
Maybe we could have a function to obtain this directory path, so we use the same when creating and when deleting it.

mrodm: Looking at the code, it was the folder package name what it was used.
But it is true, that this was not clear and easy to know from the code.

I tried to refactor the code a little in https://github.com/elastic/elastic-package/pull/2033/commits/031c2cc2cacab818e46ff2d934727ed4df721aab to help about that.

jsoriano (approved):

Issue #1971, opened 2024-07-12
Incorrect validation of documents with arrays of strings in multi-fields


jsoriano:
In https://github.com/elastic/integrations/pull/10353#issuecomment-2210478947 we found documents that were causing issues about unexpected arrays of objects, but the value is a valid array of strings. An array of strings should be considered as valid if defined as a string type.

The issue seems to happen with multi-fields and synthetic source, what seems to be frequent when logsdb mode is used.

For example:
```
          "user_agent.name.text": [
            "Firefox"
          ],
```
Seems to produce this error:
```
[1] field "user_agent.name.text" is used as array of objects, expected explicit definition with type group or nested
```

Workaround in the meantime
-----
As workaround in the meantime, package developers need to import the definition for these fields from ECS using `external: ecs`.

Tasks
-----

```[tasklist]
### Tasks (see https://github.com/elastic/elastic-package/issues/1971#issuecomment-2275300209)
- [ ] Split the rule in ecs@mappings so the fields without multifield don't get the .text multifield
- [ ] Add a test in ES that loops over all fields defined in ECS to check that ecs@mappings produces equivalent mappings
```

felixbarny:
```
          "user_agent.name.text": [
            "Firefox"
          ],
```

Is that a snippet from  `_source` or  `fields`?

jsoriano:
When synthetic is enabled `elastic-package` gets the fields from `fields`.

felixbarny:
IINM, fields always returns an array, even if the source document was sent with a scalar value.
Do we have to adjust the assertions accordingly when fetching from `fields` rather than `_source`?

Also, maybe we should consistently have the same behavior regardless of whether synthetic _source is used and always fetch from `fields`.

jsoriano:
The error is a bit misleading. It really comes from `user_agent.name.text` not having a definition. I wonder if logsdb is always adding the multifield `text` to keywords?

felixbarny:
No, this isn't coming from LogsDB. I suspect this is coming from the [`ecs@mappings`](https://github.com/elastic/elasticsearch/blob/e5c7cbc3c08cfe76c1487bea7d8004854c11b015/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json#L125) component template.

jsoriano:
Ah yes, could be, another difference on the failing build is that all packages are tested against 8.15.0-SNAPSHOT, and then they use `ecs@mappings`.

[...]

PR #2035, opened 2024-08-23
Add support for multifields added by ecs@mappings

jsoriano:
`ecs@mappings` component template adds multifields that are not defined in ECS or in packages using these mappings. These fields are present in the ingested documents and cannot be validated by `elastic-package`.

Add the definitions for these multifields when `ecs@mappings` is used and no other definition is present.

Also, improve the error reported when there is no definition for a possible multifield.

Closes https://github.com/elastic/elastic-package/issues/1971


internal/stack/resources.go
```diff
@@ -163,7 +163,7 @@ func applyResources(profile *profile.Profile, stackVersion string) error {
 		"apm_enabled":          profile.Config(configAPMEnabled, "false"),
 		"geoip_dir":            profile.Config(configGeoIPDir, "./ingest-geoip"),
 		"kibana_http2_enabled": profile.Config(configKibanaHTTP2Enabled, "true"),
-		"logsdb_enabled":       profile.Config(configLogsDBEnabled, "false"),
+		"logsdb_enabled":       profile.Config(configLogsDBEnabled, "true"),
```
jsoriano, line 166: Change in default value included here only for testing, to be removed before merging, and eventually added in a different PR.

jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
Reverting the change to enable logsdb by default, to see what would be the current status of failures by applying only the changes for the multifields added by `ecs@mappings`.

jsoriano:
test integrations

internal/fields/validate.go
```diff
@@ -383,6 +389,86 @@ func allVersionsIncludeECS(kibanaConstraints *semver.Constraints) bool {
 	// return !kibanaConstraints.Check(lastStackVersionWithoutEcsMappings)
 }
 
+func ecsPathWithMultifieldsMatch(name string) bool {
```
felixbarny, line 392: Ideally, I'd like to avoid hard coding things from `ecs@mappings` here as changing that component template will cause tests to fail somewhere else.

While we're figuring the right path of whether to adapt `ecs@mappings` or ECS itself (https://github.com/elastic/ecs/issues/2353), could we make the validation less strict and not fail if the mapping contains additional sub-fields?

jsoriano: > Ideally, I'd like to avoid hard coding things from ecs@mappings here as changing that component template will cause tests to fail somewhere else.

If `ecs@mappings` doesn't differ further from ECS we should be fine, and hopefully this should be the case.

> While we're figuring the right path of whether to adapt `ecs@mappings` or ECS itself ([elastic/ecs#2353](https://github.com/elastic/ecs/issues/2353)), could we make the validation less strict and not fail if the mapping contains additional sub-fields?

The problem is that if we relax a constraint it is difficult to put it back later. This is why I would keep this to the minimum we know.

mrodm (approved):
LGTM

PR #2038, opened 2024-08-26
Add support to use wolfi Elastic Agent image starting in 8.16.0-SNAPSHOT

mrodm:
Relates https://github.com/elastic/elastic-agent/pull/5062

Use wolfi docker images for Elastic Agent starting with 8.16.0 stack version.

This behaviour could be disabled by defining the environment variable: `ELASTIC_PACKAGE_DISABLE_ELASTIC_AGENT_WOLFI`

Examples:

```shell
# Run 8.16.0-SNAPSHOT with wolfi images
elastic-package stack up -v -d --version 8.16.0-SNAPSHOT
ELASTIC_PACKAGE_DISABLE_ELASTIC_AGENT_WOLFI=false elastic-package stack up -v -d --version 8.16.0-SNAPSHOT

# Run 8.16.0-SNAPSHTO with ubuntu images
ELASTIC_PACKAGE_DISABLE_ELASTIC_AGENT_WOLFI=true elastic-package stack up -v -d --version 8.16.0-SNAPSHOT

# Previous stack versions would keep using the same docker images as before
elastic-package stack up -v -d --version 8.15.0
elastic-package stack up -v -d --version 7.15.0
```

Build using wolfi images (failing since that docker image is not available yet):
- https://buildkite.com/elastic/elastic-package/builds/3810

Build disabling wolfi images:
- https://buildkite.com/elastic/elastic-package/builds/3811


internal/install/application_configuration.go
```diff
@@ -21,12 +21,14 @@ import (
 )
 
 const (
-	stackVersion715 = "7.15.0-SNAPSHOT"
-	stackVersion820 = "8.2.0-SNAPSHOT"
+	stackVersion715  = "7.15.0-SNAPSHOT"
+	stackVersion820  = "8.2.0-SNAPSHOT"
+	stackVersion8160 = "8.16.0-00000000-SNAPSHOT"
```
mrodm, line 26: Using `8.16.0-SNAPSHOT` here as version did not change the docker image used for `8.16.0-21bba6f5-SNAPSHOT`.

jsoriano: Maybe we can add a comment mentioning this. Or set it to `8.15.999`.

I guess we didn't have the same problem with the snapshots above because we were not using tagged versions, right?

jsoriano (commented):
LGTM, but before merging this, we need the overrides for the packages that need the complete images, right?

And maybe we need also an environment variable, or some way to disable wolfi images in case it breaks something in CI before it is GA.

internal/install/application_configuration.go
mrodm: Exactly, previously tests were using SNAPSHOT versions without the commit in the prerelease tag.

I'll add a comment :+1:

mrodm:
> And maybe we need also an environment variable, or some way to disable wolfi images in case it breaks something in CI before it is GA.

Maybe it could be used the support added here to use another docker image: https://github.com/elastic/integrations/pull/10885

But, packages that are tested with stack versions older than 8.2.0 would require a different docker image :thinking: 

I'll check to add a new environment variable to skip this wolfi image.

internal/install/application_configuration.go
```diff
@@ -150,6 +158,15 @@ func selectElasticAgentImageName(version string) string {
 		logger.Errorf("stack version not in semver format (value: %s): %v", v, err)
 		return elasticAgentImageName
 	}
+
+	disableWolfiImages := false
+	valueEnv, ok := os.LookupEnv(disableElasticAgentWolfiEnvVar)
+	if ok && strings.ToLower(valueEnv) != "false" {
```
jsoriano, line 164: Perhaps we should default to true till the image is available if we want to merge this. Wdyt?

internal/install/application_configuration.go
mrodm: Added comment in https://github.com/elastic/elastic-package/pull/2038/commits/094f898a6f5a8795cbe082c18846f6adc2386bc5

jsoriano (approved):
LGTM, but we need to wait for the image to be available, or disable this by default by now.

internal/install/application_configuration.go
mrodm: Ok, I'll push the changes to revert the condition.
Once the image is available I'll create a follow-up PR.

jsoriano (approved):

jsoriano (approved):

PR #2044, opened 2024-08-27
Add new setting in system configuration file to set Elastic Agent image type

### LINKED ISSUE #2037: Allow to run system tests with a different Elastic Agent docker image
Currently, all system tests are run using the same Elastic Agent docker image.
The specific docker image to be used depends on the Elastic stack version:
https://github.com/elastic/elastic-package/blob/main/internal/install/application_configuration.go#L27-L29

It would be helpful to allow running a different docker image depending on the requirements of the package.

Probably a new setting in the configuration file could be added to allow this. This setting could be related to the type of Elastic Agent images. As example:

```yaml
agents:
  type: "complete"
```

If that could be done, all packages could be tested for instance with `docker.elastic.co/elastic-agent/elastic-agent` except some packages that would require `docker.elastic.co/elastic-agent/elastic-agent-complete`.

mrodm:
Closes #2037
Relates https://github.com/elastic/package-spec/pull/788

Adds support for a new setting in the system configuration file (`agent.type`). This setting will indicate which kind of Elastic Agent docker image will be used for testing in the given test.

Example of this setting:
```shell
agent:
  base_image: complete
```


internal/install/application_configuration.go
```diff
@@ -165,20 +167,41 @@ func selectElasticAgentImageName(version string) string {
...
+func OptionWithStackVersion(stackVersion string) ConfigurationOption {
+	return func(opts *configurationOptions) {
+		opts.stackVersion = stackVersion
 	}
-	return elasticAgentImageName
 }
 
 // Configuration function returns the elastic-package configuration.
-func Configuration() (*ApplicationConfiguration, error) {
+func Configuration(options ...ConfigurationOption) (*ApplicationConfiguration, error) {
```
jsoriano, line 204: Does it work without a stack version? Should it be mandatory?
```suggestion
func Configuration(stackVersion string, options ...ConfigurationOption) (*ApplicationConfiguration, error) {
```

internal/testrunner/runners/system/tester.go
```diff
@@ -428,6 +428,11 @@ func (r *tester) createAgentInfo(policy *kibana.Policy, config *testConfig, runI
 		info.Agent.User = "root"
 	}
 
+	// This could be removed once package-spec adds this new field
+	if !slices.Contains([]string{"", "default", "complete"}, info.Agent.BaseImage) {
+		return agentdeployer.AgentInfo{}, fmt.Errorf("invalid value for agent.base_image")
```
jsoriano, line 433: Nit. Include the value?
```suggestion
		return agentdeployer.AgentInfo{}, fmt.Errorf("invalid value for agent.base_image: %q", info.Agent.BaseImage)
```

jsoriano (commented):
Reviewed the last 4 commits, the rest are the same as in https://github.com/elastic/elastic-package/pull/2038, right?

internal/install/application_configuration.go
mrodm: While doing these changes, there was an usage where it was not used stack version:
https://github.com/elastic/elastic-package/blob/d8572d9f9a0b378cecc67c9789a16bfbdf37173a/cmd/profiles.go#L81

In that command, it looks like that it just checked the config file in the `~/.elastic-package` directory. It's not needed there the stack versions to retrieve the images of the services.

mrodm:
> Reviewed the last 4 commits, the rest are the same as in #2038, right?

Exactly, this branch was based on that other PR so I could test everything.

jsoriano (approved):

jsoriano:
Feel free to merge after resolving the conflicts if there is nothing significant.

jsoriano (approved):

jsoriano:
> Feel free to merge after resolving the conflicts if there is nothing significant.

Oh, race condition, you already resolved them :D

PR #2050, opened 2024-08-28
Update default disable wolfi image

mrodm:
Follows #2038

Updates the default value for the ELASTIC_PACKAGE_DISABLE_ELASTIC_AGENT_WOLFI environment variable to be false. Therefore, from 8.16.0-SNAPSHOT stack versions it will be used wolfi docker images for Elastic Agent as part of the system testing.


mrodm:
Closing this PR, wolfi image already exists and it is being applied this change into #2038.

PR #2054, opened 2024-08-28
Fix matching of leaf fields for objects

jsoriano:
Validation of objects was happening only depending of the format. If there were no subobjects, the validation was happening, and if there were objects with subfields, the validation was not happening for the subfields. This leads to different results with different mapping settings for the same data. For example depending if synthetics source or `subobjects: false` is used.

Additionally, for subfields that should match the definition of the parent object, there were cases where this match was not happening, so the test could unexpectedly fail.

Even if usually it is recommended to add explicit wildcards in dynamic mappings, definitions like the following one are expected to match subfields, but currently they aren't.
```
    - name: user
      type: group
      fields:
        - name: names
          type: object
          object_type: keyword
          description: Array of user identification documents.
```
The definition above would not match for `user.names.db`.

Seen for several packages in https://buildkite.com/elastic/integrations/builds/15135 (https://github.com/elastic/integrations/pull/10919).


jsoriano:
Pending to confirm if this is actually needed or a regression in latest main, because this was not failing before in dailies.

jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
test integrations

internal/fields/validate_test.go
```diff
@@ -528,15 +539,6 @@ func Test_parseElementValue(t *testing.T) {
...
 		},
-		// long
-		{
-			key:   "bad long",
-			value: "65537",
-			definition: FieldDefinition{
-				Type: "long",
-			},
-			fail: true,
-		},
```
jsoriano, line 539: These removed cases are actually allowed by Elasticsearch, so we can probably relax the validation.

jsoriano:
test integrations

jsoriano:
Changes related to unexpected types moved to https://github.com/elastic/elastic-package/pull/2065 and https://github.com/elastic/package-spec/pull/791.

jsoriano:
test integrations

mrodm (approved):
LGTM!

PR #2055, opened 2024-08-28
Add ecs@mappings multifields to all fields, not only external ones

jsoriano:
Multifields added by `ecs@mappings` are added to all fields, so we also need to add them to all fields
that match and not only to external ones.

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -283,6 +283,12 @@ func createValidatorForDirectoryAndPackageRoot(fieldsParentDir string, finder pa
 	}
 
 	v.Schema = append(fields, v.Schema...)
+
+	// ecs@mappings adds additional multifields that are not defined anywhere.
+	// Adding them in all cases so packages can be tested in versions of the stack that
+	// add the ecs@mappings component template.
+	v.Schema = appendECSMappingMultifields(v.Schema, "")
+
 	return v, nil
 }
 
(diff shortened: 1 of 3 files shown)


jsoriano:
test integrations

internal/fields/validate.go
```diff
@@ -283,6 +283,14 @@ func createValidatorForDirectoryAndPackageRoot(fieldsParentDir string, finder pa
 	}
 
 	v.Schema = append(fields, v.Schema...)
+
+	if !v.injectFieldsOptions.DisableAppendECSMultifields {
+		// ecs@mappings adds additional multifields that are not defined anywhere.
+		// Adding them in all cases so packages can be tested in versions of the stack that
+		// add the ecs@mappings component template.
```
mrodm, line 290: I recall an issue about nested field definitions that were not added as mappings.
https://github.com/elastic/package-spec/issues/784

I'm wondering if adding this multified definitions here, those kind of errors would be hidden. Or are these missing mapping related to other errors ?

jsoriano: This is not related to nested fields, this addresses issues like this one:
```
test case failed: one or more errors found in documents stored in logs-github.dependabot-57665 data stream: [0] parsing field value failed: [0] field "github.dependabot.security_advisory.cwes.name.text" is undefined, could be a multifield
```

This `text` multifield is added by `ecs@mappings`, even if `github.dependabot.security_advisory.cwes.name` is not an ECS field.

mrodm: That field it is defined in the `github` package (`cwes` is of type nested):
https://github.com/mrodm/integrations/blob/68da185c1487ce7082693bf7ef14f54c5f70f85b/packages/github/data_stream/dependabot/fields/fields.yml#L157


If that mapping would be added by fleet, it would have precedence over the ecs mappings, component template, no ?

jsoriano: No, you are right, this should not be an issue for fields whose mapping is properly provided by the package AND installed by Fleet.

jsoriano:
Closing, the cases we have now should be rather fixed by https://github.com/elastic/package-spec/issues/784 as mentioned by Mario in https://github.com/elastic/elastic-package/pull/2055#discussion_r1735828294.

PR #2059, opened 2024-08-30
Allow to dump documents resulting of system tests

jsoriano:
Found it useful while investigating https://github.com/elastic/integrations/pull/10919#discussion_r1735806766.

What do you think about including it upstream?

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -123,7 +123,8 @@ var (
 			},
 		},
 	}
-	enableIndependentAgents = environment.WithElasticPackagePrefix("TEST_ENABLE_INDEPENDENT_AGENT")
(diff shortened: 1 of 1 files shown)


internal/testrunner/runners/system/tester.go
```diff
@@ -1528,9 +1529,36 @@ func (r *tester) runTest(ctx context.Context, config *testConfig, svcInfo servic
...
+	}
+	defer f.Close()
+
+	logger.Infof("Dumping scenario documents to %s", path)
+
+	enc := json.NewEncoder(f)
+	enc.SetIndent("", "  ")
+	enc.SetEscapeHTML(false)
+	if err := enc.Encode(docs); err != nil {
+		return fmt.Errorf("failed to encode docs: %w")
```
mrodm, line 1557:
```suggestion
		return fmt.Errorf("failed to encode docs: %w", err)
```

mrodm (approved):
LGTM, Thanks!

PR #2065, opened 2024-09-02
Accept as numbers strings that can be parsed as numbers

jsoriano:
Similar to `numeric_keyword_fields`, add a new `string_number_fields` setting to pipeline and system tests. With this setting, validation of fields can accept strings as numbers if they can be parsed as float. This mimics the behaviour of Elasticsearch, that is able to accept strings as numbers if they can be parsed like them.

Refactor `numeric_keyword_fields`, so it is applied when validating values, instead of modifying the original value.

This will require change in package-spec.

This change will be needed for https://github.com/elastic/elastic-package/pull/2054, as some fields will start failing for this reason.

Changed required in the spec: https://github.com/elastic/package-spec/pull/791


internal/fields/validate.go
```diff
@@ -1266,23 +1260,19 @@ func ensureExpectedEventType(key string, values []string, definition FieldDefini
...
-				return nil, fmt.Errorf("expected string or array of strings")
-			}
-			values = append(values, s)
+			values = append(values, fmt.Sprintf("%v", e))
 		}
-		return values, nil
+		return values
 	default:
-		return nil, fmt.Errorf("expected string or array of strings")
+		return []string{fmt.Sprintf("%v", v)}
```
mrodm, line 1276: I guess that if the value is an object (something different from string or []string), it would cause some validation error afterwards. Would that be the case?

jsoriano: It will fail earlier, yes, I have moved the type validations before these checks https://github.com/elastic/elastic-package/pull/2065/files#diff-36d325d58b4ea8957abc63576d2e04d30ec8e58435ba06d442cb7923af063ad3L1018

jsoriano:
test integrations

mrodm (approved):

jsoriano:
This change seems to break several static tests, but it should not change the behaviour, investigating...

jsoriano:
Ok, I had removed the condition based on one setting to disable numeric validation.

jsoriano:
test integrations

jsoriano:
Reverting changes in go.mod now that the change in package-spec has been released.

PR #2066, opened 2024-09-02
[system tests] Add support for systemd option in agent.base_image setting

mrodm:
This PR adds support to run system tests using the regular Elastic Agent docker image (e.g. docker.elastic.co/elastic-agent/elastic-agent).

It will take into account the Elastic stack version to use `docker.elastic.co/beats/elastic-agent` docker image when it is required.

If there is no value defined for `agent.base_image` or it has the `default` value, the behaviour would be as it was up to now. It would use the complete image where it is supported.

Changed required in the spec: https://github.com/elastic/package-spec/pull/792

### DIFF
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -28,7 +28,8 @@ const (
 	// like 8.16.0-21bba6f5-SNAPSHOT
 	stackVersion8160 = "8.16.0-00000000-SNAPSHOT"
 
-	elasticAgentImageName               = "docker.elastic.co/beats/elastic-agent"
+	elasticAgentLegacyImageName         = "docker.elastic.co/beats/elastic-agent"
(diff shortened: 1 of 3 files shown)


internal/install/application_configuration.go
```diff
@@ -167,15 +168,36 @@ func selectElasticAgentImageName(version, agentBaseImage string) string {
...
+		return selectElasticAgentCompleteImageName(v)
+	case agentBaseImage == "systemd":
+		return selectElasticAgentSystemDImageName(v)
+	default:
+		switch {
+		case !disableWolfiImages && !v.LessThan(elasticAgentWolfiVersion):
+			return elasticAgentWolfiImageName
+		default:
+			return selectElasticAgentCompleteImageName(v)
+		}
```
mrodm, line 181: Keep the same behaviour as it was previously. It means using the complete image whenever it is possible.

jsoriano (approved):

PR #2067, opened 2024-09-02
[CI] Enable wolfi Elastic Agent images in tests

mrodm:
Run tests using Elastic Agent images where it is possible.

## Author's checklist
- [x] Remove update DefaultVersion variable
- [x] Check integration tests with 8.16.0-SNAPSHOT
    - https://buildkite.com/elastic/elastic-package/builds/3914
    - https://github.com/elastic/elastic-package/pull/2067/commits/8b742988e55e0a8431f440f74ddea79fb2d17e72


internal/install/stack_version.go
```diff
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "8.15.0"
+	DefaultStackVersion = "8.16.0-SNAPSHOT"
```
mrodm, line 9: To be removed

mrodm, line 9: To be removed

test/packages/parallel/apache/data_stream/access/_dev/test/policy/test-default.expected
```diff
@@ -17,6 +17,7 @@ inputs:
             - apache-access
       type: logfile
       use_output: default
+namespaces: []
```
mrodm, line 20: It looks like that for policy tests it should also be needed to be able to mark some fields optional?
This `namespaces` field appeared when testing with `8.16.0-SNAPSHOT`.

Changes required present in these two commits:
- https://github.com/elastic/elastic-package/pull/2067/commits/ba7e1f5f05f4cae7ac91835a94800982dc62490b
- https://github.com/elastic/elastic-package/pull/2067/commits/83a8623bf19143627eec25bdf3b5c5f9bdaf6dfc

mrodm: Reverting these two commits in order to be able to merge this PR.

internal/install/application_configuration_test.go
```diff
@@ -52,8 +53,12 @@ func TestSelectElasticAgentImageName_NextStackInOwnNamespace(t *testing.T) {
 	assert.Equal(t, selected, elasticAgentCompleteImageName)
 }
 
-func TestSelectElasticAgentImageName_DefaultImage816(t *testing.T) {
+func TestSelectElasticAgentImageName_DefaultImage816_WihtoutEnvVar(t *testing.T) {
 	version := stackVersion8160
+	// Try to keep the test agnostic from the environment variables defined in CI
+	t.Setenv(disableElasticAgentWolfiEnvVar, "")
+	os.Unsetenv(disableElasticAgentWolfiEnvVar)
```
mrodm, line 60: Could this cause any issues ?
It looks like the environment variable is defined in other tests even here it is unset.

jsoriano: Setting environment variables is only going to be problematic if tests are run in parallel, as it affects the global process. `t.Setenv` recovers the previous value on cleanup, see https://pkg.go.dev/testing#B.Setenv.

mrodm: Perfect, so as these tests do not use `t.Parallel()`, I guess it is safe here. Even if it is called `os.Unsetenv` , the environment variable would be recovered aftwards thanks to the t.Setenv()

jsoriano: I think each package is executed in a different process, so we should be fine if the tests are in different packages.

jsoriano (approved):

PR #2070, opened 2024-09-02
Force raising error in provisioning script if failure

mrodm:
Force to raise an error (interrupting docker build) if there is any failure in the provisioning script executed to build custom/independent Elastic Agents.

Relates https://github.com/elastic/integrations/pull/10943
Relates https://github.com/elastic/elastic-package/pull/2067

### DIFF
--- a/internal/agentdeployer/_static/dockerfile.tmpl
+++ b/internal/agentdeployer/_static/dockerfile.tmpl
@@ -14,9 +14,10 @@ USER root
 
 {{ if ne $provisioning_script_contents "" }}
 COPY {{ $provisioning_script_filename }} .
-RUN chmod u+x {{ $provisioning_script_filename }} && \
+RUN set -eu ; \
+  chmod u+x {{ $provisioning_script_filename }} && \
   {{ $provisioning_script_language }} ./{{ $provisioning_script_filename }} && \
-  rm {{ $provisioning_script_filename }}
+  rm {{ $provisioning_script_filename }} || exit 1
 {{ end }}
 
 {{ if ne $pre_start_script_contents "" }}


internal/agentdeployer/_static/dockerfile.tmpl
```diff
@@ -14,9 +14,10 @@ USER root
 
 {{ if ne $provisioning_script_contents "" }}
 COPY {{ $provisioning_script_filename }} .
-RUN chmod u+x {{ $provisioning_script_filename }} && \
+RUN set -eu ; \
```
jsoriano, line 17: Even without this it should fail in some cases, right?

internal/agentdeployer/_static/dockerfile.tmpl
```diff
@@ -14,9 +14,10 @@ USER root
...
+  chmod u+x {{ $provisioning_script_filename }} && \
   {{ $provisioning_script_language }} ./{{ $provisioning_script_filename }} && \
-  rm {{ $provisioning_script_filename }}
+  rm {{ $provisioning_script_filename }} || exit 1
```
jsoriano, line 20: Is the exit needed if the previous commands are already failing? Or we do this to return a consistent error code?
```suggestion
  rm {{ $provisioning_script_filename }}
```

jsoriano (approved):

internal/agentdeployer/_static/dockerfile.tmpl
mrodm: As the commands are concatenated with `&&` or `||` , the RUN step would just fail if the last command fails.

If one of the first commands fail and there is no `|| exit 1` , this RUN step does not fail.

So, it is required to force failing the Docker build.

internal/agentdeployer/_static/dockerfile.tmpl
mrodm: I added this to be sure and force that it would fail the RUN step.

internal/agentdeployer/_static/dockerfile.tmpl
mrodm: Added similar changes to the oracle package: 
https://github.com/elastic/integrations/pull/10943

internal/agentdeployer/_static/dockerfile.tmpl
mrodm: I'll do some tests without the `set -eu` in this RUN step.

internal/agentdeployer/_static/dockerfile.tmpl
mrodm: Related docs:
https://manpages.ubuntu.com/manpages/jammy/man1/bash.1.html#:~:text=The%20shell%20does%20not%20exit,is%20being%20inverted%20with%20!

internal/agentdeployer/_static/dockerfile.tmpl
mrodm: As this is just a one line, it looks like it is not needed to add `set -eu`.
The problem would be if this is changed at some point to be a multiline script like:
```
RUN chmod u+x {{ $provisioning_script_filename }} ; \
  {{ $provisioning_script_language }} ./{{ $provisioning_script_filename }} ; \\
  rm {{ $provisioning_script_filename }}  
```

internal/agentdeployer/_static/dockerfile.tmpl
mrodm: It looks like the issue was more related that the provisioning scripts in oracle package was not exiting with a code different from zero.

Just tested with this RUN step 
```
{{ if ne $provisioning_script_contents "" }}
COPY {{ $provisioning_script_filename }} .
RUN chmod u+x {{ $provisioning_script_filename }} && \
  {{ $provisioning_script_language }} ./{{ $provisioning_script_filename }} && \
  rm {{ $provisioning_script_filename }}
{{ end }}

```
and this provisioning script (it adds the `|| exit 1` to ensure failing the script)
```
  ...
  echo /opt/oracle/instantclient_21_4 > /etc/ld.so.conf.d/oracle-instantclient.conf && ls filenotexist && ldconfig || exit 1
  ...
```


I think it could be closed this PR, WDYT @jsoriano ?

jsoriano: > I think it could be closed this PR, WDYT @jsoriano ?

Unless we want to add the `set -eu` I think this can be closed, yes.

mrodm: As it is just one line, I think it can be kept like this.
Closing this PR , thanks for checking this out!

PR #2072, opened 2024-09-03
[CI] Rename test scripts

mrodm:
Follows #2030 

Rename test scripts to be more specific about the purpose of each one.

Update targets in Makefile.

Added also a test in `scripts/test-build-install-zip.sh` to check whether or not the package is installed.


Makefile
```diff
@@ -127,7 +127,7 @@ test-profiles-command:
 test-check-update-version:
 	./scripts/test-check-update-version.sh
 
-test: test-go test-stack-command test-check-packages test-profiles-command test-build-zip test-check-update-version
+test: test-go test-stack-command test-check-packages test-profiles-command test-build-install-zip test-build-zip test-build-install-zip-file test-build-install-zip-file-shellinit test-check-update-version test-profiles-command test-system-test-flags
```
mrodm, line 130: Added missing targets into `test`.

jsoriano (approved):

PR #2075, opened 2024-09-03
[docs] Update system testing docs

mrodm:
Update the example of a provisioning script (oracle test package).
And add a note to make sure the script finishes with error if any command fail.

### DIFF
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -620,14 +620,20 @@ vars:
 agent:
   runtime: docker
   provisioning_script:
-    language: "bash"
+    language: "sh"
     contents: |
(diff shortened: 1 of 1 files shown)


docs/howto/system_testing.md
```diff
@@ -637,6 +643,9 @@ agent:
       cd /opt/oracle/instantclient_21_4
 ```
 
+**IMPORTANT**: The provisioning script must exit with a code different from zero in case any of the commands defined fails.
+That will ensure that the docker build step run by `elastic-package` fails too.
```
mrodm, line 647: Add a note about the importance of ensuring that the provisioning script fails if any command of the script fails.

jsoriano (approved):

PR #2084, opened 2024-09-05
Add false positive test package to detect errors in elastic agent logs

### LINKED ISSUE #1811: Add negative tests for detection of issues in elastic agent logs

mrodm:
Closes #1811 

Added a new test package into `false_positives` to be able to check if there is any error in the logs from Elastic Agent container.

This package requires to be executed in a specific Elastic stack version 8.12.0 in order to reproduce the error.

Tested failure (not detected error logs) build (using 8.12.1):
- https://buildkite.com/elastic/elastic-package/builds/3958#0191c69d-8818-4b98-95b1-34fcface0b97


scripts/test-check-false-positives.sh
```diff
@@ -59,6 +59,11 @@ function check_expected_errors() {
     cat ${results_no_spaces} | grep -E "${line}"
   done < "${expected_errors_file}"
 
+  # Copy XML files to another extension so they are not used to check jUnit tests
+  # but those files will be able to be reviewed afterwards
+  for file in $(ls $result_tests) ; do
+      cp "${file}" "${file}.txt"
```
jsoriano, line 65: Nit. Could we add some meaningful suffix?
```suggestion
      cp "${file}" "${file}-expected-errors.txt"
```

scripts/test-check-false-positives.sh
```diff
@@ -59,6 +59,11 @@ function check_expected_errors() {
     cat ${results_no_spaces} | grep -E "${line}"
   done < "${expected_errors_file}"
 
+  # Copy XML files to another extension so they are not used to check jUnit tests
+  # but those files will be able to be reviewed afterwards
```
jsoriano, line 63: :+1:

jsoriano (commented):
Nice, thanks!

scripts/test-check-false-positives.sh
mrodm: Sure!
I'll update it.

jsoriano (approved):

PR #2088, opened 2024-09-06
Fix validation of subfields of flattened objects

jsoriano:
When a document doesn't has subobjects, like when `subobjects: false` or synthetic mode is used, if an undocumented field is found, we need to check its ancestors to see if it is a member of a flattened object.

This is producing failures for `auditd_manager` package when LogsDB is enabled: https://buildkite.com/elastic/integrations/builds/15559
```
test case failed: one or more errors found in documents stored in logs-auditd_manager.auditd-20702 data stream: [0] field "auditd.data.a0" is undefined, could be a multifield
[1] field "auditd.data.a1" is undefined, could be a multifield
[2] field "auditd.data.a2" is undefined, could be a multifield
[3] field "auditd.data.a3" is undefined, could be a multifield
```

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -681,6 +681,8 @@ func (v *Validator) validateScalarElement(key string, val any, doc common.MapStr
 		switch {
 		case skipValidationForField(key):
 			return nil // generic field, let's skip validation for now
+		case isFlattenedSubfield(key, v.Schema):
+			return nil // flattened subfield, it will be stored as member of the flattened ancestor.
 		case isArrayOfObjects(val):
 			return fmt.Errorf(`field %q is used as array of objects, expected explicit definition with type group or nested`, key)
 		case couldBeMultifield(key, v.Schema):
@@ -852,6 +854,22 @@ func isArrayOfObjects(val any) bool {
 			}
 		}
 	}
+	return false
+}
+
+func isFlattenedSubfield(key string, schema []FieldDefinition) bool {
+	for strings.Contains(key, ".") {
+		i := strings.LastIndex(key, ".")
+		key = key[:i]
+		ancestor := FindElementDefinition(key, schema)
+		if ancestor == nil {
+			continue
+		}
+		if ancestor.Type == "flattened" {
+			return true
+		}
+	}
+
 	return false
 }
 
--- a/internal/fields/testdata/flattened.json
+++ b/internal/fields/testdata/flattened.json
@@ -6,6 +6,7 @@
         "userName": "Bob",
         "groupName": "admin"
       }
-    }
+    },
+    "flattened.request_parameters.userID": 1000
   }
 }


internal/fields/validate.go
```diff
@@ -855,6 +857,22 @@ func isArrayOfObjects(val any) bool {
...
+	for strings.Contains(key, ".") {
+		i := strings.LastIndex(key, ".")
+		key = key[:i]
+		ancestor := FindElementDefinition(key, schema)
+		if ancestor == nil {
+			continue
+		}
+		if ancestor.Type == "flattened" {
+			return true
+		}
```
mrodm, line 870: If it find an ancestor with a different type of "flattened" should be return false directly here?
I was thinking in the case of the ancestor being something like: `ancestor.Type = object` and `ancestor.ObjectTYpe == "keyword"`

internal/fields/validate.go
```diff
@@ -855,6 +857,22 @@ func isArrayOfObjects(val any) bool {
 	return false
 }
 
+func isFlattenedSubfield(key string, schema []FieldDefinition) bool {
+	for strings.Contains(key, ".") {
+		i := strings.LastIndex(key, ".")
+		key = key[:i]
+		ancestor := FindElementDefinition(key, schema)
```
mrodm, line 864: Could this be replaced by findParentElementDefinition ? 

```suggestion
		ancestor := findParentElementDefinition(key, schema)
```

mrodm: mmm I guess it would be needed to somehow return also the key of the field, or being able to retrieve it from the `ancestor` struct.

jlind23 (approved):

internal/fields/validate.go
jsoriano: Yep, this is the problem I found, we still need to loop over the ancestors.

internal/fields/validate.go
jsoriano: Not sure, there are packages defining mappings for subfields of flattened objects (for example [here](https://github.com/elastic/integrations/blob/2074f69f7681b1c2f04a03c228173ab425cd0283/packages/auditd_manager/data_stream/auditd/fields/fields.yml#L695)). This doesn't make a lot of sense as a mapping, but this is allowed now, and can be used for example to provide documentation.

We probably should not allow to define mappings under `flattened` fields, but this would be a different task, and checked in a different place. And we would need to provide something for cases where developers want to provide docs for flattened subfields.

mrodm: > Not sure, there are packages defining mappings for subfields of flattened objects (for example [here](https://github.com/elastic/integrations/blob/2074f69f7681b1c2f04a03c228173ab425cd0283/packages/auditd_manager/data_stream/auditd/fields/fields.yml#L695)). This doesn't make a lot of sense as a mapping, but this is allowed now, and can be used for example to provide documentation.

Ok, I didn't know about this kind of usage.

> We probably should not allow to define mappings under flattened fields, but this would be a different task, and checked in a different place. And we would need to provide something for cases where developers want to provide docs for flattened subfields.

Agreed!

mrodm (approved):
LGTM! Checked locally too with logsdb enabled for `audtid_manager` package.

PR #2090, opened 2024-09-09
Relax validation of subobject leaves for flattened fields

jsoriano:
We added some additional validations for subobjects and arrays of objects in #1498, #1489 and related PRs. These validations only apply to packages with package spec starting on 3.0.1.

These tests rely on the structure of documents. With the adoption of features like `subobjects: false` or synthetic source the structure is lost, and exceptions based on spec version are not working, so the tests fail for cases where they should not for versions of the spec older than 3.0.1. This happens for example in the `dns` data stream of the `network_traffic` package when LogsDB is enabled.

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -828,31 +828,32 @@ func skipValidationForField(key string) bool {
 // skipLeafOfObject checks if the element is a child of an object that was skipped in some previous
 // version of the spec. This is relevant in documents that store fields without subobjects.
 func skipLeafOfObject(root, name string, specVersion semver.Version, schema []FieldDefinition) bool {
-	if specVersion.LessThan(semver3_0_1) {
-		// Check if this is a subobject of an object we didn't traverse.
-		if !strings.Contains(name, ".") {
-			return false
-		}
-		key := name
-		if root != "" {
-			key = root + "." + name
-		}
-		_, ancestor := findAncestorElementDefinition(key, schema, func(key string, def *FieldDefinition) bool {
-			// Don't look for ancestors beyond root, these objects have been already traversed.
-			if len(key) < len(root) {
-				return false
-			}
-			if !slices.Contains([]string{"group", "object", "nested", "flattened"}, def.Type) {
-				return false
-			}
-			return true
-		})
-		if ancestor != nil {
-			return true
-		}
+	// We are only skipping validation of these fields on versions older than 3.0.1.
+	if !specVersion.LessThan(semver3_0_1) {
+		return false
 	}
 
-	return false
+	// If it doesn't contain a dot in the name, we have traversed its parent, if any.
+	if !strings.Contains(name, ".") {
+		return false
+	}
+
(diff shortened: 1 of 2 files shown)


jsoriano:
test integrations

jsoriano:
Testing the change with LogsDB in https://buildkite.com/elastic/integrations/builds/15648. The network traffic package tests should pass.

internal/fields/validate.go
```diff
@@ -820,6 +825,36 @@ func skipValidationForField(key string) bool {
 		isFieldFamilyMatching("event.module", key) // field is deprecated
 }
 
+// skipLeafOfObject checks if the element is a child of an object that was skipped in some previous
+// version of the spec. This is relevant in documents that store fields without subobjects.
+func skipLeafOfObject(root, name string, specVersion semver.Version, schema []FieldDefinition) bool {
+	if specVersion.LessThan(semver3_0_1) {
```
mrodm, line 831: Nit. It could be returned early here if spec is spec >= 3.0.1 and change also the latest return as `ancestor != nil`

```golang
func skipLeafOfObject(root, name string, specVersion semver.Version, schema []FieldDefinition) bool {
	if !specVersion.LessThan(semver3_0_1) {
		return false
	}
	// Check if this is a subobject of an object we didn't traverse.
	if !strings.Contains(name, ".") {
		return false
	}
	key := name
	if root != "" {
		key = root + "." + name
	}
	_, ancestor := findAncestorElementDefinition(key, schema, func(key string, def *FieldDefinition) bool {
		// Don't look for ancestors beyond root, these objects have been already traversed.
		if len(key) < len(root) {
			return false
		}
		if !slices.Contains([]string{"group", "object", "nested", "flattened"}, def.Type) {
			return false
		}
		return true
	})

	return ancestor != nil
}

File: internal/fields/validate_test.go [hunk 2]
```diff
@@ -1048,6 +1048,98 @@ func TestValidateStackVersionsWithEcsMappings(t *testing.T) {
...
+			name:     "subgroup.bar",
+			version:  semver.MustParse("3.0.1"),
+			expected: false,
+		},
+	}
+
+	// Cases we expect to skip depending on the version.
+	okRoots := []string{"flattened", "object", "group", "nested"}
+	for _, root := range okRoots {
+		t.Run("(empty root)", func(t *testing.T) {
```
mrodm, line 1111: Should the title of this `t.Run()` include the `root` too ?
If not , for each `okRoots` element this run would have the same name.

```suggestion
		t.Run("(empty root) prefix "+root, func(t *testing.T) {
```

internal/fields/validate.go
jsoriano: Yep, I organized it this way in case we have other cases like this in the future, but happy to change it and refactor later if needed.

jsoriano: Refactored, thanks.

mrodm (approved):

PR #2102, opened 2024-09-13
Add initial support for Elastic Stack 9.x

mrodm:
Part of https://github.com/elastic/package-registry/issues/1226

Add initial support for 9.0.0 stack in `elastic-package`.

Changes performed:
- By default, `elastic-package` will use wolfi images as it was for 8.16 stack versions.
- If no version is specified, it will report wolfi image instead (TBD).
- Added integration test to check `elastic-package stack command` for 9.0.0-SNAPSHOT

## Author's checklist
- [x] Enable target in updatecli configuration before merging.
- [x] Add integration test for `elastic-package stack` command with 9.0.0-SNAPSHOT
- [x] Test `elastic-package stack up -v --version 9.0.0-SNAPSHOT`.


internal/install/application_configuration_test.go
```diff
@@ -60,56 +60,56 @@ func TestSelectElasticAgentImageName_DefaultImage816_WithoutEnvVar(t *testing.T)
...
 	selected := selectElasticAgentImageName(version, "default")
-	assert.Equal(t, selected, elasticAgentCompleteImageName)
+	assert.Equal(t, elasticAgentCompleteImageName, selected)
 }
 
 func TestSelectCompleteElasticAgentImageName_ForceSystemDImage(t *testing.T) {
 	version := stackVersion8160
 	selected := selectElasticAgentImageName(version, "systemd")
-	assert.Equal(t, selected, elasticAgentImageName)
+	assert.Equal(t, elasticAgentImageName, selected)
```
mrodm, line 108: Updated the order to ensure that the values are the expected and the actual ones as required by the assert functions.
https://pkg.go.dev/github.com/stretchr/testify/assert#Equal

.github/workflows/updatecli/updatecli.d/bump-latest-9x-snapshot-version.yml
```diff
@@ -0,0 +1,44 @@
...
+targets:
+  # update-snapshot:
+  #   name: '[updatecli] Update latest snapshot to {{ source "latestSnapshot" }}'
+  #   kind: file
+  #   sourceid: latestSnapshot
+  #   scmid: default
+  #   spec:
+  #     file: Makefile
+  #     matchpattern: '(./scripts/test-stack-command.sh) 9\.[^\s]+-SNAPSHOT'
+  #     replacepattern: '$1 {{ source "latestSnapshot" }}-SNAPSHOT'
```
mrodm, line 44: To be enabled in a follow-up PR.

internal/install/application_configuration.go
```diff
@@ -153,13 +154,13 @@ func (ac *ApplicationConfiguration) SetCurrentProfile(name string) {
...
 	if version == "" { // as version is optional and can be empty
-		return elasticAgentLegacyImageName
+		return elasticAgentWolfiImageName
 	}
 
 	v, err := semver.NewVersion(version)
 	if err != nil {
 		logger.Errorf("stack version not in semver format (value: %s): %v", v, err)
-		return elasticAgentLegacyImageName
+		return elasticAgentWolfiImageName
```
mrodm, line 163: WDYT about changing these defaults values to use the wolfi image `docker.elastic.co/elastic-agent/elastic-agent-wolfi` (e.g. no version specified) instead of the legacy image (beats namespace) `docker.elastic.co/beats/elastic-agent`?

internal/install/application_configuration_test.go
```diff
@@ -14,43 +14,43 @@ import (
 func TestSelectElasticAgentImageName_NoVersion(t *testing.T) {
 	var version string
 	selected := selectElasticAgentImageName(version, "")
-	assert.Equal(t, selected, elasticAgentLegacyImageName)
+	assert.Equal(t, elasticAgentWolfiImageName, selected)
```
mrodm, line 17: This is the test changing the value, it would return the wolfi image instead.

.github/workflows/updatecli/updatecli.d/bump-latest-9x-snapshot-version.yml
```diff
@@ -0,0 +1,44 @@
...
+# targets:
+#   update-snapshot:
+#     name: '[updatecli] Update latest snapshot to {{ source "latestSnapshot" }}'
+#     kind: file
+#     sourceid: latestSnapshot
+#     scmid: default
+#     spec:
+#       file: Makefile
+#       matchpattern: '(./scripts/test-stack-command.sh) 9\.[^\s]+-SNAPSHOT'
+#       replacepattern: '$1 {{ source "latestSnapshot" }}-SNAPSHOT'
```
mrodm, line 44: To be enabled before merging

internal/install/application_configuration.go
```diff
@@ -153,13 +154,13 @@ func (ac *ApplicationConfiguration) SetCurrentProfile(name string) {
 // This is mandatory as "elastic-agent-complete" is available since 7.15.0-SNAPSHOT.
 func selectElasticAgentImageName(version, agentBaseImage string) string {
 	if version == "" { // as version is optional and can be empty
-		return elasticAgentLegacyImageName
+		return elasticAgentWolfiImageName
```
mrodm, line 157: WDYT about changing these defaults values to use the wolfi image docker.elastic.co/elastic-agent/elastic-agent-wolfi (e.g. no version specified) instead of the legacy image (beats namespace) docker.elastic.co/beats/elastic-agent?

.buildkite/scripts/stack_tests_windows.ps1
```diff
@@ -45,7 +45,7 @@ fixCRLF
 
 withGolang $env:GO_VERSION
 withDocker $env:DOCKER_VERSION
-withDockerCompose $env:DOCKER_COMPOSE_VERSION
+withDockerCompose $env:DOCKER_COMPOSE_VERSION.Substring(1)
```
mrodm, line 48: Required to remove `v` from the version string for the chocolotey tool. Currently, this environment variable has this value `v2.24.1`.

internal/install/application_configuration.go
```diff
@@ -153,13 +153,13 @@ func (ac *ApplicationConfiguration) SetCurrentProfile(name string) {
...
 
 	v, err := semver.NewVersion(version)
 	if err != nil {
-		logger.Errorf("stack version not in semver format (value: %s): %v", v, err)
-		return elasticAgentLegacyImageName
+		logger.Errorf("stack version not in semver format (value: %s): %v", version, err)
+		return elasticAgentWolfiImageName
 	}
 
 	disableWolfiImages := false
```
mrodm, line 165: As it is now, the Elastic Agent image is chosen as:
- If stack >= 8.16.0-SNAPSHOT,
    - by default it is used wolfi image_ `docker.elastic.co/elastic-agent/elastic-agent-wolfi`.
    - if ELASTIC_PACKAGE_DISABLE_ELASTIC_AGENT_WOLFI=true, then it is selected the Elastic Agent image as before. So if stack >= 7.15.0-SNAPSHOT, it will be used the complete image.
- If the configuration file of the system test has `base_image: complete`:
    - It will be used the complete version of the Elastic Agent image if the stack version allows it.
- If the configuration file of the system test has `base_image: systemd`:
    - It will be used the `systemd` version of the Elastic Agent image. That means `docker.elastic.co/elastic-agent/elastic-agent` or `docker.elastic.co/beats/elastic-agent` depending on the stack version.


Would that be ok? Specially if stack version is set to 9.0.0 and the environment variable to disable wolfi is set to true.

jsoriano: If the same set of images is going to be available in 9.0 keeping the same logic sounds good to me.

.buildkite/scripts/stack_tests_windows.ps1
jsoriano: Interesting, how has it worked in the past?

internal/install/application_configuration.go
jsoriano: We can adapt this to the new default, yes, if we are sure that in old versions the beats namespace is used.

internal/install/application_configuration_test.go
jsoriano: In principle it LGTM. Maybe another option is to return an error if the version is empty? Not sure about the implications of the change, but would look safer to me to know where/if we are passing an empty version.

jsoriano (approved):

.buildkite/scripts/stack_tests_windows.ps1
mrodm: Probably, it was failing in the past too. I realized that even if with those commands failing the execution of the script continued.

internal/install/application_configuration.go
mrodm: Yes, the same behaviour is kept.
As a reference, I add here the docker images used in different versions:
- 7.14.0
  ```
   $ docker ps --format "{{.Names}} {{.Image}}" |grep elastic-agent
  elastic-package-stack-elastic-agent-1 docker.elastic.co/beats/elastic-agent:7.14.0
  elastic-package-stack-fleet-server-1 docker.elastic.co/beats/elastic-agent:7.14.0
  ```
- 7.15.0
  ```
   $ docker ps --format "{{.Names}} {{.Image}}" |grep elastic-agent
  elastic-package-stack-elastic-agent-1 docker.elastic.co/beats/elastic-agent-complete:7.15.0
  elastic-package-stack-fleet-server-1 docker.elastic.co/beats/elastic-agent-complete:7.15.0
  ```
- 8.2.0
  ```
   $ docker ps --format "{{.Names}} {{.Image}}" |grep elastic-agent
  elastic-package-stack-elastic-agent-1 docker.elastic.co/elastic-agent/elastic-agent-complete:8.2.0
  elastic-package-stack-fleet-server-1 docker.elastic.co/elastic-agent/elastic-agent-complete:8.2.0
  ```
- 8.16.0 (SNAPSHOT)
  ```
   $ docker ps --format "{{.Names}} {{.Image}}" |grep elastic-agent
  elastic-package-stack-elastic-agent-1 docker.elastic.co/elastic-agent/elastic-agent-wolfi:8.16.0-SNAPSHOT
  elastic-package-stack-fleet-server-1 docker.elastic.co/elastic-agent/elastic-agent-wolfi:8.16.0-SNAPSHOT
  ```
- 9.0.0 (SNAPSHOT)
  ```
   $ docker ps --format "{{.Names}} {{.Image}}" |grep elastic-agent
  elastic-package-stack-elastic-agent-1 docker.elastic.co/elastic-agent/elastic-agent-wolfi:9.0.0-SNAPSHOT
  elastic-package-stack-fleet-server-1 docker.elastic.co/elastic-agent/elastic-agent-wolfi:9.0.0-SNAPSHOT
  ```

internal/install/application_configuration_test.go
mrodm: Commands like `elastic-package stack down` do not set any version, and if a error is returned here, it will fail the command.

jsoriano (approved):

PR #2114, opened 2024-09-20
feat: allow pipeline test case to define the expected output based on service versions constraints

pkoutsovasilis:
This PR essentially allows to specify a different expected output for a given pipeline test case based on given version constraints of running stack services. Such a feature is required to handle gracefully cases as the one captured [here](https://github.com/elastic/integrations/pull/11067) where the normal CI tests are passing as the stack version is defined by the minimum one required by the package manifest. However, testing it with a higher version of the stack, although not required to be the minimum supported, it results in different expected outputs

An example of an integration package that this PR could be used with can be found [here](https://github.com/pkoutsovasilis/integrations/commit/1f8141e709aed4b08d2c5d5026381285249d9f67). Here we define different expected outputs for `test-cisco-ios.log` and `test-asr920.log` depending on the elasticsearch version of the running stack

### DIFF
--- a/internal/testrunner/errors.go
+++ b/internal/testrunner/errors.go
@@ -25,5 +25,5 @@ type ErrTestCaseConstraintsSkip struct {
 
 // Error returns the message detailing the test case failure.
 func (e ErrTestCaseConstraintsSkip) Error() string {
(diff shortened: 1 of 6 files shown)


pkoutsovasilis:
cc @qcorporation @gogochan

PR #2151, opened 2024-10-09
Add error check for HTTPJSON template execution failure

efd6:
We don't know if this will result in failures in existing integrations, so running test integrations with a view to resolving them if they exist before this is merged.

Sadly, it needs to be relaxed, but it's better than nothing.

ref: https://github.com/elastic/integrations/pull/11361#discussion_r1792111556

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -120,6 +120,15 @@ var (
 						regexp.MustCompile(`Component state changed .* \(HEALTHY->DEGRADED\): Degraded: pid .* missed .* check-in`),
 					},
 				},
+				{
+					// HTTPJSON template error.
+					includes: regexp.MustCompile(`^error processing response: template: :\d+:\d+: executing "" at <`),
+					excludes: []*regexp.Regexp{
+						// Unfortunate: https://github.com/elastic/beats/issues/34544
+						// See also https://github.com/elastic/beats/pull/39929.
+						regexp.MustCompile(`: map has no entry for key`),
+					},
+				},
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


efd6:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -120,6 +120,15 @@ var (
 						regexp.MustCompile(`Component state changed .* \(HEALTHY->DEGRADED\): Degraded: pid .* missed .* check-in`),
 					},
 				},
+				{
+					// HTTPJSON template error.
+					includes: regexp.MustCompile(`^error processing response: template: :\d+:\d+: executing "" at <`),
```
jsoriano, line 125: Would it be possible to have a test case that reproduces this issue?

jsoriano:
test integrations

jsoriano (commented):
Thanks for adding this. It LGTM, would it be possible to add a test case?

jsoriano:
There seems to be still a related change in https://github.com/elastic/integrations/pull/11370

internal/testrunner/runners/system/tester.go
efd6: How is this done?

efd6:
Expanded the set of unfortunate false positives. This cannot be worked around in the HTTPJSON template.

efd6:
test integrations

jsoriano (approved):

internal/testrunner/runners/system/tester.go
jsoriano: We could maybe add a [false positives](https://github.com/elastic/elastic-package/tree/f61e253f3a81a5b9f10bb377eb95efad87fd2507/test/packages/false_positives) test case.

efd6: The current HTTPJSON false positive case is already hitting this logic.

PR #2161, opened 2024-10-14
[PoC] Use mapping validations in system tests

mrodm:
This is a PoC to use validations related to mappings in system tests: `elastic-package test system -v`.

This PoC is based on retrieving:
- the mappings from the preview of the index, and
- the final mappings from the index after being ingested some documents by the Elastic Agent,
so it can be performed some validation on them.

Queries performed:
```
# Mappings Preview
POST /_index_template/_simulate_index/<data_stream_name>

# Actual Mappings
GET /<data_stream_name>/_mapping
```

Not completed in this PoC:
- Do the actual comparison of mappings using the relevant fields from the responses of the queries.
- Compare mappings in case it is needed with the dynamic templates to ensure all fields are documented.
- Add support for other field mappings like runtime fields or histograms

### DIFF
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/cbroglie/mustache"
+	"github.com/google/go-cmp/cmp"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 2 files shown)


mrodm:
PoC completed, closing for now this PR.

PR #2167, opened 2024-10-16
Add support for value in documentation

stevengoossensB:
Update exported_fields.go - Add capability to write the value of a field as defined in the fields.yml to the Markdown documentation

### DIFF
--- a/internal/docs/exported_fields.go
+++ b/internal/docs/exported_fields.go
@@ -18,6 +18,7 @@ type fieldsTableRecord struct {
 	aType       string
 	unit        string
 	metricType  string
+	value       string
 }
 
 var escaper = strings.NewReplacer("*", "\\*", "{", "\\{", "}", "\\}", "<", "\\<", ">", "\\>")
@@ -53,6 +54,7 @@ func renderExportedFields(fieldsParentDir string) (string, error) {
 func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord) {
 	unitsPresent := areUnitsPresent(collected)
 	metricTypesPresent := areMetricTypesPresent(collected)
+	valuesPresent := areValuesPresent(collected)
 
 	builder.WriteString("| Field | Description | Type |")
 	if unitsPresent {
@@ -61,6 +63,9 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord) 
 	if metricTypesPresent {
 		builder.WriteString(" Metric Type |")
 	}
+	if valuesPresent {
+		builder.WriteString(" Value |")
+	}
 
 	builder.WriteString("\n")
 	builder.WriteString("|---|---|---|")
@@ -68,6 +73,9 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord) 
 		builder.WriteString("---|")
 	}
 	if metricTypesPresent {
+		builder.WriteString("---|")
+	}
+	if valuesPresent {
 		builder.WriteString("---|")
 	}
 
@@ -83,6 +91,9 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord) 
 		}
 		if metricTypesPresent {
 			builder.WriteString(fmt.Sprintf(" %s |", c.metricType))
+		}
+		if valuesPresent {
+			builder.WriteString(fmt.Sprintf(" %s |", c.value))
 		}
 		builder.WriteString("\n")
 	}
@@ -100,6 +111,15 @@ func areUnitsPresent(collected []fieldsTableRecord) bool {
 func areMetricTypesPresent(collected []fieldsTableRecord) bool {
 	for _, c := range collected {
 		if c.metricType != "" {
+			return true
+		}
+	}
+	return false
+}
+
+func areValuesPresent(collected []fieldsTableRecord) bool {
+	for _, c := range collected {
+		if c.value != "" {
 			return true
 		}
 	}
@@ -130,6 +150,7 @@ func visitFields(namePrefix string, f fields.FieldDefinition, records []fieldsTa
 			aType:       f.Type,
 			unit:        f.Unit,
 			metricType:  f.MetricType,
+			value:       f.Value,
 		})
 
 		for _, multiField := range f.MultiFields {


internal/docs/exported_fields.go
```diff
@@ -61,6 +63,9 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord)
 	if metricTypesPresent {
 		builder.WriteString(" Metric Type |")
 	}
+	if valuesPresent {
+		builder.WriteString(" Value |")
+	}
```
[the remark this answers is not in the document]
stevengoossensB: Making it optional makes most sense, as there's no change in the spec required for it to work. The same goes for **example** and other keys. It would be nice to see that as part of the documentation, next to the description key.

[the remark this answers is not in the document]
stevengoossensB: Cycling back to this... @jsoriano do you have a suggestion on how and where to make this optional? I've been looking at the profiles, but want to make sure it's aligned with the vision you have about this.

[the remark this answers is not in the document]
jsoriano: I see two options, each one with pros and cons, and can be complementary, so we could make both:

* Make this dependant on the `format_version` of the package. We would need to somehow pass this spec version to `renderExportedFields`, and maybe we could add a new option to `fields.InjectFieldsOptions`, to avoid loading values of constant keywords. The pro of this option is that all packages will eventually do this, the con, that packages will need to update its format version to have these fields.
* Add a build option (to `_dev/build/build.yml` for example) that enforces the generation of these fields. This would need changes in the spec, and would also need to add the setting in packages.

I would prefer the first option, but we have to take into account that this will only apply to packages using future versions of the spec.

stevengoossensB:
I added to to PR the following:
- Example key to be added to the documentation
- a requirement for the specVersion to be 3.2.3. 
- If example or value is nowhere defined in the fields, this also doesn't get added to the README.

internal/docs/exported_fields.go
```diff
@@ -5,19 +5,27 @@
...
 	"sort"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
+
 	"github.com/elastic/elastic-package/internal/fields"
+	"github.com/elastic/elastic-package/internal/packages"
 )
 
+var semver3_2_3 = semver.MustParse("3.2.3")
```
jsoriano, line 19: 3.2.3 has been already released (and 3.3.0), we would need to start doing this at least on 3.3.1.

internal/docs/exported_fields.go
```diff
@@ -46,13 +54,37 @@ func renderExportedFields(fieldsParentDir string) (string, error) {
...
+	sv, err := semver.NewVersion(manifest.SpecVersion)
+	if err != nil {
+		return "", fmt.Errorf("failed to obtain spec version from package manifest in \"%s\"", packageRoot)
+	}
+
+	if sv.LessThan(semver3_2_3) {
+		renderFieldsTable(&builder, collected, false, false)
+	} else {
+		renderFieldsTable(&builder, collected, true, true)
+	}
```
jsoriano, line 78: Nit. I would suggest something like this for self-explanatory purposes:
```suggestion
	renderExamples := true
	renderValues := true
	if sv.LessThan(semver3_2_3) {
		renderExamples = false
		renderValues = false
	}
	renderFieldsTable(&builder, collected, renderValues, renderExamples)
```

internal/docs/exported_fields.go
```diff
@@ -70,6 +108,12 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord)
 	if metricTypesPresent {
 		builder.WriteString("---|")
 	}
+	if valuesPresent && includeValues {
+		builder.WriteString("---|")
+	}
+	if examplePresent && includeExamples {
+		builder.WriteString("---|")
+	}
```
jsoriano, line 116: This is going to add two columns, right? Could you check how this looks in Kibana?

internal/docs/exported_fields.go
```diff
@@ -46,13 +54,37 @@ func renderExportedFields(fieldsParentDir string) (string, error) {
 		builder.WriteString("(no fields available)\n")
 		return builder.String(), nil
 	}
-	renderFieldsTable(&builder, collected)
+
+	packageRoot, found, err := packages.FindPackageRoot()
```
jsoriano, line 58: Wdyt about passing the spec version from `generateReadme`? This way we don't need to find the package root in different places and this method keeps depending only on the fields dir.

internal/docs/exported_fields.go
jsoriano, line 19: Would it be possible to modify a test package so it uses the new code?

jsoriano (commented):
Thanks! This approach looks better to me, I have added some questions and suggestions, let me know what you think.

internal/docs/exported_fields.go
stevengoossensB: Implemented suggestion

internal/docs/exported_fields.go
stevengoossensB: Did this in the nginx package, also build it locally and seems to work, although I'm getting this:
image

This comes from the package-spec repo I guess, as 3.3.1 is not defined there?

internal/docs/exported_fields.go
stevengoossensB: Makes sense. Adapted the code

internal/docs/exported_fields.go
stevengoossensB: While it's still okay in terms of formatting, it might make more sense to combine both in one column, putting the value for when it's a constant keyword and the example in all other cases. WDYT?

image
image

internal/docs/exported_fields.go
stevengoossensB: I've set the requirement to 3.3.1. It doesn't require a change to the spec, as both **value** and **example** are already allowed keys in the fields.yml file

internal/docs/exported_fields.go
jsoriano: Yes, I think it can be a good idea to put both in the same column.

internal/docs/exported_fields.go
stevengoossensB: Changes to 3.4.0

internal/docs/exported_fields.go
stevengoossensB: Changed the behaviour so that example gets added by default, unless it's a constant_keyword. In that case, the value is set.

We would require all fields to have the  **example** key set and all fields with type constant_keyword to have the **value** key set for the table to populate nicely. If none of it is available, the column won't be there.

New view:
image
image
image

jsoriano: > We would require **all** fields to have the example key set and all fields with type constant_keyword to have the value key set for the table to populate nicely.

Is it ok if some fields are not set, right? They would only have their cells empty, isn't it?

jsoriano: @kpollich what do you think about adding this column to this table? Could it cause any problem in Kibana?

stevengoossensB: > > We would require **all** fields to have the example key set and all fields with type constant_keyword to have the value key set for the table to populate nicely.
> 
> Is it ok if some fields are not set, right? They would only have their cells empty, isn't it?

Yes indeed. It would result in an empty cell

jsoriano:
@stevengoossensB CI is failing for some formatting issue. Try running `make check-static` and committing the changes.

stevengoossensB:
@jsoriano can you elaborate on why the tests are still failing as I don't have visibility on that.

jsoriano:
> @jsoriano can you elaborate on why the tests are still failing as I don't have visibility on that.

It looks like the change from spec 1.0.0 to 3.4.0 in the nginx package is triggering many validation issues. You can see them with:
```
go run . build -v -C ./test/packages/parallel/nginx/
```

I would suggest to test this with some test package that is already using a 3.X spec version. For example try increasing the `format_spec` to 3.4.0 in the `test/packages/parallel/ti_anomali` package, and run:
```
go run . changelog add -C test/packages/parallel/ti_anomali --link https://github.com/elastic/elastic-package/pull/2167 --description "Update format spec to 3.4.0" --version 1.23.0-rc1 --type enhancement
go run . build -v -C ./test/packages/parallel/ti_anomali/
```

And commit the changes.

stevengoossensB:
I reverted the changes for NGINX and upgraded Anomali.

PR #2192, opened 2024-10-30
Fix enabled: false validation

aleksmaus:
This change fixes `enabled: false` validation.

There is a use case where I need to set `enabled: false` for the field in order to disable indexing of the field but still preserve the source.
Defining the fields like this 
```
    - name: cisco_av_pair
      type: object
      enabled: false
 ```
 
 produces the correct mapping:
 ```
                       "cisco_av_pair": {
                        "type": "object",
                        "enabled": false
                      },
 ```
 
 but it fails the pipeline test with errors like these:
 ```
 FAILURE DETAILS:
cisco_ise/log test-pipeline-passed-authentications.log:
[0] field "cisco_ise.log.cisco_av_pair.audit-session-id" is undefined
[1] field "cisco_ise.log.cisco_av_pair.client-iif-id" is undefined
[2] field "cisco_ise.log.cisco_av_pair.coa-push" is undefined
[3] field "cisco_ise.log.cisco_av_pair.cts-device-capability" is undefined
[4] field "cisco_ise.log.cisco_av_pair.cts-environment-data" is undefined
[5] field "cisco_ise.log.cisco_av_pair.cts-environment-version" is undefined
[6] field "cisco_ise.log.cisco_av_pair.cts-pac-opaque" is undefined
[7] field "cisco_ise.log.cisco_av_pair.dc-certainty-metric" is undefined
[8] field "cisco_ise.log.cisco_av_pair.dc-device-class-tag" is undefined
[9] field "cisco_ise.log.cisco_av_pair.dc-device-name" is undefined
[10] field "cisco_ise.log.cisco_av_pair.dc-profile-name" is undefined
[11] field "cisco_ise.log.cisco_av_pair.dc-protocol-map" is undefined
[12] field "cisco_ise.log.cisco_av_pair.method" is undefined
[13] field "cisco_ise.log.cisco_av_pair.service-type" is undefined
```

The fix is to update the validation code path and pass validation if the parent `enabled: false` is set.

Screenshot 2024-10-30 at 5 02 08 PM

We need this change in CI eventually in order to be able to build and release the pending integration SDH fixes.

Let me know what else if anything needs to be added to this PR or if there any other pieces like docs need to be updated.

Thanks!

### DIFF
--- a/internal/fields/model.go
+++ b/internal/fields/model.go
@@ -28,6 +28,7 @@ type FieldDefinition struct {
 	MetricType     string            `yaml:"metric_type"`
 	External       string            `yaml:"external"`
 	Index          *bool             `yaml:"index"`
+	Enabled        *bool             `yaml:"enabled"`
 	DocValues      *bool             `yaml:"doc_values"`
 	Normalize      []string          `yaml:"normalize,omitempty"`
 	Fields         FieldDefinitions  `yaml:"fields,omitempty"`
--- a/internal/fields/validate.go
+++ b/internal/fields/validate.go
@@ -692,6 +692,8 @@ func (v *Validator) validateScalarElement(key string, val any, doc common.MapStr
 			return fmt.Errorf(`field %q is used as array of objects, expected explicit definition with type group or nested`, key)
 		case couldBeMultifield(key, v.Schema):
 			return fmt.Errorf(`field %q is undefined, could be a multifield`, key)
+		case !isParentEnabled(key, v.Schema):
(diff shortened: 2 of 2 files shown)


aleksmaus:
👋 This PR is needed for related package changes PR and is currently blocking it. Could anybody take look at it?

mjwolf (approved):

jsoriano:
@aleksmaus could you please add some test cases to validate the expected behaviour?

PR #2224, opened 2024-11-13
Add unit test for disabled field

aleksmaus:
Add unit test for disabled field.

This is a follow up on the comment to PR:
https://github.com/elastic/elastic-package/pull/2192#issuecomment-2470069146

@jsoriano I'm not too familiar with this project, let me know if this is sufficient.

### DIFF
--- a/internal/fields/testdata/disabled.json
+++ b/internal/fields/testdata/disabled.json
@@ -1,0 +1,6 @@
+{
+  "disabled": {
+    "id": "42",
+    "status": "ok"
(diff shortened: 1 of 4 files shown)


jsoriano (approved):
> @jsoriano I'm not too familiar with this project, let me know if this is sufficient.

These tests look good, yes.

We could also consider adding a field with `enabled: false` to a test package under `test/packages, but as you prefer.

Thanks!

PR #2226, opened 2024-11-13
Ignore empty namespaces when comparing policies

jsoriano:
Fixes issues found in https://github.com/elastic/elastic-package/pull/2222.

`namespaces` is an array of strings (https://github.com/elastic/kibana/blob/05a9b26d3c8c9932aa344e53182296996e707dd3/x-pack/plugins/fleet/server/types/models/agent_policy.ts#L251).


mrodm (approved):
Thanks!

PR #2242, opened 2024-11-22
System tests for transforms

jsoriano:
Check that transforms can work properly once installed, after documents are ingested. It ensures that the transform is executed at least once while monitoring its behaviour and looks for failures in transform stats and audit messages.

It doesn't check the validity of documents by now, we do this with the preview API, maybe in a future change we can do this validation at this point too, with real documents.


internal/testrunner/runners/system/tester.go
```diff
@@ -1978,6 +1985,248 @@ func (r *tester) previewTransform(ctx context.Context, transformId string) ([]co
...
+	const (
+		period  = 1 * time.Second
+		timeout = 60 * time.Second
+	)
+	lastSearchTime := 0
+	last := -1
+	running := false
+
+	// Reset transform to clean any previous state.
+	/* XXX: It fails to create the index after reset :?
```
jsoriano, line 2140: This is failing now because it tries to create indexes with a name that is matched by a data stream template, and transforms need to use indexes, not data streams.

:warning: This may be an actual issue because this means that the reset API breaks these transforms. We may need to review how transform indexes are named in packages to avoid these conflicts.

PR #2252, opened 2024-12-02
Review tear down test system handlers

mrodm:
This PR revisits the handlers run in the Tear Down test process for the system tests.

When using independent Elastic Agents for each test scenario, the following handlers can be skipped:
- Reassign policy back to the Elastic Agent
- Reset the log level to the Elastic Agent

When using independent Elastic Agents, those agents are going to be removed/uninstalled at the end of the test, so there is no need to run those two handlers. There is a special case where it needs to be kept the reassignment of the policy when using the `--no-provision` flag in system tests.


jsoriano (approved):
:+1:

PR #2277, opened 2024-12-13
Do not add as exception fields leaf objects

mrodm:
Follows #2214

Some fields were excluded (leaf objects) as exception fields, that it looks like they should not be.

That exception caused that some fields in `airflow` package were not detected as undocumented.

Compare:
- https://buildkite.com/elastic/integrations/builds/18608 (fields detected as undocumented)
- https://buildkite.com/elastic/integrations/builds/19362 (no fields detected)


mrodm:
test integrations

PR #2283, opened 2024-12-17
Archive windows releases in zip format

jsoriano:
Zip is more convenient in Windows systems, that don't support tar.gz out of the box.


mrodm (approved):
:+1:

Issue #2207, opened 2024-11-06
[system tests] Validate fields are documented with dynamic templates if needed


mrodm:
Follows #2206 

In order to try to be agnostic to the structure of the documents ingested to run the validation in system tests. It would be helpful to run validations comparing the mapping definitions instead.

For those mappings that can not be validated against the preview mappings in #2206, it is also needed to validate if they match with any of the dynamic templates found in the data stream.

Mappings and dynamic templates can be retrieved from these APIs:

- Mappings and dynamic templates installed by Fleet before ingesting any doc.
    - These preview mappings can be retrieved using this simulated API:
      ```
      POST /_index_template/_simulate/<index_template_name>

      # Example
      POST /_index_template/_simulate/logs-microsoft_dhcp.log
      ```     
- Mappings and dynamic templates that are present after ingesting the docs as part of the system tests.
    - These mappings can be retrieved using this API:
      ```
      GET /<data_stream_test>/_mapping/
      ```

For now, the errors that could be raised as part of this validation should not make fail the validation process, therefore:
- Enable via environment variable this mapping or show just just warning messages.

To be tested:
- Run these validations in stack 7.x
- Run these validations in stack 8.x
- Run these validations in input and integration packages.
- Run these validations in Stacks with LogsDB enabled (synthetics).

PR #2285, opened 2024-12-17
Validate fields based on their mappings and the dynamic templates set

mrodm:
Closes https://github.com/elastic/elastic-package/issues/2207

Add support to validate mappings after ingesting documents based on dynamic templates.

It also updates how mappings not found in the preview are compared/validated with ECS fields.

Assumptions:
- Filtered out dynamic templates created by `import_mappings`
- Filtered out dynamic templates created by `ecs@mappings` component template
- Filtered out the field definitions added via `appendECSMappingMultifields` function.
- All the mappings present in the actual mappings (after ingesting documents) must fulfill one of these conditions to be valid:
   - All the fields from the actual mapping must be present in the preview (not viceversa).
   - All the fields from the actual mapping must be present in the dynamic template (not viceversa).
- Multi-fields are just validated with the definitions in the preview (not in ECS or in dynamic templates).

## Author's Checklist

- [ ] Clean log/debug statements
- [x] Run tests on packages from integration repository


mrodm:
test integrations

mrodm:
test integrations

mrodm:
**:construction: WIP**

Looking at the errors in this build:
https://buildkite.com/elastic/integrations/builds/19707#_

There are some packages that are failing due to missing parameters:
- `auditd_manager`: `auditd.data.a0`, `auditd.data.a1`, `auditd.data.a2` and `auditd.data.a3`
- `teleport`: some examples: `teleport.audit.account_id`, `teleport.audit.aws_host`, `teleport.audit.aws_region`, `teleport.audit.instance_id`,...
- `docker`: some examples: `docker.cpu.core.0.norm.pct`, `docker.cpu.core.2.ticks`, `docker.cpu.core.3.pct`, ... 
    - incorrect field definitions? https://github.com/elastic/integrations/blob/51460fb00b85968f3f987cf73683946f0d821d37/packages/docker/data_stream/cpu/fields/fields.yml#L81-L99

There are packages that fail due to different types comparing with ECS:
- `box_events`: different type for `threat.enrichments.indicator.first_seen` and `threat.enrichments.indicator.last_seen`
- `claroty_ctd`: different type for `threat.indicator.modified_at`
- `mimecast`: different type for `threat.indicator.first_seen` and `threat.indicator.modified_at`
- `ti_anomali`: different type for `threat.indicator.modified_at`

Packages with issues related to multi_fields:
- `fortinet_fortimail`: missing multi_field (probably due to `ecs@mappings` component template)
- `cef`: missing multi_field (probably due to `ecs@mappings` component template)

internal/fields/validate.go
```diff
@@ -448,7 +450,7 @@ func appendECSMappingMultifields(schema []FieldDefinition, prefix string) []Fiel
 				{
 					Name:     "text",
 					Type:     "match_only_text",
-					External: "ecs",
+					External: externalFieldAppendedTag,
```
mrodm, line 453: Workaround to be able to filter multifieds generated here during the validation.

internal/fields/mappings.go
```diff
@@ -152,6 +153,7 @@ func createValidatorForMappingsAndPackageRoot(fieldsParentDir string, finder pac
 		return v, nil
 	}
 
+	// TODO: Should we remove this code to load external and local fields?
 	fieldsDir := filepath.Join(fieldsParentDir, "fields")
```
mrodm, line 157: This code to load the external and local fields maybe can be removed. That would mean that just the schema set (optionally) via options is taken into account.

WDYT?

internal/fields/mappings.go
```diff
@@ -624,22 +687,295 @@ func (v *MappingValidator) validateMappingsNotInPreview(currentPath string, chil
...
+		for i, v := range regexes {
+			r := strings.ReplaceAll(v, ".", "\\.")
+			r = strings.ReplaceAll(r, "*", ".*")
+			// Force to match beginning and ending
+			r = fmt.Sprintf("^%s$", r)
+
+			regexes[i] = r
+		}
+		return stringMatchesPatterns(regexes, elem)
+	}
```
mrodm, line 765: Added this as internal functions, would it be better to just move them out as regular functions ?

internal/fields/mappings.go
```diff
@@ -624,22 +687,295 @@ func (v *MappingValidator) validateMappingsNotInPreview(currentPath string, chil
...
+				// 	logger.Debugf("> Check unmatch_mapping_type: %q (type %s)", currentPath, fieldType)
+				// 	values, err := parseSetting(value)
+				// 	if err != nil {
+				// 		return fmt.Errorf("failed to check unmatch_mapping_type setting: %w", err)
+				// 	}
+				// 	logger.Debugf(">> Comparing to values: %s", values)
+				// 	if slices.Contains(values, fieldType) {
+				// 		logger.Debugf(">> Issue: matches")
+				// 		matched = false
+				// 	}
```
mrodm, line 924: Looks like that these options would be better to skip them, since they would require to have the original type before the documents are stored into Elasticsearch.

If there is no objection, this whole case could be left as empty (with a comment).

internal/fields/validate.go
mrodm: Would it be a better approach for this ?

internal/fields/mappings.go
jsoriano: Yes, sounds good.

internal/fields/mappings.go
```diff
@@ -624,22 +687,295 @@ func (v *MappingValidator) validateMappingsNotInPreview(currentPath string, chil
...
+		}
+		return stringMatchesPatterns(regexes, elem)
+	}
+
+	fieldType := mappingParameter("type", definition)
+	if fieldType == "" {
+		return fmt.Errorf("missing type parameter for field: %q", currentPath)
+	}
+
+	for _, template := range dynamicTemplates {
```
jsoriano, line 772: Nit. This loop mixes parsing of the dynamic templates and its execution. Consider unmarshalling the dynamic templates first, so this loop becomes something like:
```
for _, template := dynamicTemplates {
    matches, err := template.Matches(currentPath, definition)
    if err != nil {
        return err
    }
    if matches {
        return nil
    }
}
```
The parsing of the dynamic templates could be done once, in `validateMappingsNotInPreview`.

internal/fields/mappings.go
```diff
@@ -624,22 +687,295 @@ func (v *MappingValidator) validateMappingsNotInPreview(currentPath string, chil
...
+					matched = false
+					break
+				}
+
+				var matches bool
+				if fullRegex {
+					matches, err = stringMatchesPatterns(values, name)
+				} else {
+					matches, err = stringMatchesWildcards(values, name)
+				}
```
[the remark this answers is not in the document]
jsoriano: Yeah, the docs seem to indicate that it only applies to `match`, but it would be weird that it is not possible to set this option for `unmatch`. Maybe `match_pattern` applies to both?

In any case, do we generate templates with these options?

internal/fields/mappings.go
jsoriano: Yes, I think we can ignore this, but we should still check that the mapping type in the template matches with the actual mapping.

internal/fields/validate.go
jsoriano: Do we have many failures if we don't do this? Maybe we can leave this for a separate change.

internal/fields/mappings.go
mrodm: Ok, then I'll remove the code and add a comment there about this.

Comparing and validating the type would be achieved via the `validateObjectMappingAndParameters` function once it is found a dynamic template that matches the given parameters.

internal/fields/mappings.go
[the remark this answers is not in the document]
mrodm: > Yeah, the docs seem to indicate that it only applies to match, but it would be weird that it is not possible to set this option for unmatch. Maybe match_pattern applies to both?

According to the tests it looks like it applies to both parameters
https://github.com/elastic/elasticsearch/blob/fe3a3cb70e3dd18b3e5ddaf6a63389b4cb16b776/server/src/test/java/org/elasticsearch/index/mapper/DynamicTemplatesTests.java#L2391-L2397

And also it looks like it could apply to `path_match` and `path_unmatch`, lookign at this test
https://github.com/elastic/elasticsearch/blob/75bf27c8b69b7c32fb200afd5a244e7b1f8aea32/x-pack/plugin/core/template-resources/src/main/resources/watch-history.json#L23


Currently, I think that `math_pattern` parameter is not used. Fleet does not generate that and currently the [ECS component template](https://github.com/elastic/elasticsearch/blob/75bf27c8b69b7c32fb200afd5a244e7b1f8aea32/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json) does not use it neither.

mrodm:
test integrations

mrodm:
test integrations

internal/fields/mappings.go
mrodm: Separated that logic in two stages. First one parses the dynamic templates from `map[string]any` into a specific struct.

Once the dynamic templates are parsed, the required mappings are validated with those dynamic templates.

internal/fields/mappings.go
```diff
@@ -369,51 +300,61 @@ func isNumberTypeField(previewType, actualType string) bool {
...
-	// exceptions related to numbers
-	if isNumberTypeField(found.Type, actualType) {
-		logger.Debugf("Allowed number fields with different types (ECS %s - actual %s)", string(found.Type), string(actualType))
-		return nil
+	// Currently, validation of multi-fields with ECS is skipped
+	// Any multi-field found in the actual mapping must come from a definition from the preview or
+	// a dynamic template, therefore there is a definition for it.
+	// Example of this validation can be found at:
+	// https://github.com/elastic/elastic-package/pull/2285/commits/51656120
+	return nil
```
mrodm, line 323: @jsoriano do you think if I'm missing something in this note about why this validation is skipped/ignored ?

internal/fields/validate.go
mrodm: As talked offline, we decided to ignore the validation of these multi-fields.
Added a comment about the reasoning behind this in https://github.com/elastic/elastic-package/pull/2285/files#diff-67b718cebdeb840d99361c07166f75df079827a75df20850ef6cada861304521R318-R322

internal/testrunner/runners/system/tester.go
```diff
@@ -1528,12 +1528,10 @@ func (r *tester) validateTestScenario(ctx context.Context, result *testrunner.Re
 		logger.Warn("Validate mappings found (technical preview)")
 		exceptionFields := listExceptionFields(scenario.docs, fieldsValidator)
 
-		mappingsValidator, err := fields.CreateValidatorForMappings(r.dataStreamPath, r.esClient,
```
mrodm, line 1531: As the process to load the fields from the package and ECS has been removed, these parameters are not needed anymore.

The schema to be used to validate fields with ECS is the one set (optionally) via the method `WithMappingValidatorFallbackSchema`

mrodm:
test integrations

internal/fields/mappings.go
jsoriano: Looks good, thanks.

mrodm:
test integrations

mrodm:
test integrations

mrodm:
Follow-up of the above comment https://github.com/elastic/elastic-package/pull/2285#issuecomment-2553694568

Summary of the errors found in the build: https://buildkite.com/elastic/integrations/builds/20226

There are some packages that are failing due to missing parameters/definitions:
- `auditd_manager`: `auditd.data.a0`, `auditd.data.a1`, `auditd.data.a2` and `auditd.data.a3`
- `teleport`: some examples: `teleport.audit.account_id`, `teleport.audit.aws_host`, `teleport.audit.aws_region`, `teleport.audit.instance_id`,...
- `mongodb_atlas`: missing definition for two fields `mongodb_atlas.project.additional_info.hidden` and `mongodb_atlas.project.additional_info.is_mms_admin`
    - They do not match the dynamic template since they are boolean and not string.
- `crowdstrike`: errors related to flattened type
    - https://github.com/elastic/integrations/blob/80d18e9d05055786dbca2dcff1a9a8c0abb95fb3/packages/crowdstrike/data_stream/fdr/fields/fields.yml#L31
    - this is the exact error: `unexpected value found in mapping for field "crowdstrike.assessments.execution_blocking_custom_blocking_enabled_mac.type": preview mappings value ("flattened") different from the actual mappings value ("keyword")`
    - probably it does not match the dynamic template created:
      ```json
          {
            "crowdstrike.assessments.*": {
              "path_match": "crowdstrike.assessments.*",
              "mapping": {
                "type": "flattened"
              },
              "match_mapping_type": "object"
            }
          }
      ```
- `sublime_security`: field does not match the dynamic template `sublime_security.email_message.headers.hops.fields.position`
    - https://github.com/elastic/integrations/blob/80d18e9d05055786dbca2dcff1a9a8c0abb95fb3/packages/sublime_security/data_stream/email_message/fields/fields.yml#L644-L646
    - this is the exact error: `unexpected value found in mapping for field "sublime_security.email_message.headers.hops.fields.position.type": preview mappings value ("keyword") different from the actual mappings value ("long")`
    - It does not match the dynamic template:
      ```json
          "sublime_security.email_message.headers.hops.fields": {
              "path_match": "sublime_security.email_message.headers.hops.fields.*",
              "mapping": {
                "type": "keyword"
              },
              "match_mapping_type": "string"
            }
          }
      ```


There are packages that fail due to different types comparing with ECS:
- `box_events`: different type for `threat.enrichments.indicator.first_seen` and `threat.enrichments.indicator.last_seen`
- `claroty_ctd`: different type for `threat.indicator.modified_at`
- `mimecast`: different type for `threat.indicator.first_seen` and `threat.indicator.modified_at`
- `ti_anomali`: different type for `threat.indicator.modified_at`
- `cef`: different type for `url.original` (keyword vs wildcard)


Packages that do not have any test failing now:
- `docker`: mappings were fixed in Kibana since 8.12.0 https://github.com/elastic/kibana/pull/168842
- `fortinet_fortimail`: multi-fields are ignored now


While inspecting the mappings and dynamic templates created for the `docker` package, I found some discrepancies in the mappings created
taking into account the definition in the package:
- field definition:
```yaml
    - name: core.*.pct
      type: scaled_float
      format: percent
      unit: percent
      metric_type: gauge
      description: |
        Percentage of CPU time in this core.
```
- dynamic template created:
```json
        {
          "docker.cpu.core.*.pct": {
            "path_match": "docker.cpu.core.*.pct",
            "match_mapping_type": "*",
            "mapping": {
              "scaling_factor": 1000,
              "type": "scaled_float"
            }
          }
        },
```

cc @jsoriano

mrodm:
Among the errors mentioned above:

> cef: different type for url.original (keyword vs wildcard)

One option to solve this error in `cef` package is to update the ECS mappings that `elastic-package` adds automacally via `import_mappings` setting.

Currently, the embedded dynamic template created by elastic-package when `import_mappings` is enabled does not take any effect (it does not match `path_match`):
```json
        {
          "_embedded_ecs-url_original_to_multifield": {
            "path_match": "*.url.original",
            "mapping": {
              "fields": {
                "text": {
                  "type": "match_only_text"
                }
              },
              "type": "wildcard"
            }
          }
        },
``` 

This dynamic template could be updated as in the `ecs@mappings` component template (without `unmatch_mapping_type` parameter since it is not supported in 8.6.x):
```json
        {
          "ecs_path_match_wildcard_and_match_only_text": {
            "path_match": [
              "*.body.content",
              "*url.full",
              "*url.original"
            ],
            "unmatch_mapping_type": "object",
            "mapping": {
              "fields": {
                "text": {
                  "type": "match_only_text"
                }
              },
              "type": "wildcard"
            }
          }
        },
```

The inconvenient is that it would be needed to do a new `cef` release in order to include these new dynamic templates, since they are included in the built package.

WDYT about updating the embedded dynamic templates (in a follow-up PR)? Should it be created a new "empty" release in `cef` (with the new embedded dynamic templates) if this change is applied? Or just wait for the next `cef` version whenever it is created? @jsoriano

jsoriano:
> cef: different type for url.original (keyword vs wildcard)

Why is this issue not detected with current validation? :thinking:

> WDYT about updating the embedded dynamic templates (in a follow-up PR)?

We probably need to do that in any case.

> Should it be created a new "empty" release in cef (with the new embedded dynamic templates) if this change is applied? Or just wait for the next cef version whenever it is created?

We will need this to apply the new validation in integrations, we can wait till then.

mrodm:
> > cef: different type for url.original (keyword vs wildcard)
> 
> Why is this issue not detected with current validation? 🤔

Looking at the current logic to validate the fields (without mappings), IIUC `wildcards` are not validated or at least considered as not blocking due to this switch case (`parseSingleElementValue` function):
https://github.com/elastic/elastic-package/blob/a7d863ade9cb96980e8dd5cb598850ade278366c/internal/fields/validate.go#L1249-L1252

> > WDYT about updating the embedded dynamic templates (in a follow-up PR)?
> 
> We probably need to do that in any case.
> 
> > Should it be created a new "empty" release in cef (with the new embedded dynamic templates) if this change is applied? Or just wait for the next cef version whenever it is created?
> 
> We will need this to apply the new validation in integrations, we can wait till then.

As this requires the ECS templates file to be updated in `elastic-package`, I'll move forward with [this PR](https://github.com/elastic/elastic-package/pull/2337) to prepare it for review.

mrodm:
Created this PR https://github.com/elastic/integrations/pull/12300 to start testing different approaches/workarounds that could be followed in the failing packages found in the integrations repository.

As a note, that integrations PR is just for testing purposes, and it is not going to be merged as it is.

internal/fields/mappings.go
```diff
@@ -624,22 +576,61 @@ func (v *MappingValidator) validateMappingsNotInPreview(currentPath string, chil
...
+		logger.Debugf("> Check parameters (%q): %q", template.name, currentPath)
+		// Check that all parameters match (setting no dynamic templates to avoid recursion)
+		errs := v.validateObjectMappingAndParameters(template.mapping, definition, currentPath, []map[string]any{}, true)
+		if errs != nil {
+			// Look for another dynamic template
+			logger.Warnf(">> Invalid dynamic template for %q:\n%s", currentPath, errs.Unique())
+			continue
+		}
+
+		logger.Warnf("> Valid template for path %q: %q", currentPath, template.name)
```
jsoriano, line 623: This is printed for any field that matches with a dynamic template, right? Can it be too verbose?

internal/fields/mappings_test.go
```diff
@@ -117,9 +135,35 @@ func TestComparingMappings(t *testing.T) {
...
+							Type:     "match_only_text",
+							External: "ecs",
+						},
+					},
+				},
 			},
 			expectedErrors: []string{
 				`field "metrics" is undefined: actual mapping type (long) does not match with ECS definition type: keyword`,
+				`field "user" is undefined: field definition not found`,
+				// `field "time" is undefined: missing definition for multi-field in ECS: "other"`, // multi-fields are skipped
```
jsoriano, line 166: Leftover?

jsoriano (commented):
Looks good, and good coverage on templates matching. Added some small comments.

internal/fields/dynamic_template.go
```diff
@@ -0,0 +1,247 @@
...
+		name := fieldNameFromPath(currentPath)
+		if !slices.Contains(d.match, name) {
+			// If there is no an exact match, it is compared with patterns/wildcards
+			matches, err := stringMatchesPatterns(d.match, name, fullRegex)
+			if err != nil {
+				return false, fmt.Errorf("failed to parse dynamic template %q: %w", d.name, err)
+			}
+
+			if !matches {
+				// logger.Debugf(">> Issue matchi %q: not matches", d.name)
```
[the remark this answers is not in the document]
mrodm: I'll remove all these debug lines :+1:

internal/fields/mappings.go
mrodm: Yes, this is too verbose. I'll remove this one along with other debug messages.

internal/fields/dynamic_template.go
```diff
@@ -0,0 +1,247 @@
...
+			name: templateName,
+		}
+
+		// logger.Debugf("Checking dynamic template for %q: %q", currentPath, templateName)
+		contents, ok := rawContents.(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("unexpected dynamic template format found for %q", templateName)
+		}
+
+		for setting, value := range contents {
```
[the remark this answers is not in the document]
mrodm: Not sure how to take advantage of that library here.
Some fields need to take into account if they come from an string o a list of strings.

And looking at this example, https://pkg.go.dev/github.com/go-viper/mapstructure/v2#example-Decode-DecodeHookFunc

It looks like, I would need to do the same logic inside that function. For now, I think I would keep as it is if it is not an issue.

jsoriano (approved):

PR #2291, opened 2024-12-19
Add guidelines about the format version to use

jsoriano:
The use of spec versions for filtering packages on different Kibana versions make it harder to know the specific implications of each version.

Add some guidelines about the format version to use for a package.


mrodm (approved):
Thanks! :rocket:

Issue #1633, opened 2024-01-22
Support API Key as environment variable


ruflin:
Currently if connecting elastic-package to a remote Elastic Stack, only username / password as environment variable are supported. As we are moving more and more to API Keys, also an API key should be supported. See https://github.com/elastic/elastic-package/pull/1627 for more discussions.

Stretch goal: Ideally we would have an easy way to create an API Key either through and API call or in the UI in Kibana with the right permissions ...

Related PRs:
* https://github.com/elastic/elastic-package/pull/2298
* https://github.com/elastic/elastic-package/pull/2307
* https://github.com/elastic/elastic-package/pull/2308
* https://github.com/elastic/elastic-package/pull/2309
* https://github.com/elastic/elastic-package/pull/2317
* https://github.com/elastic/elastic-package/pull/2338

jsoriano:
There are currently issues installing packages with API keys https://github.com/elastic/kibana/issues/166473

bhapas:
@jsoriano Is this unblocked when https://github.com/elastic/kibana/pull/178120 is merged?

jsoriano:
> @jsoriano Is this unblocked when [elastic/kibana#178120](https://github.com/elastic/kibana/pull/178120) is merged?

Looks like that, yes.

andrewkroh:
We need API key support for interacting with Serverless projects. Unless you create the project through the API you won't have an admin user/pass and you cannot create local users. So API key is the only option.

jsoriano:
Updated description with related PRs.

Apart from changes similar to Andrew's (thanks!), I am adding an "environment" stack provider that can be used when elastic-package is configured with environment variables, to fill the gaps for system testing.

I have added related PRs to the description.

PR #2307, opened 2024-12-30
Remove serverless credentials reset

jsoriano:
We are using the internal `_reset-internal-credentials` API to obtain the credentials used for serverless stacks managed by `elastic-package`. But this API is only available to elastic employees.

There were two reasons to obtain credentials this way:
* Some APIs were restricted in serverless to credentials obtained this way. From these internal APIs, the only one we use now is the one to check the cluster status (`_cluster/health`). In this case, if normal credentials are used, it returns HTTP status 410 (Gone) if the cluster is alive and the credentials are valid, I think we can consider this as an OK for this case. This will be also needed in any case for API keys support, because API keys don't have access to these internal APIs.
* At some point during development the project creation API didn't return credentials with enough permissions, so we needed to request them using the reset credentials APIs. We are in GA now, and the project creation API returns user and password with admin privileges, what is enough for all `elastic-package` operations.

As these reasons are not strong enough now, I think it would be better if we just avoid these resets.

Relates to https://github.com/elastic/elastic-package/issues/1633, where it was observed that we need to make `elastic-package` work with non-internal credentials.

### DIFF
--- a/internal/elasticsearch/testdata/elasticsearch-serverless-healthy.yaml
+++ b/internal/elasticsearch/testdata/elasticsearch-serverless-healthy.yaml
@@ -1,0 +1,122 @@
+---
+version: 2
+interactions:
+    - id: 0
+      request:
+        proto: HTTP/1.1
+        proto_major: 1
+        proto_minor: 1
+        content_length: 0
+        transfer_encoding: []
+        trailer: {}
+        host: ""
+        remote_addr: ""
+        request_uri: ""
+        body: ""
+        form: {}
+        headers:
+            Authorization:
+                - Basic YWRtaW46cWN5V21PNTM5WTBlOWw5T1lxM3Y1TjIx
+            User-Agent:
+                - go-elasticsearch/7.17.10 (linux amd64; Go 1.23.4)
+            X-Elastic-Client-Meta:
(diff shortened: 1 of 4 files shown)


internal/elasticsearch/testdata/elasticsearch-serverless-healthy.yaml
```diff
@@ -0,0 +1,122 @@
...
+        method: GET
+      response:
+        proto: HTTP/2.0
+        proto_major: 2
+        proto_minor: 0
+        transfer_encoding: []
+        trailer: {}
+        content_length: 363
+        uncompressed: false
+        body: '{"error":{"root_cause":[{"type":"api_not_available_exception","reason":"Request for uri [/_cluster/health] with method [GET] exists but is not available when running in serverless mode"}],"type":"api_not_available_exception","reason":"Request for uri [/_cluster/health] with method [GET] exists but is not available when running in serverless mode"},"status":410}'
```
jsoriano, line 102: See here a response from serverless for the `_cluster/health` API.

mrodm (approved):
LGTM!
As you mentioned, it looks like we can rely on the 410 status to assume that the cluster is healthy :+1:

PR #2309, opened 2024-12-30
Ignore environment variables in tests for existing recordings

jsoriano:
When creating new recordings for testing, tests that already have recordings fail if the host doesn't match. This happens when recording from environments that don't use the usual local values, as for example in the test case added in https://github.com/elastic/elastic-package/pull/2307. This can even interfere with any other environment variable set by `elastic-package stack shellinit`.

With this change environment variables are only used to initialize the client when creating new recording files.

### DIFF
--- a/internal/elasticsearch/test/httptest.go
+++ b/internal/elasticsearch/test/httptest.go
@@ -5,10 +5,13 @@
 package test
 
 import (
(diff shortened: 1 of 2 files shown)


mrodm (approved):
:+1:

PR #2317, opened 2025-01-03
Add options for API key in clients

jsoriano:
Add support for API keys in clients and shellinit.

This is enough for some initial support of API keys (https://github.com/elastic/elastic-package/issues/1633), but many things can still fail, specially when running tests and so on. More about this in https://github.com/elastic/elastic-package/pull/2298.

Complementary to https://github.com/elastic/elastic-package/pull/2307 and https://github.com/elastic/elastic-package/pull/2309.

### DIFF
--- a/internal/stack/clients.go
+++ b/internal/stack/clients.go
@@ -21,6 +21,7 @@ import (
 func NewElasticsearchClient(customOptions ...elasticsearch.ClientOption) (*elasticsearch.Client, error) {
 	options := []elasticsearch.ClientOption{
 		elasticsearch.OptionWithAddress(os.Getenv(ElasticsearchHostEnv)),
+		elasticsearch.OptionWithAPIKey(os.Getenv(ElasticsearchAPIKeyEnv)),
 		elasticsearch.OptionWithPassword(os.Getenv(ElasticsearchPasswordEnv)),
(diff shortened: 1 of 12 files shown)


internal/stack/clients.go
```diff
@@ -158,5 +170,10 @@ func FindCACertificate(profile *profile.Profile) (string, error) {
 		caCertPath = profileConfig.CACertificatePath
 	}
 
+	// Avoid returning an empty certificate path, fallback to the default path.
+	if caCertPath == "" {
+		caCertPath = profile.Path(CACertificateFile)
+	}
```
mrodm, line 176: In case of Serverless projects that are not required to set the CA certificate path, would this cause errors ?

jsoriano: It shouldn't, this CA certificate is only additive. Leaving this empty mounts an empty file in the dockerfile.

mrodm: Ok :+1: 

I was wondering if it could affect in case it is used with a Serverless project that was created out of `elastic-package`. And the user sets the environment variables and it does not set the environment variable related to the CA to run for instance `elastic-package install`.

mrodm (approved):
:+1:

internal/stack/clients.go
jsoriano: I haven't found regressions in the serverless provider with these changes.

PR #2337, opened 2025-01-13
Update import mapping url.original

mrodm:


mrodm:
test integrations

jsoriano (approved):

internal/builder/_static/ecs_mappings.yaml
mrodm, line 1: Updated dynamic templates to be as similar as possible to the `ecs@mappings` component template: 
https://github.com/elastic/elasticsearch/blob/09df99393193b2c53d92899662a8b8b3c55b45cd/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json#L54-L69

Packages that rely on these dynamic templates should publish a new version in order to get the changes.

Issue #2341, opened 2025-01-14
[system tests] Validate fields in transforms are documented based on mappings


mrodm:
Follows #2207

In order to try to be agnostic to the structure of the documents ingested to run the validation in system tests. It would be helpful to run validations comparing the mapping definitions instead.

For those mappings that can not be validated against the preview mappings in https://github.com/elastic/elastic-package/issues/2206, it is also needed to validate if they match with any of the dynamic templates found in the data stream.

As part of the packages, there could be defined transforms with their own fields that should also be validated. 

Check and if possible validate those fields based on the available mappings and dynamic templates.

The destination index could be used in 

Mappings and dynamic templates can be retrieved from these APIs:

- Mappings and dynamic templates installed by Fleet before ingesting any doc.
    - These preview mappings can be retrieved using this simulated API:
      ```
      POST /_index_template/_simulate/<index_template_name>

      # Example
      POST /_index_template/_simulate/logs-microsoft_dhcp.log
      ```     
- Mappings and dynamic templates that are present after ingesting the docs as part of the system tests.
    - These mappings can be retrieved using this API:
      ```
      GET /<data_stream_test>/_mapping/
      ```

The destination index (`transforms.dest.index`) could be used in the above APIs:
```json
{
  "count": 1,
  "transforms": [
    {
      "id": "logs-ti_anomali.latest_intelligence-default-0.1.0",
      "dest": {
        "index": "logs-ti_anomali_latest.intelligence-1",
        "aliases": [
          {
            "alias": "logs-ti_anomali_latest.intelligence",
            "move_on_creation": true
          }
        ]
      },
```
The index template looks like it could also known in advance (suffix `-template`?)
```
GET /logs-ti_anomali_latest.intelligence-1/_mapping

# using index
POST /_index_template/_simulate_index/logs-ti_anomali_latest.intelligence-1

# using template 
POST /_index_template/_simulate/logs-ti_anomali.latest_intelligence-template
```

mrodm:
Looking at how tests are working and how transforms are tested by `elastic-package`, there are scenarios that could cause that not all documents are processed/validated by the transform.

For context, for every system test configuration defined in the package, the process is:
- the documents found in the data stream are validated (via fields/mappings),
- if there is any transform and it matches the given configuration, it is validated the documents processed by the transform (via fields/mappings).

So for a each system test, `elastic-package` requires to wait until some documents are processed by the transform to run the given validations for it.

[...]

PR #2347, opened 2025-01-16
[system tests] Validate transforms based on the mappings

mrodm:
Relates #2341

Add support to validate transforms created in packages (if any) based on their mappings.

As there are some packages that once the documents are processed are also deleted, it's required to be able to check whether or not its destination index has deleted docs.

Concerns:
- Testing packages defining transforms are going to be slower, since it is needed to wait for docs (or deleted docs).
    - This is required in order to have the destination index the actual mappings after being processed the docs.

## Author's Checklist

- [x] Run tests in packages from integrations repository
- [x] Requires to update after merging #2285


mrodm:
test integrations

mrodm:
test integrations

mrodm:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -1977,6 +1983,8 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
 			return fmt.Errorf("no documents found in preview for transform %q", transformId)
 		}
 
+		logger.Debugf("Found %d documents in preview for transform %q", len(transformDocs), transformId)
```
mrodm, line 1986: Probably, this message could be removed

internal/testrunner/runners/system/tester.go
```diff
@@ -1987,10 +1995,78 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
...
+				// Add specific timeout to ensure transform processes the documents (depends on "delay" field?)
+				timeout: 10 * time.Minute,
+				stopCondition: func(hits *hits, oldHits int) (bool, error) {
+					deleted, err := r.getDeletedDocs(ctx, destIndexTransform)
+					if err != nil {
+						return false, err
+					}
+					foundDeleted := deleted > 0
+					// Wait at least for the number of documents found after running the transform preview
+					foundHits := hits.size() >= len(transformDocs)
```
jsoriano, line 2034: Why waiting for this number of documents?

test/packages/parallel/ti_anomali/elasticsearch/transform/latest_ioc/transform.yml
```diff
@@ -22,7 +22,8 @@ description: Latest Anomali IoC data
 frequency: 30s
 sync:
   time:
-    field: "@timestamp"
+    # ensure that the field used to synchronize uses the current timestamp so it is processed
+    field: "event.ingested"
```
mrodm, line 26: Required to change this field in order to process the transforms the documents.
It looks like it requires a field whose timestamp is "now" and in order to be processed (as if they were ingested at that moment).

In the case of this package and this data stream, the value of the `@timestamp` field is a date of five years ago and it does not process the documents.

mrodm, line 26: Required to change this field in order to process the transforms the documents.
It looks like it requires a field whose timestamp is "now" and in order to be processed (as if they were ingested at that moment).

In the case of this package and this data stream, the value of the `@timestamp` field is a date of five years ago and it does not process the documents.

internal/testrunner/runners/system/tester.go
```diff
@@ -1987,10 +1995,78 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
...
+					if foundDeleted {
+						logger.Debugf("Found %d deleted docs in %s index", deleted, destIndexTransform)
+					}
+					if foundHits {
+						logger.Debugf("Found %d hits in %s index", hits.size(), destIndexTransform)
+					}
+					return foundDeleted || foundHits, nil
```
jsoriano, line 2041: Could it happen that there is only one deleted document, but we should wait for more hits found?

internal/testrunner/runners/system/tester.go
```diff
@@ -1987,10 +1995,78 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
...
+			destIndexTransform := transform.Definition.Dest.Index
+			// Index Template format is: "<type>-<package>.<transform>-template"
+			// For instance: "logs-ti_anomali.latest_intelligence-template"
+			indexTemplateTransform := fmt.Sprintf("%s-%s.%s-template",
+				ds.Inputs[0].Streams[0].DataStream.Type,
+				pkgManifest.Name,
+				transform.Name,
+			)
+
+			// TODO: use the docs obtained here for the fields validator too
```
jsoriano, line 2017: Do we need to check with the fields validator too if we are checking with the mappings? Or only if `allMethods` is true?

internal/testrunner/runners/system/tester.go
```diff
@@ -1987,10 +1995,78 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
...
+			if err != nil {
+				return fmt.Errorf("creating mappings validator for the transform index failed (index: %s): %w", destIndexTransform, err)
+			}
+
+			if errs := validateMappings(ctx, mappingsValidator); len(errs) > 0 {
+				return testrunner.ErrTestCaseFailed{
+					Reason:  fmt.Sprintf("one or more errors found in mappings in the transform (index %s)", destIndexTransform),
+					Details: errs.Error(),
+				}
 			}
```
jsoriano, line 2070: Nit. Could we move this code to another function?

test/packages/parallel/ti_anomali/elasticsearch/transform/latest_ioc/transform.yml
jsoriano: > whose timestamp is "now"

Being "now" the last 120s, right?

jsoriano: We will have to check if this happens in actual packages. I was thinking that the retention policy was also based on the `@timestamp`. Not sure if there can be cases where this change produces some kind of explosion of new docs :thinking:

jsoriano (commented):
Looks good. I am a bit worried about the required change on the time sync, though there are many packages using `event.ingested`, so it should be fine.

internal/testrunner/runners/system/tester.go
mrodm: At first, I was waiting for just any document (`> 0`).

But I thought that probably there could have at least the same number of documents that the ones obtained in the preview. Sometimes, I saw that there are even more documents.

I wanted to wait for some amount of documents here, so the mappings created in the destination index could be more accurate.

internal/testrunner/runners/system/tester.go
mrodm: Maybe this TODO is outdated (and could be removed), since there are some packages that could have just deleted documents in the destination index.

So, not sure if this would be possible to apply when the validation is based on fields :( 

WDYT ?

internal/testrunner/runners/system/tester.go
jsoriano: Not a strong opinion, just suggestions in case we need to make things faster.

internal/testrunner/runners/system/tester.go
jsoriano: As you prefer. My thought was this one, that we can just remove the TODO.
I guess fields validation is covered with the preview, and for mappings validation is enough with the changes in this PR.

internal/testrunner/runners/system/tester.go
mrodm: The issue here is that there are packages that there are just deleted documents and never receive any new document. That's the reason why it is here just an "or" condition.

I would need to check if there is any package with deleted and new documents in the index :eyes:

mrodm: Found a package `ti_eset` where this scenario happens.

When validating the transform, there are 50 deleted documents in the destination index, and just one document not deleted. However, there were 26 documents returned via the preview API.

So in that package, waiting for docs would exit with error due to timeout, if we would wait for the same amount of documents found in the preview.

```
2025/01/23 13:33:26 DEBUG Found 26 documents in preview for transform "logs-ti_eset.apt_latest_ioc-default-0.3.0"
...
2025/01/23 13:39:06 DEBUG Found 50 deleted docs in logs-ti_eset_latest.dest_apt-3 index
2025/01/23 13:39:11 DEBUG found 1 hits in logs-ti_eset_latest.dest_apt-3 index or data stream
....
```

mrodm: Found a package `ti_eset` where this scenario happens.

When validating the transform, there are 50 deleted documents in the destination index, and just one document not deleted. However, there were 26 documents returned via the preview API.

So in that package, waiting for docs would exit with error due to timeout, if we would wait for the same amount of documents found in the preview.

```
2025/01/23 13:33:26 DEBUG Found 26 documents in preview for transform "logs-ti_eset.apt_latest_ioc-default-0.3.0"
...
2025/01/23 13:39:06 DEBUG Found 50 deleted docs in logs-ti_eset_latest.dest_apt-3 index
2025/01/23 13:39:11 DEBUG found 1 hits in logs-ti_eset_latest.dest_apt-3 index or data stream
....
```

internal/testrunner/runners/system/tester.go
mrodm: Agreed, I'll remove the TODO :+1:

internal/testrunner/runners/system/tester.go
mrodm: I'll do some tests looking at the stats :eyes:

mrodm: I used as stop condition the number of processed docs, and it looks promising.

I kept the same `waitForDocs` function to keep checking the failure store too as it is performed with the regular data streams.

I was wondering to do a specific waiting method for the processed docs, but probably it is good to keep the check of the failure store. WDYT?

mrodm:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -2065,6 +2041,25 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
 	if err != nil {
 		return fmt.Errorf("loading transforms for package failed (root: %s): %w", r.packageRootPath, err)
 	}
+
+	// Before stack version 8.14.0, there are some issues generating the corresponding
+	// mappings for the fields defined in the transforms
+	// Forced to not use mappings to validate transforms before 8.14.0
+	// Related issue: https://github.com/elastic/kibana/issues/175331
```
mrodm, line 2048: Found some issues with the generated mappings for the transforms defined in some packages. If two group fields were defined (with different fields under it), they were not merged. This happened for instance in `github` package.

It looks like it was solved here https://github.com/elastic/kibana/pull/177608 and it is available/solved starting in stack version 8.14.0

Added for now a way to not run validated based on mappings (just in transforms).

mrodm:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -750,7 +752,7 @@ func (r *tester) getDocs(ctx context.Context, dataStream string) (*hits, error)
 		return &hits{}, nil
 	}
 	if resp.IsError() {
-		return nil, fmt.Errorf("failed to search docs for data stream %s: %s", dataStream, resp.String())
+		return nil, fmt.Errorf("failed to search docs for index or data stream %s: %s", dataStream, resp.String())
```
mrodm, line 755: Should we keep just data stream to avoid being too much verbose here? 
`index` refers when the process waits for the docs to appear in transforms.

mrodm:
test integrations

mrodm:
@jsoriano after some changes, this PR is ready for review again, thanks in advance!

internal/testrunner/runners/system/tester.go
```diff
@@ -2058,11 +2042,25 @@ func selectPolicyTemplateByName(policies []packages.PolicyTemplate, name string)
...
 		return fmt.Errorf("loading transforms for package failed (root: %s): %w", r.packageRootPath, err)
 	}
+
+	// Before stack version 8.14.0, there are some issues generating the corresponding
+	// mappings for the fields defined in the transforms
+	// Forced to not use mappings to validate transforms before 8.14.0
+	// Related issue: https://github.com/elastic/kibana/issues/175331
+	validationMethod := r.fieldValidationMethod
+	if stackVersion.LessThan(semver_8_14_0) {
+		logger.Debugf("Forced to validate transforms based on fields, not available for stack versions < 8.14.0")
```
jsoriano, line 2057: :+1:

internal/testrunner/runners/system/tester.go
```diff
@@ -2130,63 +2129,67 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
...
-			if errs := validateMappings(ctx, mappingsValidator); len(errs) > 0 {
-				return testrunner.ErrTestCaseFailed{
-					Reason:  fmt.Sprintf("one or more errors found in mappings in the transform (index %s)", destIndexTransform),
-					Details: errs.Error(),
-				}
+			if processed > 0 {
+				logger.Debugf("Documents processed by transform %q: %d", transformId, processed)
 			}
-		}
+			return processed >= len(transformDocs), nil
```
jsoriano, line 2160: Looks like we are comparing different things here:
* `len(transformDocs)` is the number of documents resulting from the preview.
* `processed` is the number of processed documents from the source data stream/s.

So, if I understand correctly, `processed` can be bigger than `len(transformDocs)` before all the documents in the data stream are processed.

I would expect this to stop when `processed >= number of docs in the source index at some point to be decided`.

jsoriano: Yes, I mean to compare with the number of documents in the source data stream. What I am not so sure is when to check for this number.

> This number of documents in the data stream could be added as first step into validateTransformsWithMappings.

If we are sure that we have enough documents at this point it would be perfect, yes.

mrodm: > > This number of documents in the data stream could be added as first step into validateTransformsWithMappings.
> 
> If we are sure that we have enough documents at this point it would be perfect, yes.

Not totally 100% sure, but currently I cannot think of any other better location. This could vary for each package, since there could be packages that are ingesting documents continuously or they do not have ingested all the required documents in the data stream validation (related to https://github.com/elastic/elastic-package/issues/2378).

jsoriano (approved):

internal/testrunner/runners/system/tester.go
```diff
@@ -2139,7 +2139,12 @@ func (r *tester) checkTransforms(ctx context.Context, stackVersion *semver.Versi
 	return nil
 }
 
-func (r *tester) validateTransformsWithMappings(ctx context.Context, transformId, transformName, destIndexTransform, indexTemplateTransform string, transformDocs []common.MapStr, fieldsValidator *fields.Validator) error {
+func (r *tester) validateTransformsWithMappings(ctx context.Context, sourceDataStream, transformId, transformName, destIndexTransform, indexTemplateTransform string, transformDocs []common.MapStr, fieldsValidator *fields.Validator) error {
+	logger.Debugf("Searching the number of documents found in source data stream %q before validating transform %q", sourceDataStream, transformId)
+	sourceDataStreamHits, err := r.getDocs(ctx, sourceDataStream)
```
jsoriano, line 2144: Umm, thinking about https://github.com/elastic/elastic-package/issues/2341#issuecomment-2621974145, we should probably use the source index here, because in cases where there are differences we would be also comparing different things.

mrodm: IIUC the source index in this context (where the transform is reading the documents to process) is the source data stream of the running test. If so, this would mean the data stream that the `scenario.dataStream` variable refers to.

Is that what you meant here?

jsoriano: `sourceDataStream` is the `scenario.dataStream`, right?

The source index in a transform is a pattern, that can match multiple indexes or data streams. As mentioned in https://github.com/elastic/elastic-package/issues/2341#issuecomment-2621974145 there can be several data streams matching the same transform source index. In the case of system tests, I guess this means that the source index can match other data streams apart of `sourceDataStream`/`scenario.dataStream`.

So here we are getting the docs written by the test into its data stream, and using its count to decide when to stop the test. We are comparing the number of docs in the scenario data stream with the documents of all the matching data streams.

In the case where there are more than one data streams matching, we might be continuing even if none of the expected documents has been processed.

I think we should use here the transform source index to be sure that we check the transform after all the documents in all datastreams (including the scenario one) are processed. But maybe this is quite a corner case, not sure about the cases we have.

mrodm:
Closing this PR. This approach needs to be revisited (more info at https://github.com/elastic/elastic-package/issues/2394)

PR #2355, opened 2025-01-17
Return error when profile cannot be loaded

jsoriano:
Using a non existing profile can panic because an error is being ignored.


mrodm (approved):
:+1:

PR #2356, opened 2025-01-17
Check compose status only with compose provider

jsoriano:
When creating clients from profile, check the status of the compose project only when using the compose stack provider. This prevents elastic-package trying to find local containers when using other providers.

I am not adding explicit support for other providers because they would need the client itself to check the status, what would lead to infinite recursion.


jsoriano:
test serverless

mrodm (commented):
:+1:

mrodm (approved):

Issue #2365, opened 2025-01-22
Kibana client checks local CA cert and fails on install command if it's not available.


m-a-leclercq:
Hi team, 

I'm working with `elastic-package` in an environment where I'm connecting to a remote Kibana cluster and therefore not using `stack up`

Exporting environment variables with a cluster results in the following error when running `install` : 

`Error: could not create kibana client: reading CA certificate: failed to read certificate in "/root/.elastic-package/profiles/default/certs/ca-cert.pem": open /root/.elastic-package/profiles/default/certs/ca-cert.pem: no such file or directory`

Running `stack up` even if docker is not installed creates a `ca-cert.pem` and fixes the issue. 

Thanks

m-a-leclercq:
Just in case someone finds this in the future, the correct variable to set is : `ELASTIC_PACKAGE_CA_CERT=` for the workaround.

PR #2366, opened 2025-01-22
Don't use default CA file if profile does not exist

jsoriano:
When using `elastic-package` with environment variables with a profile
without a configured stack, avoid using the default CA certificate
file if it neither exists.

Fix https://github.com/elastic/elastic-package/issues/2365

### DIFF
--- a/internal/stack/config.go
+++ b/internal/stack/config.go
@@ -43,7 +43,7 @@ func configPath(profile *profile.Profile) string {
 	return profile.Path(ProfileStackPath, configFileName)
 }
 
-func defaultConfig(profile *profile.Profile) Config {
+func defaultConfig() Config {
 	return Config{
 		Provider: DefaultProvider,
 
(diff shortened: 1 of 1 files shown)


internal/stack/config.go
```diff
@@ -52,14 +52,19 @@ func defaultConfig(profile *profile.Profile) Config {
...
 
 func LoadConfig(profile *profile.Profile) (Config, error) {
 	d, err := os.ReadFile(configPath(profile))
 	if errors.Is(err, os.ErrNotExist) {
-		return defaultConfig(profile), nil
+		config := defaultConfig()
+		caCertFile := profile.Path(CACertificateFile)
+		// Use CA file in the profile only if it exists.
+		if _, err := os.Stat(caCertFile); err == nil {
+			config.CACertFile = caCertFile
```
jsoriano, line 65: This is likely never going to happen. This can only happen with data directories created with some very old versions of elastic-package, that didn't have profile config, but had certificates.

It would be probably fine to just don't set the CA cert file, keeping it just in case.

mrodm (approved):

PR #2368, opened 2025-01-23
ci(buildkite): exclude files/folders that are not tested in Buildkite

v1v:
## What does this PR do?

Skip files that are not used in Buildkite

## Why is it important?

Run faster builds and avoid waste of CI cycles.

I decided to support the whole `.github` folder since some other files, such as CODEOWNERS or .dependabot.yml, and the rest are not used by Buildkite.

The same for the `docs`, IIUC, that's not tested in BK itself but a BK docs pipeline owned by the docs team

### DIFF
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -13,7 +13,12 @@
         "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))|^/test$",
         "skip_ci_labels": [ ],
         "skip_target_branches": [ ],
-        "skip_ci_on_only_changed": [ ],
+        "skip_ci_on_only_changed": [
+          "^docs/",
+          "^.github/",
+          "^.mergify.yml",
+          "\\.md$"
+        ],
         "always_require_ci_on_changed": [ ]
      },
      {
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -16,8 +16,7 @@
         "skip_ci_on_only_changed": [
           "^docs/",
           "^.github/",
-          "^.mergify.yml",
-          "\\.md$"
+          "^.mergify.yml"
         ],
         "always_require_ci_on_changed": [ ]
      },


.buildkite/pull-requests.json
```diff
@@ -13,7 +13,12 @@
         "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))|^/test$",
         "skip_ci_labels": [ ],
         "skip_target_branches": [ ],
-        "skip_ci_on_only_changed": [ ],
+        "skip_ci_on_only_changed": [
+          "^docs/",
+          "^.github/",
+          "^.mergify.yml",
+          "\\.md$"
```
mrodm, line 20: `README.md` in the root of the directory is generated in the CI steps.

Probably it needs to be removed from here in case someone updates it but not modifies the corresponding template:
https://github.com/elastic/elastic-package/tree/main/tools/readme

mrodm: About the others I don't recall now any usage of them in the Buildkite builds.

Shuold we remove any other regex @jsoriano?

v1v, line 20:
```suggestion
          "^.mergify.yml"
```

v1v: see [`b0a5095` (#2368)](https://github.com/elastic/elastic-package/pull/2368/commits/b0a5095e11d6bcef5aa825ae77000146f84e0997)

mrodm:
One doubt here @v1v, in this repository, the [elastic-package Buildkite pipeline](https://buildkite.com/elastic/elastic-package) is a required check status in Pull Requests to be able to merge it. The same configuration is applied in other repositories like [integrations](https://github.com/elastic/integrations) or [package-spec](https://github.com/elastic/package-storage-infra) (in their corresponding pipelines).

What would happen for those Pull Requests that do not trigger the given build? 

IIUC, if this PR is merged, all the following PRs updating just those files should be merged bypassing branch protections. Is that the expected approach?

v1v:
That's a good question. I created https://github.com/elastic/beats/pull/42437 and it passed the GH check:

image

So the GH check should be the name of the BK pipeline rather than a subset of the steps.

Is that the case?

mrodm:
> That's a good question. I created [elastic/beats#42437](https://github.com/elastic/beats/pull/42437) and it passed the GH check:
> 
> image
> So the GH check should be the name of the BK pipeline rather than a subset of the steps.
> 
> Is that the case?

Perfect! it marks the pipeline status as successful :+1: In our case, we just have one status check per pipeline, so that's good.

Thanks for checking this out!

mrodm (approved):

Issue #2342, opened 2025-01-14
Check that package installation doesn't cause deprecation warnings


jsoriano:
Elasticsearch can log warnings when deprecated features are used. We could introduce new checks to avoid the use of deprecated features in packages, and to check in daily builds if Fleet needs to be adapted to migrate packages.

For example we found cases of uses of the deprecated `_source.mode` setting: https://github.com/elastic/kibana/issues/206307.

rudolf:
See also https://github.com/elastic/dev/issues/2819

PR #2369, opened 2025-01-23
Check index deprecation warnings on system tests

jsoriano:
Use the migration APIs to get deprecation warnings for the data stream under test.

Fixes https://github.com/elastic/elastic-package/issues/2342.

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -874,18 +874,85 @@ func (r *tester) getFailureStoreDocs(ctx context.Context, dataStream string) ([]
 	return docs, nil
 }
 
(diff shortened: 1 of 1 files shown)


jsoriano:
test integrations

kpollich:
This _should_ be fixed by https://github.com/elastic/kibana/pull/207133 which will make 8.17.2

jsoriano:
Tested in 8.18.0-SNAPSHOT and it is effectively fixed: https://buildkite.com/elastic/integrations/builds/20866

So I will filter out 8.17.0 and 8.17.1 from these warnings.

mrodm (approved):

PR #2377, opened 2025-01-28
Validate deprecation warnings after checking number of docs ingested

mrodm:
Follows #2369 

Check deprecation warnings after being validated that there are at least one document in the index/data stream (so it is created).

Errors in CI daily jobs when there are no documents (index/data stream is not created): https://buildkite.com/elastic/integrations/builds/21038#0194aa90-57eb-41b1-83ee-387c64d0ba92/641-3441

Buildkite applying this change (errors is reported as "not found hits"):
https://buildkite.com/elastic/integrations/builds/21071


mrodm:
test integrations

jsoriano (approved):
Thanks!

PR #2388, opened 2025-02-04
[system tests] [transforms] Support commas in the source indices definition

mrodm:
Relates #2341 

Support to check transforms which source indices definition contains a list of comma-separated strings.

For instance ([link](https://github.com/elastic/integrations/blob/7801e00866ec0b5d4f4e4215f6b1107851083c87/packages/elasticsearch/elasticsearch/transform/index_pivot/transform.yml#L2))
```yaml
source:
  index: "metrics-elasticsearch.stack_monitoring.index*,.monitoring-es-8*,metricbeat-*"
```

`elastic-package` needs to check each source index defined in the transform independently to validate whether or not it matches with the data stream tested.

Notes:
- Discovered that Fleet creates the transforms with `logs-` prefix
- Added a configuration to skip the validation of transforms (required for `elasticsearch` package).
    - In 8.10.1, the destination index is not created:
       - Even, if the destination index is created beforehand, it fails with the same error.
       - Probably related to the aggregations ? [link](https://github.com/elastic/elasticsearch/blob/f77201fb128cbbe2c1c8920662631c54fa451b8a/x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/common/AbstractCompositeAggFunction.java#L90)
       - Error: `"Source indices have been deleted or closed."`
    - It looks like that the filters are too restrictive for the documents ingested in system test.
      Error: `"must specify at least one document in [docs]"`

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -2080,7 +2080,10 @@ func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgMan
 		// IDs format is: "<type>-<package>.<transform>-<namespace>-<version>"
 		// For instance: "logs-ti_anomali.latest_ioc-default-0.1.0"
 		transformPattern := fmt.Sprintf("%s-%s.%s-*-%s",
-			ds.Inputs[0].Streams[0].DataStream.Type,
+			// It cannot be used "ds.Inputs[0].Streams[0].DataStream.Type" since Fleet
+			// always create the transform with the prefix "logs-"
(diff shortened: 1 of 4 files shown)


mrodm:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -2057,6 +2057,10 @@ func selectPolicyTemplateByName(policies []packages.PolicyTemplate, name string)
 }
 
 func (r *tester) checkTransforms(ctx context.Context, config *testConfig, pkgManifest *packages.PackageManifest, ds kibana.PackageDataStream, dataStream string, syntheticEnabled bool) error {
+	if config.SkipTransformValidation {
+		return nil
+	}
+
```
mrodm, line 2063: Due to the errors found in the `elasticsearch` package from integrations (see PR description) when testing to check each index independently, it looks like it would be needed to add this kind of setting in the system tests.

Issue #2387, opened 2025-01-31
Unit tests are dependent on the local docker stack


m-a-leclercq:
Hi team, 

My understanding is that `test pipeline` relies heavily on the local stack as evidenced by these debug logs: 
```
2025/01/31 16:43:40 DEBUG running command: /usr/bin/docker compose -f /root/.elastic-package/profiles/default/stack/docker-compose.yml -p elastic-package-stack logs --since 2025-01-31T16:43:40Z elasticsearch
```

These logs seem to be necessary for perfect test coverage (getting a PASS does not require there be no warnings, but it's still mentioned when running the command)

```
╭─────────┬─────────────┬───────────┬────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE │ DATA STREAM │ TEST TYPE │ TEST NAME                                  │ RESULT │ TIME ELAPSED │
├─────────┼─────────────┼───────────┼────────────────────────────────────────────┼────────┼──────────────┤
│ sample  │ log         │ pipeline  │ (ingest pipeline warnings test-sample.log) │ PASS   │ 298.396643ms │
│ sample  │ log         │ pipeline  │ test-sample.log                            │ PASS   │ 100.341331ms │
╰─────────┴─────────────┴───────────┴────────────────────────────────────────────┴────────┴──────────────╯
```

A simple unit test with the bare minimum will generate the following lines: 
```
root@elastic-package:/home/elastic/integration/sample# /usr/bin/docker compose -f /root/.elastic-package/profiles/default/stack/docker-compose.yml -p elastic-package-stack logs --since 2025-01-31T16:43:40Z elasticsearch
WARN[0000] The "KIBANA_IMAGE_REF" variable is not set. Defaulting to a blank string. 
WARN[0000] The "PROFILE_NAME" variable is not set. Defaulting to a blank string. 
WARN[0000] The "PROFILE_NAME" variable is not set. Defaulting to a blank string. 
WARN[0000] The "ELASTIC_AGENT_IMAGE_REF" variable is not set. Defaulting to a blank string. 
WARN[0000] The "ELASTICSEARCH_IMAGE_REF" variable is not set. Defaulting to a blank string. 
WARN[0000] The "ELASTIC_AGENT_IMAGE_REF" variable is not set. Defaulting to a blank string. 
service "elasticsearch" has neither an image nor a build context specified: invalid compose project
```

[...]

PR #2391, opened 2025-02-10
Tolerate errors dumping logs in pipeline tests

jsoriano:
Errors are ignored in two more cases now: when the status of the stack cannot be obtained, and when using the compose provider, when the elasticsearch service doesn't exist.

This should help on executions of pipeline tests against existing servers using only environment variables on default configuration, what is a common use case according to https://github.com/elastic/elastic-package/issues/2387. 

Fixes https://github.com/elastic/elastic-package/issues/2387.


mrodm (approved):
LGTM

PR #2395, opened 2025-02-11
Refactor mapping response given by the Elasticsearch client

mrodm:
Relates #2207 

This PR creates a new struct named `Mappings` to return both dynamic templates and properties for the corresponding methods in the Elasticsearch client.

This would help in case it is needed to add more fields from the Elasticsearch API response (runtime fields, settings, etc.)


jsoriano (approved):
:+1:

Issue #1229, opened 2023-04-20
Add support for runtime fields


mrodm:
Relates https://github.com/elastic/package-spec/issues/39

Once runtime fields are allowed as part of fields in packages, it needs to be added the needed support in elastic-package:
- Check those runtime fields during the system tests (check the type and existance of those in the `fields` from the documents retrieved from Elasticsearch (https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime.html)
  > Use the fields parameter on the _search API to [retrieve the values of runtime fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-retrieving-fields.html). Runtime fields won’t display in _source, but the fields API works for all fields, even those that were not sent as part of the original _source.
- Add some label or mark in the generated Readmes to identify which fields are runtime fields. Some possible options for this:
    - Add a new column:
      | Field | Description | Type | Runtime |
      |---|---|---|---|
      | @timestamp | Event timestamp. | date | True |
    - Add `runtime` in the column **Type**:
      | Field | Description | Type |
      |---|---|---|
      | @timestamp | Event timestamp. | date (runtime) |

mrodm:
Following what commented in https://github.com/elastic/elastic-package/issues/1111#issuecomment-1571834644 , adding a new column or more information could worsen the readibily of the docs itself. And it looks like that those runtime fields could be transparent for the users of this integration/input package.

mrodm:
While working on #2208 
We have observed that if there are some failures in runtime fields, they are already reported by the tests or kibana.
Examples tested:
- Compile error: typo in function `em` instead of `emit`
    - Definition:
      ```yaml
      - name: day_of_week
        type: keyword
        runtime: "em(doc['@timestamp'].value.dayOfWeekEnum.getDisplayName(TextStyle.FULL, Locale.ENGLISH))"
      ```
    - Error:
      ```
      Error: failed to setup system runner: can't install the package: there was an apply error: installation failed: can't install the package: could not zip-install package; API status code = 500; response body = {"statusCode":500,"error":"Internal Server Error","message":"mapper_parsing_exception\n\tCaused by:\n\t\tscript_exception: compile error\n\tRoot causes:\n\t\tscript_exception: compile error"}
      ```
- Field does not exist: `@timap` instead of `@timestamp`:
     - Definition:
      ```yaml
      - name: day_of_week
        type: keyword
[...]

PR #2409, opened 2025-02-14
Add support for runtime fields and test packages

### LINKED ISSUE #2039: Report unhealthy containres as part of the jUnit test failures
Sometimes running system tests could fail because the service used for testing cannot be started or is unhealhty.

Example of this could be checked in this build:
https://buildkite.com/elastic/integrations/builds/14993#0191872e-6ce6-42fe-bea3-a48376ff96d8

Service was unhealthy after the default timeout and that was reported by elastic-package:
`Error: error running package system tests: could not complete test run: could not setup service: service is unhealthy: timeout waiting for healthy container`

The issue is that this error is not part of the test failures reported in the XML (jUnit). It would be better to add this kind of errors as another error in the corresponding test case.

Something to remark here, is that even if these errors are added into the test files, not sure if the test run could continue. Currently, this kind of errors makes fail the whole execution of tests.

### LINKED ISSUE #2208: [system tests] Take advantage of the runtime mappings to validate the runtime fields present in the docs
Relates https://github.com/elastic/elastic-package/issues/1229
Follows https://github.com/elastic/elastic-package/issues/2206

In order to try to be agnostic to the structure of the documents ingested to run the validation in system tests. It would be helpful to run validations comparing the mapping definitions instead.

Use the mappings found in the preview template or in the data stream API to validate the runtime fields present in the hits returned by Elasticsearch as part of the system tests.
```
# preview mappings API
POST /_index_template/_simulate/<index_template_name>

# Example
POST /_index_template/_simulate/logs-microsoft_dhcp.log

# actual data stream mappings API
GET /<data_stream_test>/_mapping/
```

For now, the errors that could be raised as part of this validation should not make fail the validation process, therefore:
- Ignore at this step mappings not found in the preview, to be done in the next step.
- Enable via environment variable this mapping or show just just warning messages.

To be tested:
- Run these validations in stack 7.x
- Run these validations in stack 8.x
- Run these validations in input and integration packages.
- Run these validations in Stacks with LogsDB enabled (synthetics).

mrodm:
Relates #1229
Closes #2208
Closes #2039

This PR adds support for runtime fields.
- Added test packages related to runtime fields.
- Avoid errors if dynamic templates are set with `runtime` setting.

Additional:
- All errors raised in `prepareScenario` (system tests) are now reported as errors in the xUnit file.


mrodm:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -1746,7 +1746,14 @@ func (r *tester) runTest(ctx context.Context, config *testConfig, stackConfig st
 
 	scenario, err := r.prepareScenario(ctx, config, stackConfig, svcInfo)
 	if err != nil {
-		return result.WithError(err)
+		// Known issue: do not include this as part of the xUnit results
+		// Example: https://buildkite.com/elastic/integrations/builds/22313#01950431-67a5-4544-a720-6047f5de481b/706-2459
+		if strings.Contains(err.Error(), "fork/exec /usr/bin/docker: operation not permitted") {
```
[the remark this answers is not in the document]
mrodm: Thanks!

internal/testrunner/runners/system/tester.go
```diff
@@ -1746,7 +1747,15 @@ func (r *tester) runTest(ctx context.Context, config *testConfig, stackConfig st
...
+		// Known issue: do not include this as part of the xUnit results
+		// Example: https://buildkite.com/elastic/integrations/builds/22313#01950431-67a5-4544-a720-6047f5de481b/706-2459
+		var pathErr *fs.PathError
+		if errors.As(err, &pathErr) && pathErr.Op == "fork/exec" && pathErr.Path == "/usr/bin/docker" {
+			return result.WithError(err)
+		}
+		// report all other errors as error entries in the xUnit file
+		results, _ := result.WithError(err)
+		return results, nil
 	}
```
mrodm, line 1759: This is based on https://github.com/elastic/elastic-package/blob/3aaadce01559928219d68d3c0ed0b7225b6170f1/internal/testrunner/runners/static/tester.go#L118-L122

WDYT @jsoriano ? 

This would be required to allow to catch the errors of the false positive test package introduced in this PR. I guess it could be updated the script to get the latest line something like from the stderr output (not tested).

As advantage, this would help to report unhealthy containers as errors (related to #2039).

The disadvantage of this is that any error from `prepareScenario`  would be reported as an error in xUnit. Just added here the exception for the docker error that it is a known issue. It happens from time to time in CI builds.

Should we do the same with the response of `validateTestScenario`? For this one, I think I would keep it as it is for now.
https://github.com/elastic/elastic-package/pull/2409/files#diff-0bdd696fb9abd5e4aa12a3bab634cbac31ec1a1349b3a61e8a784a6bdeb6e7d8R1766

jsoriano: It LGTM like this.

As we have the CI now, everything that goes to the xunit files sounds good to me, as this makes easier to find what failed in jobs.

jsoriano (approved):

PR #2410, opened 2025-02-14
Enable validation based on mappings by default

mrodm:
This PR enables the validation based on mappings by default.

This validation will be run along with the validation based on fields. The latter is the validation performed until now.

Additionally, CI steps related to test packages adding the validation based on mappings have been deleted, since now all the packages should be tested with both (validated based on fields and mappings).

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -288,8 +288,8 @@ func NewSystemTester(options SystemTesterOptions) (*tester, error) {
 		r.runIndependentElasticAgent = strings.ToLower(v) == "true"
 	}
 
-	// default method using just fields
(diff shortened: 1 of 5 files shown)


internal/testrunner/runners/system/tester.go
```diff
@@ -1646,7 +1646,7 @@ func (r *tester) validateTestScenario(ctx context.Context, result *testrunner.Re
 	}
 
 	if r.fieldValidationMethod == mappingsMethod {
-		logger.Warn("Validation based on mappings enabled (technical preview)")
+		logger.Debug("Performing validation based on mappings")
```
mrodm, line 1649: I thought that maybe this message could help to identify if the validation based on mappings has been executed or not. WDYT ?

jsoriano (approved):

PR #2415, opened 2025-02-17
Runtime field docs - WIP

### LINKED ISSUE #1229: Add support for runtime fields

mrodm:
Closes #1229 

This PR updates the generation of fields documentation to include the runtime flag in the corresponding field definitions.

### DIFF
--- a/internal/docs/exported_fields.go
+++ b/internal/docs/exported_fields.go
@@ -18,6 +18,7 @@ type fieldsTableRecord struct {
 	aType       string
 	unit        string
 	metricType  string
+	runtime     bool
 }
 
 var escaper = strings.NewReplacer("*", "\\*", "{", "\\{", "}", "\\}", "<", "\\<", ">", "\\>")
@@ -74,10 +75,14 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord) 
 	builder.WriteString("\n")
 	for _, c := range collected {
 		description := strings.TrimSpace(strings.ReplaceAll(c.description, "\n", " "))
+		fieldType := c.aType
+		if c.runtime {
+			fieldType = fmt.Sprintf("%s (runtime)", c.aType)
+		}
 		builder.WriteString(fmt.Sprintf("| %s | %s | %s |",
 			escaper.Replace(c.name),
 			escaper.Replace(description),
-			c.aType))
+			fieldType))
 		if unitsPresent {
 			builder.WriteString(fmt.Sprintf(" %s |", c.unit))
 		}
@@ -117,7 +122,7 @@ func collectFieldsFromDefinitions(validator *fields.Validator) []fieldsTableReco
 }
 
 func visitFields(namePrefix string, f fields.FieldDefinition, records []fieldsTableRecord) []fieldsTableRecord {
-	var name = namePrefix
+	name := namePrefix
 	if namePrefix != "" {
 		name += "."
 	}
@@ -130,6 +135,7 @@ func visitFields(namePrefix string, f fields.FieldDefinition, records []fieldsTa
 			aType:       f.Type,
 			unit:        f.Unit,
 			metricType:  f.MetricType,
+			runtime:     f.Runtime.IsEnabled(),
 		})
 
 		for _, multiField := range f.MultiFields {
--- a/internal/docs/exported_fields.go
+++ b/internal/docs/exported_fields.go
@@ -79,15 +79,15 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord) 
 		if c.runtime {
 			fieldType = fmt.Sprintf("%s (runtime)", c.aType)
 		}
-		builder.WriteString(fmt.Sprintf("| %s | %s | %s |",
+		fmt.Fprintf(builder, "| %s | %s | %s |",
 			escaper.Replace(c.name),
 			escaper.Replace(description),
-			fieldType))
+			fieldType)
 		if unitsPresent {
-			builder.WriteString(fmt.Sprintf(" %s |", c.unit))
+			fmt.Fprintf(builder, " %s |", c.unit)
 		}
 		if metricTypesPresent {
-			builder.WriteString(fmt.Sprintf(" %s |", c.metricType))
+			fmt.Fprintf(builder, " %s |", c.metricType)
 		}
 		builder.WriteString("\n")
 	}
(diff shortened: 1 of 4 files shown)


internal/fields/model.go
```diff
@@ -44,82 +43,6 @@ type ReusableConfig struct {
 	TopLevel bool `yaml:"top_level"`
 }
 
-func (orig *FieldDefinition) Update(fd FieldDefinition) {
```
mrodm, line 47: I didn't find any usages of this `Update` function around the code.

It looks like that it was removed in this PR: https://github.com/elastic/elastic-package/pull/1335/files#diff-77cc155ee037ac620b1e188302f60fdf736637a25e5021e76dbad83594edd09cL133

internal/docs/exported_fields.go
```diff
@@ -74,15 +75,19 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord)
...
+			builder.WriteString(" (runtime)")
+		}
+		builder.WriteString(" |")
 		if unitsPresent {
-			builder.WriteString(fmt.Sprintf(" %s |", c.unit))
+			fmt.Fprintf(builder, " %s |", c.unit)
 		}
 		if metricTypesPresent {
-			builder.WriteString(fmt.Sprintf(" %s |", c.metricType))
+			fmt.Fprintf(builder, " %s |", c.metricType)
```
mrodm, line 90: Update how this string is built by using fmt.Fprintf. Suggested by this static check QF1012 https://staticcheck.dev/docs/checks/#QF1012

internal/fields/model_test.go
mrodm, line 1: Removed test (file) related to the `Update` function.

mrodm:
In this PR, it is implemented that the runtime flag is added next to the type of the field:
```markdown
| day_of_week | Day of week. | keyword (runtime) |  |  |
| labels.\* |  | keyword (runtime) |  |  |
| package_registry.day_of_week | Other day of week. | keyword (runtime) |  |  |
| package_registry.day_of_week2 | Day of week 2. | keyword (runtime) |  |  |
| package_registry.labels.version | Elastic Package Registry version. | keyword (runtime) |  |  |
```

Another option would be to set the runtime flag in the first column (field name). It could be like this:
```markdown
| day_of_week (runtime) | Day of week. | keyword |  |  |
| labels.\* (runtime) |  | keyword|  |  |
| package_registry.day_of_week (runtime) | Other day of week. | keyword |  |  |
| package_registry.day_of_week2 (runtime) | Day of week 2. | keyword |  |  |
| package_registry.labels.version (runtime) | Elastic Package Registry version. | keyword |  |  |
```

Which of the two ways do you think shows better the runtime fields to the user?

jsoriano:
> Which of the two ways do you think shows better the runtime fields to the user?

I think it would be better in the type? :thinking: But not a strong opinion, as you prefer.

internal/docs/exported_fields.go
```diff
@@ -74,10 +75,14 @@ func renderFieldsTable(builder *strings.Builder, collected []fieldsTableRecord)
...
 		description := strings.TrimSpace(strings.ReplaceAll(c.description, "\n", " "))
-		builder.WriteString(fmt.Sprintf("| %s | %s | %s |",
+		builder.WriteString(fmt.Sprintf("| %s | %s | %s",
 			escaper.Replace(c.name),
 			escaper.Replace(description),
 			c.aType))
+		if c.runtime {
+			builder.WriteString(" (runtime)")
+		}
+		builder.WriteString(" |")
```
jsoriano, line 85: Nit. To keep this formatted string simpler maybe we can build the type before.
```
		fieldType := c.aType
		if c.runtime {
			fieldType += " (runtime)"
		}
		builder.WriteString(fmt.Sprintf("| %s | %s | %s |",
			escaper.Replace(c.name),
			escaper.Replace(description),
			fieldType))
```

internal/fields/model.go
jsoriano: Interesting, this used to be neccesary :thinking: But good if we don't need it anymore.

internal/fields/runtime_field.go
```diff
@@ -0,0 +1,56 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
jsoriano, line 1: Nit. Avoid using underscores in Go files.

jsoriano (approved):
Looks good, but can this break current integrations builds? Or there is no package using `runtime`?

internal/fields/runtime_field.go
mrodm: I've updated this file and I took the advantage of renaming `dynamic_template.go` and `exception_fields.go` files too

mrodm: Changes done in https://github.com/elastic/elastic-package/pull/2415/commits/37f613c84cede484c8298d992066043372ab7b64

internal/fields/model.go
mrodm: At least, that function is not called anymore...

mrodm:
test integrations

mrodm:
> * Just to be sure, triggered a new integrations PR to test all packages: [Test elastic-package#2415 - DO NOT MERGE integrations#12824](https://github.com/elastic/integrations/pull/12824)

Packages failing in that PR are the expected ones related to validation based on mappings:
https://buildkite.com/elastic/integrations/builds/22461

jsoriano (approved):

PR #2429, opened 2025-02-25
[system tests] Add new settings to wait for documents being ingested

### LINKED ISSUE #2378: [system tests] Add new options to wait for documents before validating fields/mappings
Currently, system tests wait for at least one document being ingested into Elasticsearch before validating the fields or mappings.

https://github.com/elastic/elastic-package/blob/93c7a8d7c88b000fe5c1266bc06c01b716a5e680/internal/testrunner/runners/system/tester.go#L1396

This behavior can be tuned thanks to the setting `assert.hit_count`. This setting changes the process to wait for an exact number of documents being ingested into Elasticsearch.

https://github.com/elastic/elastic-package/blob/93c7a8d7c88b000fe5c1266bc06c01b716a5e680/docs/howto/system_testing.md?plain=1#L470

These two methods solve most of the usages, but there are some scenarios where packages need to wait for a unknown number of documents:
- This would help to validate a better representation of the fields being ingested by elastic-agent (and therefore by the service).

This was observed for instance that would be interesting for [`envoyproxy` package](https://github.com/elastic/integrations/tree/4ea84a146da735090bb040c2eab6e2c3bc67bc19/packages/envoyproxy) from integrations repository. Some fields (that were failing while [testing the validation based on mappings](https://github.com/elastic/integrations/pull/12300)) just were ingested after some docs ([buildkite link](https://buildkite.com/elastic/integrations/builds/21105)). 

```
test case failed: one or more errors found in mappings in metrics-envoyproxy.stats index template: [0] field "envoy.thread_local_cluster_manager.main_thread_clusters_inflated.value" is undefined: field definition not found
[1] field "envoy.thread_local_cluster_manager.worker_0_clusters_inflated.value" is undefined: field definition not found
[2] field "envoy.thread_local_cluster_manager.worker_1_clusters_inflated.value" is undefined: field definition not found
```

It would be great to add new settings for the system tests to wait for documents being ingested into Elasticsearch:
- Wait for a minimum number of documents, but not required to be an exact number.
- Wait for some expected fields being present in the hits found.
- Any other option?

mrodm:
Closes #2378

This PR includes new system test settings to allow for different conditions when waiting for the required/needed documents to be ingested into the data stream.

Three new settings have been added:
- `assert.min_count`: minimum number of documents to be ingested in the data stream.
    - this could be useful in packages that send metrics from a service and there is no upper limit of docs to be ingested.
- `assert.fields_present`: list of fields that must be present in any document of the given data stream.
    - this could be useful for debugging to wait for a specific field that could be sent at the end.
- ~`assert.ingestion_idle_time`: minimum period of time that the data stream should not have any new document ingested.~
    - ~this could be useful in those packages that there is no exact number of documents to be ingested, every run could be a different number of documents ingested.~
    - ~Related: https://github.com/elastic/elastic-package/issues/2378#issuecomment-2674863531~

Examples of the `elastic-package` output when these settings are defined:

- ~`assert.ingestion_idle_time: 20s`:~
```
2025/02/25 13:31:34 DEBUG found 0 hits in logs-system_test_assert_conditions.test-26419 data stream
2025/02/25 13:31:35 DEBUG found 100 hits in logs-system_test_assert_conditions.test-26419 data stream
...
2025/02/25 13:31:56 DEBUG found 100 hits in logs-system_test_assert_conditions.test-26419 data stream
2025/02/25 13:31:56 DEBUG No new documents ingested in 20s
```

- `assert.fields_present: [target.file, target.expected, target.finish]`:
```
2025/02/25 13:33:34 DEBUG found 0 hits in logs-system_test_assert_conditions.test-50949 data stream
2025/02/25 13:33:35 DEBUG found 450 hits in logs-system_test_assert_conditions.test-50949 data stream
2025/02/25 13:33:35 DEBUG > Found field "target.file" in hits
2025/02/25 13:33:35 DEBUG > Found field "target.expected" in hits
2025/02/25 13:33:35 DEBUG Not found field "target.finish" in hits
2025/02/25 13:33:36 DEBUG found 492 hits in logs-system_test_assert_conditions.test-50949 data stream
2025/02/25 13:33:36 DEBUG > Found field "target.finish" in hits
```

- `assert.min_count: 50`:
```
2025/02/25 13:32:42 DEBUG checking for expected data in data stream (10m0s)...
2025/02/25 13:32:42 DEBUG found 0 hits in logs-system_test_assert_conditions.test-82487 data stream
2025/02/25 13:32:43 DEBUG found 0 hits in logs-system_test_assert_conditions.test-82487 data stream
2025/02/25 13:32:44 DEBUG found 0 hits in logs-system_test_assert_conditions.test-82487 data stream
2025/02/25 13:32:45 DEBUG found 100 hits in logs-system_test_assert_conditions.test-82487 data stream
2025/02/25 13:32:45 DEBUG Found 0 deprecation warnings for data stream logs-system_test_assert_conditions
```

Moreover, these settings can be defined simultaneously. For instance:
```yaml
assert:
  min_count: 50
  fields_present:
    - field.expected
    - field.foo
```

In that example, the validation performed by the system test will only run after at least 50 documents have been ingested into the specified data stream, and the two designated fields are prsent in any of the ingested documents.

## Author's checklist

- [x] Add test package
- [x] Add documentation

```shell

cd test/packages/other/system_test_assert_conditions

# Update system test configuration files as needed:
# - "data_stream/test/_dev/test/system/test-fields_present-config.yml"
# - "data_stream/test/_dev/test/system/test-min_count-config.yml"

elastic-package stack down -v
```


docs/howto/system_testing.md
```diff
@@ -434,6 +434,10 @@ for system tests.
 | skip_transform_validation | boolean |  | Disable or enable the transforms validation performed in system tests. |
 | vars | dictionary |  | Package level variables to set (i.e. declared in `$package_root/manifest.yml`). If not specified the defaults from the manifest are used. |
 | wait_for_data_timeout | duration |  | Amount of time to wait for data to be present in Elasticsearch. Defaults to 10m. |
+| assert.hit_count | integer |  | Exact number of documents to wait for being ingested. |
+| assert.min_count | integer |  | Minimum number of documents to wait for being ingested. |
```
mrodm, line 438: Currently, the maximum number of documents retrieved from Elasticsearch by elastic-package is 500:
https://github.com/elastic/elastic-package/blob/b6f296ce42c19fa4c6a58da623e5436dc727d1b5/internal/testrunner/runners/system/tester.go#L85

I guess it's difficult to happen, but it's worth to take this into account. If a test sets `assert.min_count: 510` , that will never finish since there will never be 510 documents in the search response.

Related: https://github.com/elastic/elastic-package/issues/2378#issuecomment-2674940199

As mentioned in that comment, should we add a new environment variable to be able to update that value (in a follow-up PR maybe)?

docs/howto/system_testing.md
```diff
@@ -434,6 +434,10 @@ for system tests.
...
+| assert.fields_present | []string|  | List of fields that must be present in the documents to stop waiting for new documents. |
+| assert.ingestion_idle_time | duration |  | Minimum time elapsed since the last document was ingested. |
```
mrodm, line 440: Do you think any of the new settings (min_count, fields_present or ingestion_idle_time) added here in this PR could be dismissed? are any of those redundant?

docs/howto/system_testing.md
jsoriano: Umm, I guess it is also possible to use some kind of pagination? Maybe in the meantime we can also limit `hit_count` and `min_count` to 500 documents, and fail if it is set to a bigger number.

docs/howto/system_testing.md
jsoriano: `ingestion_idle_time` may be redundant in many cases with `hit_count`, and basing tests on timing might also lead to flakiness. I think we can live without this option by now.

mrodm:
Depends on #2433

docs/howto/system_testing.md
mrodm: Removed and updated PR description too.

docs/howto/system_testing.md
mrodm: Removed and updated PR description too.

mrodm:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -1583,6 +1535,118 @@ func (r *tester) createServiceStateDir() error {
...
+				return true
+			}
+			if hits.size() < config.Assert.HitCount {
+				return false
+			}
+
+			ret := hits.size() == oldHits
+			if !ret {
+				time.Sleep(4 * time.Second)
+			}
```
jsoriano, line 1593: Should we also add some wait time if this assertion is satisfied but not the others?

internal/testrunner/runners/system/tester.go
```diff
@@ -1583,6 +1535,118 @@ func (r *tester) createServiceStateDir() error {
...
+					if _, err := d.GetValue(f); err == nil {
+						found = true
+						break
+					}
+				}
+				if !found {
+					logger.Debugf("Not found field %q in hits", f)
+					return false
+				}
+				logger.Debugf("> Found field %q in hits", f)
```
jsoriano, line 1622: These debug messages will appear for all fields and all documents, right? Maybe this is too much even at the debug level?

internal/testrunner/runners/system/tester.go
```diff
@@ -1583,6 +1535,118 @@ func (r *tester) createServiceStateDir() error {
...
+
+		return assertFieldsPresent && assertMinCount && assertHitCount, nil
+	}, 1*time.Second, waitForDataTimeout)
+
+	if waitErr != nil {
+		return nil, waitErr
+	}
+
+	if !passed {
+		return nil, testrunner.ErrTestCaseFailed{Reason: fmt.Sprintf("could not find hits in %s data stream", dataStream)}
```
jsoriano, line 1644: Nit. Maybe we should rewrite this error. It can fail now for different reasons, even when hits are found.

test/packages/other/system_test_assert_conditions/data_stream/test/_dev/test/system/test-fields_present-config.yml
```diff
@@ -0,0 +1,33 @@
+vars: ~
+data_stream:
+  vars:
+    paths:
+      - "/custom/paths/logs.json"
+assert:
+  fields_present:
+    - target.file
+    - target.expected
+    - target.finish
```
jsoriano, line 10: Wdyt about adding a false positives test package to check that it fails if not satisfied?

jsoriano (approved):
Nice! Added some small suggestions, but feel free to merge.

internal/testrunner/runners/system/tester.go
mrodm: That would appear just once, since it is kept a map struct for the previous fields found.

But it's true that the message about "Not found field" could be too much... I'll remove that.

internal/testrunner/runners/system/tester.go
mrodm: Something like `could not find the expected hits in %d data stream` ?

internal/testrunner/runners/system/tester.go
mrodm: Not sure I follow this. Why should it be needed to wait when `ret` is true?

In those packages where just this assertion is set, the behavior would be the same now. 
As it is, this assertion required that two consecutive times the hits size is the same once `hits.size() >= HitCount`, waiting 4 seconds when the hits size is not the same as the previous iteration.

In those packages where just this assertion is set, the behavior would be the same now. 

The issue would be when there is assertion is set along with `min_count` or `fields_present`. That time.Sleep would delay the check of the other assertions. But probably this would not cause too much trouble.

test/packages/other/system_test_assert_conditions/data_stream/test/_dev/test/system/test-fields_present-config.yml
mrodm: Added a test package (false positive) to test `assert.fields_present`
https://github.com/elastic/elastic-package/pull/2429/commits/492f6cacba390318301ef5dcf46702673dffcc29

internal/testrunner/runners/system/tester.go
jsoriano: > Why should it be needed to wait when `ret` is true?

My comment goes more in the line of why do we need to wait four seconds when `ret` is false, and if we do, why we don't need to wait when `ret` is true and some of the other conditions is false.

At least for the min count it looks like we could be interested on the same retry patterns.

internal/testrunner/runners/system/tester.go
jsoriano: Yes, something like this would be better.

The best might be to explicitly say what assertions are failing, but maybe this requires a number of changes on the current code, no need to do it by now.

test/packages/false_positives/failed_fields_present_assert.expected_errors
```diff
@@ -0,0 +1 @@
+<failure>test case failed: could not find hits in logs-failed_fields_present_assert.test-[[:digit:]]+ data stream</failure>
```
jsoriano, line 1: :+1:

jsoriano (approved):

internal/testrunner/runners/system/tester.go
mrodm: My assumption is that in the assertion of `hit_count` once the number of hits is equal or great than `config.Assert.HitCount`, that waiting time would be to ensure that there are no more documents ingested. For me, it's like an heuristic that ingestion of documents has finished.

In case of true, it would be considered/assumed that there won't be new messages, and it can be considered true that assertion (so other assertions can be validated).

In case of `min_count`, I think it would not be needed since it is just required to be greater or equal than the given number.

Would that make sense?

I think I'd keep it as it is here by now.

jsoriano: Ok :+1:

jsoriano (approved):

Issue #856, opened 2022-06-21
Improve handling of low disk watermark issues


jsoriano:
When starting a cluster with `elastic-package stack`, if the host where docker is running has little disk space available, Elasticsearch will fail to allocate shards and Kibana will fail to reach a healthy state.

The thresholds for this kind of issues to happen are the [_low and high disk watermarks_](https://www.elastic.co/guide/en/elasticsearch/reference/8.2/modules-cluster.html#disk-based-shard-allocation), they are 90% and 95% respectively by default.

Currently, `elastic-package` just waits till timeout for Kibana to reach a healthy state, and no helpful information is given to users when it finally fails, leading to issues like https://github.com/elastic/elastic-package/issues/838. When the command timeouts, this is the printed error now:
```
container for service "kibana" is unhealthy
Error: booting up the stack failed: running docker-compose failed: running command failed: running Docker Compose up command failed: exit status 1
```

This situation can be detected by checking the available disk space, or by checking Elasticsearch logs, looking for messages like `high disk watermark [95%] exceeded`.

The current workaround would be to free disk space in the docker host.

### Proposed actions

- Detect when the watermarks are exceeded during boot up (parsing logs or checking cluster state), and fail immediately, providing an error message about the availability of free space in the host.
- Modify healthcheck in elasticsearch service to fail if the status is `red`.
  - https://github.com/elastic/elastic-package/pull/1051 
  - https://github.com/elastic/elastic-package/pull/1052
- An alternative could be to disable these thresholds (` cluster.routing.allocation.disk.threshold_enabled: false`), and let Elasticsearch fail to start when there is no disk space. But this can be still misleading, specially in OSs where docker runs in a virtual machine.

efd6:
I am seeing this again.

PR #2435, opened 2025-02-27
Add garbage collection for docker images

jsoriano:
Related to https://github.com/elastic/elastic-package/issues/856.

If one is using elastic-package during enough time, Docker storage can end up full of docker images. Once the 90% threshold is reached, behavior can be unpredictable as described in https://github.com/elastic/elastic-package/issues/856.

This change introduces garbage collection for docker images that were downloaded by elastic-package. Implementation is based on two methods:
* `Track` is used to start tracking images, by recording their tags and last used times. It has to be called before pulling the images, it won't track images that are already pulled, to avoid GCing images not managed by elastic-package.
* `Run` runs the garbage collection, based on two parameters: `max_unused` can be used to remove images older than some time and `max_total_size` to trigger removals when the total size of images (from `docker system df`) goes beyond some threshold. Less recently used images are removed first.

Tracking of known images is persisted under `~/.elastic-package/cache/docker-images-gc.json`.

Garbage collection is disabled by default.

This can be configured from `~/.elastic-package/config.yml`, for example:
```
stack:
  gc:
    enabled: true
    max_unused: 336h # 2 weeks
    max_total_size: 60GB
```

### DIFF
--- a/internal/common/bytesize.go
+++ b/internal/common/bytesize.go
@@ -1,0 +1,168 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"regexp"
+	"strconv"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Common units for sizes in bytes.
+const (
+	Byte     = ByteSize(1)
+	KiloByte = 1024 * Byte
+	MegaByte = 1024 * KiloByte
+	GigaByte = 1024 * MegaByte
+)
+
+const (
+	byteString     = "B"
+	kiloByteString = "KB"
+	megaByteString = "MB"
+	gigaByteString = "GB"
+)
+
+// ByteSize represents the size of a file.
+type ByteSize uint64
+
+// Ensure FileSize implements these interfaces.
+var (
+	_ json.Marshaler   = new(ByteSize)
+	_ json.Unmarshaler = new(ByteSize)
+	_ yaml.Marshaler   = new(ByteSize)
+	_ yaml.Unmarshaler = new(ByteSize)
+)
+
+func parseFileSizeInt(s string) (uint64, error) {
+	// os.FileInfo reports size as int64, don't support bigger numbers.
+	maxBitSize := 63
+	return strconv.ParseUint(s, 10, maxBitSize)
+}
+
+// MarshalJSON implements the json.Marshaler interface for FileSize, it returns
+// the string representation in a format that can be unmarshaled back to an
+// equivalent value.
+func (s ByteSize) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + s.String() + `"`), nil
+}
+
+// MarshalYAML implements the yaml.Marshaler interface for FileSize, it returns
+// the string representation in a format that can be unmarshaled back to an
+// equivalent value.
+func (s ByteSize) MarshalYAML() (interface{}, error) {
+	return s.String(), nil
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface for FileSize.
+func (s *ByteSize) UnmarshalJSON(d []byte) error {
+	// Support unquoted plain numbers.
+	bytes, err := parseFileSizeInt(string(d))
+	if err == nil {
+		*s = ByteSize(bytes)
+		return nil
+	}
+
+	var text string
+	err = json.Unmarshal(d, &text)
+	if err != nil {
+		return err
+	}
+
+	return s.unmarshalString(text)
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for FileSize.
+func (s *ByteSize) UnmarshalYAML(value *yaml.Node) error {
+	// Support unquoted plain numbers.
+	bytes, err := parseFileSizeInt(value.Value)
+	if err == nil {
+		*s = ByteSize(bytes)
+		return nil
+	}
+
+	return s.unmarshalString(value.Value)
+}
+
+var bytesPattern = regexp.MustCompile(fmt.Sprintf(`^(\d+(\.\d+)?)(%s|%s|%s|%s|)$`, byteString, kiloByteString, megaByteString, gigaByteString))
+
+func (s *ByteSize) unmarshalString(text string) error {
+	match := bytesPattern.FindStringSubmatch(text)
+	if len(match) < 3 {
+		return fmt.Errorf("invalid format for size in bytes (%s)", text)
+	}
+
+	if match[2] == "" {
+		q, err := parseFileSizeInt(match[1])
+		if err != nil {
+			return fmt.Errorf("invalid format for size in bytes (%s): %w", text, err)
+		}
+
+		unit := match[3]
+		switch unit {
+		case gigaByteString:
+			*s = ByteSize(q) * GigaByte
+		case megaByteString:
+			*s = ByteSize(q) * MegaByte
+		case kiloByteString:
+			*s = ByteSize(q) * KiloByte
+		case byteString, "":
+			*s = ByteSize(q) * Byte
+		default:
+			return fmt.Errorf("invalid unit for filesize (%s): %s", text, unit)
+		}
+	} else {
+		q, err := strconv.ParseFloat(match[1], 64)
+		if err != nil {
+			return fmt.Errorf("invalid format for size in bytes (%s): %w", text, err)
+		}
+
+		unit := match[3]
+		switch unit {
+		case gigaByteString:
+			*s = approxFloat(q, GigaByte)
+		case megaByteString:
+			*s = approxFloat(q, MegaByte)
+		case kiloByteString:
+			*s = approxFloat(q, KiloByte)
+		case byteString, "":
+			*s = approxFloat(q, Byte)
+		default:
+			return fmt.Errorf("invalid unit for filesize (%s): %s", text, unit)
+		}
+	}
+
+	return nil
+}
+
+func approxFloat(n float64, unit ByteSize) ByteSize {
+	approx := n * float64(unit)
+	return ByteSize(math.Round(approx))
+}
+
+// String returns the string representation of the FileSize.
+func (s ByteSize) String() string {
+	format := func(q ByteSize, unit string) string {
+		return fmt.Sprintf("%d%s", q, unit)
+	}
+
+	if s >= GigaByte && (s%GigaByte == 0) {
+		return format(s/GigaByte, gigaByteString)
+	}
+
+	if s >= MegaByte && (s%MegaByte == 0) {
+		return format(s/MegaByte, megaByteString)
+	}
+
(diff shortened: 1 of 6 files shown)


internal/common/bytesize.go
```diff
@@ -0,0 +1,168 @@
...
+
+const (
+	byteString     = "B"
+	kiloByteString = "KB"
+	megaByteString = "MB"
+	gigaByteString = "GB"
+)
+
+// ByteSize represents the size of a file.
+type ByteSize uint64
```
jsoriano, line 33: Type mostly copied from https://github.com/elastic/package-spec/blob/964c4a69e024cc464c4808720ba0db9f001a82a7/code/go/internal/spectypes/filesize.go

internal/stack/compose.go
```diff
@@ -50,6 +51,34 @@ func (eb *envBuilder) build() []string {
 	return eb.vars
 }
 
+func runDockerImagesGC(ctx context.Context, project *compose.Project, opts compose.CommandOptions, appConfig *install.ApplicationConfiguration) error {
```
jsoriano, line 54: Added by now only to stack operations, it could be added too to agent and service deployers. It may be specially interesting in the service deployer.

internal/compose/compose.go
```diff
@@ -60,7 +61,23 @@ type Config struct {
 	Services map[string]service
 }
 
+// Images lists the images found in the configuration.
```
jsoriano, line 64: This only lists images found directly in the docker compose configuration, it won't find images found in Dockerfiles. Maybe we can think in other approaches to find the images to track. Ideas welcome.

internal/common/bytesize.go
```diff
@@ -0,0 +1,168 @@
...
+	// os.FileInfo reports size as int64, don't support bigger numbers.
+	maxBitSize := 63
+	return strconv.ParseUint(s, 10, maxBitSize)
+}
+
+// MarshalJSON implements the json.Marshaler interface for FileSize, it returns
+// the string representation in a format that can be unmarshaled back to an
+// equivalent value.
+func (s ByteSize) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + s.String() + `"`), nil
```
efd6, line 53:
```suggestion
	return json.Marshal(s.String())
```
?

internal/common/bytesize.go
```diff
@@ -0,0 +1,168 @@
...
+	bytes, err := parseFileSizeInt(value.Value)
+	if err == nil {
+		*s = ByteSize(bytes)
+		return nil
+	}
+
+	return s.unmarshalString(value.Value)
+}
+
+var bytesPattern = regexp.MustCompile(fmt.Sprintf(`^(\d+(\.\d+)?)(%s|%s|%s|%s|)$`, byteString, kiloByteString, megaByteString, gigaByteString))
```
efd6, line 93:
```suggestion
var bytesPattern = regexp.MustCompile(`^(\d+(\.\d+)?)(` + byteString + `|` + kiloByteString + `|` + megaByteString + `|` + gigaByteString + `|)$`)
```
(compile-time string construction instead of init-time)

but for a more efficient pattern
```suggestion
var bytesPattern = regexp.MustCompile(`^(\d+(\.\d+)?)(B|[KMG]B|)$`)
```

internal/install/application_configuration.go
```diff
@@ -85,8 +88,15 @@ type configFile struct {
 	} `yaml:"profile"`
 }
 
+type dockerGCConfig struct {
+	Enabled      *bool            `yaml:"enabled"`
+	MaxTotalSize *common.ByteSize `yaml:"max_total_size"`
+	MaxUnused    *time.Duration   `yaml:"max_unused"`
+}
+
```
efd6, line 96:
```suggestion
```
(unused)

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+type ImagesGC struct {
+	path   string
+	images []gcEntry
+	clock  func() time.Time
+	client imagesGCClient
+
+	ImagesGCConfig
+}
+
+type imagesGCClient interface {
```
efd6, line 53: Is there a reason this needs to be an interface?

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+	gc.images = images
+	return nil
+}
+
+type localImagesGCClient struct {
+}
+
+func defaultImagesGCClient() *localImagesGCClient {
+	return &localImagesGCClient{}
+}
```
efd6, line 201:
```suggestion
func defaultImagesGCClient() localImagesGCClient {
	return localImagesGCClient{}
}
```
(we're not holding anything, so we don't need to put this on a pointer)

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+
+func (c *localImagesGCClient) ListImages() ([]string, error) {
```
efd6, line 203:
```suggestion
func (localImagesGCClient) ListImages() ([]string, error) {
```

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+		result = append(result, line.Repository+":"+line.Tag)
+	}
+
+	return result, nil
+}
+
+var removeConflictRegexp = regexp.MustCompile("container [^/s]+ is using its referenced image [^/s]+")
+
+func (c *localImagesGCClient) RemoveImage(image string) error {
```
efd6, line 232:
```suggestion
func (localImagesGCClient) RemoveImage(image string) error {
```

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+		if removeConflictRegexp.MatchString(errMessage) {
+			return BusyImageErr
+		}
+		return fmt.Errorf("%w: %s", err, strings.TrimPrefix(errMessage, "Error response from daemon: "))
+	}
+
+	return nil
+}
+
+func (c *localImagesGCClient) TotalImagesSize() (common.ByteSize, error) {
```
efd6, line 249:
```suggestion
func (localImagesGCClient) TotalImagesSize() (common.ByteSize, error) {
```

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+
+type localImagesGCClient struct {
+}
+
+func defaultImagesGCClient() *localImagesGCClient {
+	return &localImagesGCClient{}
+}
+
+func (c *localImagesGCClient) ListImages() ([]string, error) {
+	cmd := exec.Command("docker", "image", "list", "--format=json")
```
efd6, line 204: Recommend x/sys/execabs over os/exec.
```suggestion
	cmd := execabs.Command("docker", "image", "list", "--format=json")
```

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+	errOutput := new(bytes.Buffer)
+	cmd.Stderr = errOutput
+
+	output, err := cmd.Output()
+	if err != nil {
+		return nil, fmt.Errorf("docker image list failed (stderr=%q): %w", errOutput.String(), err)
```
efd6, line 210:
```suggestion
		return nil, fmt.Errorf("docker image list failed (stderr=%q): %w", errOutput, err)
```

internal/docker/imagesgc.go
```diff
@@ -0,0 +1,275 @@
...
+			sizeOk = totalSize <= gc.MaxTotalSize
+		}
+		if !sizeOk || image.LastUsed.Before(maxUnused) {
+			if slices.Contains(present, image.ImageTag) {
+				err := gc.client.RemoveImage(image.ImageTag)
+				if errors.Is(err, BusyImageErr) {
+					continue
+				}
+				if err != nil {
+					gc.images = append(images, gc.images[i:]...)
```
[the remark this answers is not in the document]
jsoriano: This is done in case `gc` is used after `Run()` fails, so `gc.images` is updated with images that may have been removed.
`gc` is not reused now, but the idea is that it could be.

internal/stack/compose.go
mrodm: I've checked that running `elastic-package test system -v`  in the `mysql` package (integrations repo), it creates 36 docker images to test the service:
```shell
 $ elastic-package test system -v
...
 $ docker images |grep elastic-package-service |grep mysql  | wc -l 
36
 $ docker images |grep elastic-package-service
...
elastic-package-service-30135-mysql                      latest                   040dcd09ffba   58 minutes ago      1.34GB
elastic-package-service-13845-mysql_replica              latest                   9b7839234a41   About an hour ago   591MB
...
```

When trying with other packages using the new custom agents (e.g. oracle), `elastic-package` creates one `elastic-agent` docker image per each system test with different Dockerfile (hash of that file is part of the docker tag) [link](https://github.com/elastic/elastic-package/blob/d6915db28c9c0c87777e701dcfc91fef76702398/internal/agentdeployer/_static/docker-agent-base.yml.tmpl#L13).
```shell
 $ elastic-package test system -v --data-streams memory,performance
...
 $ docker images |grep elastic-package-test
elastic-package-test-elastic-agent                       8.17.3-1e83b6defffbb1d833d6235e1c502bad   ecbc1bc5d908   22 minutes ago   1.95GB
elastic-package-test-elastic-agent                       8.17.3-2459e61e8ebaf237cffe70dc5ad649cb   159f23420959   31 minutes ago   1.95GB
```


As most of them are built with the same base image, there could be a lot of shared layers. 

Removing all the previous docker images in my laptop (docker rmi), docker just untagged the docker images. Maybe does it need to clean the docker cache too ?

```
 $ docker images |grep elastic-package-service |grep mysql | awk '{print $3}' | xargs -I {} docker rmi {}
Untagged: elastic-package-service-40716-mysql:latest
Deleted: sha256:1fc900526a35aeb6563b6fc4a393fe88e2282b41ca95e603684bd3dc72264245
Untagged: elastic-package-service-47488-mysql:latest
Deleted: sha256:6625ce72c4b0b506cf0ad86e8b703b580d6d90707471aef59c39d2e62ab4fa82
Untagged: elastic-package-service-66823-mysql:latest
Deleted: sha256:973241860b7932f231dead77a048d8a433bd468ed1e93bb90c8e0815707cc5ee
Untagged: elastic-package-service-61615-mysql:latest
Deleted: sha256:0385a33fc301b9b417ac9db269386af710767ef7e6342cfcfccb61717857051a
...

 $ docker rmi elastic-package-test-elastic-agent:8.17.3-1e83b6defffbb1d833d6235e1c502bad 
Untagged: elastic-package-test-elastic-agent:8.17.3-1e83b6defffbb1d833d6235e1c502bad
Deleted: sha256:ecbc1bc5d908ca93d32d8af54ecc43dc067a8367abc400f2bcfc325aa37bb629

 $ docker rmi elastic-package-test-elastic-agent:8.17.3-2459e61e8ebaf237cffe70dc5ad649cb 
Untagged: elastic-package-test-elastic-agent:8.17.3-2459e61e8ebaf237cffe70dc5ad649cb
Deleted: sha256:159f234209595713fc882f7e6802c09a4b57381770610e2661defbf5dfe91618

```

internal/docker/imagesgc.go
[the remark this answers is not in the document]
efd6: YAGNI?

PR #2440, opened 2025-02-27
[pipeline_tests] Allow error.message in some cases

jrmolin:
* at times, an integration sets `error.message` for a valid reason
* add support for expected `error.message` values in  `elastic-package test pipeline`
* add a test case


taylor-swanson (commented):
I'm not a big fan of this.

I'm not thrilled that `error.message` can have dual meaning, either the event represents a failure logged by the upstream vendor product or a failure by us in the pipeline (I forget if beats failures can propagate upwards into ES or not).

However, not seeing ECS changing dramatically to remove support for the former scenario, one other thing we could do is allow `error.message` if and only if `event.kind` is NOT `pipeline_error`. The presence of these two fields should indicate a fatal condition (the second scenario I mentioned) and warrant failing a test case. Otherwise, the presence of just `error.message` would seem to indicate the first scenario I mentioned.

Perhaps a better way to look at it is just fail the test if `event.kind` is `pipeline_error`. I'll need others to chime in if the presence `error.message` WITHOUT `event.kind: pipeline_error` indicates a test failure condition.

taylor-swanson:
@andrewkroh, are there cases where `error.message` is present and `event.kind` is NOT `pipeline_error` AND we want to fail the test?

My concern with the test config ignoring error.message is that we could ignore legitimate pipeline errors.

jrmolin:
i'm going to just close this, i think. conversation in async channels convinces me not to be the one pursuing this

andrewkroh:
> are there cases where `error.message` is present and `event.kind` is NOT `pipeline_error` AND we want to fail the test?

@taylor-swanson I can't think of any instance where `error.message` is used without also setting `event.kind: pipeline_error`. The de facto meaning of `error.message` in integrations has become that it indicates a pipeline error (i.e., any type of collection or processing issue).

So I believe it's warranted to require an explicit flag stating "yes, I'm expecting `error.message`" whenever this field is present. And we might as well assert that `event.kind == pipeline_error` when `error.message` is found (this isn't strictly what ECS specifies, but it aligns with current integration practices, AFAIK).

PR #2449, opened 2025-03-04
[system test] Add remove agent handler just after setting up agent

mrodm:
Relates #2039
Follows #2409

Ensure Elastic Agent created when using Independent Agents (default method via `r.setupAgent`) is removed even if the service exists with error.

Example of Buildkite build where service containers fails (e.g. mysql) and the agent is not removed:
https://buildkite.com/elastic/integrations/builds/22965

One example of this scenario is when the service container cannot started (it exits with failure). In that case, the new Elastic Agent created was not removed in the tear down process and therefore the corresponding Agent policies were not able to be deleted.

This PR duplicates the call to `checkEnrolledAgents` just after setting up the Elastic Agent so even in the case `setupService` fails, the new Elastic Agent can be deleted in the tear down process.

It needs to be duplicated this call since `r.setupService` could create new agents in case of the kubernetes service deployer or the custom agents deployer (the latter is deprecated) if the Elastic Agent of the stack is used (disabled Independent Elastic Agents).

In this buildkite build (with the changes of this PR), `mysql` now reports the container exit as error in xUnit files:
https://buildkite.com/elastic/integrations/builds/22973

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -1179,6 +1179,16 @@ func (r *tester) prepareScenario(ctx context.Context, config *testConfig, stackC
 
 	scenario.agent = agentDeployed
 
+	if r.runIndependentElasticAgent {
+		// Ensure agent created by `r.setupAgent` is removed if service fails to start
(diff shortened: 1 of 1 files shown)


mrodm:
test integrations

PR #2454, opened 2025-03-05
[system tests] Add exceptions for leaf of objects if required for mapping bases validation

mrodm:
This PR adds the same exception for leaf of objects that was set in place for validation based on fields found in documents:
https://github.com/elastic/elastic-package/blob/b64f3748e2b513f3ee8baa4474f21b209a937162/internal/fields/exceptionfields.go#L39

These validations are based on the spec version defined in the package.

Found issue in these builds for `gcp` and `network_traffic` where those missing definitions should be ignored since those packages define spec version 3.0.0

https://buildkite.com/elastic/integrations/builds/23010

Buildkite build testing this change with all packages: https://buildkite.com/elastic/integrations/builds/23032

Buildkite build testing with stack 9.1.0-SNAPSHOT with this change:
https://buildkite.com/elastic/integrations/builds/23034

## How to test locally this PR

```shell

cd path/to/integrations

elastic-package stack up -v -d --version 9.1.0-SNAPSHOT

cd packages/gcp 
elastic-package test system -v --data-streams firewall

cd ../packages/network_traffic
elastic-package test system -v --data-streams dns

elastic-package stack down -v
```

### DIFF
--- a/internal/fields/exceptionfields.go
+++ b/internal/fields/exceptionfields.go
@@ -48,8 +48,8 @@ func (v *Validator) listExceptionFieldsMapElement(root string, elem common.MapSt
 			all = append(all, fields...)
 		}
 	}
-	all = slices.Compact(all)
 	slices.Sort(all)
+	all = slices.Compact(all)
 	return all
 }
 
--- a/internal/fields/exceptionfields.go
+++ b/internal/fields/exceptionfields.go
@@ -37,7 +37,7 @@ func (v *Validator) listExceptionFieldsMapElement(root string, elem common.MapSt
 			all = append(all, fields...)
 		default:
 			if skipLeafOfObject(root, name, v.specVersion, v.Schema) {
-				logger.Warnf("Skip validating leaf of object (key %q spec %q)", key, v.specVersion)
+				logger.Tracef("Skip validating leaf of object (key %q spec %q)", key, v.specVersion)
 				all = append(all, key)
 				// Till some versions we skip some validations on leaf of objects, check if it is the case.
 				break
@@ -88,7 +88,7 @@ func (v *Validator) parseExceptionField(key string, definition FieldDefinition, 
 	case "ip":
 	case "array":
 		if v.specVersion.LessThan(semver2_0_0) {
-			logger.Warnf("Skip validating field of type array with spec < 2.0.0 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
+			logger.Tracef("Skip validating field of type array with spec < 2.0.0 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
 			return []string{key}
 		}
 		return nil
@@ -99,7 +99,7 @@ func (v *Validator) parseExceptionField(key string, definition FieldDefinition, 
 			// This is probably an element from an array of objects,
 			// even if not recommended, it should be validated.
 			if v.specVersion.LessThan(semver3_0_1) {
-				logger.Warnf("Skip validating object (map[string]any) in package spec < 3.0.1 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
+				logger.Tracef("Skip validating object (map[string]any) in package spec < 3.0.1 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
 				return []string{key}
 			}
 			return nil
@@ -107,13 +107,13 @@ func (v *Validator) parseExceptionField(key string, definition FieldDefinition, 
 			// This can be an array of array of objects. Elasticsearh will probably
 			// flatten this. So even if this is quite unexpected, let's try to handle it.
 			if v.specVersion.LessThan(semver3_0_1) {
-				logger.Warnf("Skip validating object ([]any) because spec < 3.0.1 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
+				logger.Tracef("Skip validating object ([]any) because spec < 3.0.1 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
 				return []string{key}
 			}
 			return nil
 		case nil:
 			// The document contains a null, let's consider this like an empty array.
-			logger.Warnf("Skip validating object empty array because spec < 3.0.1 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
+			logger.Tracef("Skip validating object empty array because spec < 3.0.1 (key %q type %q spec %q)", key, definition.Type, v.specVersion)
 			return []string{key}
 		default:
 			switch {
@@ -125,7 +125,7 @@ func (v *Validator) parseExceptionField(key string, definition FieldDefinition, 
 				return v.parseExceptionField(key, definition, val)
 			case definition.Type == "object" && definition.ObjectType == "":
 				// Legacy mapping, ambiguous definition not allowed by recent versions of the spec, ignore it.
-				logger.Warnf("Skip legacy mapping: object field without \"object_type\" parameter: %q", key)
+				logger.Tracef("Skip legacy mapping: object field without \"object_type\" parameter: %q", key)
 				return []string{key}
 			}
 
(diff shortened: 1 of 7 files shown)


mrodm:
test integrations

internal/fields/mappings.go
```diff
@@ -457,7 +457,7 @@ func (v *MappingValidator) compareMappings(path string, couldBeParametersDefinit
 			}
 			return nil
 		} else if !isObject(preview) {
-			errs = append(errs, fmt.Errorf("not found properties in preview mappings for path: %q", path))
+			errs = append(errs, fmt.Errorf("undefined field mappings found in path: %q", path))
```
mrodm, line 460: Any other suggestion for this error message?

internal/fields/exceptionfields.go
```diff
@@ -37,6 +37,8 @@ func (v *Validator) listExceptionFieldsMapElement(root string, elem common.MapSt
 			all = append(all, fields...)
 		default:
 			if skipLeafOfObject(root, name, v.specVersion, v.Schema) {
+				logger.Warnf("Skip validating leaf of object (key %q spec %q)", key, v.specVersion)
```
mrodm, line 40: Should we remove these kind of messages? Do you think they are adding noise to the output ?

And, should these other warnings be removed too? 

https://github.com/elastic/elastic-package/blob/b64f3748e2b513f3ee8baa4474f21b209a937162/internal/fields/mappings.go#L564

https://github.com/elastic/elastic-package/blob/b64f3748e2b513f3ee8baa4474f21b209a937162/internal/fields/mappings.go#L575

https://github.com/elastic/elastic-package/blob/b64f3748e2b513f3ee8baa4474f21b209a937162/internal/fields/mappings.go#L440

https://github.com/elastic/elastic-package/blob/b64f3748e2b513f3ee8baa4474f21b209a937162/internal/fields/exceptionfields.go#L89

mrodm: Probably too much verbose, look at these output:
https://buildkite.com/elastic/integrations/builds/23034#01956614-a005-472f-a3aa-91d8f5940371/6-9

jsoriano: Yes, this is probably too verbose, maybe even at the debug level.

If we want to have this at the debug level, we could print the full list after collecting it instead of once per element.

internal/fields/mappings.go
jsoriano: Looks good like this.

internal/fields/exceptionfields.go
mrodm: I added a new TRACE log level here https://github.com/elastic/elastic-package/pull/2454/commits/fc6c793420c0b4e7f3ea8e7c84799f7a19f050d6 if more than `-v` is set in the command line.

This log level should not be enabled in CI.

Probably, this would also help in other scenarios.

WDYT ? @jsoriano

internal/fields/exceptionfields.go
```diff
@@ -37,6 +37,8 @@ func (v *Validator) listExceptionFieldsMapElement(root string, elem common.MapSt
 			all = append(all, fields...)
 		default:
 			if skipLeafOfObject(root, name, v.specVersion, v.Schema) {
+				logger.Tracef("Skip validating leaf of object (key %q spec %q)", key, v.specVersion)
```
jsoriano, line 40: As we are keeping the list of fields we could still log the full list after the loop instead of logging all of them. Then it could still be at the debug level.
```
logger.Debugf("Skip validating leaves of object %q (spec %q): %s", root, v.specVersion, strings.Join(all, ","))
```

jsoriano (approved):
Nice to add this trace log level.

internal/fields/exceptionfields.go
mrodm: Sure, I'll add that.

Testing it in `gcp` package, this would be the output:
```
2025/03/05 14:51:56 DEBUG Performing validation based on mappings
2025/03/05 14:51:56 DEBUG Skip validating leafs of object "" (spec "3.0.0"): gcp.source.vpc.vpc_name,source.domain,agent.type,log.logger,network.direction,data_stream.namespace,gcp.destination.instance.zone,gcp.source.vpc.subnetwork_name,agent.name,network.type,cloud.project.id,event.type,rule.name,network.name.text,rule.name.text,gcp.source.vpc.project_id,source.port,destination.address,event.kind,event.module,destination.ip,agent.ephemeral_id,event.id,event.category,gcp.destination.instance.project_id,gcp.source.instance.project_id,network.community_id,event.created,network.iana_number,gcp.destination.vpc.subnetwork_name,gcp.destination.vpc.project_id,cloud.availability_zone,network.name,gcp.firewall.rule_details.direction,gcp.firewall.rule_details.ip_port_info,gcp.firewall.rule_details.priority,gcp.source.instance.region,gcp.source.instance.zone,destination.port,cloud.provider,agent.version,data_stream.dataset,data_stream.type,source.address,gcp.firewall.rule_details.source_range,event.agent_id_status,gcp.destination.vpc.vpc_name,event.action,event.ingested,event.dataset,agent.name.text,destination.domain,gcp.destination.instance.region,source.ip,agent.id,ecs.version,gcp.firewall.rule_details.target_tag,gcp.firewall.rule_details.action,network.transport,cloud.region,related.ip
2025/03/05 14:51:56 DEBUG Skip validating leafs of object "" (spec "3.0.0"): agent.type,related.ip,gcp.destination.vpc.subnetwork_name,gcp.source.vpc.subnetwork_name,network.name.text,gcp.destination.vpc.vpc_name,event.dataset,gcp.firewall.rule_details.target_tag,source.address,event.ingested,cloud.region,agent.version,event.agent_id_status,source.port,destination.port,destination.ip,agent.ephemeral_id,rule.name,log.logger,gcp.source.instance.project_id,source.domain,network.type,gcp.destination.instance.zone,gcp.destination.instance.project_id,ecs.version,event.created,cloud.project.id,event.kind,network.community_id,cloud.provider,rule.name.text,network.name,event.action,agent.name,data_stream.type,cloud.availability_zone,destination.domain,gcp.firewall.rule_details.ip_port_info,network.iana_number,network.transport,source.ip,gcp.source.vpc.vpc_name,gcp.firewall.rule_details.action,gcp.firewall.rule_details.direction,event.type,network.direction,gcp.source.vpc.project_id,data_stream.dataset,gcp.source.instance.zone,event.module,data_stream.namespace,gcp.destination.vpc.project_id,agent.id,gcp.firewall.rule_details.priority,destination.address,agent.name.text,gcp.firewall.rule_details.source_range,gcp.destination.instance.region,gcp.source.instance.region,event.id,event.category
2025/03/05 14:51:56 DEBUG Skip validating leafs of object "" (spec "3.0.0"): cloud.availability_zone,source.domain,destination.as.number,source.ip,gcp.firewall.rule_details.priority,source.address,agent.ephemeral_id,event.category,network.name.text,destination.geo.country_name,data_stream.namespace,destination.geo.country_iso_code,agent.name.text,cloud.region,network.iana_number,event.id,gcp.source.instance.region,network.transport,event.dataset,gcp.source.instance.zone,agent.type,network.type,destination.ip,data_stream.dataset,gcp.source.instance.project_id,gcp.firewall.rule_details.target_tag,agent.id,event.module,destination.geo.continent_name,gcp.firewall.rule_details.action,data_stream.type,gcp.firewall.rule_details.ip_port_info,destination.port,log.logger,gcp.firewall.rule_details.destination_range,event.type,ecs.version,gcp.source.vpc.vpc_name,gcp.source.vpc.subnetwork_name,destination.address,event.agent_id_status,destination.geo.location,network.direction,event.action,event.ingested,rule.name,cloud.provider,event.created,agent.version,network.name,agent.name,gcp.firewall.rule_details.direction,cloud.project.id,network.community_id,event.kind,source.port,related.ip,rule.name.text,gcp.source.vpc.project_id
2025/03/05 14:51:56 DEBUG Get Mappings from data stream (logs-gcp.firewall-57633)
```

Probably with packages with more documents ingested (there would be one line per document) and more tests this could be too much.
I think it would be better to just set it as TRACE.

mrodm: The issue is that there could be quite a few elements in that list.

mrodm: Ok, I'll set it then as Trace that message with the full list of exception fields.

PR #2506, opened 2025-04-01
Upgrade go runtime to 1.24.2

marc-gr:


.go-version
```diff
@@ -1 +1 @@
-1.23.1
+1.24.1
```
jsoriano, line 1: Nit. 1.24.2 has just been released :sweat_smile:

jsoriano (approved):

PR #2511, opened 2025-04-02
Check stack subscription to use or not upload API to install packages - WIP

mrodm:
This PR checks whether or not the license of the Elastic stack is the required one to upload packages via API.
Until Elastic stack 8.8.2 the license required was `Enterprise`. In Elastic stack 8.8.2, the license required was `basic`.

### DIFF
--- a/internal/elasticsearch/client.go
+++ b/internal/elasticsearch/client.go
@@ -196,7 +196,7 @@ type Info struct {
 	Version     struct {
 		Number      string `json:"number"`
 		BuildFlavor string `json:"build_flavor"`
-	} `json:"version`
+	} `json:"version"`
 }
 
 // Info gets cluster information and metadata.
@@ -218,6 +218,33 @@ func (client *Client) Info(ctx context.Context) (*Info, error) {
 	}
 
 	return &info, nil
+}
+
+// Subscription gets cluster subscription.
+func (client *Client) Subscription(ctx context.Context) (string, error) {
+	resp, err := client.Client.License.Get(client.Client.License.Get.WithContext(ctx))
+	if err != nil {
+		return "", fmt.Errorf("error getting subscription: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("failed to get subscription: %s", resp.String())
+	}
+
+	type licenseResponse struct {
+		License struct {
+			Type string `json:"Type"`
+		} `json:"license"`
+	}
+
+	var data licenseResponse
+	err = json.NewDecoder(resp.Body).Decode(&data)
+	if err != nil {
+		return "", fmt.Errorf("error decoding subscription: %w", err)
+	}
+
+	return data.License.Type, nil
 }
 
 // IsFailureStoreAvailable checks if the failure store is available.
--- a/internal/packages/installer/factory.go
+++ b/internal/packages/installer/factory.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"path/filepath"
 
 	"github.com/Masterminds/semver/v3"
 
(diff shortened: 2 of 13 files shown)


internal/elasticsearch/client.go
```diff
@@ -196,7 +196,7 @@ type Info struct {
 	Version     struct {
 		Number      string `json:"number"`
 		BuildFlavor string `json:"build_flavor"`
-	} `json:"version`
+	} `json:"version"`
```
mrodm, line 199: Noticed while trying to get and use the Elastic stack subscription.

internal/packages/installer/factory.go
```diff
@@ -75,20 +87,40 @@ func NewForPackage(options Options) (Installer, error) {
...
+		return true, nil
+	}
+
+	err := kbnClient.EnsureZipPackageCanBeInstalled(ctx)
+	if errors.Is(err, kibana.ErrNotSupported) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
```
jsoriano, line 119: Nit.
```suggestion
	if kibanaVersion.LessThan(semver8_8_2) {
		err := kbnClient.EnsureZipPackageCanBeInstalled(ctx)
		if errors.Is(err, kibana.ErrNotSupported) {
			return false, nil
		}
		if err != nil {
			return false, err
		}
	}
```

internal/kibana/packages.go
```diff
@@ -27,6 +30,47 @@ func (c *Client) InstallPackage(ctx context.Context, name, version string) ([]pa
...
+
+	// If the stack allows to use the upload API, the response is like this one:
+	// {
+	//   "statusCode":400,
+	//   "error":"Bad Request",
+	//   "message":"Error during extraction of package: Error: end of central directory record signature not found. Assumed content type was application/zip, check if this matches the archive type."
+	// }
+	if resp.Error == "Forbidden" && resp.Message == "Requires Enterprise license" {
+		return ErrNotSupported
+	}
```
jsoriano, line 70: This request should never return `StatusOK`, right? Should we handle unexpected statuses more explicitly?

```suggestion
	switch statusCode {
	case http.StatusBadRequest:
		// If the stack allows to use the upload API, the response is like this one:
		// {
		//   "statusCode":400,
		//   "error":"Bad Request",
		//   "message":"Error during extraction of package: Error: end of central directory record signature not found. Assumed content type was application/zip, check if this matches the archive type."
		// }
		return nil
	case http.StatusForbidden:
		var resp struct {
			Message    string `json:"message"`
		}
		if err := json.Unmarshal(respBody, &resp); err != nil {
			return fmt.Errorf("could not unmarhsall response to JSON: %w", err)
		}
		if resp.Message == "Requires Enterprise license" {
			return ErrNotSupported
		}
	}

	return fmt.Errorf("unexpected response (status code %d): %s", statusCode, string(respBody))
```

internal/kibana/packages.go
```diff
@@ -27,6 +30,47 @@ func (c *Client) InstallPackage(ctx context.Context, name, version string) ([]pa
 	return processResults("install", statusCode, respBody)
 }
 
+// EnsureZipPackageCanBeInstalled checks whether or not it can be installed a package using the upload API.
+func (c *Client) EnsureZipPackageCanBeInstalled(ctx context.Context) error {
```
jsoriano, line 34: This is a bit hacky and would be prone to fail if this API would change, maybe add a comment mentioning that is only intended to be used between 8.7.0 and 8.8.2, and is only safe to be used on these versions.

internal/packages/installer/factory.go
```diff
@@ -54,11 +57,20 @@ func NewForPackage(options Options) (Installer, error) {
...
 	if options.ZipPath != "" {
-		if !supportsZip {
-			return nil, fmt.Errorf("not supported uploading zip packages in Kibana %s (%s required)", version, semver8_7_0)
+		if !supportsUploadZip {
+			if version.LessThan(semver8_7_0) {
+				return nil, fmt.Errorf("not supported uploading zip packages in Kibana %s (%s required)", version, semver8_7_0)
+			}
+			if version.LessThan(semver8_8_2) {
+				return nil, fmt.Errorf("not supported uploading zip packages in Kibana %s (%s required or Enteprise license)", version, semver8_8_2)
+			}
```
jsoriano, line 71: Nit. Maybe `isAllowedInstallationViaApi` could also return the reason, as we are checking with the same versions as here.

jsoriano (approved):
Looks good, only some nitpicking.

Maybe the only problematic thing is that receiving `StatusOK` when checking for the installation API should not be expected.

internal/kibana/packages.go
mrodm: > This request should never return StatusOK, right? Should we handle unexpected statuses more explicitly?

That's right, that request should just return Bad Request or Forbidden. Other status, like `StatusOK` would not be possible.

I'll apply your suggestion, thanks!

internal/kibana/packages.go
mrodm: Added comment here https://github.com/elastic/elastic-package/pull/2511/commits/4da2bd995f732cf64b12a87999225f74aa9eab5a

internal/kibana/packages.go
mrodm: Applied suggestion here https://github.com/elastic/elastic-package/pull/2511/commits/538149c0e2123d7d0176ce03c1fc864974d5bff9

internal/packages/installer/factory.go
mrodm: Applied suggestion here https://github.com/elastic/elastic-package/pull/2511/commits/b2fbf7d459e93118c2530889473d4e8600baa9a0

internal/packages/installer/factory.go
mrodm: Added a new return parameter that indicates the reason, just in the case there is an expected failure.

mrodm: Applied here https://github.com/elastic/elastic-package/pull/2511/commits/05e082b9311a22029c672ccb3af4365be22d7870

jsoriano (approved):
:+1:

PR #2529, opened 2025-04-09
Allow new documentation ranges in the validation of IPs

mrodm:
This PR adds the required ranges to the `isDocumentation` so they can be validated successfully in `elastic-package` tests.

If these ranges a test including for instance `3fff::1` would fail with this error:
```
[0] the IP "3fff::1" is not one of the allowed test IPs (see: https://github.com/elastic/elastic-package/blob/main/internal/fields/_static/allowed_geo_ips.txt)
```
But, that IP should be allowed since it is part of the documentation range `3fff::/20` ([RFC9637](https://datatracker.ietf.org/doc/rfc9637/))


docs/howto/ingest_geoip.md
```diff
@@ -2,7 +2,19 @@
 
 Elasticsearch provides default GeoIP databases that can be downloaded in runtime and which weights ~70 MB. This can be
 a root cause of flakiness of package tests, so elastic-package embeds small samples of GeoIP databases, that can identify
-accurately only few ranges of IP addresses included [here](../../internal/fields/_static/allowed_geo_ips.txt)
+accurately only few ranges of IP addresses included [here](../../internal/fields/_static/allowed_geo_ips.txt).
```
jsoriano, line 5: We want to eventually use documentation IP ranges only. I think we can remove the mention to `allowed_geo_ips.txt` in docs.

jsoriano (approved):

internal/fields/validate.go
```diff
@@ -1263,29 +1263,35 @@ func (v *Validator) parseSingleElementValue(key string, definition FieldDefiniti
...
 	}
 	// Following RFC 3849, Section 2. Documentation IPv6 Address Prefix which
 	// says:
 	//   The prefix allocated for documentation purposes is 2001:DB8::/32
-	return len(ip) == net.IPv6len && ip[0] == 32 && ip[1] == 1 && ip[2] == 13 && ip[3] == 184
+	// Following RFC 9637, a new address block 3fff::/20 is registered for documentation purposes
+	return len(ip) == net.IPv6len &&
+		(ip[0] == 32 && ip[1] == 1 && ip[2] == 13 && ip[3] == 184) ||
+		(ip[0] == 63 && ip[1] == 255 && ip[2] <= 15)
 }
```
mrodm, line 1288: Got these values from this code 
```go
package main

import (
	"fmt"
	"log"
	"net"
)
func main() {
	showBlocks("3fff:0000:0000:0000:0000:0000:0000:0000")
	showBlocks("3fff:0fff:ffff:ffff:ffff:ffff:ffff:ffff")
	fmt.Println()
	showBlocks("2001:0db8:0000:0000:0000:0000:0000:0000")
	showBlocks("2001:0db8:ffff:ffff:ffff:ffff:ffff:ffff")

}

func showBlocks(s string) {
	ip := net.ParseIP(s)
	if ip == nil {
		log.Fatalf("wrong ip: %s", s)
	}
	for i, k := range ip {
		fmt.Printf("ip[%d] == %d\n", i, k)
	}
}
```

PR #2547, opened 2025-04-16
Support failure store only from 8.18

jsoriano:
As found in https://github.com/elastic/elastic-package/pull/2544#issuecomment-2808825892, failure store APIs have changed since 8.18.

As this feature is not documented, and not GA, drop support for older versions so we don't need to maintain legacy implementations.


internal/elasticsearch/ingest/testdata/elasticsearch-8-18-enable-failure-store-noop.yaml
```diff
@@ -0,0 +1,311 @@
...
+        proto: HTTP/1.1
+        proto_major: 1
+        proto_minor: 1
+        content_length: 91
+        transfer_encoding: []
+        trailer: {}
+        host: ""
+        remote_addr: ""
+        request_uri: ""
+        body: '{"index_patterns": ["metrics-eptest.failurestore-*"],"data_stream": {"failure_store":true}}'
```
jsoriano, line 73: New index template is created, with `data_stream.failure_store: true`.

internal/elasticsearch/ingest/testdata/elasticsearch-8-18-enable-failure-store-noop.yaml
```diff
@@ -0,0 +1,311 @@
...
+        method: GET
+      response:
+        proto: HTTP/1.1
+        proto_major: 1
+        proto_minor: 1
+        transfer_encoding: []
+        trailer: {}
+        content_length: 202
+        uncompressed: false
+        body: '{"index_templates":[{"name":"ep-test-index-template","index_template":{"index_patterns":["metrics-eptest.failurestore-*"],"composed_of":[],"data_stream":{"hidden":false,"allow_custom_routing":false}}}]}'
```
jsoriano, line 135: But when GETing the index template, it doesn't have the setting.

jsoriano:
Replaced by https://github.com/elastic/elastic-package/pull/2553

PR #2559, opened 2025-04-24
Update default stack version to 8.18.0

jsoriano:


jsoriano:
Self-approving and merging because this change uses to be automatic.

PR #2564, opened 2025-04-30
Replace archiver dependency

mrodm:
As [github.com/mholt/archiver](https://github.com/mholt/archiver) has been archived, this dependency has been replaced by [github.com/mholt/archives](github.com/mholt/archives) as suggested in its README.


internal/files/compress.go
```diff
@@ -5,38 +5,32 @@
...
-
-	z := archiver.Zip{
-		CompressionLevel:       flate.DefaultCompression,
-		MkdirAll:               false,
-		SelectiveCompression:   true,
-		ContinueOnError:        false,
-		OverwriteExisting:      true,
-		ImplicitTopLevelFolder: false,
-	}
+func Zip(ctx context.Context, sourcePath, destinationFile string) error {
```
mrodm, line 20: If it is added the context as parameter, it requires to update different functions in the base code.

The main issue is that `elastic-package check` has some difference and it does not have any context. So, it requires to create a new one if that happens:
https://github.com/elastic/elastic-package/pull/2564/files#diff-5794c40aa4c9301aaf9c39eacf7d3c8511f2511cb24bdd95e0a08704ba583a5fR89

internal/files/compress.go
```diff
@@ -47,7 +41,27 @@ func Zip(sourcePath, destinationFile string) error {
...
+	out, err := os.Create(destinationFile)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	z := archives.Zip{
+		SelectiveCompression: true,
+		ContinueOnError:      false,
+	}
```
mrodm, line 62: New struct from `archives` has fewer options than the previous dependency:
https://github.com/mholt/archives/blob/4393d8832e1d36b9e188cba6d6e2fd3396362ae6/zip.go#L60-L78

internal/files/compress.go
mrodm: Another option could be create the context inside this method, so the signature does not change.

internal/files/compress.go
jsoriano: `Compression` looks similar to the old `CompressionLevel`, or at least seems to be used to control compression level/algorithm. We should check if we are compressing at the same level with old and new implementation, at least confirm that we are not only archiving without compression.

jsoriano: Umm, I have checked and it looks like both implementations do the same, they are storing without compression :thinking:

This is what `file` says, compression method store:
```
Zip archive data, at least v2.0 to extract, compression method=store
```

internal/files/compress.go
jsoriano: I think it'd be better to add support for context, it seems to be used in this package for cancellation.

cmd/build.go
```diff
@@ -82,7 +83,12 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 		cmd.Printf("%s file rendered: %s\n", fileName, target)
 	}
 
-	target, err := builder.BuildPackage(builder.BuildOptions{
+	ctx := cmd.Context()
+	if ctx == nil {
+		// FIXME: when executging elastic-package check , context returned is nil
+		ctx = context.Background()
```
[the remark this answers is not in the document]
mrodm: I'll give a try to fix that function to set the context from its "parent".

internal/files/compress.go
mrodm: Comparing the zip file resulting of using `elastic-package` v0.111.0 and the one from this PR with the `file` command, both zip files have the same output and the same size:

```shell
 $ elastic-package build -v
...
 $ file /home/user/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zipuser/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zip: Zip archive data, at least v2.0 to extract, compression method=store
 $ ls -l /home/user/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zip
-rw-rw-r-- 1 user user 1305990 may  6 16:17 /home/user/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zip

 $ elastic-package-v0.111.0 build -v 
...
 $ file /home/user/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zip
/home/user/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zip: Zip archive data, at least v2.0 to extract, compression method=store
 $ ls -l /home/user/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zip
-rw-rw-r-- 1 user user 1305990 may  6 16:18 /home/user/Coding/work/integrations/build/packages/elastic_package_registry-0.3.1.zip
```

Looking at the code, it looks like there is no compression:
- Previously, it was set `CompressionLevel: flate.DefaultCompression`, so it looks like there is no compression.
https://github.com/mholt/archiver/blob/cc194d2e4af2dc09a812aa0ff61adc4813ea6c69/zip.go#L371-L389
- In both deps, if `zip` file extension is used and `SelectiveCompression` is true, the method used is `Store` in both cases:
    - https://github.com/mholt/archiver/blob/cc194d2e4af2dc09a812aa0ff61adc4813ea6c69/zip.go#L416
    - https://github.com/mholt/archives/blob/4393d8832e1d36b9e188cba6d6e2fd3396362ae6/zip.go#L156
- Looking at the compression methods:
    - Previously, the [`FileMethod` field from zip](https://github.com/mholt/archiver/blob/cc194d2e4af2dc09a812aa0ff61adc4813ea6c69/zip.go#L83) struct was not set, so IIUC this should be [`Store` method](https://github.com/mholt/archiver/blob/cc194d2e4af2dc09a812aa0ff61adc4813ea6c69/zip.go#L28).
    - In the new dep, [`Compression` field](https://github.com/mholt/archives/blob/4393d8832e1d36b9e188cba6d6e2fd3396362ae6/zip.go#L67) is not set neither.
        - And these are the compression methods available: https://github.com/mholt/archives/blob/4393d8832e1d36b9e188cba6d6e2fd3396362ae6/zip.go#L333-L341


Do you know if we could do some other test to ensure that the resulting zip files keep the same behavior?

jsoriano (approved):

PR #2572, opened 2025-05-06
Increase TestRunID suffix to 6 digits

kcreddy:
While debugging integrations flaky tests, in [comment](https://github.com/elastic/integrations/issues/13432#issuecomment-2841453546) it is observed that the system tests are failing because there is existing S3 bucket deployed by Terraform that is clashing with current test run.

Increasing TestRunID suffix to 6 digits to avoid clash with existing test resources.

Closes: https://github.com/elastic/integrations/issues/13432

### DIFF
--- a/internal/common/helpers.go
+++ b/internal/common/helpers.go
@@ -14,8 +14,8 @@ import (
 )
 
 const (
-	testRunMaxID = 99999
-	testRunMinID = 10000
+	testRunMaxID = 999999
+	testRunMinID = 100000
 )
 
 // TrimStringSlice removes whitespace from the beginning and end of the contents of a []string.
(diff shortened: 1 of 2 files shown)


mrodm:
@kcreddy It would be required to update the script related to system flags in order to take into account those 6 digits instead of 5 in the regexex here:

https://github.com/elastic/elastic-package/blob/0965fdd104c2462e2b6cf79e937450c140d59c2f/scripts/test-system-test-flags.sh#L269-L271

jsoriano:
Is this change required at the end?

kcreddy:
> Is this change required at the end?

@jsoriano, not completely sure if this fixes the problem in https://github.com/elastic/integrations/issues/13432 (and all related issues). We are not able to find the root cause, and @mrodm (in [this comment](https://github.com/elastic/integrations/issues/13432#issuecomment-2854078836)) also verified the builds running at the same time and confirmed there is no retry. 
Would appreciate your input in resolving this flaky test.

cc: @narph

mrodm:
Closing in favor of https://github.com/elastic/elastic-package/pull/2744 cc @kcreddy

PR #2567, opened 2025-05-05
bk: use OIDC to tear-down the cloud resources

v1v:
### What

Use OIDC to access the AWS cloud resources and run the tear-down.

This will run in parallel while we migrate.

### Why

Avoid using service accounts and use the Keyless/OIDC approach

### Further details

Uses https://github.com/elastic/oblt-aws-auth-buildkite-plugin

PR #2573, opened 2025-05-06
bk: remove cloud cleanup without AWS OIDC

v1v:
https://github.com/elastic/elastic-package/pull/2567 enabled the OIDC support for the tear-down. In this case, let's remove the old fashion way.

This can be merged only when `integrations` are moved to OIDC.

Requires https://github.com/elastic/integrations/pull/13790#issuecomment-2863301372


.buildkite/hooks/pre-exit
```diff
@@ -17,7 +17,5 @@ cleanup
 unset_secrets
 
 # integrations-parallel-aws and intregrations-parallel-aws_logs
-unset ELASTIC_PACKAGE_AWS_ACCESS_KEY
-unset ELASTIC_PACKAGE_AWS_SECRET_KEY
 unset AWS_ACCESS_KEY_ID
 unset AWS_SECRET_ACCESS_KEY
```
mrodm, line 21: I guess this comment should be just for the cloud-cleanup step now ? 
In elastic-package, there is no other usage of creating AWS resources via Terraform.

Or could those two commands to unset AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY be removed too now ? IIRC those two variables were defined/exported just in `.buildkite/scripts/cloud-cleanup.sh`

v1v: Let me delete those two variables and the comment too

.buildkite/hooks/pre-exit
```diff
@@ -17,7 +17,5 @@ cleanup
 unset_secrets
 
 # integrations-parallel-aws and intregrations-parallel-aws_logs
```
v1v, line 19:
```suggestion
```

mrodm (approved):
:+1:

Issue #1722, opened 2024-03-13
[Feature] elastic-package export pipelines


nicpenning:
As an integration developer, it requires external tooling to use ingest pipelines in Kibana that have been developed and convert them to usable files in an integration (convert to YAML).

Similar to the Kibana export command where the data is exported in a usable format, I am purposing we can do that with `elastic-package`
 
https://github.com/elastic/elastic-package/pull/670#issuecomment-1036403036

It can be very easy for some to use Kibana to develop lengthy and worthwhile pipelines with a large sample set. So having the ability to run something like: `elastic-package export pipelines` and then be able to select the pipelines for export, this would save time having to convert them from JSON and ensure it's usable in the integration.

efd6:
This should be fixed. Please take a look at a recent build to confirm.

MichelLosier:
I'm working through the implementation on this one, and have been looking at the prior work with the `export dashboards` and `dump` commands. So far I have implemented:
* Retrieval of ingest pipelines from ES, along with dependent ones in `processors`
* Multi-selection of ingest pipelines if ids are not provided via flag

But in terms of writing the pipeline yaml files into the package, I do have a couple questions to consider:

**Where to write the exported ingest pipeline files?**

A. In a data_stream directory like: `<package-root>/data_stream/<name>/elasticsearch/ingest_pipeline`
- If so, seems we should also prompt for selecting an existing data stream directory to write to, then follow that up with the multi-select for the pipelines to export
- Does a bit more for the user

B. In `<package-root>/elasticsearch/ingest_pipeline`
  - I saw this was permitted in the [package spec here](https://github.com/elastic/package-spec/blob/main/spec/integration/spec.yml#L64-L68).
- I've seen more examples of ingest pipelines defined in `data_streams` than here. Is this basically a shared directory that other data_stream pipelines can reference?
- We can just dump them in here, and they can stay there or the user can move them to where they need to be.

C. Some `tmp/exports` directory.
- No sanitization, just as they are dumped into a staging area for the user to edit and move as needed. Would not be defined in package spec, since we don't want that directory staying around. 

**File naming / sanitization**

Much like dashboards exports, may want to consider renaming the ingest pipelines to remove details specific to their original context
- It seems ingest pipeline id/names follow this pattern `<package-name>.<datastream-name>-<package-semver>-<pipeline-name>` and if pipeline name is omitted, then it is assumed the `default` pipeline of that data stream? 
[...]

PR #2574, opened 2025-05-06
Add export ingest-pipelines command

MichelLosier:
Resolves: https://github.com/elastic/elastic-package/issues/1722

* Adds ingest pipeline export from elasticsearch to a package with the command `elastic-package export ingest-pipelines`
* User is prompted with a multi-selection of pipelines to export
   *  References to pipelines in a pipeline processor definitions will also be pulled automatically
* User is prompted to select an existing data stream in the package, or the package root as destinations to export to.
   * Dependent pipeline references are not prompted for, they'll be written to the same destination as selected for the parent.
* Pipelines are written to yaml files unchanged

  
Screenshot 2025-05-09 at 12 01 29 PM
Screenshot 2025-05-09 at 12 02 27 PM


cmd/export.go
```diff
@@ -38,111 +27,26 @@ func setupExportCommand() *cobraext.Command {
...
 		Long:  exportLongDescription,
 	}
 	cmd.AddCommand(exportDashboardCmd)
+	cmd.AddCommand(exportIngestPipelinesCmd)
 	cmd.PersistentFlags().StringP(cobraext.ProfileFlagName, "p", "", fmt.Sprintf(cobraext.ProfileFlagDescription, install.ProfileNameEnvVar))
 
 	return cobraext.NewCommand(cmd, cobraext.ContextPackage)
-}
-
-func exportDashboardsCmd(cmd *cobra.Command, args []string) error {
```
MichelLosier, line 52: Moves export dashboards functionality into own file now that we are adding support for ingest-pipelines

internal/dump/ingestpipelines.go
```diff
@@ -6,122 +6,17 @@ package dump
...
-	"io"
-	"net/http"
-	"slices"
 
 	"github.com/elastic/elastic-package/internal/elasticsearch"
+	"github.com/elastic/elastic-package/internal/elasticsearch/ingest"
 )
 
-// IngestPipeline contains the information needed to export an ingest pipeline.
-type IngestPipeline struct {
```
MichelLosier, line 19: A lot of this functionality was moved to the `internal/elasticsearch` package since it shares a lot of retrieval needs with the new export functionality

cmd/export_dashboards.go
```diff
@@ -0,0 +1,122 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
MichelLosier, line 1: This file was extracted from `cmd/export.go`

cmd/export_ingest_pipelines.go
```diff
@@ -0,0 +1,213 @@
...
+
+func promptIngestPipelineIDs(ctx context.Context, api *elasticsearch.API) ([]string, error) {
+	ingestPipelineNames, err := ingest.GetRemotePipelineNames(ctx, api)
+	if err != nil {
+		return nil, fmt.Errorf("finding ingest pipelines failed: %w", err)
+	}
+
+	ingestPipelineNames = slices.DeleteFunc(ingestPipelineNames, func(name string) bool {
+		// Filter out system pipelines that start with dot "." or global@
+		return strings.HasPrefix(name, ".") || strings.HasPrefix(name, "global@")
```
MichelLosier, line 146: Any other pipeline types / patterns we should exclude here?

internal/export/ingest_pipelines.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	yamlBytes, err := yaml.Marshal(jsonPipeline)
+	if err != nil {
+		return fmt.Errorf("marshalling ingest pipeline json to yaml failed (ID: %s): %w", pipeline.Name(), err)
+	}
+
+	// requirement: https://github.com/elastic/package-spec/pull/54
+	documentStartDashes := []byte("---\n")
+	documentBytes := append(documentStartDashes, yamlBytes...)
```
jsoriano, line 106: Nit. To avoid appending, you can use a bytes buffer. Something like this:

```suggestion
	var documentBytes bytes.Buffer
	// requirement: https://github.com/elastic/package-spec/pull/54
	documentBytes.WriteString("---\n")
	err := yaml.NewEncoder(&documentBytes).Encode(jsonPipeline)
	if err != nil {
		return fmt.Errorf("marshalling ingest pipeline json to yaml failed (ID: %s): %w", pipeline.Name(), err)
	}
```

internal/export/ingest_pipelines.go
```diff
@@ -0,0 +1,131 @@
...
+	ParentPath string
+}
+
+func (p PipelineWriteLocation) WritePath() string {
+	return filepath.Join(p.ParentPath, "elasticsearch", "ingest_pipeline")
+}
+
+type PipelineWriteAssignments map[string]PipelineWriteLocation
+
+func IngestPipelines(ctx context.Context, api *elasticsearch.API, writeAssignments PipelineWriteAssignments) error {
```
jsoriano, line 40: Could we have a test case for this?

internal/export/ingest_pipelines.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func writePipelineToFile(pipeline ingest.RemotePipeline, writeLocation PipelineWriteLocation) error {
+	var jsonPipeline map[string]any
+	err := json.Unmarshal(pipeline.JSON(), &jsonPipeline)
+	if err != nil {
+		return fmt.Errorf("unmarshalling ingest pipeline failed (ID: %s): %w", pipeline.Name(), err)
+	}
+
+	yamlBytes, err := yaml.Marshal(jsonPipeline)
```
jsoriano, line 99: I wonder if we should do some formatting here. Some things:
* Remove `_meta`, specially the `managed*` fields should not be needed here.
* Remove `version`, I think this is not being used in packages.
* Indentation. Looking to pipelines they use to be indented by two spaces, default marshalling seems to use four spaces.

Maybe we should make `elastic-package format` to automatically format these files too. But that is another story.

jsoriano (commented):
Thanks! This is looking quite good, and works well. Added some comments.

cmd/export_ingest_pipelines.go
```diff
@@ -0,0 +1,213 @@
...
+	var questions []*survey.Question
+
+	for _, pipelineName := range pipelineNames {
+		question := &survey.Question{
+			Name: pipelineName,
+			Prompt: &survey.Select{
+				Message: fmt.Sprintf("Select a location to export ingest pipeline '%s'", pipelineName),
+				Options: options,
+				Description: func(value string, index int) string {
+					idx := slices.IndexFunc(writeLocations, func(p export.PipelineWriteLocation) bool { return p.Name == value })
```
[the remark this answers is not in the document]
MichelLosier: ah good catch, you're right -- `[]options` are derived from `[]writeLocations` so positions should be the same

internal/dump/installedobjects_test.go
```diff
@@ -39,18 +43,20 @@ func TestDumpInstalledObjects(t *testing.T) {
 		&installedObjectsDumpSuite{
 			// To reproduce the scenario:
 			// - Start the stack with version 7.16.2.
-			// - Install apache package (1.3.4).
+			// - Install apache package (1.3.5).
```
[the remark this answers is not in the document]
MichelLosier: A changed that was introduced here was having `getIngestPipelines` make bulk requests for selected pipelines and their dependent processor pipelines. That functionality was moved to a shared location, so the diff is harder to notice.

Original
https://github.com/elastic/elastic-package/pull/2574/files#diff-e2e0dfd36e0729c7bbcf8c5653e42269861139aeaac9f15c51f2b350989842cdL50

New
https://github.com/elastic/elastic-package/pull/2574/files#diff-a4f552962ddca8d0f873f854bec0d949905c2c07ca8a932d7d72a4c364e308d5R157-R161

Since this changed the shape of the request / interactions, I needed to re-record them. The versions though these fixtures were originally recorded on no longer exist in EPR, so I had to record on the next available version.

internal/export/ingest_pipelines.go
MichelLosier: I didn't come across cases where the `version` field showed up in a pipeline while exporting, I assume you are seeing this as a top-level field in some instances?

internal/export/ingest_pipelines_test.go
```diff
@@ -0,0 +1,194 @@
...
+			Type:       PipelineWriteLocationTypeRoot,
+			Name:       pipelineID,
+			ParentPath: outputDir,
+		}
+	}
+
+	return writeAssignments
+}
+
+func countFiles(t *testing.T, dir string) (count int) {
```
MichelLosier, line 120: Fair bit of testing utils borrowed from the `dump` package tests. I didn't come across an existing package where shared testing utils might have a home, but since this is the only place this code is repeated, I figured this is fine for now. Otherwise, I could create an `internal/testutils` package if we expect to see more of this type of testing pattern of comparing file output from the CLI.

internal/export/ingest_pipelines.go
jsoriano: Yes, it appears at the root level. After your latest changes it doesn't appear anymore. Thanks.

internal/export/ingest_pipelines_test.go
jsoriano: Yep, this is fine by now, but creating a shared package sounds good too, as you prefer.

internal/export/ingest_pipelines.go
```diff
@@ -96,23 +97,27 @@ func writePipelineToFile(pipeline ingest.RemotePipeline, writeLocation PipelineW
...
-	yamlBytes, err := yaml.Marshal(jsonPipeline)
+	delete(jsonPipeline, "_meta")
+	delete(jsonPipeline, "version")
+
+	var documentBytes bytes.Buffer
+	// requirement: https://github.com/elastic/package-spec/pull/54
+	documentBytes.WriteString("---\n")
+	yamlEncoder := yaml.NewEncoder(&documentBytes)
+	yamlEncoder.SetIndent(2)
+	err = yamlEncoder.Encode(jsonPipeline)
```
jsoriano, line 108: Btw, we have a `formatter` package with shared formatters for JSON and YAML. They do basically what we are doing here, with the exception of filtering out fields.

Maybe we can add a method `FormatWithFilter(content []byte, filter func(map[string]any))`

But current approach is fine too, so as you prefer.

jsoriano (approved):
Nice, thanks for adding the test cases!

PR #2581, opened 2025-05-07
Set the expected COMMAND when DRY_RUN is false

mrodm:
Use different DRY_RUN env. vars for each step.


.buildkite/scripts/cloud-cleanup.sh
```diff
@@ -25,26 +25,43 @@ export DELETE_RESOURCES_BEFORE_DATE
...
 else
     COMMAND="plan"
 fi
 
+buildkite-agent annotate \
+  "[${BUILDKITE_STEP_KEY}] Running DRY_RUN (${DRY_RUN}) using cloud-reaper command \"${COMMAND}\"${redshift_message}" \
+  --context "ctx-cloud-reaper-info-deprecated" \
+  --style "info"
+
 any_resources_to_delete() {
```
jsoriano, line 48: Are these methods duplicated? Could we move them to some shared file?

jsoriano (commented):
I don't fully understand this PR :thinking: How would we run each step independently with DRY_RUN false?

.buildkite/scripts/cloud-cleanup.sh
mrodm: This file is going to be deprecated by the new file introduced in #2567 once integrations also use the new system cc @v1v.

So I thought it's not needed to do refactors fow now here.

mrodm:
> I don't fully understand this PR 🤔 How would we run each step independently with DRY_RUN false?

Sorry, I did not add that explanation. These changes introduce a new selector in the Buildkite UI, so the user can select which step is going to be run in DRY_RUN mode.
- DRY_RUN would just be honored by the new step (OIDC)
- DRY_RUN_DEPRECATED would just be honored by the previous step.

Input variables selector

jsoriano:
Oh ok, I misunderstood, I thought that we were fully deprecating the current variable, but I was missing what was replacing it. Thanks.

jsoriano (approved):

Issue #2258, opened 2024-12-05
Managed flag is set to false when exporting dashboards


taylor-swanson:
When exporting dashboards, either brand new dashboard or an existing managed one made editable by `elastic-package`, the managed flag on the dashboard (within the exported JSON) is set to `false`. This may be desirable in some situations, but in most cases, these dashboards are meant for packages managed by fleet, so the desired value should be `true`.

![Image](https://github.com/user-attachments/assets/397dbb1a-1c62-49e7-b38c-4a9614b6f197)

At the moment, this doesn't seem to cause any issues, since it appears Fleet either overrides this value or ignores it completely. However, many (if not all) of our integrations have `managed` set to `false` in checked-in code. Another option would be to not include `managed` in the JSON at all, but not sure if that's the equivalent to `"managed": false`.

Another thing to note, adding the `managed` tag on a dashboard doesn't appear to set the managed flag in the JSON, so as of right now, the only way to set it to true is to edit the JSON.

jsoriano:
I think we could remove `managed` when exporting dashboards, as Fleet will set this value.

PR #2586, opened 2025-05-09
Remove 'managed' key from exported dashboards

MichelLosier:
Resolves: https://github.com/elastic/elastic-package/issues/2258

Removes the "managed" key from dashboards when exported.


jsoriano (approved):
Good one, thanks!

PR #2590, opened 2025-05-13
Migrate to github.com/olekukonko/tablewriter 1.0

jsoriano:
Tablewriter introduced several breaking changes in its 1.0 version, both in its APIs and in its behaviour.

Adapt the code to the new major, configuring renderer to behave as similar as possible as in v0.

Take the opportunity to refactor the related code to reuse table settings where possible.

Supersedes https://github.com/elastic/elastic-package/pull/2591, https://github.com/elastic/elastic-package/pull/2597, https://github.com/elastic/elastic-package/pull/2600.


cmd/testdata/status-local-version-stage
```diff
@@ -1,11 +1,11 @@
...
-|             | 1.0.2)                         |                   |       |                 |
-+-------------+--------------------------------+-------------------+-------+-----------------+
++-------------+----------------------------+-------------------+-------+-----------------+
+| ENVIRONMENT |          VERSION           |      RELEASE      | TITLE |   DESCRIPTION   |
++-------------+----------------------------+-------------------+-------+-----------------+
+| Local       | 2.0.0-rc1                  | Release Candidate | Foo   | Foo integration |
++-------------+----------------------------+-------------------+-------+-----------------+
+| Production  | 1.1.0-beta1 (1.0.0, 1.0.1, | Beta              | Foo   | Foo integration |
+|             | 1.0.2)                     |                   |       |                 |
++-------------+----------------------------+-------------------+-------+-----------------+
```
jsoriano, line 11: Wrapping seems to change in 1.0, I haven't managed to mimic old behaviour, but the new looks better.

cmd/status.go
```diff
@@ -255,37 +249,35 @@ func renderPackageVersions(p *status.PackageStatus, w io.Writer, extraParameters
...
+	table.Render()
+}
 
-	table.SetColumnColor(columnColors...)
+var formatHeadersReplacer = strings.NewReplacer("_", " ", ".", " ")
 
-	table.SetRowLine(true)
-	table.AppendBulk(environmentTable)
-	table.Render()
+func formatHeaders(headers []string) []string {
```
jsoriano, line 275: This was done automatically in v0, I haven't found how to do this in v1.

cmd/tables.go
```diff
@@ -0,0 +1,57 @@
...
+		Header: renderer.Tint{
+			FG: renderer.Colors{color.Bold},
+		},
+		Column: renderer.Tint{
+			Columns: []renderer.Tint{
+				{FG: renderer.Colors{color.Bold, color.FgCyan}},
+			},
+		},
+		Settings:  defaultTableRendererSettings,
+		Symbols:   tw.NewSymbols(tw.StyleASCII),
```
jsoriano, line 53: We could change this to the default style in v1, that would render tables like this (in my shell all lines are continuous):
```
$ elastic-package status -C ./packages/apache --info kibana.version
Package: apache
Owner: elastic/obs-infraobs-integrations
Package Versions:
┌─────────────┬─────────┬─────────┬────────────────────┬───────────────────────────────┬───────────────────┐
│ ENVIRONMENT │ VERSION │ RELEASE │       TITLE        │          DESCRIPTION          │  KIBANA VERSION   │
├─────────────┼─────────┼─────────┼────────────────────┼───────────────────────────────┼───────────────────┤
│ Local       │ 2.0.0   │ GA      │ Apache HTTP Server │ Collect logs and metrics from │ ^8.13.0 || ^9.0.0 │
│             │         │         │                    │ Apache servers with Elastic   │                   │
│             │         │         │                    │ Agent.                        │                   │
├─────────────┼─────────┼─────────┼────────────────────┼───────────────────────────────┼───────────────────┤
│ Production  │ 2.0.0   │ GA      │ Apache HTTP Server │ Collect logs and metrics from │ ^8.13.0 || ^9.0.0 │
│             │         │         │                    │ Apache servers with Elastic   │                   │
│             │         │         │                    │ Agent.                        │                   │
└─────────────┴─────────┴─────────┴────────────────────┴───────────────────────────────┴───────────────────┘
```

FTR, this is the output with ASCII style:
```
$ go run github.com/elastic/elastic-package status -C ./packages/apache --info kibana.version
Package: apache
Owner: elastic/obs-infraobs-integrations
Package Versions:
+-------------+---------+---------+--------------------+--------------------------------+-------------------+
| ENVIRONMENT | VERSION | RELEASE |       TITLE        |          DESCRIPTION           |  KIBANA VERSION   |
+-------------+---------+---------+--------------------+--------------------------------+-------------------+
| Local       | 2.0.0   | GA      | Apache HTTP Server | Collect logs and metrics from  | ^8.13.0 || ^9.0.0 |
|             |         |         |                    | Apache servers with Elastic    |                   |
|             |         |         |                    | Agent.                         |                   |
+-------------+---------+---------+--------------------+--------------------------------+-------------------+
| Production  | 2.0.0   | GA      | Apache HTTP Server | Collect logs and metrics from  | ^8.13.0 || ^9.0.0 |
|             |         |         |                    | Apache servers with Elastic    |                   |
|             |         |         |                    | Agent.                         |                   |
+-------------+---------+---------+--------------------+--------------------------------+-------------------+
```

jsoriano: It is easy to change :slightly_smiling_face: We have a `-generate` flag. Changed and tests updated.

cmd/status.go
```diff
@@ -225,20 +226,13 @@ func renderPendingChanges(p *status.PackageStatus, w io.Writer) {
...
-	table := tablewriter.NewWriter(w)
-	table.SetHeader([]string{"Type", "Description", "Link"})
-	table.SetHeaderColor(
-		twColor(tablewriter.Colors{tablewriter.Bold}),
-		twColor(tablewriter.Colors{tablewriter.Bold}),
-		twColor(tablewriter.Colors{tablewriter.Bold}),
+	colorCfg := defaultColorizedConfig()
+	table := tablewriter.NewTable(w,
+		tablewriter.WithRenderer(renderer.NewColorized(colorCfg)),
+		tablewriter.WithConfig(defaultTableConfig),
```
mrodm, line 232: Just found that `elastic-package stack status` and other commands use another dependency to render data into tables: `"github.com/jedib0t/go-pretty/table"`

```
cmd/stack.go:   "github.com/jedib0t/go-pretty/table"
internal/benchrunner/runners/pipeline/format.go:        "github.com/jedib0t/go-pretty/table"
internal/benchrunner/runners/pipeline/format.go:        "github.com/jedib0t/go-pretty/text"
internal/benchrunner/runners/rally/report.go:   "github.com/jedib0t/go-pretty/table"
internal/benchrunner/runners/rally/report.go:   "github.com/jedib0t/go-pretty/text"
internal/benchrunner/runners/system/report.go:  "github.com/jedib0t/go-pretty/table"
internal/benchrunner/runners/system/report.go:  "github.com/jedib0t/go-pretty/text"
internal/testrunner/reporters/formats/human.go: "github.com/jedib0t/go-pretty/table"
```

Probably, it could be used eventually the same dependency everywhere (in follow-ups), WDYT ?

mrodm (approved):
:+1:

cmd/status.go
jsoriano: Good one, not sure which one is better.

By now I have done some changes for consistency in https://github.com/elastic/elastic-package/pull/2590/commits/80f55791f0a694ad142f3f9ec310dd0d1500ccf7:
- `stack status` uses tablewriter now, so at least everything under `cmd` has consistent tables. This changes `stack status` to show colors, bold headers and lines between rows:
imagen
- Default style modified to "Rounded", it looks nicer, and will be consistent with the other tables.

From this point, not sure what to do, do you have an opinion? `go-pretty` looks simpler, but also we are not using many formatting options, so at the end it may be similar in complexity.

Basically we have three options:
* Migrate everything to go-pretty.
* Migrate everything to tablewriter.
* Leave things as they are now, using tablewriter for commands, and go-pretty for test reports.

If we go for using the same library, I can create an issue to delay the decision (maybe forever? :sweat_smile:)

mrodm: > From this point, not sure what to do, do you have an opinion? `go-pretty` looks simpler, but also we are not using many formatting options, so at the end it may be similar in complexity.
> 
> Basically we have three options:
> 
> * Migrate everything to go-pretty.
> * Migrate everything to tablewriter.
> * Leave things as they are now, using tablewriter for commands, and go-pretty for test reports.

I guess, it would be good to just use one dependency for rendering tables. But I don't have an strong opinion in which one to choose. 

For me, it could be kept as it is now in this PR. Trying to keep the usage of `tablewriter` in `cmd` and `go-pretty` for other usages.

> If we go for using the same library, I can create an issue to delay the decision (maybe forever? 😅)

We could create an issue as a change that would be nice to have (using just one library).

> I think I prefer the previous, more compact format for `stack status`, I have modified it to:

Comparing both versions, I also prefer the previous one. The one shown in the screenshot here: https://github.com/elastic/elastic-package/pull/2590#discussion_r2098426814

jsoriano: Follow up issue created https://github.com/elastic/elastic-package/issues/2604.

cmd/stack.go
```diff
@@ -355,12 +357,15 @@ func printStatus(cmd *cobra.Command, servicesStatus []stack.ServiceStatus) {
 		cmd.Printf(" - No service running\n")
 		return
 	}
-	t := table.NewWriter()
-	t.AppendHeader(table.Row{"Service", "Version", "Status"})
-
+	config := defaultColorizedConfig()
+	config.Settings.Separators.BetweenRows = tw.Off
+	table := tablewriter.NewTable(cmd.OutOrStderr(),
```
jsoriano, line 362: I am using `cmd.OutOrStderr()` here because it is what `cmd.Println` does. I found out by seeing that `scripts/test-stack-command.sh` uses `2>` to redirect its output for comparison.

We should also work on consistency of outputs, I would say that all these tables should go to stdout.

Not to change in this PR in any case.

mrodm (approved):

PR #2593, opened 2025-05-14
Make is_ready container-image configurable

maximilianpohle:
Hi everybody,
due to the reduced Docker rate limits, it may happen that the stack can no longer be downloaded because of the is_ready Docker image.

Therefore, I have created an additional environment variable that allows overriding the image.

Best regards
Max

### DIFF
--- a/internal/install/application_configuration.go
+++ b/internal/install/application_configuration.go
@@ -102,7 +102,7 @@ func (s stack) ImageRefOverridesForVersion(version string) ImageRefs {
 		Elasticsearch: checkImageRefOverride("ELASTICSEARCH_IMAGE_REF_OVERRIDE", stringOrDefault(appConfigImageRefs.Elasticsearch, "")),
 		Kibana:        checkImageRefOverride("KIBANA_IMAGE_REF_OVERRIDE", stringOrDefault(appConfigImageRefs.Kibana, "")),
 		Logstash:      checkImageRefOverride("LOGSTASH_IMAGE_REF_OVERRIDE", stringOrDefault(appConfigImageRefs.Logstash, "")),
-		is_ready:      checkImageRefOverride("IS_READY_IMAGE_REF_OVERRIDE", stringOrDefault(appConfigImageRefs.is_ready, "")),
+		IsReady:       checkImageRefOverride("IsReady_IMAGE_REF_OVERRIDE", stringOrDefault(appConfigImageRefs.IsReady, "")),
 	}
 }
 
@@ -112,7 +112,7 @@ type ImageRefs struct {
 	Elasticsearch string `yaml:"elasticsearch"`
 	Kibana        string `yaml:"kibana"`
 	Logstash      string `yaml:"logstash"`
-	IsReady      string `yaml:"is_ready"`
+	IsReady       string `yaml:"is_ready"`
 }
 
 // AsEnv method returns key=value representation of image refs.
@@ -122,7 +122,7 @@ func (ir ImageRefs) AsEnv() []string {
 	vars = append(vars, "ELASTICSEARCH_IMAGE_REF="+ir.Elasticsearch)
 	vars = append(vars, "KIBANA_IMAGE_REF="+ir.Kibana)
 	vars = append(vars, "LOGSTASH_IMAGE_REF="+ir.Logstash)
-	vars = append(vars, "IS_READY_IMAGE_REF="+ir.is_ready)
+	vars = append(vars, "IsReady_IMAGE_REF="+ir.IsReady)
 	return vars
 }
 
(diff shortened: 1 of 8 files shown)


internal/install/application_configuration.go
```diff
@@ -110,6 +112,7 @@ type ImageRefs struct {
 	Elasticsearch string `yaml:"elasticsearch"`
 	Kibana        string `yaml:"kibana"`
 	Logstash      string `yaml:"logstash"`
+	is_ready      string `yaml:"is_ready"`
```
jsoriano, line 115:
```suggestion
	IsReady      string `yaml:"is_ready"`
```

internal/stack/_static/docker-compose-stack.yml.tmpl
jsoriano, line 1: Please review other compose files in this directory that may be using this image.

internal/stack/_static/docker-compose-stack.yml.tmpl
```diff
@@ -28,7 +28,7 @@ services:
       - "127.0.0.1:9200:9200"
 
   elasticsearch_is_ready:
-    image: tianon/true:multiarch
+    image: "${IS_READY_IMAGE_REF}"
```
mrodm, line 31: Should it be updated also the docker-compose for local services ? 
https://github.com/elastic/elastic-package/blob/f00e430124b1e8bb43f234fcc7e756461a107786/internal/stack/_static/local-services-docker-compose.yml.tmpl

To use that environment variable, this file uses facts instead. Here it is the method where those facts are set:
https://github.com/elastic/elastic-package/blob/f00e430124b1e8bb43f234fcc7e756461a107786/internal/stack/localresources.go#L55

internal/install/application_configuration.go
```diff
@@ -37,6 +37,7 @@ const (
 	elasticsearchImageName              = "docker.elastic.co/elasticsearch/elasticsearch"
 	kibanaImageName                     = "docker.elastic.co/kibana/kibana"
 	logstashImageName                   = "docker.elastic.co/logstash/logstash"
+	isreadyImageName                    = "tianon/true:multiarch"
```
mrodm, line 40: Nit.
```suggestion
	isReadyImageName                    = "tianon/true:multiarch"
```

internal/stack/_static/local-services-docker-compose.yml.tmpl
```diff
@@ -40,7 +40,7 @@ services:
     - "host.docker.internal:host-gateway"
 
   fleet-server_is_ready:
-    image: tianon/true:multiarch
+    image: "${ISREADY_IMAGE_REF}"
```
mrodm, line 43: Here in this docker-compose file template, it can be used the function `fact` to get the value of the `is_ready` docker image for those 3 containers instead of the environment variable to follow the same pattern as the other services:

```suggestion
    image: "{{ fact "isready_image" }}"
```

Other docker-compose files must keep using the environment variable.

mrodm (commented):
Thanks! 
Just added minor comments.
It would be needed to check if the environment variable can be used to define the docker-compose scenario in the `sql_input` test package.

test/packages/parallel/sql_input/_dev/deploy/docker/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       - ${SERVICE_LOGS_DIR}/sql_input:/var/log/mysql
       - mysqldata:/var/lib/mysql
   sql_input_is_ready:
-    image: tianon/true:multiarch
+    image: "${ISREADY_IMAGE_REF}"
```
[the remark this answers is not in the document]
mrodm: The test for that package failed with this error:
``` 
could not setup service: could not boot up service using Docker Compose: running Docker Compose up command failed: exit status 15: service "sql_input_is_ready" has neither an image nor a build context specified: invalid compose project
```

Those environment variables are available for the stack docker-compose, not for the services defined in system tests. So, here for the `sql_input`, it needs to be set back to `tianon/true:multiarch` docker image:

```suggestion
    image: tianon/true:multiarch
```

internal/stack/_static/local-services-docker-compose.yml.tmpl
```diff
@@ -76,7 +76,7 @@ services:
     - "host.docker.internal:host-gateway"
 
   elastic-agent_is_ready:
-    image: tianon/true:multiarch
+    image: "${ISREADY_IMAGE_REF}"
```
mrodm, line 79: This image should be using also fact as the previous service:

```suggestion
    image: "{{ fact "isready_image" }}"
```

jsoriano (approved):

maximilianpohle:
@mrodm @jsoriano  Thanks for approving! Can we merge?

mrodm:
> @mrodm @jsoriano Thanks for approving! Can we merge?

Sure, I'll merge this PR. 
Thanks for your contribution @maximilianpohle !

Issue #1302, opened 2023-06-12
Support deploying multiple services in system tests


bhapas:
This is a feature request to support multiple ServiceDeployers in a testrunner. When a [ServiceDeployer](https://github.com/elastic/elastic-package/blob/main/docs/howto/system_testing.md#defining-a-system-test) is defined for system testing , it is only allowed to have only [one serviceDeployer](https://github.com/elastic/elastic-package/blob/main/internal/testrunner/runners/system/servicedeployer/factory.go#L102-L103) per testrunner.

### Usecase

When terraform service is to be tested against another docker based service like [localstack](https://docs.localstack.cloud/getting-started/installation/#docker) it should be possible to setup a docker service in one container and terraform service in another container.

cc : @andrewkroh @jsoriano

efd6:
Related to this is that when we have an integration that has both API and cloud storage inputs, we are not able to test both. This kind of situation is becoming more frequent where vendors are providing, e.g., an S3 path and an API endpoint to obtain data from. When we have an integration like this we have to make a choice about which we are going to test. In a not insignificant number of cases this has left us open to regressions in the untested case, wasting time and impacting on user experience.

bhapas:
> Related to this is that when we have an integration that has both API and cloud storage inputs, we are not able to test both. This kind of situation is becoming more frequent where vendors are providing, e.g., an S3 path and an API endpoint to obtain data from. When we have an integration like this we have to make a choice about which we are going to test. In a not insignificant number of cases this has left us open to regressions in the untested case, wasting time and impacting on user experience.

@efd6 Thanks for pointing this out. This scenario also lead to testing most common inputs in system tests leaving out some inputs untested for a production grade integration until it comes back as an SDH. https://github.com/elastic/integrations/issues/8439 has some examples of integrations where multiple inputs needed testing , but could not. I am sure this list should be much more now.

mrodm:
> I think the first one solves some issues like testing all inputs in the same system test suite within the integration.. Like running the system test for API based / cloud based inputs by passing some sort of flag in the system test command.

@bhapas Perfect, thanks! I start then to check how to allow multiple service deployers into one data stream, but just run one of them per test.

efd6:
Agree with @bhapas.

mrodm:
[...]

PR #2638, opened 2025-06-05
Allow to define more than one service deployer and choose the deployer via configuration setting

mrodm:
Relates https://github.com/elastic/elastic-package/issues/1302

This PR allows us to define more than one service deployer in the `_dev/deploy` folder (at package or data stream level).

If there are more than one service deployer defined, it must be added to the test system configuration file a new setting `deployer`. This new setting helps us to choose which specific specific service deployer will be used by each test. Example:
- Test using `docker` deployer:
```yaml
deployer: docker
service: nginx
vars: ~
data_stream:
  vars:
    paths:
      - "{{SERVICE_LOGS_DIR}}/access.log*"
```
- Test using `terrafirn` deployer:
```yaml
deployer: tf
wait_for_data_timeout: 2m
vars: ~
data_stream:
  vars:
    paths:
      - "{{SERVICE_LOGS_DIR}}/file.log*"
```

If this new setting is not present in the configuration file, there must be zero or one service deployers defined (same behavior as before).

Added a new test package where this new feature is used: `nginx_multiple_services`
This test package is a subset of the `nginx` test package (just access and error data-stream and no kibana assets).

### Author's checklist
- [x] Test packages in integrations repositiory with this `elastic-package`: https://github.com/elastic/integrations/pull/14180
- [x] Add a test package (`nginx_multiple_services`) defining both docker and terraform service deployers.


mrodm:
test integrations

internal/testrunner/runners/system/test_config.go
```diff
@@ -129,6 +135,11 @@ func newConfig(configFilePath string, svcInfo servicedeployer.ServiceInfo, servi
 		c.Agent.PreStartScript.Language = agentdeployer.DefaultAgentProgrammingLanguage
 	}
 
+	// Not included in package-spec validation for deployer name
+	if c.Deployer != "" && !slices.Contains(allowedDeployerNames, c.Deployer) {
+		return nil, fmt.Errorf("invalid deployer name %q in system test configuration file %q, allowed values are: %s", c.Deployer, configFilePath, strings.Join(allowedDeployerNames, ", "))
+	}
```
mrodm, line 141: Added validation here instead of package-spec. In package-spec this file has `additionalProperties: true`

internal/testrunner/runners/system/test_config.go
```diff
@@ -24,7 +25,10 @@ import (
 	"github.com/elastic/elastic-package/internal/testrunner"
 )
 
-var systemTestConfigFilePattern = regexp.MustCompile(`^test-([a-z0-9_.-]+)-config.yml$`)
+var (
+	systemTestConfigFilePattern = regexp.MustCompile(`^test-([a-z0-9_.-]+)-config.yml$`)
+	allowedDeployerNames        = []string{"docker", "k8s", "tf"}
```
jsoriano, line 30: Do we need to abbreviate? :slightly_smiling_face: And I think `compose` is more accurate than `docker`.
```suggestion
	allowedDeployerNames        = []string{"compose", "kubernetes", "terraform"}
```

Though I guess you are using the directory names we support?

internal/agentdeployer/factory.go
```diff
@@ -111,43 +115,25 @@ func selectAgentDeployerType(options FactoryOptions) (string, error) {
...
-	} else if err != nil && !errors.Is(err, os.ErrNotExist) {
-		return "", fmt.Errorf("stat failed for package (path: %s): %w", packageDevDeployPath, err)
-	}
-
-	return "", fmt.Errorf("\"%s\" %w", options.DevDeployDir, os.ErrNotExist)
-}
-
-func findAgentDeployers(devDeployPath string) ([]string, error) {
-	fis, err := os.ReadDir(devDeployPath)
+func findAgentDeployers(devDeployPath, expectedDeployer string) ([]string, error) {
```
jsoriano, line 118: Nit. This method returns now a single deployer or none, right? We could change it to return a single value, and we could return a named error in case of no deployer found.
```suggestion
func findAgentDeployer(devDeployPath, expectedDeployer string) (string, error) {
```

test/packages/parallel/nginx_multiple_services/manifest.yml
jsoriano, line 1: Nit. Could we also have a failing case, with a package that defines multiple deployers but only defines one?

jsoriano (approved):
Looks good.

test/packages/parallel/nginx_multiple_services/manifest.yml
jsoriano: Oh sorry, I mispelled, I meant a package that defines multiple deployers, but uses none. So that this error is reproduced: `expected to find only one agent deployer in...`

mrodm: Added a test package `nginx_missing_deployer` with two errors:
- data_stream access requires `tf` deployer but it is not present in the `_dev/deploy` folder from the data stream.
- data_stream error uses the `_dev/deploy` folder (from package level) where there are two deployers defined, but it does not set `deployer` field in the system test configuration.

jsoriano (approved):
Looks good, thanks for adding the test cases!

PR #2642, opened 2025-06-06
Dump all service logs even if there is any error

mrodm:
Allow to dump all service logs even if one of them fails (e.g. container was not created).

Example of build where all the container logs were dumped:
https://buildkite.com/elastic/elastic-package/builds/5434#019745c5-7e0d-4065-b34f-af4121b8f5b7

Example of build where just one the container logs was dumped (`elastic-agent` container was not created):
https://buildkite.com/elastic/elastic-package/builds/5389#0197201e-faa5-4f25-b4c8-da59e732e12c

```
 Container elastic-package-stack-fleet-server-1  Waiting
 Container elastic-package-stack-fleet-server-1  Waiting
 Container elastic-package-stack-fleet-server-1  Error
 Container elastic-package-stack-fleet-server-1  Error
 Container elastic-package-stack-kibana_is_ready-1  Started
dependency failed to start: container elastic-package-stack-fleet-server-1 is unhealthy
Error: booting up the stack failed: running docker-compose failed: running command failed: running Docker Compose up command failed: exit status 1: Container elastic-package-stack-package-registry-1  Running
...
2025/05/30 07:44:33 DEBUG run command: /usr/bin/docker cp elastic-package-stack-elastic-agent-1:/usr/share/elastic-agent/state/data/logs/ build/elastic-stack-dump/stack/9.1.0-4fd38a91-SNAPSHOT/logs/elastic-agent-internal
Error: dump failed: can't dump Elastic stack logs: can't copy internal logs (service: elastic-agent): docker copy failed: could not copy files from the container (stderr="Error response from daemon: Could not find the file /usr/share/elastic-agent/state/data/logs/ in container elastic-package-stack-elastic-agent-1\n"): exit status 1
```

### DIFF
--- a/scripts/test-build-install-zip.sh
+++ b/scripts/test-build-install-zip.sh
@@ -19,21 +19,11 @@ cleanup() {
   exit $r
 }
 
+trap cleanup EXIT
+
 testype() {
   basename "$(dirname "$1")"
 }
-
-stackVersion() {
-  curl -s \
-    -u "${ELASTIC_PACKAGE_ELASTICSEARCH_USERNAME}:${ELASTIC_PACKAGE_ELASTICSEARCH_PASSWORD}" \
-    --cacert "${ELASTIC_PACKAGE_CA_CERT}" \
-    -H 'content-type: application/json' \
-    -H 'kbn-xsrf: true' \
-    -f "${ELASTIC_PACKAGE_KIBANA_HOST}/api/status" | yq -r '.version.number'
-}
-
-
-trap cleanup EXIT
 
 OLDPWD=$PWD
 # Build packages
@@ -61,8 +51,6 @@ elastic-package stack up -d -v
 
 eval "$(elastic-package stack shellinit)"
 
-stack_version=$(stackVersion)
-
 # Install packages from working copy
 for d in test/packages/*/*/; do
   # Packages in false_positives can have issues.
@@ -72,11 +60,6 @@ for d in test/packages/*/*/; do
   package_name=$(yq -r '.name' "${d}/manifest.yml")
   package_version=$(yq -r '.version' "${d}/manifest.yml")
 
-  PACKAGE_NAME_VERSION="${package_name}-${package_version}"
-  if [[ "${stack_version}" =~ ^9 ]]; then
-    PACKAGE_NAME_VERSION="${package_name}/${package_version}"
-  fi
-
   elastic-package install -C "$d" -v
 
   # check that the package is installed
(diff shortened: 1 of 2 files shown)


scripts/test-build-install-zip.sh
```diff
@@ -51,15 +61,21 @@ elastic-package stack up -d -v
...
   fi
-  package_name=$(cat "${d}/manifest.yml" | yq -r .name)
-  package_version=$(cat "${d}/manifest.yml" | yq -r .version)
+  package_name=$(yq -r '.name' "${d}/manifest.yml")
+  package_version=$(yq -r '.version' "${d}/manifest.yml")
+
   PACKAGE_NAME_VERSION="${package_name}-${package_version}"
+  if [[ "${stack_version}" =~ ^9 ]]; then
+    PACKAGE_NAME_VERSION="${package_name}/${package_version}"
+  fi
```
mrodm, line 78: Should we use just the latest syntax for this API `<name>/<version>` ?
These scripts are just used with the default stack version in elastic-package, they should run successfully.

mrodm: Or this could be moved to another PR...

mrodm: Changes in shell-scripts under `scripts` folder were added since I did some tests running test packages with Elastic stack version 9.1.0-SNAPSHOT.

It looks like that in 8.x version, there were two ways of requesting information of a given package (mainly it is used here to know if the package is installed or not):

```
${ELASTIC_PACKAGE_KIBANA_HOST}/api/fleet/epm/packages/<package>-<version>"
${ELASTIC_PACKAGE_KIBANA_HOST}/api/fleet/epm/packages/<package>/<version>"
```

But in 9.x version, just the latter is valid (was deprecated/removed the first one?):
```
${ELASTIC_PACKAGE_KIBANA_HOST}/api/fleet/epm/packages/<package>/<version>
```

Checking these scripts, they should be running always with the default stack version that currently is 8.18.2. That means that probably we could use just the latter version to query about the information of a given package.

As a note, there are more changes required once the default Elastic stack version is set to 9.x:
- Update Kibana constraints manifests in some test packages.
- Review errors in packages using the deprecated custom agents.
    - Should we remove the support for this ?
- Errors found in Elastic Agent logs (related to https://github.com/elastic/integrations/issues/14022).

kpollich: Got it - thanks for confirming. I think it makes sense to remove the conditional at this point since it should never be triggered against the latest stack version. The APIs are equivalent (same code path) and since we will never be hitting an early 8.x stack or 7.x we can just default to the non-deprecated API path.

PR #2707, opened 2025-07-03
Update new package README template

mjwolf:
Update the README templates and generator used in new package creation to match the new standardized README structure. The updated template should be better at helping users write the readme for new packages, by explaining what is needed in each section of the template.

This also makes a few other small changes which will make it easier for new users to create and use integration packages. It moves the generated readme to `_dev/build/docs`, automatically adds a "Do not modify" message to the README generated during `build`, and adds a default sample-event (which is needed to build the update readme template, since it now includes `{{event}}`).


mjwolf:
The new README template in Kibana, with no changes made to it:

mw-dev-e3609a kb canadacentral azure staging elastic cloud_app_integrations_detail_hello-0 0 9_overview

internal/docs/readme.go
```diff
@@ -204,6 +209,9 @@ func renderReadme(fileName, packageRoot, templatePath string, linksMap linkMap)
 			}
 			return linksMap.RenderLink(args[0], options)
 		},
+		"header": func(args ...string) (string, error) {
```
mjwolf, line 212: This depends on the user leaving ``{{header}}` in the readme in order to add the "Do not modify" string to the generated files.

I could also inject the string into all generated READMEs, regardless of if this {{header}} is in it, but I don't know if this would cause problems with other package types, so I didn't do that. If it won't cause problems, I could change to add to all generated files.

mjwolf:
> Could you modify one of the packages under `test/packages` to see the expected usage and result?

The `panw` integration has already been published with the README updated to use the new template: https://github.com/elastic/integrations/tree/main/packages/panw/docs

I can also modify one of the test packages in elastic-package, so there's an example of new template here.

mjwolf:
I've changed the readme in the `test/packages/parallel/apache` package to use the new format.  This isn't used in the real, published `apache` integration yet though. I've just added it here so there's an example of the new format in elastic-package.

internal/docs/readme.go
jsoriano: Yep, it could make sense to inject it in all generated files, but we would need to check that validation doesn't fail for packages that don't include this header.

test/packages/parallel/apache/_dev/build/docs/README.md
```diff
@@ -1,34 +1,129 @@
...
+This integration is compatible with all Apache server versions >= 2.4.16 and >= 2.2.31.
 
-### Access Logs
+### How it works
 
-Access logs collects the Apache access logs.
+Access and error logs data stream read from a logfile. You will configure Apache server to write log files to a location that is readable by elastic-agent, and 
+elastic-agent will monitor and ingest events written to these files.
 
-{{fields "access"}}
```
jsoriano, line 17: It looks like `fields` are not rendered anymore. In the rendered README I can only see `{{fields}}`, in https://github.com/elastic/elastic-package/pull/2707/files#diff-21c266b51e67f0cc833e6b48a09867796a652d1b2d9abe69119b3f36d7975d99R114.

internal/packages/archetype/package.go
```diff
@@ -54,6 +54,14 @@ func createPackageInDir(packageDescriptor PackageDescriptor, cwd string) error {
 		return fmt.Errorf("can't render package README: %w", err)
 	}
 
+	if packageDescriptor.Manifest.Type == "integration" {
```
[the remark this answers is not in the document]
kgeller: yeah I think so, dropping the if check

internal/packages/archetype/_static/package-docs-readme.md.tmpl
```diff
@@ -1,84 +1,105 @@
...
+<!-- This template can be used as a starting point for writing documentation for your new integration. For each section, fill in the details
+described in the comments.
 
-# {{.Manifest.Title}}
+Find more detailed documentation guidelines in https://www.elastic.co/docs/extend/integrations/documentation-guidelines
+-->
 
-<!-- The {{.Manifest.Title}} integration allows you to monitor {name of service}. {name of service} is {describe service}.
+<!-- Do not remove header -->
+{{ `{{header}}` }}
```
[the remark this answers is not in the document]
jsoriano: Adding test that would have detected this issue https://github.com/elastic/elastic-package/pull/2796.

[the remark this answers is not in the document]
kgeller: Thanks for adding that test! 

I intended to drop that function all together, but I somehow missed it being called in the template 🥲 

Pushing up some changes now, but I also tested it locally as well with the create and then the build, and it is now successful

internal/packages/archetype/_static/package-docs-readme.md.tmpl
```diff
@@ -1,84 +1,118 @@
...
+
+For each data_stream_name, include an optional summary of the datastream, the exported fields reference table and the sample event. -->
+
+<!-- The fields template function will be replaced by a generated list of all fields from the `fields/` directory of the data stream when building the integration. -->
+<!--
+#### {data stream name} fields
+
+To include a generated list of fields from the `fields/` directory, uncomment and use:
+(Remove the spaces between curly braces when using)
+{ { fields "data_stream_name" } }
```
jsoriano, line 95: We can directly generate the expected curly braces.
```suggestion
{{ "{{ fields \"data_stream_name\" }}" }}
```

internal/packages/archetype/_static/package-docs-readme.md.tmpl
```diff
@@ -1,84 +1,118 @@
...
-<!-- Optional -->
-<!-- #### Example
+-->
+
+<!-- The event template function will be replace by a sample event, taken from `sample_event.json`, when building this integration. -->
+<!--
+
+To include a sample event from `sample_event.json`, uncomment and use:
+(Remove the spaces between curly braces when using)
+{ { event "data_stream_name" } }
```
jsoriano, line 104:
```suggestion
{{ "{{ event \"data_stream_name\" }}" }}
```

internal/packages/archetype/_static/package-docs-readme.md.tmpl
jsoriano: > Are you saying we should use the Go template comments for both the scenarios above, and then try and modify the readme generation (`elastic-package build`) to strip out all comments? I tried this, but the go template comments get stripped out for both readme's when I run `create package` (unless there's something with the rendering I can also update).

Yes, but we could use different delimiters for each step. So let's say that we use `<<`/`>>` as delimiters for the `create` step, and `{{`/`}}` for the build step, then we could have comments with `{{/*`/`*/}}` that would stay in the template, but won't be visible in the final README.

I see how this complicates things and may require even a small refactor to allow the different delimiters, we may leave this for a different PR. I think it is worth it because this way comments can stay in the template and are easier to incorporate by package developers.

jsoriano (approved):
LGTM, the new content is much more useful and complete, thanks!

We can leave the refactor of comments/delimiters in templates for a different iteration.

PR #2716, opened 2025-07-08
109 docs validate readme template structure in elastic package

haetamoudi:
Changes in package-specs that need to occur before https://github.com/elastic/package-spec/pull/924
Tests will fail until then.

[Initial discussion on elastic-package](https://github.com/elastic/package-spec/issues/920)
[RFC](https://docs.google.com/document/d/1suPRUVDYAqFa1LL-2RIWI7cnj9zI8-LYG0i1eiBfIII/edit?tab=t.0#heading=h.pfm1mse74cxb)

### Goal
Enforce the structure of the docs (README.md) according to the template created by the docs team.

### Implementation
The template will be enforced only on new integrations. When not set in `validation.yml` the `docs_structure_enforced` is `false` by default.
The template will evolve with time, so we need to be able to handle versions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -184,5 +184,3 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
-
-replace github.com/elastic/package-spec/v3 => github.com/haetamoudi/package-spec/v3 v3.0.0-20250710152742-0e3bdbef3665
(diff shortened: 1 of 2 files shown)


cmd/build.go
```diff
@@ -72,6 +72,11 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 	}
 	logger.Debugf("Use build directory: %s", buildDir)
 
+	err = docs.ValidateDocsStructure(packageRoot)
+	if err != nil {
+		return fmt.Errorf("validating docs structure failed: %w", err)
+	}
```
jsoriano, line 78: Should we do this after updating readmes below?

internal/docs/readme.go
```diff
@@ -258,3 +264,78 @@ func readmePath(fileName, packageRoot string) string {
...
+	if err != nil {
+		return fmt.Errorf("structure validation: %w", err)
+	}
+
+	return nil
+}
+
+func validateReadmeStructure(packageRoot string) (error) {
+	requiredHeaders := []string{"Overview", "Troubleshooting", "Scaling"}
+	fileName := "README.md"
```
jsoriano, line 296: There are packages with a main README.md, and multiple .md files, one for each policy template. An example of this is the AWS package. Do we want to apply the same validation in these cases?

test/packages/other/readme_structure/_dev/build/build.yml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  ecs:
+    reference: git@1.10
+  docs_structure_enforced: true
```
jsoriano, line 4: Maybe we can make this an object, so it doesn't need to be an all or nothing? This could allow for example to introduce some additional setting to force only a set of headers.
```
  docs_structure_enforced:
    enabled: true
    skip:
      - title: Performance and scaling
        reason: System metrics, no performance tuning available.
```

jsoriano, line 4: This setting should not be under `dependencies`.

It could also belong to the `validation.yml` file at the root of the package.

PR #2744, opened 2025-07-17
[system tests] Allow setting an optional prefix for the service TestRunID

mrodm:
This PR allows to customize the Service TestRunID used in system tests to include a prefix string. This prefix is defined via environment variable:

```shell
export ELASTIC_PACKAGE_PREFIX_SERVICE_TEST_RUN_ID=1a2b
```

```shell
cd test/packages/parallel/terraform_local

export ELASTIC_PACKAGE_PREFIX_SERVICE_TEST_RUN_ID=1a2b

```

While the system tests are running you can check the files created by terraform at:
```shell
ls -l ~/.elastic-package/profiles/default/service_logs/agent-terraform_local-local-*/file*
```
it should countain a prefix of 9 digits (e.g. `file-1a2b21464.log`).

or via the Discover view in Kibana filtering by `data_stream.dataset : "terraform_local.local"`, and it should contain a `log` field similar to:

```
    "log": {
      "file": {
        "path": "/tmp/service_logs/file-1a2b21464.log"
      },
      "offset": 0
    },
```

## Author's Checklist
- [x] Test with a package creating resources in Terraform
- [x] Test with packages defined in the integrations repository

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -136,6 +136,7 @@ var (
 	enableIndependentAgentsEnv   = environment.WithElasticPackagePrefix("TEST_ENABLE_INDEPENDENT_AGENT")
 	dumpScenarioDocsEnv          = environment.WithElasticPackagePrefix("TEST_DUMP_SCENARIO_DOCS")
 	fieldValidationTestMethodEnv = environment.WithElasticPackagePrefix("FIELD_VALIDATION_TEST_METHOD")
+	prefixServiceTestRunIDEnv    = environment.WithElasticPackagePrefix("PREFIX_SERVICE_TEST_RUN_ID")
 )
 
 type fieldValidationMethod int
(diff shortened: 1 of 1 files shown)


internal/testrunner/runners/system/tester.go
```diff
@@ -483,7 +483,12 @@ func (r *tester) createServiceInfo() (servicedeployer.ServiceInfo, error) {
 	svcInfo.Name = r.testFolder.Package
 	svcInfo.Logs.Folder.Local = r.locationManager.ServiceLogDir()
 	svcInfo.Logs.Folder.Agent = ServiceLogsAgentDir
-	svcInfo.Test.RunID = common.CreateTestRunID()
+
+	prefix := ""
+	if v, found := os.LookupEnv("ELASTIC_PACKAGE_PREFIX_SERVICE_TEST_RUN_ID"); found && v != "" {
+		prefix = v
+	}
+	svcInfo.Test.RunID = common.CreateTestRunIDWithPrefix(prefix)
```
mrodm, line 491: In case of using the terraform service deployer, a part of updating the RunID used in as terraform variable, the alias used to render the policy templates in tests needs to be set to the same value as the terraform variable. This is needed for instance for `aws.ec2_metrics` that filters by the some resource name.

Issue #2751, opened 2025-07-21
Fix error message when file linked is not found


jsoriano:
In some cases, when a link file points to a file that doesn't exist, the error reported is about the link referencing a file out of the package root. The message should be about the file that doesn't exist.

See https://github.com/elastic/integrations/pull/14594#discussion_r2218498742.

PR #2752, opened 2025-07-21
Rely on os.Root for checking paths

jsoriano:
Fix https://github.com/elastic/elastic-package/issues/2751.

Remove `pathIsInRepositoryRoot`, and rely on implicit `os.Root` operations to check if a path is under the repository root.

### DIFF
--- a/internal/files/linkedfiles.go
+++ b/internal/files/linkedfiles.go
@@ -40,7 +40,12 @@ func CreateLinksFSFromPath(workDir string) (*LinksFS, error) {
 		return nil, fmt.Errorf("opening repository root: %w", err)
 	}
 
-	return NewLinksFS(root, workDir)
+	absWorkDir, err := filepath.Abs(workDir)
+	if err != nil {
+		return nil, fmt.Errorf("obtaining absolute path of working directory: %w", err)
+	}
+
+	return NewLinksFS(root, absWorkDir)
 }
 
 var _ fs.FS = (*LinksFS)(nil)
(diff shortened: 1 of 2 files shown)


internal/files/linkedfiles_test.go
```diff
@@ -845,13 +855,13 @@ func TestLinksFS_WorkDirValidation(t *testing.T) {
 			name:        "invalid absolute workDir outside repo",
 			workDir:     outsideDir,
 			expectError: true,
-			errorMsg:    "is outside the repository root",
+			errorMsg:    "path escapes from parent",
```
jsoriano, line 858: Error from the stdlib is directly returned now, but we cannot check for it to convert it as it is not public (https://github.com/golang/go/blob/e5502e0959bb54ec70ca500e8d2b6f5ac5efbc53/src/os/file.go#L421).

We could check the error message and replace it, but not sure if it is worth it.

internal/files/linkedfiles_test.go
```diff
@@ -785,13 +790,18 @@ func TestLinksFS_ErrorConditions(t *testing.T) {
 		{
 			name:     "broken link to non-existent file",
 			fileName: "broken.txt.link",
-			errorMsg: "escapes the repository root",
+			errorMsg: "no such file or directory",
```
jsoriano, line 793: This is the issue solved by this PR.

internal/files/linkedfiles.go
```diff
@@ -56,39 +61,28 @@ type LinksFS struct {
...
 	if filepath.IsAbs(workDir) {
 		absWorkDir = workDir
+		relative, err := filepath.Rel(repoRoot.Name(), absWorkDir)
+		if err != nil {
+			return nil, fmt.Errorf("invalid working directory %s: %w", absWorkDir, err)
+		}
+		workDir = relative
 	} else {
-		// First try: assume workDir is relative to the repository root
 		absWorkDir = filepath.Join(repoRoot.Name(), workDir)
```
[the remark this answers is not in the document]
jsoriano: Yep, better, thenks.

mrodm (approved):
LGTM

PR #2753, opened 2025-07-21
Start using 9.x versions of the stack by default

jsoriano:


jsoriano:
We could consider removing tests for 7.x, but I would wait a little bit, we can keep them till they start to be problematic.

mrodm (approved):
:rocket:

PR #2766, opened 2025-07-23
[CI] Refactor script to check and test packages in CI

mrodm:
This PR applies a refactor in `scripts/test-check-packages.sh` script to allow understand better the for loop that test all packages.

### Author's checklist
- [x] Validate that parallel and false positive targets run each package in a different Buildkite step
- [x] Validate that other targets test all the corresponding packages in the same Buildkite step.
- [x] Validate that `use_pipeline_tests` is also part of the `benchmarks` step.
- [x] Validate that Serverless pipeline works testing all the packages in the same Buildkite step. https://buildkite.com/elastic/elastic-package-test-serverless/builds/488

### DIFF
--- a/scripts/test-check-packages.sh
+++ b/scripts/test-check-packages.sh
@@ -6,14 +6,19 @@ source "${SCRIPT_DIR}/stack_parameters.sh"
 
 set -euxo pipefail
 
+# Add default values
+export SUFFIX_FOLDER_DUMP_LOGS="${PACKAGE_UNDER_TEST:-${PACKAGE_TEST_TYPE:-any}}"
+export PACKAGE_TEST_TYPE="${PACKAGE_TEST_TYPE:-"other"}"
+export PACKAGE_UNDER_TEST="${PACKAGE_UNDER_TEST:-*}"
+
 cleanup() {
   r=$?
 
   # Dump stack logs
   elastic-package stack dump -v \
-      --output "build/elastic-stack-dump/check-${PACKAGE_UNDER_TEST:-${PACKAGE_TEST_TYPE:-any}}"
+      --output "build/elastic-stack-dump/check-${SUFFIX_FOLDER_DUMP_LOGS}"
 
-  if [ "${PACKAGE_TEST_TYPE:-other}" == "with-kind" ]; then
+  if [ "${PACKAGE_TEST_TYPE}" == "with-kind" ]; then
     # Dump kubectl details
     kubectl describe pods --all-namespaces > build/kubectl-dump.txt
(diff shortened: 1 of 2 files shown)


scripts/test-check-packages.sh
```diff
@@ -98,39 +140,33 @@ for d in test/packages/${PACKAGE_TEST_TYPE:-other}/${PACKAGE_UNDER_TEST:-*}/; do
   package_to_test=$(basename "${d}")
 
   if [ "${PACKAGE_TEST_TYPE:-other}" == "benchmarks" ]; then
-    # It is not used PACKAGE_UNDER_TEST, so all benchmark packages are run in the same loop
-    if [ "${package_to_test}" == "pipeline_benchmark" ]; then
-      rm -rf "${PWD}/build/benchmark-results"
-      elastic-package benchmark pipeline -C "$d" -v --report-format xUnit --report-output file --fail-on-missing
+    # FIXME: There are other packages in test/packages/benchmarks folder that are not tested like rally_benchmark
```
mrodm, line 143: `rally_benchmark` package is located in `test/packages/benchmarks` folder but it is not tested at all. 
I don't know if this could be tested as it is here. IIUC according to the docs it is required also `esrally` tool.

mrodm:
test serverless

scripts/test-check-packages.sh
```diff
@@ -98,39 +140,35 @@ for d in test/packages/${PACKAGE_TEST_TYPE:-other}/${PACKAGE_UNDER_TEST:-*}/; do
...
-        continue
-    fi
-    # Run all tests
-    # defer-cleanup is set to a short period to verify that the option is available
-    elastic-package test -C "$d" -v --report-format xUnit --report-output file --defer-cleanup 1s --test-coverage --coverage-format=generic
+    # FIXME: There are other packages in test/packages/benchmarks folder that are not tested like rally_benchmark
+    case "${package_to_test}" in
+      pipeline_benchmark|use_pipeline_tests)
+        run_pipeline_benchmark "${package_to_test}" "$d"
+        ;;
```
mrodm, line 147: Until now just `pipeline_benchmark` package was tested, other packages under `test/packages/benchmarks/` folder were not tested. One of them is `use_pipeline_tests` package.

The difference that I see between `pipeline_benchmark` and `use_pipeline_tests` is that:
- `pipeline_benchmark` has the benchmark configuration for the data stream at `data_stream/test/_dev/benchmark/pipeline/*`. 
- `use_pipeline_tests` no benchmark configuration.

Tested locally with `use_pipeline_tests` package and it does not fail running the pipeline benchmarks and it creates a benchmark report successfully.

I think it can be added here, WDYT ?

mrodm:
test serverless

PR #2779, opened 2025-07-28
[Spacetime] Enhance errors

Supplementing:
### Summary 
Closes https://github.com/elastic/ingest-dev/issues/5837

This is an exploratory PR aimed to improve errors displayed to the user during common commands. 

The check added checks the status of Docker and ensures the daemon is running before trying to continue, which resolves issues where the CLI could 'hang' with no error output unless the user explicitly passed in the -d and or -v flags. Now we will perform a precheck and log to the user the issue. 

I also explored adding a precheck for packages to be built before running install, but realized that the install command checks that already and then performs the build automatically, so the check is not needed. 

### Testing Instructions: 

1. Create a new package
2. Build the package
3. Ensure docker isnt already running (kill the desktop application and any instances)
4. Try installing the package using `elastic-package install`
5. Note that the errors are surfaced to the user telling them that Docker isnt running and it fails immediately.


internal/stack/compose.go
```diff
@@ -77,6 +89,11 @@ func dockerComposeBuild(ctx context.Context, options Options) error {
 }
 
 func dockerComposePull(ctx context.Context, options Options) error {
+	// Check if Docker daemon is running
+	if err := CheckDockerRunning(); err != nil {
+		return err
+	}
```
jsoriano, line 95: Instead of having to check it on every operation, I think we could create a new `newComposeProvider() (*composeProvider, error)` constructor that could run the check, when initializing it in the factory, here:
https://github.com/elastic/elastic-package/blob/af5ac62d414f18c9bf442b83618f0499ca5bdc41/internal/stack/providers.go#L59

internal/docker/docker.go
```diff
@@ -190,3 +190,20 @@ func Copy(containerName, containerPath, localPath string) error {
...
+func CheckDaemonRunning() error {
+	cmd := exec.Command("docker", "info")
+	errOutput := new(bytes.Buffer)
+	cmd.Stderr = errOutput
+
+	logger.Debugf("running command to check Docker daemon: %s", cmd)
+	if err := cmd.Run(); err != nil {
+		daemonError := fmt.Errorf("docker daemon is not running (stderr=%q): %w", errOutput.String(), err)
+		logger.Error(": Docker daemon is not running or not accessible")
+		logger.Error(": Please make sure Docker is installed and running before executing commands that require Docker")
```
jsoriano, line 205: This is an internal method. I would prefer to log closer to the user.
```suggestion
```

internal/docker/docker.go
```diff
@@ -190,3 +190,20 @@ func Copy(containerName, containerPath, localPath string) error {
...
+		return daemonError
```
jsoriano, line 206: Nit. You can directly return the error.
```suggestion
		logger.Error(": Docker daemon is not running or not accessible")
		logger.Error(": Please make sure Docker is installed and running before executing commands that require Docker")
		return fmt.Errorf("docker daemon is not running (stderr=%q): %w", errOutput.String(), err)
```

internal/docker/docker.go
```diff
@@ -190,3 +190,20 @@ func Copy(containerName, containerPath, localPath string) error {
...
 }
+
+// CheckDaemonRunning checks if the Docker daemon is running.
+// Returns nil if the daemon is running, or an error if it's not.
+func CheckDaemonRunning() error {
+	cmd := exec.Command("docker", "info")
+	errOutput := new(bytes.Buffer)
+	cmd.Stderr = errOutput
+
+	logger.Debugf("running command to check Docker daemon: %s", cmd)
```
mrodm, line 201: I'd use Trace level for this message

```suggestion
	logger.Tracef("running command to check Docker daemon: %s", cmd)
```

internal/stack/compose.go
mrodm: I think the other providers (Environment and Serverless) could take advantage of this check too. IIRC these two providers run also docker (docker-compose) commands.

Just thinking that doing this check in the providers makes that elastic-package test or install commands do not run this check.
Maybe it's not needed, since all these commands will require to be the Elastic stack up... WDYT?

internal/docker/docker.go
```diff
@@ -190,3 +190,20 @@ func Copy(containerName, containerPath, localPath string) error {
 	}
 	return nil
 }
+
+// CheckDaemonRunning checks if the Docker daemon is running.
+// Returns nil if the daemon is running, or an error if it's not.
+func CheckDaemonRunning() error {
+	cmd := exec.Command("docker", "info")
```
mrodm, line 197: This commands has a large output. It should not be an issue, but maybe it could be format the output to just shown a specific value (e.g. `ServerVersion`). 

```suggestion
	cmd := exec.Command("docker", "info", "--format", ""{{ .ServerVersion }}")
```

Supplementing:
Got a little bogged down during the final two weeks of this iteration, and this took a backburner after ON week. Will revisit it sometime during next sprint to address requested changes! @mrodm @jsoriano

Supplementing:
@mrodm I was looking at my PR list and this one has just been sitting forever. Its super outdated at this point and honestly with the advancements in AI, we could probably make a more robust PR in no time with better changes. If you're okay with it, im going to close it but we can keep the issue open and maybe Ill revisit it during the next spacetime or something 🙂

mrodm:
> @mrodm I was looking at my PR list and this one has just been sitting forever. Its super outdated at this point and honestly with the advancements in AI, we could probably make a more robust PR in no time with better changes. If you're okay with it, im going to close it but we can keep the issue open and maybe Ill revisit it during the next spacetime or something 🙂

Sounds good to me, feel free to close it and keep the issue opened to be revisited later 👍

PR #2792, opened 2025-07-31
Update default stack version to 9.1.0

jsoriano:
Policy templates ignore namespaces when they are `default`.

This was originally done in https://github.com/elastic/elastic-package/pull/2784, but updatecli overwrote and close it.

### DIFF
--- a/internal/install/stack_version.go
+++ b/internal/install/stack_version.go
@@ -6,5 +6,5 @@ package install
 
 const (
 	// DefaultStackVersion is the default version of the stack
-	DefaultStackVersion = "9.0.4"
+	DefaultStackVersion = "9.1.0"
 )
--- a/internal/testrunner/runners/policy/policy.go
+++ b/internal/testrunner/runners/policy/policy.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/pmezard/go-difflib/difflib"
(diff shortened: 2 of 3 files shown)


jsoriano:
The failure in sql package is related to a fix included in this change: https://github.com/elastic/kibana/pull/214216, and may affect other input packages.

Testing with integrations to assess what to do. I think it would be good to validate that these fields are in the policy, as they should. So if it doesn't break many packages I would prefer to update the test files than adding more exceptions.

mrodm:
> The failure in sql package is related to a fix included in this change: [elastic/kibana#214216](https://github.com/elastic/kibana/pull/214216), and may affect other input packages.
> 
> Testing with integrations to assess what to do. I think it would be good to validate that these fields are in the policy, as they should. So if it doesn't break many packages I would prefer to update the test files than adding more exceptions.

Currently, in the integrations repository just one integration package defines policy tests:
```
 $ find packages -type d -name policy | grep "_dev/test/policy"
packages/auditd_manager/data_stream/auditd/_dev/test/policy
```

Would this issue raise just for input packages?

jsoriano:
> Would this issue raise just for input packages and with 9.1.0 onwards?

Yes.

I am going to remove these fields at the end. As it is now, this is going to cause issues for packages supporting versions < 9.1 and >= 9.1.

mrodm (approved):

jsoriano:
Trying yet another time in https://github.com/elastic/elastic-package/pull/2804.

PR #2796, opened 2025-08-01
Test building archetypes

jsoriano:
After creating packages from archetypes, they may contain errors that are only discovered on update or build time. Perform a full build of the created packages to detect these issues.

This would have detected a missing function in a template in https://github.com/elastic/elastic-package/pull/2707.

This change also allows to build packages in a target build directory.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -72,7 +72,7 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 	}
 	logger.Debugf("Use build directory: %s", buildDir)
 
-	targets, err := docs.UpdateReadmes(packageRoot)
+	targets, err := docs.UpdateReadmes(packageRoot, "")
 	if err != nil {
 		return fmt.Errorf("updating files failed: %w", err)
 	}
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -72,7 +72,7 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 	}
 	logger.Debugf("Use build directory: %s", buildDir)
 
-	targets, err := docs.UpdateReadmes(packageRoot, "")
+	targets, err := docs.UpdateReadmes(packageRoot, buildDir)
 	if err != nil {
 		return fmt.Errorf("updating files failed: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


cmd/build.go
```diff
@@ -72,7 +72,7 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 	}
 	logger.Debugf("Use build directory: %s", buildDir)
 
-	targets, err := docs.UpdateReadmes(packageRoot)
+	targets, err := docs.UpdateReadmes(packageRoot, "")
```
mrodm, line 75: Could `buildDir` be used here as that folder has been created above? 

```suggestion
	targets, err := docs.UpdateReadmes(packageRoot, buildDir)
```

internal/packages/archetype/data_stream_test.go
```diff
@@ -80,5 +68,5 @@ func createAndCheckDataStream(t *testing.T, pd PackageDescriptor, dd DataStreamD
 	err = CreateDataStream(dd)
 	require.NoError(t, err)
 
-	checkPackage(t, pd.Manifest.Name, valid)
+	checkPackage(t, packageRoot, valid)
```
mrodm, line 71: Previously, it was working because of the `os.Chdir`, am I right?
In any case, better to use here `packageRoot` :+1:

mrodm (approved):

internal/builder/packages.go
```diff
@@ -73,10 +74,13 @@ func findBuildDirectory() (string, bool, error) {
 }
 
 // BuildPackagesDirectory function locates the target build directory for the package.
-func BuildPackagesDirectory(packageRoot string) (string, error) {
-	buildDir, err := buildPackagesRootDirectory()
-	if err != nil {
-		return "", fmt.Errorf("can't locate build packages root directory: %w", err)
+func BuildPackagesDirectory(packageRoot string, buildDir string) (string, error) {
+	if buildDir == "" {
```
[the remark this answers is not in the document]
jsoriano: Yes, I am leaving the responsibility to the caller, but maybe better safe than sorry. Adding checks in https://github.com/elastic/elastic-package/pull/2796/commits/5f094a0ae95f2bc68d98268e21254ec03b11e597.

cmd/build.go
jsoriano: Umm, seeing that actually maybe we don't need all the magic below to create the build dir, as is being created explicitly here. Maybe I can remove some more code.

mrodm (approved):

PR #2798, opened 2025-08-01
Refactor survey validators to don't need chdirs

jsoriano:
After https://github.com/elastic/elastic-package/pull/2796, these would be the last use of chdir in tests.

Chdir can be problematic because it is global, and might affect other tests, or other functions tested in the same test. It is also error-prone as one may forget to go back to the previous working directory.

An object is created now that can keep as context the working directory, and validators are added as members of this object.


mrodm (approved):
Nice! :rocket:

PR #2826, opened 2025-08-14
Generate data stream input configuration and documentation

mjwolf:
This change introduces a new interactive prompt to the create data-stream command. When a user creates a data stream of type logs, they can now select from a list of common input types (e.g., aws-s3, gcp-pubsub, udp, tcp).

Based on the user's selection, the command now automatically populates the data_stream/<name>/manifest.yml with the corresponding input streams and their required variables. This streamlines the creation process for new log-based data streams by providing sensible defaults and reducing manual configuration.

This also updates the package README generation step to support a new function "inputDocs".  When this is used in the readme template, all inputs used in the package will be found, and all will be listed in the readme. For inputs which have additional documentation added to elastic-package, the input documentation will also be inserted into the rendered readme.

*Key changes include:*

Added a multi-select prompt to the create data-stream command for choosing log input types.

Added `internal/packages/_static/inputs/*.yml` files, which contains information on each input type supported by the create command, with information on default name, title, description and variables.

Introduced internal/packages/archetype/data_stream_inputs.go to manage the logic for populating input variables from the new static inputs yml files.

Updated the dataStream-manifest.yml.tmpl to dynamically render streams and their associated variables.

Refactored the packages.DataStreamManifest to use a new packages.Stream struct for better organization.

Add "inputDocs" to the readme render functions, which will find all inputs used in the package, and render documentation for them.

Add `docs/_static/inputs/*.yml` for keeping input documentation (this is separate from archetype yml file, to avoid a circular dependancy between the packages). 

*How to test:*

**Create command**
Run the elastic-package create data-stream command.

Select logs as the data stream type.

Select one or more input types from the list.

Verify that the generated data_stream/<name>/manifest.yml contains the correct streams configuration for the inputs you selected.

**Build readme**
Update a package readme to have the {{inputDocs}} function in the readme template

Build the package

Check the generated readme for the input documentation.


cmd/create_data_stream.go
```diff
@@ -81,7 +82,7 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
 			Name: "subobjects",
 			Prompt: &survey.Confirm{
 				Message: "Enable creation of subobjects for fields with dots in their names?",
-				Default: false,
+				Default: true,
```
mjwolf, line 85: @jsoriano I changed this for two reasons.
1. True is the default value in the [package-spec](https://github.com/elastic/package-spec/blob/561d1bc653724d549c527fba82f64f6b25d8cff1/spec/integration/data_stream/manifest.spec.yml#L479)
2. When set to false, the created manifest can't be installed, and I wanted the default options to create a working package.  I get an error like this when trying to install a package with a datastream with this option set to false.
```
Error: package installation failed: failed to build package: invalid content found in built zip package: found 1 validation error:
   1. file "/home/mwolf/git/integrations/build/packages/endace-0.3.0.zip/data_stream/stream2/manifest.yml" is invalid: field elasticsearch.index_template.mappings: Additional property subobjects is not allowed```

This isn't something introduced here, it also happens with elastic-package 0.114.0. I don't know enough about what this option should do to fix it properly

**Review by kgeller — COMMENTED** [review]
Some general feedback/questions: 

1) It feels a bit awkward without the changes to the agent/stream template since the manifest will align with what's there. I know we have that slotted for a subtask, but I feel it would make more sense to include all together. Unless we can guarantee they'll be merged together around the same time and will definitely make it into the same release

2) I appreciate that the user doesn't have to select an input, but then we get a pretty empty generated manifest file (and later no agent/stream files). I feel it could be beneficial if no input is selected, we fall back to what gets generated today with the currently existing logfile details 

3) Not sure how possible this is, but I think it could also be great if the manifest sections could be broken out into their own files. The one big massive file feels less ideal for maintaining

4) Is there anything we can do sync up the package manifest with the new datastream details?

File: cmd/create_data_stream.go [hunk 2]
```diff
@@ -126,6 +127,41 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
...
 	}
 
+	if answers.Type == "logs" {
+		qs := []*survey.Question{
+			{
+				Name: "inputs",
+				Prompt: &survey.MultiSelect{
+					Message: "Select input types which will be used in this data stream",
+					Options: []string{
+						"aws-cloudwatch",
```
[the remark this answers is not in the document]
mjwolf: I thought about that too. It would be nice, but I don't think it's possible with the survey module we're using.

It might be nice to switch to another TUI module which can do more in the future, but that'll be a big job.  For now I can add a link to the input documentation in the first prompt.

I think that's the best that can be done with this module.

mjwolf:
> Some general feedback/questions:
> 
> 1. It feels a bit awkward without the changes to the agent/stream template since the manifest will align with what's there. I know we have that slotted for a subtask, but I feel it would make more sense to include all together. Unless we can guarantee they'll be merged together around the same time and will definitely make it into the same release

I agree with that now. I was thinking that it'll be a big PR to have both changes in the same PR. But now that I'm working on the 2nd, a lot of the changes are closely related, and I think it makes sense to do them both at the same time.

So I converted this PR to draft, and I'll keep working on it to do them both at the same time.

cmd/create_data_stream.go
```diff
@@ -126,6 +127,41 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
...
+						"azure-blob-storage",
+						"azure-eventhub",
+						"cel",
+						"entity-analytics",
+						"etw",
+						"filestream",
+						"gcp-pubsub",
+						"gcs",
+						"http_endpoint",
+						"httpjson",
```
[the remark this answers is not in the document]
kgeller: I still see it when testing locally and in the PR?

mjwolf:
> > This is what the `panw` integration rendered README looks like in Kibana when using this markdown in the "References" section:
> 
> @mjwolf is it possible to take advantage of the collapsable sections here with md now that we have support for that in kibana? I worry these big sections at the bottom will create unnecessary noise for the users when browsing the page


I've changed to use collapsible sections for each input's doc (and updated the screenshot above)

cmd/create_data_stream.go
```diff
@@ -126,6 +127,42 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
...
+						"journald",
+						"netflow",
+						"redis",
+						"tcp",
+						"udp",
+						"winlog",
```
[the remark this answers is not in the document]
mjwolf: I don't know of a source of truth that could be used for this. The inputs here were selected because they are the most-used. I think the idea of composable inputs could be good to become the source of truth.  It would be good if that had some concept of deprecated/discouraged inputs, since there's now some inputs that we don't want to add new usages.

cmd/create_data_stream.go
```diff
@@ -126,6 +127,80 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
...
+							return "Netflow"
+						case "redis":
+							return "Redis"
+						case "tcp":
+							return "TCP"
+						case "udp":
+							return "UDP"
+						case "winlog":
+							return "WinLogBeat"
+						}
```
jsoriano, line 192: To ensure that we keep this list updated with the options of the multiselect, we could keep this info in a map, and use the ordered keys as options for the multiselect.
```suggestion
						return inputOptions[value]
```

internal/docs/resources.go
```diff
@@ -0,0 +1,84 @@
...
+	inputGcs,
+	inputHttpEndpoint,
+	inputHttpJson,
+	inputJournald,
+	inputNetflow,
+	inputRedis,
+	inputTcp,
+	inputUdp,
+	inputWinlog,
+}
```
jsoriano, line 84: This list is duplicated in `internal/packages/archetype/resources.go`? Could we reuse it?

internal/docs/resources.go
```diff
@@ -0,0 +1,84 @@
...
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package docs
+
+import _ "embed"
+
+// Input definitions
+
+//go:embed _static/inputs/aws-cloudwatch.yml
```
jsoriano, line 11: Is it also possible to embed a whole directory. In case this helps, maybe when adding new inputs in the future.
```
//go:embed _static/inputs
var inputResources embed.FS
```

cmd/create_data_stream.go
mjwolf: Thanks, it's a good idea to use a map. I've done that, so there's now just one map for the names and descriptions, instead of the duplicated lists.

mjwolf:
> 4. Is there anything we can do sync up the package manifest with the new datastream details?

@kgeller, I had updated this to add a new policy_template to the package manifest when a new data stream is created, but it then caused some unit tests to fail. I think there's some interaction with the package-spec that makes it a bit more complicated then I thought. What do you think of leaving this out of this PR, since it's already pretty big.

We could look at updating the package spec later.

(This is what I did to update the package manifest when a data stream is added: https://github.com/elastic/elastic-package/pull/2826/commits/05acf8772b849e2543ff6ebf936604fc41aa7ea4)

internal/docs/_static/inputs/cel.yml
```diff
@@ -0,0 +1,397 @@
...
+    description: Client ID used for OAuth2 authentication
+    show_user: true
+    required: false
+  - name: oauth_secret
+    type: password
+    title: OAuth2 Client Secret
+    description: Client secret used for OAuth2 authentication
+    show_user: true
+    required: false
+    secret: true
```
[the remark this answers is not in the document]
mjwolf: The Secret field wasn't in the updated manifest template, so it wasn't added into the resulting manifest, even though it was in the inputs yaml definition.

I've added secrets, and other missing fields to the manifest template now.

jsoriano (approved):

Issue #2704, opened 2025-07-03
[tests] Expand coverage results


kcreddy:
Following paths are not covered in the coverage reports for each `<package>`:

```
./data_stream/*/lifecycle.yml
./data_stream/*/elasticsearch/ilm/*
./elasticsearch/transform/*
./kibana/tags.yml
./kibana/tag/*
./validation.yml
```

This leads to low coverage of packages inside PRs and scheduled CI runs.
Example: https://github.com/elastic/integrations/pull/14290#issuecomment-2998896369 which has 18% coverage on new code.

Increase coverage on these paths. Alternately add these paths into [exclusion list](https://github.com/elastic/integrations/blob/main/sonar-project.properties#L10) as temporary measure.

PR #2831, opened 2025-08-18
Update coverage information with tags and transforms

jsoriano:
Related to https://github.com/elastic/elastic-package/issues/2704.

Add some information for coverage reports:
* `validation.yml` is ignored.
* Kibana tags are added as covered on asset reports.
* Transform files are added as covered on system tests.

### DIFF
--- a/test/packages/other/tags_without_assets/data_stream/foo/agent/stream/filestream.yml.hbs
+++ b/test/packages/other/tags_without_assets/data_stream/foo/agent/stream/filestream.yml.hbs
@@ -1,0 +1,44 @@
+paths:
+{{#each paths as |path|}}
+  - {{path}}
+{{/each}}
+{{#if exclude_files}}
+prospector.scanner.exclude_files:
+{{#each exclude_files as |pattern f|}}
+  - {{pattern}}
+{{/each}}
+{{/if}}
+{{#if multiline_json}}
+multiline.pattern: '^{'
+multiline.negate: true
+multiline.match: after
+multiline.max_lines: 5000
+multiline.timeout: 10
+{{/if}}
+{{#if custom}}
+{{custom}}
+{{/if}}
+
+{{#if tags.length}}
+tags:
+{{#each tags as |tag|}}
+- {{tag}}
+{{/each}}
+{{#if preserve_original_event}}
+- preserve_original_event
+{{/if}}
+{{else}}
+{{#if preserve_original_event}}
+tags:
+- preserve_original_event
+{{/if}}
+{{/if}}
+
+{{#contains "forwarded" tags}}
+publisher_pipeline.disable_host: true
+{{/contains}}
+
+{{#if processors}}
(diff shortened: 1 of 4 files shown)


jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
> So I think that we would need to do at least:
> 
>     * Introduce the wait mechanism.
> 
>     * Change detection to look for tags by name, or expect ids like the ones generated by [getPackageSpecTagId](https://github.com/elastic/kibana/blob/f7add87c5a13d37ebcd60d100bfcee06b6960459/x-pack/platform/plugins/shared/fleet/server/services/epm/kibana/assets/tag_assets.ts#L68).

Doing this in https://github.com/elastic/elastic-package/pull/2831/commits/dadf4413474439f968ff58b1b6ec091ff433bf47, I will leave further improvements for future changes.

mrodm:
> I am starting to think that these tags are installed asynchronously, and then we may have some kind of race condition.
> For the rest of assets we are not actually checking if they are installed, we check that they are listed in the package response from the API, we might have the same problem if we would check for them.

That scenario looks likely to happen :+1: . 


> I think the problem is that they are not there _yet_, so we would need to introduce some kind of wait until mechanism in any case.
> 
> So I think that we would need to do at least:
> 
> * Introduce the wait mechanism.
> * Change detection to look for tags by name, or expect ids like the ones generated by [getPackageSpecTagId](https://github.com/elastic/kibana/blob/f7add87c5a13d37ebcd60d100bfcee06b6960459/x-pack/platform/plugins/shared/fleet/server/services/epm/kibana/assets/tag_assets.ts#L68).

It sounds a good plan to follow.

But I still think that tags defined in `kibana/tags.yml` should not be checked for their existence if there are no Kibana assets defined in the package (e.g. `azure_blob_storage` or `cel` packages). In those packages, those tags are not going to be created.

internal/testrunner/runners/asset/tester.go
```diff
@@ -191,20 +211,67 @@ func (r *tester) TearDown(ctx context.Context) error {
...
 			return true
 		}
 	}
 
+	if expectedAsset.Type == "tag" && expectedAsset.ID == "" {
+		// If we haven't found the asset, and it is a tag, it could be some of the shared
+		// tags defined in tags.yml, whose id can be unpredictable, so check by name.
+		if len(actualAssets) == 0 {
+			// If there are no assets, the tag may not be installed, so assume it would have been.
+			// TODO: More accurately we should check if any of the listed tags in `tags.yml` is present.
```
jsoriano, line 226: This would be actually wrong in the package, and can require quite some changes to be supported as we would need to bring here the whole `tags.yml`. So if we find a package with this I would prefer to fix this in the package.

jsoriano:
test integrations

jsoriano:
test integrations

Issue #2835, opened 2025-08-19
Ensure input packages with `otelcol` input type (e.g. "OTel integrations") can be built, tested, and published with `elastic-package`


kpollich:
The new OTel integrations input package type introduced in https://github.com/elastic/package-spec/issues/906 should be fully supported in `elastic-package`. Integration authors should be able to build, test, and publish new versions of their OTel integrations.

A detail to take into account is that data streams have an additional `otel` suffix, see https://github.com/elastic/kibana/issues/232628.

mrodm:
After doing some tests with the current status of these PRs:
- https://github.com/elastic/kibana/pull/233090
- https://github.com/elastic/fleet-server/pull/5469

It is required to update how policy tests are validated in case of the `otelcol` input. Working on that https://github.com/elastic/elastic-package/pull/2886

Draft PR to add support to run system tests on input packages defining an `otelcol` input:  https://github.com/elastic/elastic-package/pull/2885

kpollich:
Yes let's go ahead and close this. Thank you @mrodm ❤️

PR #2873, opened 2025-09-04
Support policy tests in input packages with input type otelcol

mrodm:
Supersedes #2799
Relates #2835 

Add support to run policy tests in input packages defining a policy template with the input type `otelcol`. Continuing the work done in #2799 

In order to support policy tests in those packages, it is needed to add filters to remove the ids that may appear in policies with OTel configuration. Some examples:
```yaml
receivers:
    httpcheck/b0f518d6-4e2d-4c5d-bda7-f9808df537b7:
        collection_interval: 1m
        targets:
            - endpoints:
                - https://epr.elastic.co
              method: GET
secret_references: []
service:
    pipelines:
        logs:
            receivers:
                - httpcheck/b0f518d6-4e2d-4c5d-bda7-f9808df537b7
```

This PR adds a new test package `httpcheck` that includes this new input type (copied from https://github.com/elastic/integrations/pull/14315).

```shell
elastic-package stack up -v -d --version 9.2.0-SNAPSHOT

elastic-package -C test/packages/parallel/httpcheck test policy -v

elastic-package stack down -v
```


internal/stack/_static/kibana.yml.tmpl
```diff
@@ -161,7 +170,3 @@ xpack.fleet.outputs:
 {{- if eq $version "9.0.0-SNAPSHOT" }}
 xpack.fleet.internal.registry.kibanaVersionCheckEnabled: false
 {{- end }}
-
-logging.loggers:
-  - name: plugins.fleet
-    level: debug
```
mrodm, line 167: Reverting the change introduced in https://github.com/elastic/elastic-package/pull/2501/ related to the log level of fleet plugin.

internal/stack/_static/kibana.yml.tmpl
```diff
@@ -4,7 +4,7 @@ server.host: "0.0.0.0"
 {{- $kibana_http2_enabled := fact "kibana_http2_enabled" -}}
 {{- if (and (eq $kibana_http2_enabled "true") (not (semverLessThan $version "8.15.0-SNAPSHOT"))) }}
 server.protocol: http2
-{{ end -}}
+{{- end }}
```
mrodm, line 7: This change fixes an error when rendering the kibana.yml file for versions older than 8.15.0-SNAPSHOT

Tested with stack version 8.9.0

```
 $ yq -r ~/.elastic-package/profiles/default/stack/kibana.yml
Error: bad file '/home/mariorodriguez/.elastic-package/profiles/default/stack/kibana.yml': yaml: line 2: mapping values are not allowed in this context
```

This was the rendered file:
```yaml

server.name: kibana
server.host: "0.0.0.0"server.ssl.enabled: true
server.ssl.certificate: "/usr/share/kibana/config/certs/cert.pem"
server.ssl.key: "/usr/share/kibana/config/certs/key.pem"

```
Even that, kibana was able to start successfully.

test/packages/parallel/httpcheck.stack_version
```diff
@@ -0,0 +1 @@
+9.2.0-SNAPSHOT
```
mrodm, line 1: Forced to run this package with 9.2.0-SNAPSHOT.

This should be removed once the default version is 9.2.0 or higher.

test/packages/parallel/httpcheck/manifest.yml
```diff
@@ -0,0 +1,76 @@
...
+  license: "Elastic-2.0"
+description: "Perform HTTP checks using the HTTP Check receiver of the OTel Collector."
+type: input
+categories:
+  - monitoring
+  - network
+  - web
+conditions:
+  kibana:
+    version: "^9.2.0"
```
mrodm, line 15: Set as constraint 9.2.0, where the input type `otelcol` was added.

internal/testrunner/runners/policy/policy_test.go
```diff
@@ -285,6 +285,72 @@ secret_references:
...
+                  privileges:
+                    - auto_configure
+                    - create_doc
+receivers:
+    httpcheck/b0f518d6-4e2d-4c5d-bda7-f9808df537b7:
+        collection_interval: 1m
+        targets:
+            - endpoints:
+                - https://epr.elastic.co
+              method: GET
```
mrodm, line 343: @jsoriano in these policies, could it be more than one receiver/exporter/extensions/processors?
Focusing on receiver as example, if there is another receiver defined, would it start also with `httpcheck` ? 

``` 
receivers:
    httpcheck/b0f518d6-4e2d-4c5d-bda7-f9808df537b7:
        collection_interval: 1m
        targets:
            - endpoints:
                - https://epr.elastic.co
              method: GET
    httpcheck/b0f518d6-4e2d-4c5d-bda7-11111111111:
        collection_interval: 2m
        targets:
            - endpoints:
                - https://example.com
              method: GET
```

If that could happen, the resulting yaml would contain just one receiver `httpcheck/componentid`

test/packages/parallel/httpcheck/manifest.yml
```diff
@@ -0,0 +1,76 @@
+format_version: 3.4.0
```
jsoriano, line 1: This will need to be `3.5.0` as of https://github.com/elastic/package-spec/pull/975.

internal/stack/_static/kibana.yml.tmpl
mrodm: Moved to https://github.com/elastic/elastic-package/pull/2875

internal/testrunner/runners/policy/policy_test.go
```diff
@@ -285,6 +285,178 @@ secret_references:
...
+                  privileges:
+                    - auto_configure
+                    - create_doc
+processors:
+    batch/11c35ad0-4351-49d4-9c78-fa679ce9d950:
+        send_batch_size: 10
+        timeout: 1s
+    batch/e6e379c5-6446-4090-af10-a9e5f8fc4640:
+        send_batch_size: 10000
+        timeout: 10s
```
mrodm, line 312: With the current implementation, this could fail in some test executions:
```
--- FAIL: TestComparePolicies (0.00s)
    --- FAIL: TestComparePolicies/otel_ids (0.00s)
        policy_test.go:472: 
            	Error Trace:	/home/mariorodriguez/Coding/work/elastic-package-otel/internal/testrunner/runners/policy/policy_test.go:472
            	Error:      	Should be empty, but was --- want
            	            	+++ got
            	            	@@ -17,4 +17,4 @@
            	            	     batch/componentid:
            	            	-        send_batch_size: 10
            	            	-        timeout: 1s
            	            	+        send_batch_size: 10000
            	            	+        timeout: 10s
            	            	 receivers:
            	Test:       	TestComparePolicies/otel_ids
FAIL
FAIL	github.com/elastic/elastic-package/internal/testrunner/runners/policy	0.061s
FAIL
```

Working to replace the OTEL ids (directly on the `[]byte` variable) before running the cleanup process.

internal/testrunner/runners/policy/policy.go
```diff
@@ -184,6 +214,46 @@ func cleanPolicy(policy []byte, entriesToClean []policyEntryFilter) ([]byte, err
...
+
+					// store the otel ID replaced without the space indentation and the colon to be replaced later
+					// (e.g. http_check/4391d954-1ffe-4014-a256-5eda78a71828 replaced by http_check/componentid-0)
+					otelID := strings.SplitN(strings.TrimSpace(stringLine), ":", 2)[0]
+					replacementsDone[otelID] = strings.SplitN(strings.TrimSpace(string(lines[i])), ":", 2)[0]
+				}
+			}
+		}
+		return bytes.Join(lines, []byte("\n"))
+	})
```
jsoriano, line 242: Is it neccesary to take into account the root level? Couldn't we directly replace the matching ids?
```suggestion
	policy = uniqueOtelComponentIDReplace.regexp.ReplaceAllStringFunc(policy, func(match string) string {
		replacement := fmt.Sprintf(uniqueOtelComponentIDReplace.replace, strconv.Itoa(count))
		replaced := uniqueOtelComponentIDReplace.regexp.ReplaceAllString(match, replacement)
		otelID, _, _ := strings.Cut(strings.TrimSpace(match), ":")
		replacementsDone[otelID], _, _ = strings.Cut(strings.TrimSpace(string(replaced)), ":")
	})
```

internal/testrunner/runners/policy/policy.go
```diff
@@ -184,6 +214,46 @@ func cleanPolicy(policy []byte, entriesToClean []policyEntryFilter) ([]byte, err
...
+			if strings.Contains(stringLine, ":") {
+				if uniqueOtelComponentIDReplace.regexp.MatchString(stringLine) {
+					replacement := fmt.Sprintf(uniqueOtelComponentIDReplace.replace, strconv.Itoa(count))
+					count++
+					lines[i] = []byte(uniqueOtelComponentIDReplace.regexp.ReplaceAllString(stringLine, replacement))
+
+					// store the otel ID replaced without the space indentation and the colon to be replaced later
+					// (e.g. http_check/4391d954-1ffe-4014-a256-5eda78a71828 replaced by http_check/componentid-0)
+					otelID := strings.SplitN(strings.TrimSpace(stringLine), ":", 2)[0]
+					replacementsDone[otelID] = strings.SplitN(strings.TrimSpace(string(lines[i])), ":", 2)[0]
```
jsoriano, line 237: Nit. I think these `strings.SplitN` could be replaced by `strings.Cut`.
```suggestion
					otelID, _, _ := strings.Cut(strings.TrimSpace(stringLine), ":")
					replacementsDone[otelID], _, _ = strings.Cut(strings.TrimSpace(string(lines[i])), ":")
```

internal/testrunner/runners/policy/policy.go
```diff
@@ -184,6 +214,46 @@ func cleanPolicy(policy []byte, entriesToClean []policyEntryFilter) ([]byte, err
...
+// It also replaces references to those IDs in service.extensions and service.pipelines.
+func replaceOtelComponentIDs(policy []byte) []byte {
+	replacementsDone := map[string]string{}
+
+	policy = otelComponentIDsRegexp.ReplaceAllFunc(policy, func(match []byte) []byte {
+		count := 0
+		lines := bytes.Split(match, []byte("\n"))
+		for i, line := range lines {
+			line = bytes.TrimRight(line, "\r\n")
+			stringLine := string(line)
```
jsoriano, line 227: Nit. Instead of keeping an array with the lines that we need to join later, we could use an `Scanner` to read the lines, and a `strings.Builder` to build the result. In principle the `Scanner` would take care of the newline characters and so on.

jsoriano (commented):
Looks good overall, added some suggestions. In any case we will need to make changes here as the E2E implementation consolidates, so we will have opportunities to revisit this.

internal/testrunner/runners/policy/policy.go
```diff
@@ -184,6 +214,46 @@ func cleanPolicy(policy []byte, entriesToClean []policyEntryFilter) ([]byte, err
...
+			if strings.Contains(stringLine, ":") {
+				if uniqueOtelComponentIDReplace.regexp.MatchString(stringLine) {
```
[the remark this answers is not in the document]
mrodm: That's right. I'll remove the first if with `strings.Contains` function.
Thanks!

internal/testrunner/runners/policy/policy.go
mrodm: That approach looks better, thanks!
Updated code in https://github.com/elastic/elastic-package/pull/2873/commits/0c5483c9ba7c95b7d7bc5b0f147ba8dfa65ad326

mrodm:
test integrations

test/packages/parallel/httpcheck/_dev/test/system/test-local-config.yml
```diff
@@ -0,0 +1,7 @@
+skip:
+  reason: Not supported system tests with input type otelcol.
+  link: https://github.com/elastic/elastic-package/issues/2835
+service: web
+vars:
```
jsoriano, line 5: Nit. I found out that setting a period here makes tests faster :sweat_smile: It defaults to 60s.
```suggestion
vars:
  period: 1s
```

jsoriano (approved):
:+1:

PR #2911, opened 2025-09-15
Conditional subobjects based on package version

### LINKED ISSUE #2832: Create data-stream command doesn't create a valid manifest with mappings.subobject = false
When creating a new data stream, with the mapping.subobject set to false, the generated manifest will cause an error building the package. There's no error when mappings.subobjects = true. Happens with (at least) elastic-package v0.114.0

## Steps to reproduce:
1. Create new datastream by running `elastic-package datastream create` in  a package dir.
2. In the wizard, create "logs" type data-stream and answer "N" to "Enable creation of subobjects for fields with dots in their names?" Give any answer to the other questions.
3. Build the package with `elastic-package build`. There will be this build error:
`.../data_stream/new_data_stream/manifest.yml" is invalid: field elasticsearch.index_template.mappings: Additional property subobjects is not allowed`

This is the manifest file that was created:
```
title: "New Data Stream"
type: logs
streams:
  - input: logfile
    title: Sample logs
    description: Collect sample logs
    vars:
      - name: paths
        type: text
        title: Paths
        multi: true
        default:
          - /var/log/*.log
elasticsearch:
  index_template:
    mappings:
      subobjects: false
```

teresaromero:
Fix #2832 

The way to reproduce this bug is then the package has a spec version lower than 3.2.
Subobjects support was included after 3.2 so having this spec under the data-stream manifest in a package with lower version does not recognize the property of the data-stream manifest

Solution proposed is to skip the question for subobjects when a package spec version is under 3.2. The default value (false) should not be written at the manifest of data-stream

### DIFF
--- a/cmd/create_data_stream_test.go
+++ b/cmd/create_data_stream_test.go
@@ -90,8 +90,5 @@ func TestCreateDataStreamDescriptorFromAnswers_SubobjectsTrueForSpecVersionGTE3_
 	}
 	descriptor := createDataStreamDescriptorFromAnswers(answers, "/tmp/package", specVersion)
 
-	require.NotNil(t, descriptor.Manifest.Elasticsearch)
-	require.NotNil(t, descriptor.Manifest.Elasticsearch.IndexTemplate)
-	require.NotNil(t, descriptor.Manifest.Elasticsearch.IndexTemplate.Mappings)
-	assert.False(t, descriptor.Manifest.Elasticsearch.IndexTemplate.Mappings.Subobjects)
+	require.Nil(t, descriptor.Manifest.Elasticsearch)
 }


cmd/create_data_stream.go
```diff
@@ -80,14 +83,21 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
 			},
 			Validate: survey.Required,
 		},
-		{
+	}
+
+	sv, err := semver.NewVersion(manifest.SpecVersion)
+	if err != nil {
+		return fmt.Errorf("failed to obtain spec version from package manifest in \"%s\"", packageRoot)
```
jsoriano, line 90: Wrap the error to have more context if this fails.
```suggestion
		return fmt.Errorf("failed to obtain spec version from package manifest in \"%s\": %w", packageRoot, err)
```

cmd/create_data_stream.go
```diff
@@ -183,14 +193,14 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func createDataStreamDescriptorFromAnswers(answers newDataStreamAnswers, packageRoot string) archetype.DataStreamDescriptor {
+func createDataStreamDescriptorFromAnswers(answers newDataStreamAnswers, packageRoot string, specVersion *semver.Version) archetype.DataStreamDescriptor {
```
jsoriano, line 196: Perhaps we could unit test this function, checking the manifest it generates for given answers.

Further testing could be added to `internal/packages/archetype/data_stream_test.go`, but we wouldn't be really testing if the behaviour changes for the format version used in the package.

teresaromero: I've added some unit testing as suggested, also for the survey questions i've encapsulated this logic to at least check that the question is added or not based on the version.

mrodm (commented):
Looks great!
Just added some comments in tests to avoid possible errors.

cmd/create_data_stream_test.go
```diff
@@ -0,0 +1,97 @@
...
+		Title:      "Test Stream",
+		Type:       "logs",
+		Subobjects: true,
+	}
+	descriptor := createDataStreamDescriptorFromAnswers(answers, "/tmp/package", specVersion)
+
+	require.NotNil(t, descriptor.Manifest.Elasticsearch)
+	require.NotNil(t, descriptor.Manifest.Elasticsearch.IndexTemplate)
+	require.NotNil(t, descriptor.Manifest.Elasticsearch.IndexTemplate.Mappings)
+	assert.False(t, descriptor.Manifest.Elasticsearch.IndexTemplate.Mappings.Subobjects)
```
mrodm, line 96: I think in my previous suggestion, this assert was incorrect, sorry

```suggestion
	assert.True(t, descriptor.Manifest.Elasticsearch.IndexTemplate.Mappings.Subobjects)
```

mrodm: Or taking into account this condition:
```golang
	if !specVersion.LessThan(semver3_2_0) && !answers.Subobjects {
		manifest.Elasticsearch = &packages.Elasticsearch{
			IndexTemplate: &packages.ManifestIndexTemplate{
				Mappings: &packages.ManifestMappings{
					Subobjects: false,
				},
			},
		}
	}
```
And as subobjects is `true` in this test case, then it should be just nil `descriptor.Manifest.Elasticsearch`? :
```golang
	descriptor := createDataStreamDescriptorFromAnswers(answers, "/tmp/package", specVersion)

	require.Nil(t, descriptor.Manifest.Elasticsearch)
}
```

WDYT?

jsoriano (approved):

PR #2923, opened 2025-09-17
Use Bubbletea TUI library

mjwolf:
This replaces the TUI library used in elastic-package with "github.com/charmbracelet/bubbletea".

The previously used library has more limited functionality, and is also currently unmaintained.  Switching to bubbletea will support some upcoming improvements to elastic-package, which will involve more user interaction, and will benefit from the improved UI elements that are available with bubbletea.  This adds a `tui` package which contains the TUI components, and converts the existing users of `survey` to use the `tui` package.

I've tested all the command which used `survey` (`create package`, `create data-stream`, `export ingest-pipelines` and `export dashboards`) and confirmed they continue to work properly, and the created manifests contain the information selected by the user.

---

# Screenshots

## create data-stream
https://github.com/user-attachments/assets/182f7330-7336-4a63-8849-d50003a44b5d

## create package
https://github.com/user-attachments/assets/ce4ff743-1d3c-4d7f-9dd0-4972dfe6b22e

## export dashboards
https://github.com/user-attachments/assets/82b9ce97-09db-4104-bc82-9f7fa91d78a8

## export ingest-pipelines
https://github.com/user-attachments/assets/e5e67e07-04e0-4e68-ba04-4d948ad41721

### DIFF
--- a/cmd/create_data_stream.go
+++ b/cmd/create_data_stream.go
@@ -202,17 +202,17 @@ func getInitialSurveyQuestionsForVersion(specVersion *semver.Version) []*tui.Que
 	qs := []*tui.Question{
 		{
 			Name:     "name",
-			Prompt:   tui.NewInput("Data stream name:", "new_data_stream"),
+			Prompt:   tui.NewInput("Data stream name", "new_data_stream"),
 			Validate: tui.ComposeValidators(tui.Required, validator.DataStreamDoesNotExist, validator.DataStreamName),
 		},
 		{
 			Name:     "title",
-			Prompt:   tui.NewInput("Data stream title:", "New Data Stream"),
+			Prompt:   tui.NewInput("Data stream title", "New Data Stream"),
 			Validate: tui.Required,
 		},
 		{
 			Name:     "type",
-			Prompt:   tui.NewSelect("Type:", []string{"logs", "metrics"}, "logs"),
+			Prompt:   tui.NewSelect("Type", []string{"logs", "metrics"}, "logs"),
 			Validate: tui.Required,
 		},
 	}
--- a/internal/tui/models.go
+++ b/internal/tui/models.go
@@ -40,82 +40,32 @@ type ValidatorFunc func(interface{}) error
 
 // ANSI 16 color constants
 const (
-	ansiBlack         = "0"
-	ansiRed           = "1"
-	ansiGreen         = "2"
-	ansiYellow        = "3"
-	ansiBlue          = "4"
-	ansiMagenta       = "5"
-	ansiCyan          = "6"
-	ansiWhite         = "7"
-	ansiBrightBlack   = "8" // Gray
-	ansiBrightRed     = "9"
-	ansiBrightGreen   = "10"
-	ansiBrightYellow  = "11"
-	ansiBrightBlue    = "12"
-	ansiBrightMagenta = "13"
-	ansiBrightCyan    = "14"
-	ansiBrightWhite   = "15"
+	ansiBlack         = lipgloss.Color("0")
+	ansiRed           = lipgloss.Color("1")
+	ansiGreen         = lipgloss.Color("2")
+	ansiYellow        = lipgloss.Color("3")
+	ansiBlue          = lipgloss.Color("4")
+	ansiMagenta       = lipgloss.Color("5")
+	ansiCyan          = lipgloss.Color("6")
+	ansiWhite         = lipgloss.Color("7")
+	ansiBrightBlack   = lipgloss.Color("8") // Gray
+	ansiBrightRed     = lipgloss.Color("9")
+	ansiBrightGreen   = lipgloss.Color("10")
+	ansiBrightYellow  = lipgloss.Color("11")
+	ansiBrightBlue    = lipgloss.Color("12")
+	ansiBrightMagenta = lipgloss.Color("13")
+	ansiBrightCyan    = lipgloss.Color("14")
+	ansiBrightWhite   = lipgloss.Color("15")
 )
 
-// colorSupported checks if color output is supported based on environment variables
-func colorSupported() bool {
-	// Check NO_COLOR environment variable (https://no-color.org/)
-	if os.Getenv("NO_COLOR") != "" {
-		return false
-	}
-
-	// Check TERM environment variable
-	term := os.Getenv("TERM")
-	switch {
-	case term == "":
-		return false
-	case term == "dumb":
-		return false
-	default:
-		// Default to supporting color for most modern terminals
-		return true
-	}
-}
-
-// getColor returns the color if colors are supported, empty string otherwise
-func getColor(ansiColor string) lipgloss.Color {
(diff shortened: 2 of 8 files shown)


internal/tui/runner.go
```diff
@@ -0,0 +1,293 @@
...
+
+func (m *questionnaireModel) View() string {
+	if m.finished {
+		return ""
+	}
+
+	var b strings.Builder
+
+	// Header with progress
+	progress := fmt.Sprintf("Question %d of %d", m.currentQuestion+1, len(m.questions))
```
jsoriano, line 117: This is nice, but in some cases we ask at different rounds, so the total count is not known beforehand. For example when creating data streams we add a question if the user selected the logs type on the first round of questions.

With this code after "Question 4 of 4" there is a "Question 1 of 1".

As the final number of questions can be unpredictable, I wonder if it should be shown :thinking:

Or I wonder if there is some way to skip questions depending on the value of previous questions, then maybe we can represent the status better in these cases.

internal/tui/runner.go
jsoriano, line 1: It looks like this library is quite low level and we need to implement a lot of helpers, would it be possible to unit test them?

Or would it be possible to use this library in some more-vanilla way that doesn't require us to maintain all this code?

cmd/create_data_stream.go
```diff
@@ -150,24 +130,23 @@ func createDataStreamCommandAction(cmd *cobra.Command, args []string) error {
...
-						}
-						return ""
-					},
-				},
+				Name:   "inputs",
+				Prompt: multiSelect,
 			},
 		}
-		err = survey.Ask(qs, &answers)
+		err = tui.Ask(qs, &answers)
```
jsoriano, line 149: Thanks for keeping the same semantics, this helps on the migration :+1:

cmd/create_data_stream.go
```diff
@@ -10,13 +10,12 @@ import (
...
 	"slices"
 
-	"github.com/AlecAivazis/survey/v2"
-
 	"github.com/spf13/cobra"
 
 	"github.com/elastic/elastic-package/internal/packages"
 	"github.com/elastic/elastic-package/internal/packages/archetype"
 	"github.com/elastic/elastic-package/internal/surveyext"
+	"github.com/elastic/elastic-package/internal/tui"
```
jsoriano, line 18: Nit. Given that this is an internal package we could even keep it calling `survey` :slightly_smiling_face:

cmd/create_data_stream.go
```diff
@@ -10,13 +10,12 @@ import (
 	"maps"
 	"slices"
 
-	"github.com/AlecAivazis/survey/v2"
-
 	"github.com/spf13/cobra"
 
 	"github.com/elastic/elastic-package/internal/packages"
 	"github.com/elastic/elastic-package/internal/packages/archetype"
 	"github.com/elastic/elastic-package/internal/surveyext"
```
jsoriano, line 17: Nit. Now that we have an internal package to provide the `survey` functionality we don't need this `...ext` package, this could be merged into the `tui internal package.

internal/tui/runner.go
```diff
@@ -0,0 +1,293 @@
...
+	_, ok := prompt.(*Select)
+	return ok
+}
+
+// Ask runs multiple questions and stores answers in the provided struct
+func Ask(questions []*Question, answers interface{}) error {
+	model := newQuestionnaireModel(questions)
+	program := tea.NewProgram(model)
+
+	finalModel, err := program.Run()
```
jsoriano, line 167: With the previous implementation users can see what they have answered in the questions before, but now one can only see the current question. Would it be possible to have this behavior with this library?
imagen

jsoriano (commented):
Thanks for replacing the unmaintained library!

This one looks nice, but I am a bit concerned that it seems to be quite lower level, having to implement even basic UI components, and things like scrolling and moving the cursor :sweat_smile: I wonder if there are other libraries that can help with this.

As the interfaces are well isolated in its own package, this looks fine to me though. I would only like to see some tests for this code if we are going to need to maintain it.

internal/tui/runner.go
```diff
@@ -0,0 +1,293 @@
...
+		if m.finished {
+			return m, tea.Quit
+		}
+
+		switch msg.String() {
+		case "ctrl+c":
+			m.err = fmt.Errorf("cancelled by user")
+			return m, tea.Quit
+
+		case "enter":
```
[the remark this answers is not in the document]
mjwolf: I'm not too sure which way is best here. I would think that comparing the type is better too. But more of the bubbletea examples use string comparison, and since the documentation says it's encouraged, I think it probably really is better to keep using strings

cmd/create_data_stream.go
mjwolf: I think it makes sense to leave it as tui, since it could do more than just ask survey questions in the future

internal/tui/models.go
```diff
@@ -67,3 +143,124 @@ func Required(val interface{}) error {
...
+	v, _ := ver.SetPrerelease("")
+	return "^" + v.String()
+}
+
+// Compile-time interface checks to ensure all prompt types implement the Prompt interface
+var (
+	_ Prompt = &Input{}
+	_ Prompt = &Select{}
+	_ Prompt = &Confirm{}
+	_ Prompt = &MultiSelect{}
```
jsoriano, line 265: Nit. I would expect these checks to be closer to the definitions of each type.

internal/tui/models.go
```diff
@@ -28,21 +35,90 @@ type Prompt interface {
...
+	ansiBrightCyan    = "14"
+	ansiBrightWhite   = "15"
+)
+
+// colorSupported checks if color output is supported based on environment variables
+func colorSupported() bool {
+	// Check NO_COLOR environment variable (https://no-color.org/)
+	if os.Getenv("NO_COLOR") != "" {
+		return false
+	}
```
jsoriano, line 66: Nit. It looks like lipgloss uses [`temenvs.EnvNoColor()`](https://pkg.go.dev/github.com/muesli/termenv#EnvNoColor), that seems to take other variables into account.
```suggestion
	if termenv.EnvNoColor() {
		return false
	}
```

internal/tui/runner.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	var b strings.Builder
+
+	// Display previous questions and their answers
+	for i := 0; i < m.currentQuestion; i++ {
+		question := m.questions[i]
+		if answer, exists := m.answers[question.Name]; exists {
+			answerStr := m.formatAnswer(answer)
+			questionLine := fmt.Sprintf("? %s: %s", question.Prompt.Message(), answerStr)
```
jsoriano, line 168: Nit. If we want to keep the previous style, the `?` should be shown in a different color.

jsoriano (commented):
Looks good, added a couple more of small comments.
It still looks to me that with this library we have to maintain much more code, but it is definitely much more flexible.

PR #2939, opened 2025-09-22
[System tests] Refactor mapping validation

mrodm:
Part of #2835 
Relates #2885

Apply some refactors to the validation mapping logic run in system tests.


mrodm:
test integrations

PR #2946, opened 2025-09-23
[System tests] [input packages] Ensure dataStreamManifest is not filled with values from package manifest

mrodm:
This PR ensures that the `dataStreamManifest` field from the system tester is not filled with the values from the package manifest when testing with input packages.

If it is tested input packages, `r.dataStreamPath` is empty and the file that is read in `packages.ReadDataStreamManifest()` would be just `manifest.yml` and therefore for input packages it would correspond to the package manifest itself.

As now `r.dataStreamManifest` could be nil, a few changes have been performed to avoid errors while running the system tests.

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -277,9 +277,15 @@ func NewSystemTester(options SystemTesterOptions) (*tester, error) {
 		return nil, fmt.Errorf("reading package manifest failed: %w", err)
 	}
 
-	r.dataStreamManifest, err = packages.ReadDataStreamManifest(filepath.Join(r.dataStreamPath, packages.DataStreamManifestFile))
-	if err != nil {
-		return nil, fmt.Errorf("reading data stream manifest failed: %w", err)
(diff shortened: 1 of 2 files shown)


internal/testrunner/runners/system/tester.go
```diff
@@ -1870,14 +1890,17 @@ func createPackageDatastream(
...
 	config testConfig,
 	suffix string,
-) kibana.PackageDataStream {
+) (kibana.PackageDataStream, error) {
 	if pkg.Type == "input" {
-		return createInputPackageDatastream(kibanaPolicy, pkg, policyTemplate, config, suffix)
+		return createInputPackageDatastream(kibanaPolicy, pkg, policyTemplate, config, suffix), nil
+	}
+	if ds == nil {
+		return kibana.PackageDataStream{}, fmt.Errorf("data stream manifest is required for integration packages")
```
jsoriano, line 1901: This error assumes that by not being an input package this is an integration package.
This is fine by now, but may not be in the future. In any case we will need to review this code if we introduce system tests in other types of packages.

jsoriano (approved):

PR #2956, opened 2025-09-25
[System tests] Skip dynamic templates that do not contain match conditions

mrodm:
Relates #2885

This PR skips dynamic templates that do not contain any match condition:
- `match` and `unmatch`
- `path_match`and `path_unmatch`
- `match_mapping_type` and `unmatch_mapping_type`

Related docs: https://www.elastic.co/docs/manage-data/data-store/mapping/dynamic-templates#match-mapping-type


mrodm:
test integrations

PR #3008, opened 2025-10-20
Ensure errors from boolean flags in CLI are managed

mrodm:
Ensure that errors from boolean flags are managed appropriately.
This PR focuses on flags: `--tls-skip-verify` and `--allow-snapshot`


jsoriano (approved):
👍

Issue #2944, opened 2025-09-23
implement script testing


efd6:
In order to allow testing unhappy path as proposed in #1941 I suggest that we implement script testing based on [github.com/rogpeppe/go-internal/testscript](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript). This package was chosen because it is the only well maintained and extensively used script testing package available with the features that we need. Alternatives such as JS, CEL, starlark, or yaml are not chosen because they are either too strong, too weak or impose a learning cost that is higher than should be expected for a testing system.

Closes #2868

PR #3012, opened 2025-10-21
internal/testrunner/script: add script testing package

efd6:
This adds script testing for data streams. This is the MVP, future versions can be generalised to operate on input packages.

The current implementation supports:
- pipeline testing
- system testing
- package upgrade testing (only to latest, not to arbitrary versions)
- shared stack
- independent stack
- docker services

Not supported:
- test coverage
- report output configuration
- k8s services
- tf services (ish, these can be shimmed via docker)

Depends on a v3.5.1 of package-spec (currently shimmed in via a go.mod replace directive, to be removed)

Please take a look.

For #2944

### DIFF
--- a/internal/testrunner/script/data_stream.go
+++ b/internal/testrunner/script/data_stream.go
@@ -113,7 +113,7 @@ func addPackagePolicy(ts *testscript.TestScript, neg bool, args []string) {
 	ts.Check(decoratedWith("reading policy", err))
 	ts.Check(decoratedWith("assigning policy", stk.kibana.AssignPolicyToAgent(ctx, installed.enrolled, *pol)))
 
-	dsName := system.BuildDataStreamName(templ.Input, pds, pkgMan.Type, config.Vars)
+	dsName := system.BuildDataStreamName(pds, templ, pkgMan.Type, config.Vars)
 	ts.Setenv(dsNameLabel, dsName)
 	dataStreams[dsName] = struct{}{}
 


.go-version
```diff
@@ -1 +1 @@
-1.24.2
+1.25.3
```
efd6: This change should be a separate PR, but included here to allow the tests to actually run.

jsoriano: Will be done in https://github.com/elastic/elastic-package/pull/3015.

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+
+- `CONFIG_ROOT`: the `elastic-package` configuration root path
+- `CONFIG_PROFILES`: the `elastic-package` profiles configuration root path
+- `HOME`: the user's home directory path
+- `PKG`: the name of the running package
+- `PKG_ROOT`: the path to the root of the running package
+- `CURRENT_VERSION`: the current version of the package
+- `PREVIOUS_VERSION`: the previous version of the package
+- `DATA_STREAM`: the name of the data stream
+- `DATA_STREAM_ROOT`: the path to the root of the data stream
```
jsoriano: Should `NETWORK_NAME` be added to this list?

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+- package commands:
+  - `add_package`: add the current package's assets
+  - `remove_package`: remove assets for the current package
+  - `add_package_zip`: add assets from a Zip-packaged integration package
+  - `remove_package_zip`: remove assets for Zip-packaged integration package
+  - `upgrade_package_latest`: upgrade the current package or another named package to the latest version
+
+- data stream commands:
+  - `add_data_stream`: add a data stream policy
+  - `remove_data_stream`: remove a data stream policy
```
jsoriano: We should probably refer to package policies here instead of referring to data streams. Historically this has been called data streams, but in fleet this is called package policy. In `elastic-package` there are two different types and associated methods because of this.

Lets converge to Fleet API terminology.

See https://github.com/elastic/elastic-package/issues/3038.
```suggestion
  - `add_package_policy`: add a package policy for a data stream
  - `remove_package_policy`: remove a package policy
```

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+
+The `elastic-package test script` command has the following sub-command-specific
+flags:
+
+- `--continue`: continue running the script if an error occurs
+- `--data-streams`: comma-separated data streams to test
+- `--external-stack`: use external stack for script tests (default true)
+- `--run`: run only tests matching the regular expression
+- `--scripts`: path to directory containing test scripts (advanced use only)
+- `--update`: update archive file if a cmp fails
```
jsoriano: Does it mean to update the txtar file itself, for the part with the file?

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+- `--verbose-scripts`: verbose script test output (show all script logging)
```
jsoriano: Would it be possible to use the general `-v` flag? It can be used twice (`-vv`) to enable a trace mode.

internal/testrunner/script/script.go
```diff
@@ -0,0 +1,688 @@
...
+		return err
+	}
+	update, err := cmd.Flags().GetBool(cobraext.UpdateScriptTestArchiveFlagName)
+	if err != nil {
+		return err
+	}
+	cont, err := cmd.Flags().GetBool(cobraext.ContinueOnErrorFlagName)
+	if err != nil {
+		return err
+	}
```
jsoriano: Flags should be get from `cmd` as done in other sub-commands.

internal/testrunner/script/script.go
```diff
@@ -0,0 +1,688 @@
...
+		return fmt.Errorf("could not make work space root: %w", err)
+	}
+	var workdirRoot string
+	if work {
+		// Only create a work root and pass it in if --work has been requested.
+		// The behaviour of testscript is to set TestWork to true if the work
+		// root is non-zero, so just let testscript put it where it wants in the
+		// case that we have not requested work to be retained. This will be in
+		// os.MkdirTemp(os.Getenv("GOTMPDIR"), "go-test-script") which on most
+		// systems will be /tmp/go-test-script. However, due to… decisions, we
```
jsoriano: What decisions are you referring to? 🙂

test/packages/other/with_script/data_stream/first/_dev/test/scripts/agent_up_down.txt
```diff
@@ -0,0 +1,61 @@
...
+stack:
+    image_ref_overrides: {}
+profile:
+    current: default
+-- profiles/default/profile.json --
+{
+  "name": "default",
+  "date_created": "2025-09-15T03:46:37.979361939Z",
+  "version": "1"
+}
```
jsoriano: Where are these files created and how are we telling the commands to use them?

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Expressing tests
+
+Tests are written as [txtar format](https://pkg.go.dev/golang.org/x/tools/txtar#hdr-Txtar_format)
+files in a data stream's \_dev/test/scripts directory. The logic for the test is
+written in the txtar file's initial comment section and any additional resource
+files are included in the txtar file's files sections.
+
+The standard commands and behaviors for testscript scripts are documented in
+the [testscript package documentation](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript).
```
jsoriano: Just for the record. Could you comment, maybe in the description of the PR, why you chose `testscript` instead of some other embedded language?

jsoriano (commented):
Thanks! This is a pretty interesting feature.

I haven't checked the implementation in detail, added by now some comments and questions about the overall behavior.

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+# Start the service.
+docker_signal test-hits SIGHUP
+
+# Wait for the service to exit.
+docker_wait_exit -timeout 5m test-hits
+
+# Check that we can see our policy.
+get_policy -profile ${CONFIG_PROFILES}/default -timeout 1m ${DATA_STREAM_NAME}
+cp stdout got_policy.json
+exec jq '.name=="'${DATA_STREAM_NAME}'"' got_policy.json
```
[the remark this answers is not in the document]
efd6: As you say, this can already be done. Tests are not run without authorisation so it (as is always the case) up to the authoriser to verify that externally provided changes are safe to run. Note also that even an integration that only uses CEL, and possibly even filestream (with imagination), can be bent to malicious use in CI.

If is really is a significant concern I can write a local version of `exec` that takes an allow list and for, but this would not have the full set of features of the built in testscript version since there is internal state registration that the built in command takes advantage of.

docs/howto/script_testing.md
efd6: I'm not convinced, but I will do this if necessary. The currently used name clearly defines what is being done, the fleet term does not.

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+
+# Start the service.
+docker_signal test-hits SIGHUP
+
+# Wait for the service to exit.
+docker_wait_exit -timeout 5m test-hits
+
+# Check that we can see our policy.
+get_policy -profile ${CONFIG_PROFILES}/default -timeout 1m ${DATA_STREAM_NAME}
+cp stdout got_policy.json
```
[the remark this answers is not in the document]
efd6: All work is relative to `$WORK`. This is in ~/.elastic-package/tmp/script_tests/\<slug>/.

docs/howto/script_testing.md
efd6: Yes. When there is an e.g. `cmp stdout want.text` and the want.text file in the txtar does not match, it is updated.

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+  - `uninstall_pipelines`: remove installed ingest pipelines
+
+
+## Environment variables
+
+- `CONFIG_ROOT`: the `elastic-package` configuration root path
+- `CONFIG_PROFILES`: the `elastic-package` profiles configuration root path
+- `HOME`: the user's home directory path
+- `PKG`: the name of the running package
+- `PKG_ROOT`: the path to the root of the running package
```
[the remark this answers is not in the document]
efd6: I don't see that this adds any clarity at the cost of additional text.

test/packages/other/with_script/data_stream/first/_dev/test/scripts/agent_up_down.txt
efd6: Relative to $WORK.

docs/howto/script_testing.md
efd6: It's the only well maintained and extensively used script testing package available with the features that we need. It's not a language, so I'm not sure what competing systems you are referring to.

internal/testrunner/script/script.go
```diff
@@ -0,0 +1,688 @@
...
+				"dump_logs":              dumpLogs,
+				"match_file":             match,
+				"get_policy":             getPolicyCommand,
+			},
+			Setup: func(e *testscript.Env) error {
+				e.Setenv("CONFIG_ROOT", loc.RootDir())
+				e.Setenv("CONFIG_PROFILES", loc.ProfileDir())
+				e.Setenv("HOME", home)
+				if pkgRoot != "" {
+					e.Setenv("PKG", filepath.Base(pkgRoot))
```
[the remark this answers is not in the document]
jsoriano: Packages declare their name in the manifest, with the `name` field. In most of the cases, this is the same value as `filepath.Base(pkgRoot)`, but not in all of them. See for example the packages for universal profiling or keeper in the integrations repository. The apm package used to live in an `apmpackage` directory in the apm-server repository. And I think there were other cases.

In general it cannot be assumed that the package is named like the directory that contains it.

This is relevant for example when creating package policies, because the package name that must be used is the one in the manifest.

internal/testrunner/script/script.go
jsoriano: Maybe this should be part of the coment 🙂

docs/howto/script_testing.md
jsoriano: > The currently used name clearly defines what is being done

Not really. This action is creating a package policy and assigning it to an agent policy, and in some cases it will create also an index template. Then if the agent policy is assigned to some agent, and it sends data using this config, a data stream will be created.

docs/howto/script_testing.md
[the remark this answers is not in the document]
efd6: Cost is subjective. I will relent, but this is an ongoing cost that I will be paying whenever I'm reading tests.

internal/testrunner/script/script.go
efd6: This will be done in the follow up, #3053.

docs/howto/script_testing.md
jsoriano: > The functions that are called are `system.CreatePackageDatastream` and `stk.es.Indices.DeleteDataStream`.

Yes, this is part of the problem described in https://github.com/elastic/elastic-package/issues/3038.
In the kibana client we actually have functions with both names that use the same Fleet APIs.
* https://github.com/elastic/elastic-package/blob/e84244e66b81db61c7a9cf1e5237b22338bd279c/internal/kibana/policies.go#L242
* https://github.com/elastic/elastic-package/blob/e84244e66b81db61c7a9cf1e5237b22338bd279c/internal/kibana/policies.go#L287

But the ones with "DataStream" are using a format in the body request that is deprecated according to the [documentation](https://www.elastic.co/docs/api/doc/kibana/operation/operation-get-fleet-package-policies#operation-post-fleet-package-policies-body-application-json).

At some point we should migrate everything to the "PackagePolicy" functions, to avoid the deprecation and to be coherent with the naming used in the API that these functions are using.

If you prefer to keep `add_data_stream` and `remove_data_stream` by now because they are using the "DataStream" functions, take into account that we may rename them in the future when we apply these refactors, or we will need to duplicate the functions or keep these legacy names to avoid breaking changes in script tests...

Makefile
```diff
@@ -97,6 +97,9 @@ test-check-packages-with-kind:
 test-check-packages-other:
 	PACKAGE_TEST_TYPE=other ./scripts/test-check-packages.sh
 
+test-check-packages-independent-script:
+	elastic-package test script -C test/packages/other/with_script --external-stack=false --defer-cleanup 1s
```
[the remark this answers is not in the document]
efd6: When `external-stack=false`, the script brings up a stack. This would fail if a stack is already running due to port collision. The external and non-external stack tests need to be in the same package to ensure that the conditional is correctly implemented.

internal/testrunner/script/script.go
```diff
@@ -0,0 +1,699 @@
...
+	}
+	loc, err := locations.NewLocationManager()
+	if err != nil {
+		return err
+	}
+	work, err := cmd.Flags().GetBool(cobraext.WorkScriptTestFlagName)
+	if err != nil {
+		return err
+	}
+	workRoot := filepath.Join(home, loc.TempDir(), "script_tests")
```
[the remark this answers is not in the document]
efd6: Thanks, originally, I used `os.UserHomeDir()`/`filepath.FromSlash(".elastic-package/tmp")` but was asked to change it, and failed to update that.

cmd/testrunner.go
```diff
@@ -600,6 +604,46 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
...
+	cmd.Flags().String(cobraext.ScriptsFlagName, "", cobraext.ScriptsFlagDescription)
+	cmd.Flags().Bool(cobraext.ExternalStackFlagName, true, cobraext.ExternalStackFlagDescription)
+	cmd.Flags().StringSliceP(cobraext.DataStreamsFlagName, "d", nil, cobraext.DataStreamsFlagDescription)
+	cmd.Flags().String(cobraext.RunPatternFlagName, "", cobraext.RunPatternFlagDescription)
+	cmd.Flags().BoolP(cobraext.UpdateScriptTestArchiveFlagName, "u", false, cobraext.UpdateScriptTestArchiveFlagDescription)
+	cmd.Flags().BoolP(cobraext.WorkScriptTestFlagName, "w", false, cobraext.WorkScriptTestFlagDescription)
+	cmd.Flags().Bool(cobraext.ContinueOnErrorFlagName, false, cobraext.ContinueOnErrorFlagDescription)
+	cmd.Flags().Bool(cobraext.VerboseScriptFlagName, false, cobraext.VerboseScriptFlagDescription)
+
+	cmd.MarkFlagsMutuallyExclusive(cobraext.DataStreamsFlagName, cobraext.DataStreamsFlagName)
```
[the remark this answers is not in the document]
efd6: Thanks, yes, I did.

efd6:
> 1. When we run dump_logs, it captures logs like $WORK/logs/elastic-agent-internal/elastic-agent-YYYYMMDD.ndjson and $WORK/logs/elastic-agent-internal/cel/cel-UUID-URL.ndjson. Will these less predictable file names make it difficult to write assertions on their file contents from testscript?

It will to a degree; from what I can see elastic-package's testing was not written with a concern for bit-reproducible testing. The filenames there are a consequence of the way that logs are written by the underlying routines in elastic-agent though AFAICS. At the moment, the way to handle these would be to have a helper shell script or maybe `jq` over the glob. If this is seen as too complex, we can maybe harmonise all the files in those directories into single file collections that have been temporally ordered and written to a file with a known name.

The current situation does not make it impossible, but depending on the threshold for "difficult", yes, no or maybe.
 
> 2. Can we add a mechanism to capture the registry files from the Elastic Agent container? I would like to be able to run assertions on their contents, like checking that the httpjson cursor contains a specific timestamp or pagination token after test completion.

We can; in fact that is one of my unchecked TODO boxes. Would you like that added in this change, or can it be added in a follow-up?

efd6:
> If this is seen as too complex, we can maybe harmonise all the files in those directories into single file collections that have been temporally ordered and written to a file with a known name.

I've done this.

docs/howto/script_testing.md
```diff
@@ -0,0 +1,209 @@
...
+- `--verbose-scripts`: verbose script test output (show all script logging)
+- `--work`: print temporary work directory and do not remove when done
+
+
+## Limitations
+
+While the testscript package allows reference to paths outside the configuration
+root and the package's root, the backing `elastic-package` infrastructure does
+not, so it is advised that tests only refer to paths within the `$WORK` and
+`$PKG_ROOT` directories.
```
mrodm: Looking into the environment variables listed above, should this be `PACKAGE_ROOT`?
```suggestion
`$PACKAGE_ROOT` directories.
```

mrodm (approved):
LGTM
Just added a few comments.

It would be interesting to wait also for @jsoriano's review before merging, since he had some comments here too.

PR #3023, opened 2025-10-24
internal/testrunner/runners/system: supress absent Records complaint by awss3 input

efd6:
The awss3 input incorrectly complains about the absence of a Records field in SQS messages. Valid SQS messages are possible without a Records field as described in the AWS documentation[1].

[1]https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html#notification-content-structure-examples

See also https://github.com/elastic/integrations/issues/15472#issuecomment-3439970619.

### DIFF
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -153,9 +153,12 @@ var (
 				{
 					includes: regexp.MustCompile("->(FAILED|DEGRADED)"),
 
-					// this regex is excluded to avoid a regresion in 8.11 that can make a component to pass to a degraded state during some seconds after reassigning or removing a policy
 					excludes: []*regexp.Regexp{
+						// this regex is excluded to avoid a regresion in 8.11 that can make a component to pass to a degraded state during some seconds after reassigning or removing a policy
 						regexp.MustCompile(`Component state changed .* \(HEALTHY->DEGRADED\): Degraded: pid .* missed .* check-in`),
+						// This is excluded because the awss3 input incorrectly complains about missing Records fields in s3:TestEvent messages.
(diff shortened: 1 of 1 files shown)


internal/testrunner/runners/system/tester.go
```diff
@@ -153,9 +153,12 @@ var (
 				{
 					includes: regexp.MustCompile("->(FAILED|DEGRADED)"),
 
-					// this regex is excluded to avoid a regresion in 8.11 that can make a component to pass to a degraded state during some seconds after reassigning or removing a policy
 					excludes: []*regexp.Regexp{
+						// this regex is excluded to avoid a regresion in 8.11 that can make a component to pass to a degraded state during some seconds after reassigning or removing a policy
 						regexp.MustCompile(`Component state changed .* \(HEALTHY->DEGRADED\): Degraded: pid .* missed .* check-in`),
+						// This is excluded because the awss3 input incorrectly complains about missing Records fields in s3:TestEvent messages.
+						// See https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html#notification-content-structure-examples.
+						regexp.MustCompile(`state changed .* \(HEALTHY->DEGRADED\): .* the message is an invalid S3 notification: missing Records field`),
```
mrodm: One doubt here, is this exception required even if applying one of the suggestions mentioned in https://github.com/elastic/integrations/issues/15472#issuecomment-3439970619 ? 

It looks like that if one of those suggestions mentioned in the integrations issue is applied, this exception/exclude would not be needed:
- with the first suggestion, it would imply that there is no error anymore
- with the second suggestion, IIUC the error reported in Elastic Agent logs would be a valid error since it won't be an `s3:TestEvent` message.

Would it be needed to wait to merge this PR depending on the action taken in the integrations issue ? 
cc @jsoriano

PR #3027, opened 2025-10-27
Add find and foreach sub commands for bulk operations

### LINKED ISSUE #2327: Add subcommand for bulk actions
In repositories with multiple packages, and in elastic-package development itself, it is usual to run actions on multiple packages. For this, the developer usually needs to craft some kind of shell pipeline.

It could be useful to introduce in elastic-package a subcommand that can discover packages under certain rules and run some other subcommand for the discovered packages. For example the following command would add a changelog to all the modified packages under the packages directory.
```
elastic-package foreach --modified --packages ./packages/* changelog add ...
```
Or to test elastic-package while development with a set of packages:
```
elastic-package foreach --packages ./test/packages/parallel/* test
```

In general, its usage would be like this:
```
elastic-package foreach [discovery flags] [other flags] [subcommand] [subcommand flags]
```

Discovery flags
----
The subcommand would contain a set of flags used to discover packages, that can be combined, for example:
* `--modified` to select only packages with modified files.
* `--packages [glob]` to select packages matching with a glob pattern.
* `--auto` to walk the repository looking for compatible manifest files.
* `--constraint [constraint]` to filter packages based on their constraints (like `--constraint kibana.version=8.17.0`).

Not all these flags would be needed in a first MVP, they could be added later, as well as other flags.

Technically, the `foreach` subcommand would call the specified subcommand for each one of the discovered packages, setting for the called subcommands the `-C` flag.

Other flags
----
There could be support for other flags that control execution of subcommands, for example `--parallel 4` would allow the execution of four subcommands in parallel.

Available subcommands
----
In principle all subcommands would be supported, we would need to investigate what is the best implementation so the `--help` makes sense and avoid loops between subcommands.

Additionally, we could add some kind of `query` subcommand that could help in shell pipelines or just listing info from packages. For example the following to list the name of modified packages:
```
elastic-package foreach --modified query --path manifest.package.name
```

vinit-chauhan:
Elastic-package has been extensively used for integrations related tasks. However, most of the commands now are targeted to be run on one package at a time. We don’t have an option if we want to repeat a certain operation across multiple integrations.

This pull request adds two subcommands to the `elastic-package` to allow bulk operations from the elastic-package.

1. `find`
2. `foreach`

Note: Both commands are expected to be run from the integration repository.

## Find subcommand

Find command adds the ability to filter and return a list of integrations based on specified criteria.

Available Filters:

- **Category** (`--categories`)
- **Codeowner** (`--code-owners`)
- **Input** (`--inputs`)
- **Package Dir Name** (`--package-dirs`) (supports glob patterns)
- **Package Name** (`--packages`) (supports glob patterns)
- **Package Type** (`--package-types`) ( eg., `integration`, `input` )
- **Spec Version** (`--spec-version`)

You can chain multiple filters and each filter can have multiple comma-separated values.
Matching:
All filters must match.
At least one of the values match.

> Currently, the find command runs sequentially and reads all the manifest files at once and keeps it in memory. We can update the code to use a buffered channel and have a producer and consumer to reduce memory footprint.

The current elastic-package and spec does not enforce the same package_name and directory name in the repo. Which leads to some integrations having different package_name and directory name.

The find command by default returns the absolute path of the package in the integration repo. However, it also provides a flag `--output-info` to make it return the package name or directory name.

Moreover, it also provides a flag `--output` to select the output format json, yaml or new line separated paths (default).

> If no filter flag is provided, the command will return a list of all the integrations.

```shell
elastic-package find --inputs tcp,udp --code-owners elastic/integration-experience --packages cisco_*
```

## Foreach subcommand

The foreach command leverages the filter registry. Therefore all the flags available in the filter are directly available to foreach commands without any code changes.

~Additionally, foreach has 1 flag `--parallel` which allows the user to run commands parallelly using worker pool.~ (Will be added in a follow-up PR)

> default is 1 ( runs sequentially )

The `elastic-package` command you want to run goes after -- with all of its flags.

> Note: You are only allowed to run allowed elastic-package subcommands ["build","check","changelog","clean","format","install","lint","test","uninstall"] (cmd/foreach.go)

```shell
elastic-package foreach --inputs tcp,udp --code-owners elastic/integration-experience -- test system -g
```

### File changes:

`internal/packages/packages.go`: Added function to find the integrations repo root dir and read all manifests.
`cmd/find.go`: Find command implementation
`cmd/foreach.go`: Foreach command implementation
`internal/filter/*`: Filter interface and implementation for each filter flag.

## Related Issues

- Closes #2327
- Closes https://github.com/elastic/integration-experience/issues/341
- Related https://github.com/elastic/integration-experience/issues/280

## AI Tools used

- `Cursor` With `Claude-4.5-Sonnet`
- `Antigravity` With `Gemini-3-Pro`

### DIFF
--- a/cmd/foreach.go
+++ b/cmd/foreach.go
@@ -18,11 +18,11 @@ import (
 )
 
 const foreachLongDescription = `[Technical Preview]
-Execute a command for each package matching the given filter criteria.
+Execute a command for each package matching the given query flags.
 
-This command combines filtering capabilities with command execution, allowing you to run any elastic-package subcommand across multiple packages in a single operation.
+This command combines query capabilities with command execution, allowing you to run any elastic-package subcommand across multiple packages in a single operation.
 
-The command uses the same filter flags as the 'filter' command to select packages, then executes the specified subcommand for each matched package.`
+The command uses the same query flags as the 'find' command to select packages, then executes the specified subcommand for each matched package.`
 
 // getAllowedSubCommands returns the list of allowed subcommands for the foreach command.
 func getAllowedSubCommands() []string {
@@ -50,7 +50,7 @@ func setupForeachCommand() *cobraext.Command {
 		Args: cobra.MinimumNArgs(1),
 	}
 
-	// Add filter flags
+	// Add query flags
 	filter.SetFilterFlags(cmd)
 
 	return cobraext.NewCommand(cmd, cobraext.ContextPackage)
@@ -61,8 +61,8 @@ func foreachCommandAction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("validating sub command failed: %w", err)
 	}
 
-	// reuse filterPackage from cmd/filter.go
-	filtered, err := filterPackage(cmd)
+	// reuse findPackage from cmd/find.go
+	filtered, err := findPackage(cmd)
 	if err != nil {
 		return fmt.Errorf("filtering packages failed: %w", err)
 	}
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ Use this command to export ingest pipelines with referenced pipelines from the E
 
 Use this command to download selected ingest pipelines and its referenced processor pipelines from Elasticsearch. Select data stream or the package root directories to download the pipelines. Pipelines are downloaded as is and will need adjustment to meet your package needs.
 
-### `elastic-package filter [flags]`
+### `elastic-package find [flags]`
 
 _Context: package_
 
@@ -382,11 +382,11 @@ Use --change-directory to change the working directory and --depth to change the
 _Context: package_
 
 [Technical Preview]
-Execute a command for each package matching the given filter criteria.
+Execute a command for each package matching the given query flags.
 
-This command combines filtering capabilities with command execution, allowing you to run any elastic-package subcommand across multiple packages in a single operation.
+This command combines query capabilities with command execution, allowing you to run any elastic-package subcommand across multiple packages in a single operation.
 
-The command uses the same filter flags as the 'filter' command to select packages, then executes the specified subcommand for each matched package.
+The command uses the same query flags as the 'find' command to select packages, then executes the specified subcommand for each matched package.
 
 Allowed subcommands:
 build, check, changelog, clean, format, install, lint, test, uninstall
--- a/cmd/find.go
+++ b/cmd/find.go
@@ -16,24 +16,24 @@ import (
 	"github.com/elastic/elastic-package/internal/packages"
 )
 
-const filterLongDescription = `[Technical Preview]
+const findLongDescription = `[Technical Preview]
 This command gives you a list of all packages based on the given query.
 
 The command will search for packages in the working directory for default depth of 2 and return the list of packages that match the given criteria. 
 
 Use --change-directory to change the working directory and --depth to change the depth of the search.`
 
-const filterExample = `  elastic-package filter --inputs tcp,udp --categories security --depth 3 --output json
-  elastic-package filter --packages 'cisco_*,fortinet_*' --output yaml`
+const findExample = `  elastic-package find --inputs tcp,udp --categories security --depth 3 --output json
+  elastic-package find --packages 'cisco_*,fortinet_*' --output yaml`
 
-func setupFilterCommand() *cobraext.Command {
+func setupFindCommand() *cobraext.Command {
 	cmd := &cobra.Command{
-		Use:     "filter [flags]",
-		Short:   "Filter integrations based on given flags [Technical Preview]",
-		Long:    filterLongDescription,
+		Use:     "find [flags]",
+		Short:   "Find integrations based on given flags [Technical Preview]",
+		Long:    findLongDescription,
 		Args:    cobra.NoArgs,
-		RunE:    filterCommandAction,
-		Example: filterExample,
+		RunE:    findCommandAction,
+		Example: findExample,
 	}
 
 	// add filter flags to the command (input, code owner, kibana version, categories)
@@ -46,10 +46,10 @@ func setupFilterCommand() *cobraext.Command {
 	return cobraext.NewCommand(cmd, cobraext.ContextPackage)
 }
 
-func filterCommandAction(cmd *cobra.Command, args []string) error {
-	filtered, err := filterPackage(cmd)
+func findCommandAction(cmd *cobra.Command, args []string) error {
+	found, err := findPackage(cmd)
 	if err != nil {
-		return fmt.Errorf("filtering packages failed: %w", err)
+		return fmt.Errorf("finding packages failed: %w", err)
 	}
 
 	outputFormatStr, err := cmd.Flags().GetString(cobraext.FilterOutputFlagName)
@@ -67,14 +67,14 @@ func filterCommandAction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("creating output options failed: %w", err)
 	}
 
-	if err = printPkgList(filtered, outputOptions, os.Stdout); err != nil {
+	if err = printPkgList(found, outputOptions, os.Stdout); err != nil {
 		return fmt.Errorf("printing JSON failed: %w", err)
 	}
 
 	return nil
 }
 
-func filterPackage(cmd *cobra.Command) ([]packages.PackageDirNameAndManifest, error) {
+func findPackage(cmd *cobra.Command) ([]packages.PackageDirNameAndManifest, error) {
 	depth, err := cmd.Flags().GetInt(cobraext.FilterDepthFlagName)
 	if err != nil {
 		return nil, fmt.Errorf("getting depth flag failed: %w", err)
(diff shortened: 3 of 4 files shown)


cmd/foreach.go
```diff
@@ -0,0 +1,162 @@
...
+	if errs.Error() != "" {
+		return fmt.Errorf("errors occurred while executing command for packages: \n%s", errs.Error())
+	}
+
+	return nil
+}
+
+func executeCommand(args []string, path string) error {
+	// Look up the elastic-package binary in PATH
+	execPath, err := exec.LookPath("elastic-package")
```
teresaromero: as this is a command run on top of other commands of this binary, could we target directly the commands by its functions, instead of looking for the path binary?
what if the binary is not updated or needs to run specific version? just having some thoughts here on how would be a better aproach without depending of having the binary installed as its own dependency.

jsoriano: I think it should be possible to do something like this:
```go
cmd := cmd.RootCmd()
cmd.SetArgs(args)
return cmd.Execute()
```

We could also create our own command that is like `cmd.RootCmd()`, but where we only add the commands with `AddCommand` that we allow.

internal/cobraext/flags.go
```diff
@@ -133,8 +133,37 @@ const (
...
+
+	FilterKibanaVersionFlagName        = "kibana-version"
+	FilterKibanaVersionFlagDescription = "kibana version to filter by (semver)"
+
+	FilterOutputPackageNameFlagName        = "output-package-name"
+	FilterOutputPackageNameFlagShorthand   = "p"
+	FilterOutputPackageNameFlagDescription = "print the package name instead of the directory name in the output"
+
+	FilterPackagesFlagName        = "package-name"
+	FilterPackagesFlagDescription = "package names to filter by (comma-separated values)"
```
jsoriano: Maybe we can call this flag directly `packages`, or `package-path` to avoid confusion between package paths and their names.

jsoriano (commented):
Thanks for working on this! This will be helpful. Added some questions and suggestions.

jrmolin:
are you going to pull this out of draft?

internal/packages/packages.go
```diff
@@ -419,6 +425,87 @@ func ReadPackageManifest(path string) (*PackageManifest, error) {
...
+
+		// Check if this is a manifest file
+		if d.Name() != PackageManifestFile {
+			return nil
+		}
+
+		// Validate it's a package manifest
+		ok, err := isPackageManifest(path)
+		if err != nil {
+			// Log the error but continue searching
```
mjwolf: This doesn't actually log the error here. You should either remove the comment or add logging

cmd/foreach.go
```diff
@@ -0,0 +1,96 @@
...
+	// reuse filterPackage from cmd/filter.go
+	filtered, err := filterPackage(cmd)
+	if err != nil {
+		return fmt.Errorf("filtering packages failed: %w", err)
+	}
+
+	errors := multierror.Error{}
+
+	for _, pkg := range filtered {
+		rootCmd := cmd.Root()
```
jsoriano: We could define another root command with the specific commands that are supported and use it here, then we could have different sets of subcommands, and maybe we could even plug it directly as subcommands of foreach.

This could be left for a future refactor too.

Something like this:
```go
import (
        "github.com/spf13/cobra"

        "github.com/elastic/elastic-package/internal/cobraext"
)

var forEachCommands = []*cobraext.Command{
        setupBuildCommand(),
        setupCheckCommand(),
        setupCleanCommand(),
        setupFormatCommand(),
        setupInstallCommand(),
        setupLintCommand(),
        setupTestCommand(),
        setupUninstallCommand(),
}

// ForEachRootCmd creates and returns root cmd for elastic-package
func ForEachRootCmd() *cobra.Command {
        forEachCmd := &cobra.Command{
                SilenceUsage: true,
                PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
                        return cobraext.ComposeCommandActions(cmd, args,
                                processPersistentFlags,
                                checkVersionUpdate,
                        )
                },
        }
        forEachCmd.PersistentFlags().CountP(cobraext.VerboseFlagName, cobraext.VerboseFlagShorthand, cobraext.VerboseFlagDescription)
        forEachCmd.PersistentFlags().StringP(cobraext.ChangeDirectoryFlagName, cobraext.ChangeDirectoryFlagShorthand, "", cobraext.ChangeDirectoryFlagDescription)

        for _, cmd := range forEachCommands {
                forEachCmd.AddCommand(cmd.Command)
        }
        return forEachCmd
}
```

That could be used here like this:

```suggestion
        rootCmd := ForEachRootCmd()
        rootCmd.SetContext(cmd.Context())
```

jrmolin (commented):
there's a startling lack of testing, but it looks reasonable otherwise

cmd/foreach.go
vinit-chauhan: This is something that we need to do to add parallel execution. That's why I left it out of this PR, and would like to revisit it while I work on the other PR. 

With that said, we do have a list of allowed sub-commands. [Here](https://github.com/vinit-chauhan/elastic-package/blob/implement-filter-registry/cmd/foreach.go#L29-L40)

Let me know if I missed any command. - I'll also add `changelog` command in allow list.

vinit-chauhan:
Updated the code to support multiple output formats. 
```yaml
- Newline separated list ( default )
- JSON
- YAML 
```

cmd/foreach.go
```diff
@@ -0,0 +1,162 @@
...
+		return fmt.Errorf("elastic-package binary not found in PATH: %w", err)
+	}
+
+	cmd := &exec.Cmd{
+		Path:   execPath,
+		Args:   append([]string{execPath}, args...),
+		Dir:    path,
+		Stdout: io.Discard,
+		Stderr: os.Stderr,
+		Env:    os.Environ(),
```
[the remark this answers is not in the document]
teresaromero: i like to pass specific values in order to make it transparent to the user, so its explicit what is passed to the cmd as environment. I also see it beneficial when debugging to know what values are passed  and control what envs are being uses/required, etc... 

its true that the tool already has access to the environ running locally, it was just for the sake of "explicitness". its fine 👍🏻

internal/cobraext/flags.go
```diff
@@ -133,8 +133,46 @@ const (
...
+	FilterKibanaVersionFlagName        = "kibana-version"
+	FilterKibanaVersionFlagDescription = "kibana version to filter by (semver)"
+
+	FilterOutputFlagName        = "output"
+	FilterOutputFlagDescription = "select the output format [pkgname, dirname, absolute]"
+	FilterOutputFlagDefault     = "dirname"
+	FilterOutputFlagShorthand   = "o"
+
+	FilterPackageDirNameFlagName        = "package-dirs"
+	FilterPackageDirNameFlagDescription = "package directories to filter by (comma-separated values)"
```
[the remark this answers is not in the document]
jsoriano: I was trying again the command today and I struggled again to properly find packages per directory.

I find also a bit confusing to have to use `-C` for this, whose purpose is to change directory to do the operation, and in the case of this command limits the option to look for packages in multiple directories.

I think it would be nice to have an explicit flag to indicate where to look for packages.

internal/filter/type.go
```diff
@@ -0,0 +1,148 @@
...
+		format:   format,
+	}
+	if err := cfg.validate(); err != nil {
+		return nil, err
+	}
+	return cfg, nil
+}
+
+func (o *OutputOptions) validate() error {
+	validInfo := []string{"pkgname", "dirname", "absolute"}
```
jsoriano: Nit. As this is exposed as user options I would use `name` or `package_name` for the package name, and maybe `absolute_path` instead of `absolute`.
```suggestion
	validInfo := []string{"package_name", "dir_name", "absolute_path"}
```

internal/cobraext/flags.go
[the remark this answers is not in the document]
vinit-chauhan: By Default, it looks for package in current dir + 2 depth. 

So you can either increase `--depth/-d` you can look through more directories or run the command in the directory you want to search in. 

```bash
CWD: Users/vinit.chauhan/github.com/elastic/elastic-package
❯ elastic-package filter -d 4 --output-info absolute_path
2025/11/26 10:44:41  INFO Found 65 matching package(s)
/Users/vinit.chauhan/github.com/elastic/elastic-package/internal/fields/testdata
/Users/vinit.chauhan/github.com/elastic/elastic-package/internal/files/testdata/links
/Users/vinit.chauhan/github.com/elastic/elastic-package/internal/files/testdata/testpackage
/Users/vinit.chauhan/github.com/elastic/elastic-package/new_package
/Users/vinit.chauhan/github.com/elastic/elastic-package/test/new_package2
/Users/vinit.chauhan/github.com/elastic/elastic-package/test/packages/benchmarks/pipeline_benchmark
/Users/vinit.chauhan/github.com/elastic/elastic-package/test/packages/benchmarks/rally_benchmark
```

```bash
❯ cd test/packages/parallel 
❯ elastic-package filter -d 4 --output-info absolute_path -v
2025/11/26 10:58:43  INFO Found 19 matching package(s)
/Users/vinit.chauhan/github.com/elastic/elastic-package/test/packages/parallel/apache
/Users/vinit.chauhan/github.com/elastic/elastic-package/test/packages/parallel/apache_basic_license
/Users/vinit.chauhan/github.com/elastic/elastic-package/test/packages/parallel/auditd_manager
/Users/vinit.chauhan/github.com/elastic/elastic-package/test/packages/parallel/auth0_logsdb
```

The only reason to have a flag for filtering package_directories is because package spec allows different package_name and directory name of the package. 

User might want to search all Cisco packages, they can do it using following command. 

```
CWD: .../integrations/
❯ elastic-package filter --package-dirs 'cisco*'
2025/11/26 11:08:44  INFO Found 12 matching package(s)
cisco_aironet
cisco_asa
cisco_duo
cisco_ftd
cisco_ios
cisco_ise
cisco_meraki
cisco_meraki_metrics
cisco_nexus
cisco_secure_email_gateway
cisco_secure_endpoint
cisco_umbrella
```

jsoriano (approved):
Hey, sorry for the delayed review, I thought we had already merged this and I missed it today 🙂

I think we can merge it and keep iterating on it.

Thanks!

PR #3065, opened 2025-11-11
Fix `build` updateReadme inside zip package

teresaromero:
Bug introduced at #2957 

When a package is built, the readme is not being updated at the zip bundle. When working on the mentioned PR, the readme updater was moved after the built so the linked fields where already there when doing the update. The bundle zip is already created when this happens, so the readme inside the zip is the one at source before running the updater.

~~This changes propose to inject the updater function in order to run it after the linked files and the external fields are resolved. This way all the files in the bundle are the updated ones.~~

~~I've also explored the option of removing from the Builder function the link resolver, but still, the external fields have to be resolved inside the Builder, and if the readme updater runs before this, the readme wont have the external fields documented.~~

The PR adds a new option to update readmes inside the builder function.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/elastic/elastic-package/internal/builder"
 	"github.com/elastic/elastic-package/internal/cobraext"
-	"github.com/elastic/elastic-package/internal/docs"
 	"github.com/elastic/elastic-package/internal/files"
 	"github.com/elastic/elastic-package/internal/logger"
 	"github.com/elastic/elastic-package/internal/packages"
@@ -86,7 +85,8 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 		SignPackage:     signPackage,
 		SkipValidation:  skipValidation,
 		RepositoryRoot:  repositoryRoot,
-	}, docs.UpdateReadmes)
+		UpdateReadmes:   true,
+	})
 	if err != nil {
 		return fmt.Errorf("building package failed: %w", err)
 	}
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/magefile/mage/sh"
 
+	"github.com/elastic/elastic-package/internal/docs"
 	"github.com/elastic/elastic-package/internal/environment"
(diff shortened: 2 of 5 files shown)


internal/builder/packages.go
```diff
@@ -210,32 +212,41 @@ func BuildPackage(ctx context.Context, options BuildOptions) (string, error) {
...
+		return "", nil, fmt.Errorf("adding dynamic mappings: %w", err)
 	}
 
 	err = resolveTransformDefinitions(builtPackageDir)
 	if err != nil {
-		return "", fmt.Errorf("resolving transform manifests failed: %w", err)
+		return "", nil, fmt.Errorf("resolving transform manifests failed: %w", err)
+	}
+
+	updatedReadmesTargets, err := readmeUpdater(options.RepositoryRoot, options.PackageRootPath, options.BuildDir)
```
jsoriano: Could we directly use `docs.UpdateReadmes` instead of having to pass this function? At the end it makes sense to make the update of readmes part of the build.
```suggestion
	updatedReadmesTargets, err := docs.UpdateReadmes(options.RepositoryRoot, options.PackageRootPath, options.BuildDir)
```

teresaromero: at the installer this step is not done, that is why i passed the function, and at the installer its a noop one to not run the update readme. 

we were discussing offline with @mrodm why the installer wasn't updating the readmes. some ideas where to decouple the build function and have one for installing a package and an other for building the package itself.

cmd/build.go
```diff
@@ -80,24 +80,19 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 	}
 	logger.Debugf("Use build directory: %s", buildDir)
 
-	target, err := builder.BuildPackage(cmd.Context(), builder.BuildOptions{
+	target, updatedReadmeTargets, err := builder.BuildPackage(cmd.Context(), builder.BuildOptions{
```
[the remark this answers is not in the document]
teresaromero: yup, i dont like either but as i saw this mas a cmd log i felt it was a solution... i will check this out and log inside the builder 👍🏻

internal/builder/packages.go
teresaromero: updated on the last commit to pass a bool option.

Issue #991, opened 2022-09-25
feature: render package, stack and elastic-package version into test reports


efd6:
Currently elastic-package test builds do not explicitly render the package, stack and elastic-package executable versions into the test report. Having these available can be very helpful when attempting to debug issues relating to changes in the versions of system components.

When `elastic-package test` is run the package version *is* incidentally rendered into the test output in the asset tests in both the human and XML report outputs, but the stack version and elastic-stack executable versions are not. See details (all run with elastic-package as at ba7e6801f6944617e5ca33ebd60402385bf079fc with the default stack version 8.4.1 in this version of ep.

Human output

<details>
<pre>
$ elastic-package test
2022/09/25 10:10:19  WARN CommitHash is undefined, in both .../.elastic-package/version and the compiled binary, config may be out of date.
Run test suite for the package
Run system tests for the package
2022/09/25 10:11:10  INFO Write container logs to file: .../go/src/github.com/elastic/integrations/build/container-logs/meraki-webhook-https-1664066470434862000.log
2022/09/25 10:11:57  INFO Write container logs to file: .../go/src/github.com/elastic/integrations/build/container-logs/meraki-webhook-http-1664066517438873000.log
2022/09/25 10:12:36  INFO Write container logs to file: .../go/src/github.com/elastic/integrations/build/container-logs/cisco_meraki-log-logfile-1664066556401479000.log
2022/09/25 10:13:23  INFO Write container logs to file: .../go/src/github.com/elastic/integrations/build/container-logs/cisco_meraki-log-tcp-1664066603425999000.log
2022/09/25 10:14:06  INFO Write container logs to file: .../go/src/github.com/elastic/integrations/build/container-logs/cisco_meraki-log-udp-1664066646381401000.log
--- Test results for package: cisco_meraki - START ---
╭──────────────┬─────────────┬───────────┬──────────────┬────────┬───────────────╮
│ PACKAGE      │ DATA STREAM │ TEST TYPE │ TEST NAME    │ RESULT │  TIME ELAPSED │
├──────────────┼─────────────┼───────────┼──────────────┼────────┼───────────────┤
│ cisco_meraki │ events      │ system    │ meraki-https │ PASS   │ 35.901381537s │
│ cisco_meraki │ events      │ system    │ meraki-http  │ PASS   │ 30.872360032s │
│ cisco_meraki │ log         │ system    │ logfile      │ PASS   │ 23.100860977s │
[...]

PR #3101, opened 2025-11-24
New test result format `json`

teresaromero:
Related to #991 

New request for format the test report with json


efd6 (commented):
nit only

PR #3104, opened 2025-11-24
Fix documentation of ECS fields in linked files

jsoriano:
Fields validator is used for many things, including looking for fields for documentation.
In any case, it needs to resolve linked fields files, so the fields the linked files contain are included in documentation, built packages and validation processes.
This change explicitly includes links resolution in the validator, what makes it depend on a repository root. Most of the changes here are related to passing the repository root to the validator.


jsoriano:
test integrations

Issue #2993, opened 2025-10-14
Support `elastic-package` usage for integration development on air-gapped environments


jsoriano:
Users may be developing on restricted environments, where they have to access external resources through mirrors or proxies. Support this use case by allowing to configure the access to these resources. This will translate to allow the configuration of any currently hard-coded external URL.

For convenience with these use cases, the configuration to access these resources should be supported at the profile level. `elastic-package` should allow the configuration of any hard-coded external URL.

Some identified resources at the moment:
* Package Registry managed by elastic-package uses `EPR_PROXY_TO` to connect to `https://epr.elastic.co`.
  * Done in https://github.com/elastic/elastic-package/pull/3113. 
* ECS schemas are retrieved from the hard-coded `ecsSchemaURL`.
* `elastic-package status`:
  * It looks for packages published in `https://epr.elastic.co`, this could be parameterizable to use a custom registry.
  * It looks for serverless config in github, this should be probably optional.

nimarezainia:
@jsoriano do we have a full list of these hard-coded settings? (trying to gauge the extent of this effort)

jsoriano:
> [@jsoriano](https://github.com/jsoriano) do we have a full list of these hard-coded settings? (trying to gauge the extent of this effort)

The ones in the description are the only ones we have identified, I have added another reference to `epr.elastic.co` that is probably not so blocker.

There are some other hard-coded URLs used in the serverless stack provider, but we can probably ignore them as it doesn't make sense to use serverless in air-gapped environments.

nimarezainia:
@jsoriano couldn't users use DNS to get the hard-coded urls to resolve to a local address?

nimarezainia:
thanks @jsoriano Would you be able to provide us some small description for documentation as to how the user can deploy EPR in air-gap environments and what configuration to apply? (cc: @jstrassb )

jlind23:
@kpollich Assigning this to you for now until we find the right assignee.

nimarezainia:
@mrodm I was wondering if the changes here were documented? thx

mrodm:
> [@mrodm](https://github.com/mrodm) I was wondering if the changes here were documented? thx

@nimarezainia It was added some documentation in the Readme of this repository, mainly at https://github.com/elastic/elastic-package/blob/main/README.md#elastic-package-configuration

And, there is also a reference about `stack.epr.base_url` setting in  https://github.com/elastic/elastic-package/blob/main/README.md#elastic-package-profiles-1

PR #3114, opened 2025-11-26
[WIP] Allow to use custom URLs to retrieve ECS schemas

jsoriano:
Part of https://github.com/elastic/elastic-package/issues/2993.

Includes base URL for ECS field in the configuration.

There are many components (and tests) assuming that the schemas are available on the local home or in ECS, we should probably add some refactors so retrieving these schemas is more explicit.


jsoriano:
Work on this continues in https://github.com/elastic/elastic-package/pull/3220, closing.

PR #3163, opened 2025-12-16
Implement shared tags functionality to remove tags when dashboard export

teresaromero:
- Resolves https://github.com/elastic/package-spec/issues/977 : package-spec is now validating if the package contains duplicate tags (json files that belong to shared tags declared at `tags.yml` file)

When exporting a dashboard, tags that are shared (defined at kibana/tags.yml) where being also exported into `kibana/tag` folder.

Shared tags are managed by fleet when reading the `tags.yml` so exporting them as `json` is duplicating this tags in the UI.

This PR adds a filter layer when exporting a dashboard, removing shared tag objects and dashboard references to them.
Identification of tags is done based on fleet criteria on creating the json objects when reading the ´tags.yml´ file.

Updated unit tests to include shared tags object and check the filter is working as expected


internal/export/transform_remove_fleet_managed_tags.go
```diff
@@ -80,6 +86,42 @@ func isTagFleetManaged(aId, packageName string) bool {
 	return ok
 }
 
+func isSharedTag(aId string, sharedTags []string) bool {
+	for _, tag := range sharedTags {
+		id := fmt.Sprintf("tag-ref-%s-default", strings.ReplaceAll(strings.ToLower(tag), " ", "-"))
```
teresaromero: does the shared tag have more formatting than this?

internal/export/transform_remove_shared_tags.go
jsoriano: Nit. It is usually discouraged to use underscores in go files or package names, as some of them have special meanings (`_test`, `_linux`...).
Maybe this file could be called just `removesharedtags.go`, or we could merge all these files into a single `tags.go`.
Though I see there are already other files here with underscores.

jsoriano (commented):
Approach looks mostly good to me. I only wonder if we are relying on ids, when we are interested on the names, that is what users see.

Tried to re-export all Crowdstrike dashboards and the changes make sense to me 👍

internal/export/transform_remove_fleet_managed_tags.go
```diff
@@ -80,6 +86,42 @@ func isTagFleetManaged(aId, packageName string) bool {
...
+		if aId == id {
```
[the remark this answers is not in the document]
teresaromero: yup, i missed to rename this when i became aware that the key was "name", the value is the name one. i will fix this

internal/export/transform_remove_shared_tags.go
teresaromero: i followed the naming i saw.. but agree, i think we can merge into a single one 👍🏻

teresaromero: consolidated tag functions into new ´tag.go´ file

internal/export/tags.go
```diff
@@ -111,3 +121,40 @@ func filterOutFleetManagedTags(ctx *transformationContext, references []interfac
 	}
 	return newReferences, nil
 }
+
+// isSharedTag checks if the given tag ID is a shared tag for the specified package.
+// Shared tags ids are created by fleet form the tags.yml file in the package.
+// https://github.com/elastic/kibana/blob/5385f96a132114362b2542e6a44c96a697b66c28/x-pack/platform/plugins/shared/fleet/server/services/epm/kibana/assets/tag_assets.ts#L67
+func isSharedTag(aId string, packageName string) bool {
+	defaultSharedTagTemplate := "fleet-shared-tag-%s"
+	securitySolutionTagTemplate := "%s-security-solution"
```
jsoriano: I guess with the test cases you are adding we would detect changes on these patterns from Fleet, right?

jsoriano (approved):
👍

internal/export/tags.go
jsoriano: Well, if this pattern is changed in Fleet, the exported objects will change and I guess we will detect it there.
In any case no need to worry a lot about this, I guess this is not going to change, at least too frequently.

PR #3220, opened 2026-01-21
Allow to use custom URLs to retrieve ECS schemas

mrodm:
Part of https://github.com/elastic/elastic-package/issues/2993.
Follows https://github.com/elastic/elastic-package/pull/3114

Includes base URL for ECS field in the configuration file `~/.elastic-package/config.yml`.
Example:
```yaml
stack:
    image_ref_overrides: {}
profile:
    current: default
schema_urls:
  ecs_base: "https://raw.githubusercontent.com/elastic/ecs"
```

If not defined the `schema_urls` in that configuration file or this field is not set in the corresponding functions/structs, `elastic-package` will use the default URL `https://raw.githubusercontent.com/elastic/ecs`.

Commands updated to take into account the new setting from `~/.elastic-package/config.yml`:
- `elastic-package build`
- `elastic-package lint`
- `elastic-package install`
- `elastic-package test <assert|static|policy|pipeline|system|script>`

## Author's checklist
- [x] Ensure all commands requiring to download ECS schemas define the required configuration/parameters.
- [x] Test setting a different cache folder per operation: https://github.com/elastic/elastic-package/pull/3220/commits/d11ccbfb0e1d0739d3dec72302e2a197e361190e
- [x] Ensure to set the default URL value
- [x] Test removing the local cache files at `~/.elastic-package/cache/fields/ecs`
- [x] Update documentation
- [x] Remove debug changes
    - [x] https://github.com/elastic/elastic-package/pull/3220/commits/d11ccbfb0e1d0739d3dec72302e2a197e361190e 

- Edit `~/.elastic-package/config.yml` including:
```yaml
schema_urls:
  ecs_base: "https://raw.githubusercontent.com/mrodm/ecs"
```
- Remove the cache folder and run unit tests
```shell
rm -rf "~/.elastic-package/cache/fields/ && make test-go
```
- Run some tests and run tests for a test package:
```shell
rm -rf "~/.elastic-package/cache/fields/ && \
  go run github.com/elastic/elastic-package -C test/packages/parallel/nginx test system -v --data-streams access
```

### DIFF
--- a/internal/stack/serverless.go
+++ b/internal/stack/serverless.go
@@ -232,7 +232,6 @@ func (sp *serverlessProvider) createClients(project *serverless.Project, appConf
 		return fmt.Errorf("failed to create kibana client: %w", err)
 	}
 
-	logger.Tracef(">> Using package registry URL (TO BE REMOVED): %s", packageRegistryProxyToURL(sp.profile, appConfig))
 	sp.registryClient = registry.NewClient(appConfig.PackageRegistryBaseURL())
 
 	return nil


internal/docs/readme_test.go
```diff
@@ -51,6 +52,7 @@ Introduction to the package`,
 			expected:               "",
 		},
 	}
+	urls := fields.SchemaURLs{}
```
mrodm: These tests do not require any ecs schema URL because there is no external field to be resolved.

mrodm: > we can have X string defined as url, if the are no external fields, does it fail?

If there are no external fields defined in the package, `elastic-package` is not going to try to download any ECS schema.

About the other questions, I think there would not be any issue since the methods to get the ECS base URL take into account the scenario where the string is empty

https://github.com/elastic/elastic-package/pull/3220/changes#diff-87c6c2397b75252e87ef19b597b96a3973f47f86588782fd693b64ab3de079faR69-R75

Is that what you were referring to ?

cmd/build.go
```diff
@@ -78,6 +79,11 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 	}
 	logger.Debugf("Use build directory: %s", buildDir)
 
+	config, err := install.Configuration()
```
jsoriano: Nit. In https://github.com/elastic/elastic-package/pull/3232 `appConfig` is used as name for this variable. Should we use the same name in both places for consistency?

PR #3268, opened 2026-02-10
Fix issue that removed entries from lists in policy tests

jsoriano:
`memberReplace`, when used on slices, was removing them instead of replacing its members. Add explicit support for them.

Also fix a type check whose condition was reversed.

Add tests that reproduce the issues that this change solves.

Change prepared with assistance by copilot.


jsoriano:
test integrations

PR #3285, opened 2026-02-16
Fix Kibana data_stream.dataset variable for input package policies

### LINKED ISSUE #3275: Package policy dataset mismatch for otelcol input packages
## Summary

When elastic-package creates Fleet package policies for **input packages** whose policy template uses `input: otelcol`, the dataset value it sends does not match the value Fleet uses when the same integration is added via the Fleet UI. As a result, policy tests that create a policy via elastic-package and compare it to the downloaded policy fail, because the downloaded policy reflects Fleet’s compilation (e.g. `statsdreceiver`) while the test was written to expect that value, and elastic-package is currently sending a different value (e.g. `statsd_input_otel.statsdreceiver`).

## Background

- **OTel input packages** are packages with `type: input` and policy templates that declare a single `otelcol` `input`.
- For such packages, the **dataset** is sent to Fleet as part of the package policy (in the data stream and in vars such as `data_stream.dataset`). Fleet compiles and stores the policy; the **download agent policy API** returns the compiled policy.
- **Policy tests** in elastic-package create a Fleet policy (via the same APIs), download it, and compare the result to an expected file. The expected file is often generated or updated to match what the Fleet download API returns when the integration is added through the Fleet UI.

## Current behavior (wrong for otelcol)

- For every input package, elastic-package builds the dataset as **package name + policy template name** (e.g. `statsd_input_otel.statsdreceiver`), sends it in the package policy request, and Fleet stores/uses that value.
- For **otelcol** input packages, when the integration is added via the **Fleet UI**, Fleet compiles the dataset as **policy template name only** (e.g. `statsdreceiver`). The download API then returns that compiled value.

So for otelcol:

- **Fleet UI / download API:** dataset = `statsdreceiver` (policy template name only).
- **elastic-package–created policy:** dataset = `statsd_input_otel.statsdreceiver` (package name + policy template name).

The policy created by the policy test therefore differs from the policy created in the UI: same package, different dataset in the compiled agent policy. Once the expected file is updated to match Fleet (e.g. `statsdreceiver`), the test fails until elastic-package stops sending `statsd_input_otel.statsdreceiver` for that package.

Relevant code: https://github.com/elastic/elastic-package/blob/b7fb0463c604fa46e1741bc4c5d6e93731930597/internal/resources/fleetpolicy.go#L245-L310

StatsD Input Package sample event: https://github.com/elastic/integrations/blob/592c26bfe301286231d416606ca20d3f11de44c9/packages/statsd_input_otel/sample_event.json#L5-L9

Mismatched dataset value in expected policy: https://github.com/elastic/integrations/blob/592c26bfe301286231d416606ca20d3f11de44c9/packages/statsd_input_otel/_dev/test/policy/test-default.expected#L27-L32

## Expected behavior

- For input packages whose policy template uses **`input: otelcol`**: the dataset sent by elastic-package (and thus the compiled policy) should match Fleet’s behavior when the integration is added via the UI—i.e. dataset should be **policy template name only** (e.g. `statsdreceiver`), so that the download API and the expected file align and policy tests pass.
- For **other** input packages (non-otelcol): existing behavior may still be correct (dataset = package name + policy template name); that should be confirmed or left as-is depending on product expectations.

## Impact

- **Policy tests** for otelcol input packages (e.g. `statsd_input_otel`) fail when the expected policy file is set to Fleet’s compiled value (policy template name only), because elastic-package sends a different dataset.
- Consistency between “add via UI” and “add via elastic-package / policy test” is broken for otelcol packages.

## Additional context

Here's an example of Fleet's compiled policy for the statsd input package:

Image

teresaromero:
For input packages, createInputPackagePolicy now sets the stream-level variable `data_stream.dataset` to the policy template name (e.g. `sql_query`) instead of the full dataset string (e.g. `sql_input.sql_query`). Kibana expects this variable to hold the template name so it can correctly resolve the dataset; using the full string caused the dataset to be incorrect.

Add `TestCreateInputPackagePolicy_DatasetVariable` to assert:
- The variable is set to the policy template name when not defined by the template or user
- Template default and user override (`packagePolicy.Vars`) are preserved
- An error is returned when `DataStreamName` is set for an input package

System tests already use the correct behavior and were not modified in this PR.

- System test config is loaded from YAML with a top-level **`vars`** (and optional **`data_stream.vars`**). These are passed as `cfgVars` into the system runner (`internal/testrunner/runners/system/tester.go`).
- For **input packages**, the system runner already resolves the dataset as:
  1. **`vars["data_stream.dataset"]`** from the test config if set
  2. Otherwise the default from the policy template’s var definition
  3. Otherwise **`policyTemplate.Name`**
- So system tests already default to the **policy template name** and only override when the test config sets `data_stream.dataset` in `vars`. No code change was needed there; this PR only brings **policy tests** in line with that same behavior.

Resolves #3275 

Made with [Cursor](https://cursor.com)

### DIFF
--- a/test/packages/parallel/mongodb/_dev/deploy/docker/Dockerfile
+++ b/test/packages/parallel/mongodb/_dev/deploy/docker/Dockerfile
@@ -1,4 +1,4 @@
-ARG MONGODB_VERSION=5.0
-FROM mongo:${MONGODB_VERSION}
-RUN apt update && yes | apt install netcat
+ARG SERVICE_VERSION=7.0
+FROM mongo:${SERVICE_VERSION}
+RUN apt-get update && apt-get install -y netcat-openbsd
 HEALTHCHECK --interval=1s --retries=90 CMD nc -z localhost 27017
--- a/test/packages/parallel/mongodb/_dev/deploy/docker/creator/Dockerfile
+++ b/test/packages/parallel/mongodb/_dev/deploy/docker/creator/Dockerfile
@@ -1,6 +1,0 @@
-ARG SERVICE_VERSION=${SERVICE_VERSION:-5.0}
-FROM mongo:${SERVICE_VERSION}
-RUN sed -i "/jessie-updates/d" /etc/apt/sources.list
-RUN apt-get update && apt-get install -y netcat
-HEALTHCHECK --interval=1s --retries=90 \
-  CMD echo 'db.runCommand({serverStatus:1}).ok' | mongo
--- a/test/packages/parallel/mongodb/_dev/deploy/docker/docker-compose.yml
+++ b/test/packages/parallel/mongodb/_dev/deploy/docker/docker-compose.yml
@@ -1,16 +1,13 @@
 version: '2.3'
 services:
   mongodb:
-    build: creator
-    ports:
-      - 27017:27017
-  mongodb-log:
-    image: docker.elastic.co/integrations-ci/beats-mongodb:${MONGODB_VERSION:-5.0}-1
-    user: root
     build:
       context: .
-    volumes:
-      - ${SERVICE_LOGS_DIR}:/var/log/mongodb
-    entrypoint: >
-      bash -c "chmod a+wx /var/log/mongodb && chmod a+r -R /var/log/mongodb && touch /var/log/mongodb/mongod.log && chmod 644 /var/log/mongodb/mongod.log && mongod --replSet beats --logpath /var/log/mongodb/mongod.log --logappend"
-
+      args:
+        - SERVICE_VERSION=${SERVICE_VERSION:-7.0}
+    ports:
+      - 27017:27017
+    healthcheck:
+      test: ["CMD", "mongosh", "--eval", "db.runCommand({serverStatus:1}).ok"]
+      interval: 1s
+      retries: 90
--- a/test/packages/parallel/mongodb/_dev/build/docs/README.md
+++ b/test/packages/parallel/mongodb/_dev/build/docs/README.md
@@ -4,10 +4,7 @@ This integration is used to fetch logs and metrics from [MongoDB](https://www.mo
 
 ## Compatibility
 
-The `log` dataset is tested with logs from versions v3.2.11 and v4.4.4 in
-plaintext and json formats.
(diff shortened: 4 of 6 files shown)


teresaromero:
When this fix is released, is expected to have some policy tests failing. New snapshots should be generated with the version bump

test/packages/parallel/mongodb/_dev/deploy/docker/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     ports:
       - 27017:27017
   mongodb-log:
-    image: docker.elastic.co/integrations-ci/beats-mongodb:${MONGODB_VERSION:-5.0}-1
+    # image: docker.elastic.co/integrations-ci/beats-mongodb:${MONGODB_VERSION:-7.0}-1
```
teresaromero: test uses the image or builds from the given context dockerfile. Can we safely remove this reference to the beats image? 

cc @mrodm @jsoriano

test/packages/parallel/mongodb/_dev/deploy/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG MONGODB_VERSION=5.0
+ARG MONGODB_VERSION=7.0
```
mrodm: Should we use 6.0 instead to ensure it could run with the previous major ?


```suggestion
ARG MONGODB_VERSION=6.0
```

test/packages/parallel/mongodb/_dev/deploy/docker/creator/Dockerfile
```diff
@@ -1,6 +1,4 @@
-ARG SERVICE_VERSION=${SERVICE_VERSION:-5.0}
+ARG SERVICE_VERSION=${SERVICE_VERSION:-7.0}
```
mrodm: In case it is decided to use the previous major

```suggestion
ARG SERVICE_VERSION=${SERVICE_VERSION:-6.0}
```

teresaromero:
Moved all mongodb changes into https://github.com/elastic/elastic-package/pull/3288

Issue #3038, opened 2025-10-29
Use inputs as objects when creating package policies


jsoriano:
`elastic-package` is using a legacy arrays-based format when creating package policies (around [here](https://github.com/elastic/elastic-package/blob/fbe01ff32b62908a8bbc3dae26a4d7ffc5bd5528/internal/testrunner/runners/system/tester.go#L1915)).

According to the [documentation](https://www.elastic.co/docs/api/doc/kibana/operation/operation-get-fleet-package-policies#operation-post-fleet-package-policies-body-application-json), this format is deprecated and a new format based on objects should be used.

We should check since what version this new format exists, in case we need to support both formats for old versions.

Update: `elastic-package` is actually using both formats, there is a type `kibana.PackageDataStream` that uses the arrays format, while the `kibana.PackagePolicy` uses the object format. They have equivalent methods that at the end use the same `/package_policies` endpoints. See for example:

* https://github.com/elastic/elastic-package/blob/e84244e66b81db61c7a9cf1e5237b22338bd279c/internal/kibana/policies.go#L242
* https://github.com/elastic/elastic-package/blob/e84244e66b81db61c7a9cf1e5237b22338bd279c/internal/kibana/policies.go#L287

`kibana.PackageDataStream` type and related helpers should be removed in favor of `kibana.PackagePolicy`.

jsoriano:
New API cannot be used in all packages, because it was introduced between 8.0 and 8.3. We would need to drop support for packages supporting older versions.

We could do it if we assume that we can drop support for 7.x, but there can still be problems on the first 8.x versions.

Probably the best path forward is to try to use the new APIs, but build the body depending on the stack version. This will let us migrate some things, but would still require us to maintain some legacy code.

PR #3307, opened 2026-02-25
Use new format in calls to policy API

jsoriano:
Fixes https://github.com/elastic/elastic-package/issues/3038.

Use the recommended "simplified" format of the Fleet API to create policies. This format is based on objects instead of lists for variables, and automatically manages defaults.

To keep supporting old versions of the stack that don't support this API, a conversion layer is added. This conversion layer will need to be maintained while we support 8.5.0. 

A new setting is added to system tests to allow forcing the use of an specific format, `policy_api_format`. This is added as a escape hatch for cases that cannot be solved with the simplified API, as seen for example in https://github.com/elastic/kibana/issues/255976. Definition to package spec to be added in https://github.com/elastic/package-spec/pull/1103.

The new "simplified" API enables all data streams for the configured inputs by default, so some new code is needed to disable the data streams that are not relevant for each test case.

Refactor the code related to policy creation so it is reused in the different places where the functionality is needed. Two main methods are created for that, `BuildIntegrationPackagePolicy` for integration packages, and `BuildInputPackagePolicy` for input packages.

A couple of additional enhancements:
* Variables at the package level are supported now for input packages.
* Handling `use_apm`, co-authored by @mrodm, mostly copied from https://github.com/elastic/elastic-package/pull/3312.

New tests are added for the methods used to create package policies.

_Change assisted by Copilot/Claude_

### DIFF
--- a/internal/testrunner/runners/policy/testconfig.go
+++ b/internal/testrunner/runners/policy/testconfig.go
@@ -28,7 +28,7 @@ type testConfig struct {
 	// "" (auto-detect from Kibana version, default).
 	// Use "legacy" as a workaround when a Fleet simplified-API bug prevents the
 	// test from passing (e.g. select vars with "false"/"true" option values).
-	PolicyAPIFormat string `config:"policy_api_format" yaml:"policy_api_format,omitempty"`
+	PolicyAPIFormat string `config:"policy_api_format,omitempty" yaml:"policy_api_format,omitempty"`
 }
 
 func readTestConfig(testPath string) (*testConfig, error) {


jsoriano:
test integrations

jsoriano:
test integrations

jsoriano:
I am going to open this for review.

@elastic/ecosystem sorry for the big change, let me guide a little bit, as the changes are mostly reorganizations and not so extensive:
* `internal/kibana`:
  * `packagepolicy.go` contains the most significant part of this change. You can find there the actual code to generate policies, what is reused in other parts of the code.
  * `legacypolicy.go` contains the code to convert from the new objects to the legacy format. The structs there were moved from existing ones.
  * `packagepolicy_test.go` contains a new set of tests that compare the policies generated with golden files. I found this very useful to compare requests generated by Fleet with the requests generated with elastic-package.
  * `policies.go` most code there are the shared structs and the glue used for callers of the methods used to generate policies. Perhaps some of this code could be moved to `packagepolicy.go`.
* `internal/packages`:
  * `packages.go`, new helpers to obtain information from packages required by the methods used to create policies.
* `internal/resources`:
  * `fleetpolicy.go`, mostly removing code that is now shared from `internal/kibana`.
* `internal/testrunner/runners/system`:
  * `tester.go`, mostly removing code that is now shared from `internal/kibana`. Also removed the code to obtain expected datasets, as this is directly removed for the methods to create policies.

Other changes are to adapt the calls to the new interfaces.

internal/kibana/packagepolicy_test.go
```diff
@@ -0,0 +1,324 @@
...
+			inputVars:          common.MapStr{},
+			dsVars: common.MapStr{
+				"host":       "0.0.0.0",
+				"port":       2055,
+				"endace_url": "http://test.elastic.co",
+			},
+			goldenSimplified: "testdata/endace_netflow_pkg_var_in_dsvars.json",
+			goldenLegacy:     "testdata/endace_netflow_pkg_var_in_dsvars_legacy.json",
+		},
+		{
```
mrodm: Could you add a new unit test for an OTel input package defining traces as signal ? That would help to ensure that the `use_apm` variable is set even there is no variable in manifest nor set by the test config

I've tested locally and it works, but it would be helpful to add it here too to run as part of the tests.

There is an example of this package that you could use in https://github.com/elastic/elastic-package/pull/3312/changes#diff-6f7b1cb2d2b713ecbefd49fdffaef24b4c188e565d62d49e30bfb39e01e4a82f

This would require an stack 9.4.0-SNAPSHOT.

mrodm (approved):
Tested locally these changes and they worked successfully in different packages. Great coverage added to ensure correctness too!

Thanks!

PR #3331, opened 2026-03-03
Revert go-ucfg dependency to v0.8.8

mrodm:
This PR reverts the `go-ucfg` upgrade to v0.9.0 https://github.com/elastic/elastic-package/pull/3294

It keeps the version v0.8.0

The issue found is related to variables containing dots like `data_stream.dataset` that cannot be retrieve from the config.
Given this configuration file for system tests
```yaml
vars:
  data_stream.data_set: custom.foo
```

This call would return key not found (if err was validated):
https://github.com/elastic/elastic-package/blob/a3061de9f1e8eb9e1bc2700069adcdecbb08de8a/internal/testrunner/runners/system/tester.go#L2073-L2076

## Author's Checklist
- [x] Add tests to validate that variables with dots can be read (e.g. `data_stream.dataset`)
    - Failure with v0.9.0: https://buildkite.com/elastic/elastic-package/builds/7316 
    - Success with v0.8.8: https://buildkite.com/elastic/elastic-package/builds/7317


internal/testrunner/runners/system/testconfig_test.go
```diff
@@ -0,0 +1,173 @@
...
+  data_stream.dataset: other.name
+`), 0644)
+		require.NoError(t, err)
+
+		cfg, err := newConfig(configPath, servicedeployer.ServiceInfo{}, "")
+		require.NoError(t, err)
+		require.NotNil(t, cfg)
+
+		v, err := cfg.Vars.GetValue("data_stream.dataset")
+		require.NoError(t, err)
```
mrodm: This was not found in v0.9.0

```
Error: Received unexpected error: key not found
Test: TestNewConfig/vars_with_data_stream.dataset_are_detected 
```

PR #3337, opened 2026-03-04
render alert rule templates as collapsible tables

jakubgalecki0:
Render alert rule templates as collapsible table 

image


mrodm (commented):
To try to avoid being breaking change for the current packages that define alert rule templates in the integrations repository, I added some comments to suggest an approach to just apply the new rendering for those packages that use a spec version >= 3.6.0.

3.6.0 spec version is not released yet, but it will be soon. So for those packages updating to 3.6.0 , they will have that new rendering using collapsible tables.

internal/docs/alert_rule_template.go
```diff
@@ -11,8 +11,18 @@ import (
...
+	"github.com/Masterminds/semver/v3"
+
+	"github.com/elastic/elastic-package/internal/packages"
 )
 
+// alertRuleTemplatesCollapsibleTableMinSpecVersion is the minimum package spec
+// version from which alert rule templates are rendered as a collapsible table.
+// For earlier spec versions, the previous rendering (name and description as a
+// list) is preserved to avoid introducing a breaking change.
+var alertRuleTemplatesCollapsibleTableMinSpecVersion = semver.MustParse("3.6.0")
```
mrodm, line 24: As this package-spec has already been released, should we use now 3.7.0 ? @jsoriano @teresaromero 

I'm thinking that maybe some package using `format_version: 3.6.0` starts using `{{ alertRuleTemplates }}` helper in their documentation before a new elastic-package release is published with these changes.

mrodm (commented):
Thanks, LGTM!

But I’d like to get @jsoriano’s and @teresaromero’s opinion on which spec version should be the cutoff for rendering the new format. Taking into account that 3.6.0 is already published and is likely to be used by some packages before the next elastic-package release.

internal/docs/alert_rule_template.go
jsoriano: In principle yes, we should apply this only starting on a future version, like `3.7.0`.

But `3.6.0` has been recently released and there is no package using `alertRuleTemplates` with this format version yet, so maybe we can keep 3.6.0 here so this is adopted earlier.

teresaromero: i had to look back at package-spec for some context. i see this [was introduced](https://github.com/elastic/package-spec/blob/43bfae1ae31830d178cecfff099d88485b345ec6/spec/changelog.yml#L177) with 3.5.0, so packages from this version could be subject of using this syntax, isnt it? i don't see the difference between allowing it on 3.6.0 or 3.7.0 🤔

jsoriano: The syntax in the template effectively doesn't change, but the documentation generated from the template changes. This change happens in a way that cannot be controlled by users and would be detected in CI and produce a failure.

Including this logic based on the version we control what is generated, so for the same package source elastic-package generates the same documentation.
Only after updating the format version the documentation is generated with the new format, but this becomes a controlled change in the package.

mrodm: > But 3.6.0 has been recently released and there is no package using alertRuleTemplates with this format version yet, so maybe we can keep 3.6.0 here so this is adopted earlier.

Ok, let's keep 3.6.0 then :+1:

mrodm (approved):

muthu-mps (approved):

Issue #3380, opened 2026-03-16
Replace input packages in stream definitions of composable packages


jsoriano:
In the context of the composable packages project, integration packages can start depending on input packages.

When resolving these input packages on build time, all the information of the input package must be bundled into integration package streams, so the streams in a built composable package are indistinguishable from integration package streams.

In this context, https://github.com/elastic/elastic-package/issues/3278 and https://github.com/elastic/elastic-package/issues/3279 bundle some parts of it.

As a final step, remove the references to the input package from inputs and streams, and copy metadata from the input package:
* In the input in the main manifest, set at least `type`, `title` and `description` from the information in the input package.
* In the stream in the data stream manifest, set the `type` and `title`.
* In both cases remove the `package` reference.

Issue #3280, opened 2026-02-12
Bundle field definitions of required input packages


jsoriano:
In the context of the composable packages project, integration packages are going to start depending on input packages.

When building integration packages with dependencies, if the dependencies include field definitions, they should be merged with the fields in the data streams that use it.

When a field is defined in both packages, the definition in the integration package should be used, and the one in the input package should be discarded.

Issue #3279, opened 2026-02-12
Merge list of variables when building packages with required input packages


jsoriano:
In the context of the composable packages project, integration packages are going to start requiring other packages.

When they require input packages, as part of policy templates and data streams, the variables of both packages must be bundled into the integration package.

Merging policy template variables
----
When using composable packages, that depend on input packages, there are three places where variables can be defined:
* Policy templates in the input packages.
* Policy templates in the composable package main manifest.
* Data stream manifests in the composable package.

After resolving the variables, there will be two lists of them, a list of data stream variables, and a list of input variables. As with integration packages, data stream variables are specific to each data stream, while input variables are used for all the data streams with the same input type.

In order to resolve the final definitions for a given policy template, the following steps should be followed:
* Obtain the variables from the policy template of the input package. These variables are considered data stream variables.
* Obtain the variables from the policy template in the composable package. These variables are considered input variables.
  * If any of them was defined in the policy template of the input package, the variable is moved from data stream to input variables, and merged with the one in the composable package.
* Obtain the variables from the data stream manifest in the composable package. Merge these variables with the remaining data stream variables.
* Check both lists of variables. Fail if there are multiple variables with the same name.

The merging process should always use the variables in the input package as base, and override with the defined attributes in the composable packages. Variable definitions in composable packages don't need to include all attributes, only the name is required.

Once both lists are compiled, during build time, elastic-package will replace the defined lists of variables with the compiled ones for inputs and data streams.

Issue #3278, opened 2026-02-12
Bundle templates of required input packages


jsoriano:
In the context of the composable packages project, integration packages can start depending on input packages.

On build time, the templates of the referenced input packages must be bundled in the integration package, by copying it to the directory of the data streams where it is referenced.

Apart from that, if the integration is also defining a template, both templates must be listed in the corresponding `template_paths`, to indicate the rendering order. The templates in the input packages must be placed first, and the templates in the input package later, taking precedence.

PR #3459, opened 2026-04-14
Support building integration packages with required input dependencies

### LINKED ISSUE #3277: Support building integration packages with dependencies
As part of the composable packages project, integration packages can start including dependencies, see https://github.com/elastic/package-spec/issues/1083.

The way of handling these dependencies is going to depend on the type of packages.
Content packages will be installed along with the integration package, and won't require any handling in build time.
On the other hand, input packages will be bundled into the integration package, and some processing will be needed.

teresaromero:
## Summary

Composable (integration) packages can declare **`requires.input`** so that, at **build time**, elastic-package downloads those **input packages** from the Elastic Package Registry (or the configured `package_registry.base_url`), then updates the built package: **agent templates** are bundled with Fleet-correct `template_paths` order, **variables** are merged into the integration and data stream manifests, **field definitions** are merged into data stream `fields/`, and **`package:` references** on policy inputs and data stream streams are replaced with the **concrete input type** from the resolved input package. This aligns with the composable packages direction in [package-spec#1083](https://github.com/elastic/package-spec/issues/1083).

## What ships (sub-issues / phases)

1. **Agent template bundling** — [#3278](https://github.com/elastic/elastic-package/issues/3278): download required input packages, copy policy and data stream templates, order `template_paths` (input templates first, integration last).
2. **Variable merge** — [#3279](https://github.com/elastic/elastic-package/issues/3279): merge variable definitions from input packages into the composable manifest and data stream manifests (promotion, overrides, duplicates).
3. **Field bundling** — [#3280](https://github.com/elastic/elastic-package/issues/3280): copy field definitions from input packages into built data streams; integration-defined fields win.
4. **Stream / input resolution** — [#3380](https://github.com/elastic/elastic-package/issues/3380): replace `package:` with resolved `type:` / `input:` so the built output matches hand-authored packages.

## Implementation

- **`internal/requiredinputs/`** — `RequiredInputsResolver` implements `Bundle(buildPackageRoot)` (nil resolver is a no-op). Pipeline: templates → variables → fields → stream resolution.
- **`internal/registry/`** — `DownloadPackage`, TLS options, tests.
- **`internal/files/`** — optional detached-signature verification (env-configured, off by default).
- **`internal/packages/`** — manifest types for `requires.input`.
- **`internal/builder/packages.go`** — wires `RequiredInputsResolver` into the build.
- **Commands** — profile-aware registry base URL for **install**, **test**, **benchmark**, and testrunner **script** paths so behavior matches install.

## How to test

```bash
go test ./...
make build format lint licenser gomod update
```

- Closes https://github.com/elastic/elastic-package/issues/3277


teresaromero:
# Review Comments Addressed

This PR consolidates the work from #3329 and addresses all review comments. Summary below.

---

## @jsoriano

**File naming — no underscores in Go filenames**
The builder code was fully refactored into a new `internal/requiredinputs/` package with clean filenames (`requiredinputs.go`, `copy.go`, `fields.go`, `policytemplates.go`, `streams.go`, `streamdefs.go`, `variables.go`, `yamlutil.go`).

**Rename `processDSManifest`**
Renamed to `processDataStreamManifest` in `internal/requiredinputs/streams.go`.

**Rename `resolveInputPackage`**
Renamed to `mapRequiredInputPackagesPaths` in `internal/requiredinputs/requiredinputs.go`, which better reflects that it downloads and maps multiple packages.

**Reuse core logic from `ReadAllPackageManifests`**
Introduced `openPackageFS()` as a shared helper (handles both directory and zip packages) and reused `packages.ReadDataStreamManifestBytes` / `packages.ReadPackageManifestBytes` across `copy.go`, `fields.go`, and `streams.go`.

**Variable naming matching YAML key**
`PackageDependency.Package` now uses `yaml:"package"` tag in `internal/packages/packages.go`.

**YAML format preservation vs. struct approach**
Tried the struct approach — marshalling caused data loss (comments, ordering, unknown fields). Keeping the `yaml.Node` tree manipulation to preserve the manifest faithfully. The built package manifest is modified in-place with targeted key upserts/removals.

**Use `slices` package instead of custom remove helper**
`removeKey` in `yamlutil.go` uses `slices.IndexFunc` + `slices.Delete(idx, idx+2)`. Using `slices.DeleteFunc` directly is not applicable here because YAML mapping nodes store key-value pairs as _two consecutive_ elements in `Content`; you need to delete both atomically, which requires the index-based `Delete(idx, idx+2)` form.

**Zip reader as filesystem instead of extracting**
`DownloadPackage` in `internal/registry/client.go` now saves the zip file as-is (no extraction). `openPackageFS()` in `requiredinputs.go` opens it with `zip.OpenReader` + `fs.Sub`, presenting it as an `fs.FS` to all consumers.

**Content packages in dependency docs**
`docs/howto/dependency_management.md` now covers both input packages (resolved at build time) and content packages (resolved at runtime by Fleet).

**Higher-level abstraction in `factory.go`**
`installer.Options.RequiredInputsResolver` is typed as the `requiredinputs.Resolver` interface. `NoopRequiredInputsResolver` is used where no bundling is needed. The TODO for local-path override is noted in the struct comment.

**Local cache for packages** — deferred to a follow-up as agreed.

**Converge local registry serving methods** — noted; left as a longer-term architectural refactor.

**Methods with `PackageRoot` as struct members** — noted; left for a future refactor.

---

## @mrodm

**EPR URL should account for profile settings**
`stack.PackageRegistryBaseURL(profile, appConfig)` is now a public function in `internal/stack/registry.go` and is called from `cmd/install.go`. The build command does not use a profile, so it correctly reads from application config only.

**Docker image reference in docs**
Updated to `docker.elastic.co/package-registry/package-registry:v1.37.0` in `docs/howto/local_package_registry.md`.

**Port conflict warning**
`local_package_registry.md` now explicitly warns to use a port other than `8080` to avoid conflicting with `elastic-package stack up`.

**`os.RootFS` vs `os.DirFS`**
`os.OpenRoot()` is used in `requiredinputs.go` (both for the build root and for directory-based input packages). `root.FS()` returns a sandboxed `fs.FS` that blocks symlink traversal outside the root directory.

**Linked files / symlinks in policy templates**
All file reads go through `fs.ReadFile(inputPkgFS, ...)` on the FS returned by `openPackageFS`. For directory packages `os.OpenRoot` prevents escaping the package directory. For downloaded packages the FS is derived from `zip.OpenReader`, which has no real symlinks.

**Package signature validation**
PGP signature verification is implemented in `internal/registry/client.go`. When `ELASTIC_PACKAGE_VERIFY_PACKAGE_SIGNATURE=true`, the client downloads the detached `.sig` file and verifies it via `files.VerifyDetachedPGP()` before returning the package path. The verifier public key is configurable via `ELASTIC_PACKAGE_VERIFIER_PUBLIC_KEYFILE`.

---

## Known follow-ups (not in scope for this PR)

These were flagged by @jsoriano during testing and agreed to be addressed separately:

- Input keys in generated package policies are incomplete (missing input type).
- When multiple OTel input packages declare `otelcol`, the same input appears multiple times in a policy, confusing the Fleet UI. https://github.com/elastic/ingest-dev/issues/7380

internal/registry/client.go
```diff
@@ -5,32 +5,76 @@
...
 // NewClient creates a new instance of the client.
-func NewClient(baseURL string) *Client {
-	return &Client{
-		baseURL: baseURL,
+func NewClient(baseURL string, opts ...ClientOption) *Client {
+	c := &Client{baseURL: baseURL}
+	for _, opt := range opts {
+		opt(c)
+	}
+	c.httpClient, _ = c.newHTTPClient()
```
jsoriano: Error here is silently ignored, it should be propagated. If we think that errors in `certs.SystemPoolWithCACertificate` can be ignored, please ignore the error explicitly when calling the method, and add an explaining comment.

internal/requiredinputs/streamdefs.go
```diff
@@ -0,0 +1,191 @@
...
+	m, err := packages.ReadPackageManifestBytes(manifestBytes)
+	if err != nil {
+		return inputPkgInfo{}, fmt.Errorf("parsing manifest: %w", err)
+	}
+
+	if len(m.PolicyTemplates) == 0 {
+		return inputPkgInfo{}, fmt.Errorf("input package %q has no policy templates", m.Name)
+	}
+	if len(m.PolicyTemplates) > 1 {
+		logger.Debugf("Input package %q has multiple policy templates; using input identifier %q from first", m.Name, m.PolicyTemplates[0].Input)
```
jsoriano: This should be probably printed at the Warning level, or even fail. We want to catch these cases as I think we are not supporting input packages with multiple policy templates yet.

internal/requiredinputs/yamlutil.go
```diff
@@ -0,0 +1,78 @@
...
+	"slices"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/elastic/elastic-package/internal/formatter"
+)
+
+// mappingValue returns the value node for the given key in a YAML mapping node,
+// or nil if the key is not present.
+func mappingValue(node *yaml.Node, key string) *yaml.Node {
```
jsoriano: We have `internal/yamledit`, that has methods for getting, setting, deleting and printing yaml nodes. They could likely be reused here.

internal/requiredinputs/variables.go
```diff
@@ -0,0 +1,487 @@
...
+// fields win when explicitly specified.
+//
+// Input-level vars: vars declared in policy_templates[].inputs[].vars are
+// "promoted" — they become input-level variables in the merged manifest.
+//
+// Data-stream-level vars: all remaining (non-promoted) base vars are placed at
+// the data-stream level, merged with any stream-level overrides the composable
+// package declares.
+//
+//nolint:gocognit // multi-step merge pipeline (promotion, DS manifests, policy templates)
```
jsoriano: This method is really too complex, there are already some steps defined, could they be split on different methods?

internal/requiredinputs/variables.go
```diff
@@ -0,0 +1,487 @@
...
+	manifestBytes, err := buildRoot.ReadFile("manifest.yml")
+	if err != nil {
+		return fmt.Errorf("reading manifest: %w", err)
+	}
+	var doc yaml.Node
+	if err := yaml.Unmarshal(manifestBytes, &doc); err != nil {
+		return fmt.Errorf("parsing manifest YAML: %w", err)
+	}
+
+	// Step B — Build a promotedIndex: (pkg, dsName) → map[varName]overrideNode.
```
jsoriano: What is a "promoted index"? Is it an index with variables that have been promoted?

cmd/build.go
```diff
@@ -84,15 +90,25 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("can't load configuration: %w", err)
 	}
 
+	baseURL := appConfig.PackageRegistryBaseURL()
```
jsoriano: Other commands seem to use `stack.PackageRegistryBaseURL` to get the registry URL. Should we do the same here?

internal/requiredinputs/requiredinputs.go
```diff
@@ -0,0 +1,176 @@
...
+// It returns a map of package name to zip path.
+func (r *RequiredInputsResolver) mapRequiredInputPackagesPaths(manifestInputRequires []packages.PackageDependency, tmpDir string) (map[string]string, error) {
+	inputPkgPaths := make(map[string]string, len(manifestInputRequires))
+	errs := make([]error, 0, len(manifestInputRequires))
+	for _, inputDependency := range manifestInputRequires {
+		if _, ok := inputPkgPaths[inputDependency.Package]; ok {
+			// skip if already downloaded
+			continue
+		}
+		path, err := r.eprClient.DownloadPackage(inputDependency.Package, inputDependency.Version, tmpDir)
```
jsoriano, line 128: Would we have any way to resolve local packages? I remember we discussed this in the past, but I don't remember if we decided to do this in a follow up. I think the complicated part was to build the required input package.

It would be useful to allow developing changes in related composable and input packages that live in the same repository.

internal/files/verify.go
```diff
@@ -0,0 +1,68 @@
...
+
+// PackageSignatureVerificationFromEnv reports whether detached PGP verification should run
+// for registry package downloads. When verify is true, publicKeyPath is the path to an
+// armored public key and has been checked for existence. A non-nil err means the environment
+// is inconsistent (e.g. verify enabled but no key path or inaccessible file).
+func PackageSignatureVerificationFromEnv() (verify bool, publicKeyPath string, err error) {
+	raw := os.Getenv(verifyPackageSignatureEnv)
+	if raw == "" {
+		return false, "", nil
+	}
```
mrodm: As this process is going to be used to build the packages (downloading other packages from EPR) and publish those into package-registry (https://buildkite.com/elastic/integrations-publish), would it be possible to be enabled by default this verification ? So, the env variable could be like `ELASTIC_PACKAGE_DISABLE_VERIFY_PACKAGE_SIGNATURE`.

If so, probably it would be needed to ensure that the integrations repository (and elastic-package repo?) has its own copy of the public gpg key in a known location?

Or maybe, bundle the elastic public gpg key into elastic-package but allow to use other via an environment variable.

WDYT @teresaromero  @jsoriano ?

internal/registry/client.go
```diff
@@ -83,5 +139,43 @@ func (c *Client) DownloadPackage(name, version, destDir string) (string, error)
...
+		return "", fmt.Errorf("downloading package signature %s-%s: unexpected status code %d", name, version, sigCode)
+	}
+
+	zipFile, err := os.Open(zipPath)
+	if err != nil {
+		_ = os.Remove(zipPath)
+		return "", fmt.Errorf("opening downloaded package zip %s: %w", zipPath, err)
+	}
+	verifyErr := files.VerifyDetachedPGP(zipFile, sigBody, pubKey)
+	closeErr := zipFile.Close()
```
mrodm: Why not using `defer zipFile.Close()` after opening the zip file 

```
	zipFile, err := os.Open(zipPath)
	if err != nil {
		_ = os.Remove(zipPath)
		return "", fmt.Errorf("opening downloaded package zip %s: %w", zipPath, err)
	}
	defer zipFile.Close()
```

internal/registry/client.go
```diff
@@ -83,5 +139,43 @@ func (c *Client) DownloadPackage(name, version, destDir string) (string, error)
...
+	if verifyErr != nil {
+		_ = os.Remove(zipPath)
```
mrodm: Not tested, but maybe it could be used a defer function to ensure that the zip file path is removed/deleted if there is any error.

```golang
defer func() {
  if err != nil {
    _ = os.Remove(zipPath)
  }
}()
```

At least `verifyErr` should be renamed to `err` if this is implemented.

internal/requiredinputs/fields_test.go
```diff
@@ -0,0 +1,314 @@
...
+		},
+	}
+	resolver := NewRequiredInputsResolver(epr)
+
+	err = resolver.Bundle(buildPackageRoot)
+	require.NoError(t, err)
+
+	bundledPath := filepath.Join(buildPackageRoot, "data_stream", "ci_composable_logs", "fields", "ci_input_pkg-fields.yml")
+	_, statErr := os.Stat(bundledPath)
+	assert.True(t, os.IsNotExist(statErr), "bundled fields file should not be created when all fields already exist")
```
mrodm:
```suggestion
	assert.True(t, errors.Is(statErr, os.ErrNotExist), "bundled fields file should not be created when all fields already exist")
```

internal/requiredinputs/fields_test.go
```diff
@@ -0,0 +1,314 @@
...
+          - /var/log/ci/*.log
+`), 0644))
+
+	resolver := NewRequiredInputsResolver(epr)
+	err := resolver.Bundle(buildPackageRoot)
+	require.NoError(t, err)
+
+	bundledPath := filepath.Join(buildPackageRoot, "data_stream", "ci_composable_logs", "fields", "no_fields_pkg-fields.yml")
+	_, statErr := os.Stat(bundledPath)
+	assert.True(t, os.IsNotExist(statErr), "no fields file should be created when input package has no fields")
```
mrodm: Nit.
```suggestion
	assert.True(t, errors.Is(statErr, os.ErrNotExist), "no fields file should be created when input package has no fields")
```

scripts/test-build-install-zip-file.sh
```diff
@@ -116,16 +137,32 @@ for d in test/packages/*/*/; do
...
+echo "--- Phase-2 build: composable integration"
+elastic-package build -C "${COMPOSABLE_INTEGRATION_DIR}"
+
 # Remove unzipped built packages, leave .zip files
 rm -r build/packages/*/
 
-if [ ${USE_SHELLINIT} -eq 1 ]; then
-  eval "$(elastic-package stack shellinit)"
-else
+if [ ${USE_SHELLINIT} -eq 0 ]; then
```
mrodm: With this change even if `USE_SHELLINIT` is zero, it has been already loaded the environment variables in line 148.

It would be better to try not to load the environment variables to ensure that just the ones defined in this "if" are the ones defined.

scripts/test-build-install-zip-file.sh
```diff
@@ -116,16 +137,32 @@ for d in test/packages/*/*/; do
...
+  PREV_REGISTRY_URL=$(yq '.package_registry.base_url // ""' "${ELASTIC_PACKAGE_CONFIG_FILE}")
+  yq eval --inplace '.package_registry.base_url = "https://127.0.0.1:8080"' "${ELASTIC_PACKAGE_CONFIG_FILE}"
+else
+  mkdir -p "$(dirname "${ELASTIC_PACKAGE_CONFIG_FILE}")"
+  yq -n '.package_registry.base_url = "https://127.0.0.1:8080"' > "${ELASTIC_PACKAGE_CONFIG_FILE}"
+fi
+PACKAGE_REGISTRY_CI_OVERRIDE=1
+
+echo "--- Phase-2 build: composable integration"
+elastic-package build -C "${COMPOSABLE_INTEGRATION_DIR}"
```
mrodm: Would it be better to create a new Buildkite step for these packages ? 

If a new step is added , it could be added some flag in the script to execute these commands. Ensuring that the other packages are tested with the same local stack scenario.

internal/registry/client.go
```diff
@@ -83,5 +139,43 @@ func (c *Client) DownloadPackage(name, version, destDir string) (string, error)
 	if err := os.WriteFile(zipPath, body, 0o644); err != nil {
 		return "", fmt.Errorf("writing package zip to %s: %w", zipPath, err)
 	}
+
+	if !verify {
+		return zipPath, nil
+	}
```
jsoriano: Nit. Move verification to its own method.
```suggestion
	if verify {
		err := c.verifyPackage(pubKeyPath, name, version, zipPath)
		if err != nil ...
	}
	
	return zipPath, nil
```

internal/requiredinputs/requiredinputs.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	return inputPkgPaths, errors.Join(errs...)
+}
+
+// openPackageFS returns an fs.FS rooted at the package root (manifest.yml at
+// the top level) and a close function that must be called when done. For
+// directory packages it closes the os.Root; for zip packages it closes the
+// underlying zip.ReadCloser.
+func openPackageFS(pkgPath string) (fs.FS, func() error, error) {
```
jsoriano, line 147: We might support only packages in zip format.

By now we are only supporting packages from the registry, that will come in this format. If we support local packages we can build them too.

internal/requiredinputs/streamdefs.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// resolveStreamInputTypes replaces all package: <pkg-name> references in the
+// composable package's manifest.yml (policy_templates[].inputs) and in each
+// data_stream/*/manifest.yml (streams[]) with the actual input type identifier
+// from the referenced input package, then removes the package: key.
+//
+// This step must run last, after mergeVariables, because that step uses
+// stream.Package and input.Package to identify which entries to process.
+func (r *RequiredInputsResolver) resolveStreamInputTypes(
```
jsoriano, line 34: This method has multiple steps, that can probably be split in smaller methods.

internal/requiredinputs/streamdefs.go
```diff
@@ -0,0 +1,191 @@
...
+				return fmt.Errorf("input package %q referenced in policy_templates[%d].inputs[%d] not found in required inputs", input.Package, ptIdx, inputIdx)
+			}
+
+			inputNode, err := getInputMappingNode(&doc, ptIdx, inputIdx)
+			if err != nil {
+				return fmt.Errorf("getting input node at pt[%d].inputs[%d]: %w", ptIdx, inputIdx, err)
+			}
+
+			upsertKey(inputNode, "type", &yaml.Node{Kind: yaml.ScalarNode, Value: info.identifier})
+
```
jsoriano, line 96: We will need to set a unique name too, and use the corresponding name in the data stream. But we can do this as a follow up.

Related to https://github.com/elastic/ingest-dev/issues/7380.

jsoriano (commented):
Tested with the package in https://github.com/elastic/elastic-package/pull/3432 and it seems to work.

As the change is pretty big, and it seems to work, maybe we can fix the most urgent things, merge it, and leave the rest for smaller follow ups.

internal/requiredinputs/requiredinputs.go
teresaromero: it is pending, still. i will open an issue to address this as a follow up and keep track

internal/requiredinputs/streamdefs.go
teresaromero: I will open a followup issue to address this

internal/requiredinputs/streamdefs.go
teresaromero: we are not indeed, and this should never appear. its a guard for that happening

internal/requiredinputs/variables.go
teresaromero: the "promoted" concept is a variable declared at the composable(integration) that is requiring the input, with that variable. so it brings the variable from the input or overrides its value if the composable declares a value

scripts/test-build-install-zip-file.sh
teresaromero: I've extracted the tests into its own step

internal/requiredinputs/streamdefs.go
jsoriano: Yeah, as this is not supposed to happen I'd say that we should be noisier when it does.

So we don't silently produce unexpected situations if a developer occurs to start using multiple policy templates, or if we start supporting it.

cmd/build.go
```diff
@@ -84,15 +90,28 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("can't load configuration: %w", err)
 	}
 
+	prof, err := profile.LoadProfile(appConfig.CurrentProfile())
+	if err != nil {
+		return fmt.Errorf("could not load profile: %w", err)
+	}
+	baseURL := stack.PackageRegistryBaseURL(prof, appConfig)
```
teresaromero, line 97: @jsoriano with the use of profile at the build command, we are coupling the stack with build. the command itself does not have the flag to select profile.

internal/files/verify.go
teresaromero: >it would be needed to ensure that the integrations repository (and elastic-package repo?) has its own copy of the public gpg key in a known location

after investigating the integrations repo:
integrations repository does not store any gpg key. this is delegated to a custom service that is centralized in buildkite. when the package is [built](https://github.com/elastic/integrations/blob/b47ae059e315c5e869c62225583a17fb9283cf7c/.buildkite/scripts/build_packages.sh#L131) it is sent to the signing service and then the artifact is stored with the `.sig` file. 

> Or maybe, bundle the elastic public gpg key into elastic-package

I think that storing the key "hard-copy" at the repo can cause some trouble, when for example, the key is rotated.

overall, I think this process is not trivial, as we are injecting the download+verification dependency into the build process. 
Security-wise, would this be a problem to leave the download part without signature verification? 
I mean this to address this on a follow-up issue and keep the download package step without the verification. If ok i would revert the verification code and move it to a follow-up where we can discuss a  better aproach cc @mrodm @jsoriano

internal/requiredinputs/requiredinputs.go
teresaromero: https://github.com/elastic/elastic-package/issues/3464

internal/files/verify.go
jsoriano: Agree with moving verification out of this PR. We can merge the basic functionality without this.

cmd/build.go
jsoriano: We need it to support the local package registry, right?

What it is true is that it should remain working even if there is no stack started, this should be optional.

internal/requiredinputs/streamdefs.go
teresaromero: https://github.com/elastic/elastic-package/issues/3465

internal/files/verify.go
teresaromero: https://github.com/elastic/elastic-package/issues/3466

cmd/build.go
teresaromero: it does, the fallback using this function is stack -> profile -> prod; so the variable is set regardless and if the build does not require download, it just does not use it

internal/testrunner/runners/system/runner.go
```diff
@@ -295,11 +296,12 @@ func (r *runner) Type() testrunner.TestType {
...
-			Absent:         !opts.installedPackage,
-			Force:          opts.installedPackage, // Force re-installation, in case there are code changes in the same package version.
-			RepositoryRoot: r.repositoryRoot,
-			SchemaURLs:     r.schemaURLs,
+			PackageRoot:            r.packageRoot,
+			Absent:                 !opts.installedPackage,
+			Force:                  opts.installedPackage, // Force re-installation, in case there are code changes in the same package version.
+			RepositoryRoot:         r.repositoryRoot,
+			SchemaURLs:             r.schemaURLs,
+			RequiredInputsResolver: &requiredinputs.NoopRequiredInputsResolver{},
```
mrodm, line 304: To run the system tests of a composable package, this package should be installed via this code as a requisite. I guess that if needed the package should be bundled before being installed, is that right ? Is it intended to be done this in a follow-up PR ?

I see that for policy tests there is a parameter in the runner `r.requiredInputsResolver`.

I think it would be needed to do something similar as in policy tests here to be able to build the composable packages in system tests.

internal/testrunner/runners/asset/tester.go
```diff
@@ -88,11 +89,12 @@ func (r *tester) Run(ctx context.Context) ([]testrunner.TestResult, error) {
...
-			Absent:         !installedPackage,
-			Force:          installedPackage, // Force re-installation, in case there are code changes in the same package version.
-			RepositoryRoot: r.repositoryRoot,
-			SchemaURLs:     r.schemaURLs,
+			PackageRoot:            r.packageRoot,
+			Absent:                 !installedPackage,
+			Force:                  installedPackage, // Force re-installation, in case there are code changes in the same package version.
+			RepositoryRoot:         r.repositoryRoot,
+			SchemaURLs:             r.schemaURLs,
+			RequiredInputsResolver: &requiredinputs.NoopRequiredInputsResolver{},
```
mrodm, line 97: As mentioned for system tests, as this tests require to install the package it looks to me that it would be needed to be able to use a `r.requiredInputsResolver` parameter as in policy tests. WDYT ? Maybe it is expected to be done in a follow-up ?

internal/testrunner/script/package.go
```diff
@@ -124,10 +127,11 @@ func removePackage(ts *testscript.TestScript, neg bool, args []string) {
...
 	_, err = m.ApplyCtx(ctx, resources.Resources{&resources.FleetPackage{
-		PackageRoot:    pkgRoot,
-		Absent:         true,
-		Force:          true,
-		RepositoryRoot: root, // Apparently not required, but adding for safety.
+		PackageRoot:            pkgRoot,
+		Absent:                 true,
+		Force:                  true,
+		RepositoryRoot:         root, // Apparently not required, but adding for safety.
+		RequiredInputsResolver: &requiredinputs.NoopRequiredInputsResolver{},
```
mrodm, line 134: As `Absent` is set to true here, I think it could be set here the Noop struct

internal/testrunner/script/package.go
```diff
@@ -71,11 +73,12 @@ func addPackage(ts *testscript.TestScript, neg bool, args []string) {
...
-		Absent:         false,
-		Force:          true,
-		RepositoryRoot: root,
-		SchemaURLs:     fields.NewSchemaURLs(fields.WithECSBaseURL(ecsBaseSchemaURL)),
+		PackageRoot:            pkgRoot,
+		Absent:                 false,
+		Force:                  true,
+		RepositoryRoot:         root,
+		SchemaURLs:             fields.NewSchemaURLs(fields.WithECSBaseURL(ecsBaseSchemaURL)),
+		RequiredInputsResolver: &requiredinputs.NoopRequiredInputsResolver{},
```
mrodm, line 81: Would it require here to set an inputs resolver created beforehand as in policy tests?

internal/testrunner/runners/asset/tester.go
teresaromero: i've narrowed the implementation of the bundling on build of the composable package. The noop resolver is here as a placeholder but it should be resololved in the followup, probably here https://github.com/elastic/elastic-package/issues/3464 ; when we also enable the overriding of the depencency

internal/testrunner/script/package.go
teresaromero: as mentioned on the comment above, this is a placeholder to address when testing is enabled for composable

internal/testrunner/script/package.go
teresaromero: same as above

internal/testrunner/runners/system/runner.go
teresaromero: i wont be addressing the testing of composable packages on this pr. just the bundling (build); following this comment https://github.com/elastic/elastic-package/pull/3459#discussion_r3093495118

internal/requiredinputs/yamlutil.go
jsoriano: I see you have migrated the code to `goccy/go-yaml`. Did you consider the higher-level yamledit? We can also refactor this in a follow up, or look at how we can reuse helpers among the different cases of yaml edition.

teresaromero: I migrated to align with the library used but the methods where not common, so still there are "custom" methods used on the requiredinputs package. I think we could use some refactor for handling the yaml edits. I will open an issue for that too.

internal/testrunner/runners/system/runner.go
mrodm: Ok! Agreed, it makes sense to address that in a follow-up :+1:

PR #3467, opened 2026-04-15
Fix policy tests failing on 9.4.0+ due to beatsauth OTel fields injected by Fleet

### LINKED ISSUE #3463: Policy tests fail on 9.4.0+ due to new beatsauth OTel fields injected by Fleet
Since https://github.com/elastic/kibana/pull/259308 (present in 9.4.0-SNAPSHOT onwards), Fleet injects a new `beatsauth` authentication extension into every OTel policy that uses an Elasticsearch output. This     
causes policy tests to fail when run against a 9.4.0+ stack because the downloaded policy contains fields that existing fixtures don't include.                                                                 
                                                                                                                                                                                                                
The three new fields injected are:

```yaml
# 1. auth reference added to the elasticsearch exporter
exporters:
  elasticsearch/componentid-0:
    auth:                                                                                                                                                                                                       
      authenticator: beatsauth/default
                                                                                                                                                                                                                
# 2. new top-level extension with deployment-specific SSL data
extensions:
  beatsauth/componentid-0:
    ssl:
      ca_trusted_fingerprint: 2611966E95E1F6B5A4C0A3094B8BFBB83B7AD1A72F195C35A54432C8B563A73A # <dynamic hash>
                                                                                                                                                                                                                
# 3. extension registered in service
service:                                                                                                                                                                                                        
  extensions:   
    - beatsauth/componentid-0
```

Packages tested against older stacks (e.g. 9.2.0) don't have these fields, so fixtures created on older stacks fail on newer ones and vice versa.                                                               
 
A concrete example of this breakage is visible in https://github.com/elastic/integrations/pull/18321.                                                                                                                                   
                
## Proposed direction

These fields are deployment-specific (the `ca_trusted_fingerprint` changes per installation, and the auth wiring is controlled by Fleet, not the package) and should be stripped by `cleanPolicy` before comparison,
following the existing pattern in `policyEntryFilters`.
                                                                                                                                                                                                                
- `exporters[*].auth` — can be filtered with the existing `mapValues` mechanism (add `{name: "auth"}` to the exporters filter).                                                                                       
- `extensions` keys matching `beatsauth/*` — requires a new capability to remove map entries by key pattern.
- `service.extensions` list entries matching `beatsauth/*` — requires a new capability to remove list elements by value pattern (or by prefix).                                                                     
                
The existing `onlyIfEmpty`/`ignoreValues` mechanism is close but not sufficient for pruning specific entries from a list while keeping others.

## Related issues
- https://github.com/elastic/integrations/pull/18321                                                              
- https://github.com/elastic/kibana/pull/259308

---

_Generated with the assistance of Claude_

mrodm:
Since https://github.com/elastic/kibana/pull/259308 (present in 9.4.0+), Fleet injects a `beatsauth` authentication extension into every OTel policy that uses an Elasticsearch output. This caused policy tests to fail when run against a 9.4.0+ stack because the downloaded policy contains fields that existing fixtures don't include.

The three new fields injected are:

- `exporters[*].auth` — authentication reference added to each exporter
- `extensions[beatsauth/*]` — new top-level extension with deployment-specific SSL data (dynamic `ca_trusted_fingerprint`)
- `service.extensions` entries matching `beatsauth/*` — extension registered in the service block

These fields are deployment-specific and should be stripped by `cleanPolicy` before comparison, following the existing pattern in `policyEntryFilters`.

### Changes

- Added a new `deletePattern *regexp.Regexp` field to `policyEntryFilter` to support removing map entries by key pattern and list entries by value pattern
- Extended the existing `exporters` filter to strip `auth` (present in any exporter since 9.4.0, removed for backwards compatibility)
- Added filters to strip `beatsauth/*` entries from `extensions` and `service.extensions`, and to remove those keys entirely when left empty
- Re-enabled previously skipped policy and system tests for `zipkin_input_otel` and `sql_server_input_otel` (also broken by the same Kibana change)
- Updated `zipkin_input_otel` policy fixtures to reflect the new `data_stream.dataset` processor statements introduced in 9.4.0

## Related issues

- Closes #3463

## Author's Checklist

- [x] Verified that policy tests pass on a 9.4.0+ stack
    - Tested in https://buildkite.com/elastic/integrations-test-stack/builds/65
    - Failing packages are not related to this issue, and they are already reported as flaky tests. 
- [x] Verified that policy tests on an older stack (e.g. 9.2.0) still pass — fixtures created before 9.4.0 should not require regeneration
- [x] Tested a package that requires a minimum stack version of 9.2.0 (e.g. `filelog_otel`) against a 9.4.0 stack to confirm policy tests no longer fail due to beatsauth fields
- [x] Confirmed that `zipkin_input_otel` and `sql_server_input_otel` tests are no longer skipped and pass correctly

---

> _This PR was created with the assistance of [Claude](https://claude.ai)._

### DIFF
--- a/internal/testrunner/runners/policy/policy.go
+++ b/internal/testrunner/runners/policy/policy.go
@@ -162,6 +162,8 @@ var policyEntryFilters = []policyEntryFilter{
 			regexp:  regexp.MustCompile(`^https?://.*$`),
 			replace: "https://elasticsearch:9200",
 		}},
+		// auth is injected by Fleet since 9.4.0 and may appear in any exporter, not just
+		// elasticsearch. Removed for backwards compatibility with older stacks.
 		{name: "auth"},
 	}},
 
--- a/internal/testrunner/runners/policy/policy.go
+++ b/internal/testrunner/runners/policy/policy.go
@@ -118,6 +118,9 @@ type policyEntryReplace struct {
 	replace string
 }
 
+// beatsauthPattern matches OTel component IDs for the beatsauth extension injected by Fleet.
+var beatsauthPattern = regexp.MustCompile(`^beatsauth/`)
+
 // policyEntryFilter includes a list of filters to do to the policy. These filters
 // are used to remove or control fields whose content is not relevant for the package
 // test.
@@ -163,9 +166,9 @@ var policyEntryFilters = []policyEntryFilter{
 	}},
 
 	// Fields injected by Fleet into OTel policies since 9.4.0 (beatsauth extension).
-	{name: "extensions", deletePattern: regexp.MustCompile(`^beatsauth/`)},
+	{name: "extensions", deletePattern: beatsauthPattern},
 	{name: "extensions", onlyIfEmpty: true},
-	{name: "service.extensions", deletePattern: regexp.MustCompile(`^beatsauth/`)},
+	{name: "service.extensions", deletePattern: beatsauthPattern},
 	{name: "service.extensions", onlyIfEmpty: true},
 
 	// Signatures that change from installation to installation.
--- a/internal/testrunner/runners/policy/policy.go
+++ b/internal/testrunner/runners/policy/policy.go
@@ -108,6 +108,7 @@ type policyEntryFilter struct {
 	mapValues          []policyEntryFilter
 	memberReplace      *policyEntryReplace
 	stringValueReplace *policyEntryReplace
+	deletePattern      *regexp.Regexp
 	onlyIfEmpty        bool
 	ignoreValues       []any
 }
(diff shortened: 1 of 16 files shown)


mrodm:
test integrations

jsoriano (approved):
Thanks!

teresaromero (approved):
thanks;  traces are already arriving at logs then?

mrodm:
> traces are already arriving at logs then?

@teresaromero That would be solved via this kibana PR https://github.com/elastic/kibana/pull/263415

Once that change is available in the 9.4.0-SNAPSHOT, it might require to update the policy tests in the test packages from elastic-package (expected permissions).

PR #3474, opened 2026-04-17
Support source overrides for composable integration test packages

### LINKED ISSUE #3464: Support local `source` for `requires.input` (align with package-spec `package_requirement`)
Part of #3277.

## Summary

Implement **local resolution** for required input packages so a composable integration and its input packages can be developed **in the same repository without publishing to EPR first**.

This is the follow-up tracked from PR #3459 review

## Spec alignment

Package spec defines `package_requirement` as a **oneOf**:

- `package` + `version` → exact version from the registry, or
- `source` → **relative path to the source of the required package**, which **overrides** downloading that package from EPR for tests.

Reference: [`config.spec.yml` — `package_requirement`](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/_dev/test/config.spec.yml#L30).

For elastic-package, we want the **same override idea** for the build/install path: when `source` is present for a required input, resolve that path (relative to the integration package root) and use it instead of `DownloadPackage` from the registry.

## Current behavior

- `RequiredInputsResolver.mapRequiredInputPackagesPaths` in `internal/requiredinputs/requiredinputs.go` only downloads via the EPR client.
- `openPackageFS` already supports directory and zip once a filesystem path is available.

## Scope / acceptance criteria (proposal)

1. **Model:** Extend `requires.input` dependency representation (e.g. `internal/packages.PackageDependency`) so each entry can be either `{ package, version }` **or** `{ source }`, consistent with package-spec’s `package_requirement` oneOf (mutually exclusive; validate errors if both or neither where inappropriate).

2. **Resolve `source`:** Interpret `source` as a path relative to the **composable package root** (same spirit as spec: relative path to required package source). Resolve to an absolute path, then either use it as a built package directory/zip or **build the input package from source** into a temp output if the path is source (team decision: require pre-built vs auto-`elastic-package build`).

3. **Wire into resolver:** `mapRequiredInputPackagesPaths` (or equivalent) chooses EPR download vs local path per entry; downstream bundling unchanged aside from the map values.

4. **Spec / publishing:** Ensure published `manifest.yml` behavior remains sane (e.g. `source` only for local dev, stripped or disallowed in published artifacts if required)—coordinate with package-spec if `manifest.yml` schema for `requires.input` must be updated to include `source`.

5. **Tests:** Add coverage with a fixture under `test/packages/` or `test/manual_packages/` that uses `source` and does not hit EPR for that dependency.

## Related code

- `internal/requiredinputs/requiredinputs.go` — `mapRequiredInputPackagesPaths`, `openPackageFS`, `Bundle`
- `internal/packages/packages.go` — `PackageDependency`, `Requires.Input`
- `cmd/build.go` (and other commands that construct `RequiredInputsResolver`) — may need profile/registry only when `source` is absent

cc @jsoriano (original review)

teresaromero:
## Summary

Closes #3464.

Implements local `source` resolution for `requires.input` dependencies so a composable integration and its required input packages can be developed and tested **in the same repository without publishing to EPR first**.

This is the follow-up work tracked from the review of #3459.

### Changes

- **Global test config**: Extended `globaltestconfig.go` to parse and expose `source` overrides from `_dev/test/config.yml` package requirements.
- **Test runners**: Updated `asset`, `system`, and `policy` test runners to resolve `source` overrides from the global test configuration instead of downloading from EPR.
- **`FleetPolicyPackageRoot`**: New function in `internal/builder` to determine the correct package root when resolving manifests for composable integrations.
- **`requiredinputs`**: Refactored `RequiresSourceOverrides` and `MergedRequiresSourceOverrides` to support local path resolution per entry.
- **CI**: Added `test-composable-packages.sh` script and `test-check-packages-composable` Makefile target; wired into the Buildkite integration test pipeline.
- **Test fixtures**: Added `01_ci_input_pkg` and `02_ci_composable_integration` packages under `test/packages/composable/` to exercise the new path without hitting EPR.

Co-authored-by: Cursor and Claude


internal/builder/fleet_policy_package_root.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return "", fmt.Errorf("read package manifest: %w", err)
+	}
+	if manifest.Type != "integration" {
+		return sourcePackageRoot, nil
+	}
+	if manifest.Requires == nil || len(manifest.Requires.Input) == 0 {
+		return sourcePackageRoot, nil
+	}
+
```
jsoriano: It would be actually safer to use the built manifest in all cases, right? Even if it is the same as the source package for non-composable packages.

Then we could maybe directly use `BuildPackagesDirectory` instead of `FleetPolicyPackageRoot`, it already reads the package manifest.

internal/builder/fleet_policy_package_root.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return "", fmt.Errorf("locate built package directory: %w", err)
+	}
+	builtManifest := filepath.Join(builtRoot, packages.PackageManifestFile)
+	if _, err := os.Stat(builtManifest); err != nil {
+		if os.IsNotExist(err) {
+			return "", fmt.Errorf("built package manifest not found at %s (build or install the package before creating Fleet policies for composable integrations): %w", builtManifest, err)
+		}
+		return "", fmt.Errorf("stat built package manifest %s: %w", builtManifest, err)
+	}
```
jsoriano: `BuildPackagesDirectory` already ensures that the built package has a manifest, it is required to get the name and version of the package.

```suggestion
```

internal/builder/packages.go
```diff
@@ -234,13 +234,11 @@ func BuildPackage(options BuildOptions) (string, error) {
...
-	}
-	err = resolver.Bundle(buildPackageRoot)
-	if err != nil {
-		return "", fmt.Errorf("bundling input package templates failed: %w", err)
+	if options.RequiredInputsResolver != nil {
+		err = options.RequiredInputsResolver.Bundle(buildPackageRoot)
+		if err != nil {
+			return "", fmt.Errorf("bundling input package templates failed: %w", err)
+		}
 	}
```
jsoriano, line 242: 👍

internal/resources/fleetpolicy.go
```diff
@@ -158,28 +159,34 @@ func (f *FleetAgentPolicy) Create(ctx resource.Context) error {
 }
 
 func createPackagePolicy(policy FleetAgentPolicy, packagePolicy FleetPackagePolicy) (*kibana.PackagePolicy, error) {
-	manifest, err := packages.ReadPackageManifestFromPackageRoot(packagePolicy.PackageRoot)
+	root, err := builder.FleetPolicyPackageRoot(packagePolicy.PackageRoot)
```
jsoriano: As mentioned in other comment, we might use the same code for all packages, though then we need to ensure that it is built first. Though this is probably fine.
```suggestion
	root, err := builder.BuildPackagesDirectory(packagePolicy.PackageRoot, "")
```

internal/testrunner/runners/system/tester.go
```diff
@@ -2176,25 +2177,52 @@ func CreatePackagePolicy(
...
+	if err != nil {
+		return kibana.PackagePolicy{}, "", "", err
+	}
+	usePkg := pkg
+	useDs := ds
+	useTempl := policyTemplate
+	if root != packageRoot {
+		// Composable integrations: input types are resolved on the built tree (requiredinputs.Bundle);
+		// reload manifests so Fleet policy keys match the package installed in setup (see FleetPackage).
+		pm, err := packages.ReadPackageManifestFromPackageRoot(root)
```
jsoriano: As mentioned in other comments, we can probably read the package manifest from the built package already in the caller, so we don't need to handle this here.

We need to build packages in any case for system tests, and for non-composable the built and source manifest will be the same.

Using the built manifest is safer in any case, we might introduce in the future other features that modify it in build time.

scripts/test-build-install-zip-file.sh
```diff
@@ -116,6 +116,11 @@ for d in test/packages/*/*/; do
   if [ "${packageTestType}" == "false_positives" ]; then
     continue
   fi
+  # Composable packages require the local registry to be running during build;
```
jsoriano, line 119: When using `source` the built packages could be retrieved from the local build directory.
Is this required for packages referenced by name/version?

jsoriano (commented):
We can probably use built packages in all cases for system testing, this would likely simplify things, and would ensure better that we test what the user receives.

scripts/test-build-install-zip-file.sh
teresaromero: no, if the test does not configure override or the override is a package it should download from the registry. i am reviewing this case as its not covered

internal/builder/fleet_policy_package_root.go
teresaromero: yes, agree. i did this change to narrow for composable but i think is safer to use the built version in any case. i spotted this when testing as `type` was not at the source package. i will change this

scripts/test-build-install-zip-file.sh
teresaromero: the override is only applicable at package tests. when running these test of build and install command there is no override and the registry is required. in this case local.

teresaromero: this is a guard to keep current build and install tests as is and have a dedicated pipeline for composable cases

jsoriano (approved):
👍 👍

scripts/test-composable-packages.sh
```diff
@@ -0,0 +1,93 @@
...
+# The local registry container serves packages from build/packages/, including the
+# newly built input package.
+elastic-package stack up -d -v ${stack_args}
+elastic-package stack status
+
+# Point elastic-package commands to the local registry started by the stack.
+echo "stack.epr.base_url: https://127.0.0.1:8080" \
+  >> ~/.elastic-package/profiles/composable/config.yml
+
+eval "$(elastic-package stack shellinit)"
```
mrodm: I guess this would work even without this `eval` command, is that right?

Could it be removed ?

.buildkite/pipeline.trigger.integration.tests.sh
```diff
@@ -64,6 +64,17 @@ for test in "${CHECK_PACKAGES_TESTS[@]}"; do
...
+echo "        command: ./.buildkite/scripts/integration_tests.sh -t test-build-install-packages-composable"
+echo "        agents:"
+echo "          provider: \"gcp\""
+echo "          image: \"${UBUNTU_X86_64_AGENT_IMAGE}\""
+echo "        artifact_paths:"
+echo "          - build/test-results/*.xml"
+echo "          - build/elastic-stack-dump/composable/logs/*.log"
+echo "          - build/elastic-stack-dump/composable/logs/fleet-server-internal/**/*"
+echo "          - build/test-coverage/coverage-*.xml"
+
```
mrodm: Nit. I would move this Buildkite step closer to the other buildkite steps that test the build and install process.
Around https://github.com/elastic/elastic-package/pull/3474/changes#diff-4e2e09cf3f17805ad30f828b15ea267ba7a186ff3f5129b9ecba76ef7317718bR188-R206

mrodm (approved):
LGTM!
Added a couple of questions/suggestions related to the CI scripts.

teresaromero:
@mrodm after addressing your feedback i am mergin this. if any comment on the changes for the review i can address them on the next followup

PR #3432, opened 2026-04-06
Nginx composable package

jsoriano:
Test package for https://github.com/elastic/elastic-package/pull/3329

PR #3480, opened 2026-04-20
Composable packages: input `name` qualifiers and stream `input` refs

### LINKED ISSUE #3465: Composable `requires.input`: set input `name` and stream `input` per package-spec qualified inputs (#1135)
Part of #3277.

## Origin

- [@jsoriano on PR #3459](https://github.com/elastic/elastic-package/pull/3459#discussion_r3081447971): set a **unique name** and use the **same name** on data streams when resolving `package:` references in composable builds.
- Product context: [elastic/ingest-dev#7380](https://github.com/elastic/ingest-dev/issues/7380).

## Spec / Fleet expectation

[elastic/package-spec#1135](https://github.com/elastic/package-spec/pull/1135) (**spec 3.6.0+**) introduces optional **`policy_templates[].inputs[].name`** (input **qualifier**). When present, **data streams reference the input by that `name` instead of by `type`**, so multiple inputs with the same `type` (e.g. multiple `otelcol`) are unambiguous. If a policy template has **more than one input with the same `type`**, package-spec requires **`name`** on those inputs (semantic validator `SVR00010`).

Reference implementation for Fleet/UI: [elastic/kibana#262138](https://github.com/elastic/kibana/pull/262138).

Example from package-spec test package `good_input_qualifier`: two inputs both `type: otelcol` with `name: filelog_otel` and `name: nginx_otel`; streams use `input: filelog_otel` / `input: nginx_otel` respectively—not `input: otelcol`.

## Problem (elastic-package today)

In [`internal/requiredinputs/streamdefs.go`](https://github.com/elastic/elastic-package/blob/main/internal/requiredinputs/streamdefs.go), resolution sets **`inputs[].type`** and **`streams[].input`** from a single derived `identifier` (today the input package's input id). For multiple required input packages sharing the same type, this does **not** emit spec-compliant **`name`** + stream-by-name linkage and matches the Fleet confusion described above.

## Acceptance criteria (proposal)

1. **Policy template inputs:** Keep **`type`** as the resolved real input type from the input package (e.g. `otelcol`). When a composable policy template ends up with **multiple inputs of the same `type`** after merging required inputs (or whenever disambiguation is required), set a stable, unique **`name`** on each such input per [package-spec#1135](https://github.com/elastic/package-spec/pull/1135) (derive from required package name / policy template / agreed convention—document choice in PR).
2. **Data stream streams:** Set **`streams[].input`** to the **`name`** of the corresponding policy input when using qualifiers; it must **match** the `name` on the integration manifest input that serves that stream (same correlation as the spec example).
3. **Spec compatibility:** Built integration packages that use `name` should target **`format_version` ≥ 3.6.0** (or otherwise align with when the `name` field is valid per package-spec), so `elastic-package` validation agrees with package-spec.
4. **Tests:** Fixture or unit test with two required input packages sharing the same input `type`, asserting built `manifest.yml` and `data_stream/*/manifest.yml` match the **name + type** pattern above and would pass package-spec's duplicate-type / missing-name rules.

## Likely implementation area

- [`internal/requiredinputs/streamdefs.go`](https://github.com/elastic/elastic-package/blob/main/internal/requiredinputs/streamdefs.go) — extend `inputPkgInfo` / YAML upserts to handle **`name`** on inputs and **`input`** on streams per spec, not only `type`.

cc @jsoriano

teresaromero:
## Why

Composable packages with `requires.input` can have **multiple required inputs of the same `type`**. [package-spec#1135](https://github.com/elastic/package-spec/issues/1135) (3.6.0+) requires distinct **`policy_templates[].inputs[].name`** and **`streams[].input`** that reference that name—not a single derived id for both policy and streams. elastic-package previously did not match that pattern, which caused ambiguous Fleet policies and errors (e.g. missing stream templates).

This branch also builds policies from the **built** package manifest (so composable `package:` refs resolve for Fleet), adds **CI/fixtures** (nginx composable, dual same-type inputs, bootstrap input pkgs), and includes **`data_stream.type`** for input-package system/policy tests ([#3509](https://github.com/elastic/elastic-package/pull/3509)) plus docs for Kibana/Fleet version behavior.

## What changed

- **`internal/requiredinputs` / `internal/kibana` / `internal/packages` / `internal/resources/fleetpolicy`:** Stable input **`name`** when types collide; **`type`** stays the agent type; streams use **`inputRef`** (name or type); **`PackagePolicyInput.inputType`** is the Fleet input type; `FindInput` matches name or type. Policy built from **built** tree (`ReadBuiltPackageManifest`, etc.).
- **Test runners & scripts:** Policy/system/script updates for composable flows, overrides, and data stream naming; composable Makefile/script and packages under `test/packages/composable/` and `test/packages/parallel/nginx_composable/`.

## How to verify

`go build ./...` / `go test ./...` (see [AGENTS.md](AGENTS.md) for narrower scopes). Composable: Makefile + `scripts/test-composable-packages.sh`. Example: `go run . test -C ./test/packages/parallel/nginx_composable` on a stack with the Fleet fixes you need. `go test ./internal/kibana/... ./internal/testrunner/...` for golden/policy coverage.

## Related

- Closes https://github.com/elastic/elastic-package/issues/3465

- [#3465](https://github.com/elastic/elastic-package/issues/3465) / [package-spec#1135](https://github.com/elastic/package-spec/issues/1135)
- Fleet: [kibana#264366](https://github.com/elastic/kibana/pull/264366); nginx/system discussion: [kibana#266101](https://github.com/elastic/kibana/pull/266101)
- [#3509](https://github.com/elastic/elastic-package/pull/3509); Fleet vars / legacy API: [kibana#214216](https://github.com/elastic/kibana/pull/214216), [kibana#266321](https://github.com/elastic/kibana/issues/266321)

## Depends on

**Stack:** Snapshots (or pins) that include the Fleet fixes your integration tests assume—align CI stack version with what you document. **`data_stream.type`:** Use **`policy_api_format: legacy`** where the simplified API still rejects the var ([kibana#266321](https://github.com/elastic/kibana/issues/266321)); see howto docs.


teresaromero:
test integrations

internal/packages/packages.go
```diff
@@ -851,6 +852,20 @@ func (pt *PolicyTemplate) FindInputByType(inputType string) *Input {
...
+// input's Name qualifier first (when set) and then its Type. Use this when the
+// identifier may be a name qualifier (set on inputs that share a type) rather
+// than a bare type string.
+func (pt *PolicyTemplate) FindInput(identifier string) *Input {
+	for i := range pt.Inputs {
+		input := &pt.Inputs[i]
+		if (input.Name != "" && input.Name == identifier) || input.Type == identifier {
+			return input
+		}
+	}
```
mrodm, line 855: Q: Should this loop duplicated to first look for input.Name and if it does not find any , then look for input.Type ? 

Is it ok to search at the same time for both ?

scripts/test-composable-packages.sh
```diff
@@ -52,12 +78,18 @@ elastic-package profiles use composable
...
-elastic-package build -C "${COMPOSABLE_PACKAGES_PATH}/${COMPOSABLE_INPUT_PKG}" -v
+for bootstrap_name in "${COMPOSABLE_BOOTSTRAP_PKGS[@]}"; do
+  echo "--- Building bootstrap input package: ${COMPOSABLE_PACKAGES_PATH}/${bootstrap_name}"
+  # Bootstrap packages have no (or satisfied) requires.input; no registry yet.
+  # After build, artifacts land in build/packages/ for the stack local registry.
+  elastic-package build -C "${COMPOSABLE_PACKAGES_PATH}/${bootstrap_name}" -v
+done
 
 echo "--- Prepare Elastic stack"
+export PACKAGE_TEST_TYPE=composable
```
mrodm: PACKAGE_TEST_TYPE should be defined in the Makefile target, as it is done in the other targets. For instance:

```
test-check-packages-with-custom-agent:
	PACKAGE_TEST_TYPE=with-custom-agent ./scripts/test-check-packages.sh
```

scripts/test-composable-packages.sh
```diff
@@ -52,12 +78,18 @@ elastic-package profiles use composable
...
+if [[ -z "${PACKAGE_UNDER_TEST:-}" ]]; then
+  export PACKAGE_UNDER_TEST="${COMPOSABLE_STACK_PIN_PACKAGE:-${COMPOSABLE_INPUT_PKG}}"
+fi
```
mrodm: As this test script is intended to test all packages at the same Buildkite step, I was wondering that it could be set directly in the elastic-package stack up command the required verion.

Helpers for doing this task, just ignore the targets that do not set PACKAGE_UNDER_TEST
```
stack_version_args() {
  if [[ -z "${PACKAGE_UNDER_TEST:-""}" ]]; then
    # Don't force stack version if we are testing multiple packages.
    return
  fi
  ...
```


It would be needed to add a TODO note to remove that version once the default one in elastic-package is the right one (or create an issue for that ?)

test/packages/composable/01_ci_input_pkg.stack_version
mrodm, line 1: Given how the Buildkite step works for this new test step, just one file should be needed because there is just one Elastic stack started.

Given that, I was wondering if it could be just pinned in the shellscript adding a TODO to remove it (or maybe an issue). WDYT?

internal/packages/packages.go
```diff
@@ -851,6 +852,20 @@ func (pt *PolicyTemplate) FindInputByType(inputType string) *Input {
 	return nil
 }
```
mrodm: This function looks is not used anymore.

internal/kibana/packagepolicy.go
```diff
@@ -48,17 +48,20 @@ func BuildIntegrationPackagePolicy(
...
+			ref := inputRef(input)
+			inputKey := fmt.Sprintf("%s-%s", pt.Name, ref)
+			if ref == streamInput && pt.Name == policyTemplate.Name {
 				// The target input: enabled with user-provided vars.
 				streams := buildStreamsForInput(streamInput, manifest, dsManifest, enabled, dsVars, datastreams)
 				inputEntry := PackagePolicyInput{
 					Enabled:        enabled,
 					Streams:        streams,
-					inputType:      streamInput,
+					inputType:      input.Type, // always the real agent type, not the name qualifier
```
mrodm: Q: So then here, would be "metrics", "logs", "traces" or "otelcol"?

test/packages/composable/01_ci_input_pkg/manifest.yml
mrodm, line 1: Would it be possible to add an example of composable package using an otel package (input type `otelcol`)?

internal/packages/packages.go
teresaromero: as defiened on spec https://github.com/elastic/package-spec/blob/61cee23beb811b6c93ddb71caf822914ee30afba/code/go/internal/validator/semantic/validate_integration_input_qualifier.go#L68 the propoerty name is required when inputs share the same type. so i think is safe to make this comparison, so when name is found, its due to the fact that type is the same. when name is not found (empty) then we can trust type

test/packages/composable/01_ci_input_pkg/manifest.yml
teresaromero: i have cherry-picked nginx_composabe from jaimes draft https://github.com/elastic/elastic-package/pull/3432 to have it as a test fixture for this cases too

internal/kibana/packagepolicy.go
teresaromero: these are the agent input types, not the data stream types, added a comment for clarification

internal/packages/packages.go
teresaromero: removed!

scripts/test-composable-packages.sh
teresaromero: i've added the version flag to the command directly on the script. does it make more sense?

scripts/test-composable-packages.sh
teresaromero: changed to makefile

test/packages/composable/01_ci_input_pkg/manifest.yml
teresaromero: to address this example, i have been working with the nginx_composable but i cant seem to make system tests work. would like some help here to understand if this is in the scope of the pr change or is the test package itself that needs a fix on configuration. 

system test get stuck when looking for hits on the dataset so it never finds hits

@jsoriano could you help me unblock this?

jsoriano: Discussed over slack, https://github.com/elastic/kibana/pull/266101 will likely help.

teresaromero:
reported errors are expected as a fix was backported to 9.4 which is expected for this pr. snapshot does not contain the fix yet

scripts/test-composable-packages.sh
mrodm: Yes, let's keep it as that for now.

scripts/test-composable-packages.sh
```diff
@@ -2,24 +2,44 @@
...
+# if you run this script directly, use the same prefix
+# (e.g. PACKAGE_TEST_TYPE=composable ./scripts/test-composable-packages.sh).
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+# Prefer the elastic-package binary built in this repo (see Makefile `build`) so
+# composable install uses the same code as the checkout, not an older global install.
+if [[ -x "${REPO_ROOT}/elastic-package" ]]; then
+  PATH="${REPO_ROOT}:${PATH}"
+fi
```
mrodm: This should not be needed. `elastic-package` is installed here https://github.com/elastic/elastic-package/blob/7ba5943de91f6f3c8455399eb6dc5be896c12fcd/.buildkite/scripts/integration_tests.sh#L153

And it should be available already in the PATH.

Here it is the installation in the Buildkite build: https://buildkite.com/elastic/elastic-package/builds/7852#019dd956-4795-49ef-b207-d1d2b25e1155/L436

scripts/test-composable-packages.sh
```diff
@@ -52,13 +72,17 @@ elastic-package profiles use composable
...
+  # Bootstrap packages have no (or satisfied) requires.input; no registry yet.
+  # After build, artifacts land in build/packages/ for the stack local registry.
+  elastic-package build -C "${COMPOSABLE_PACKAGES_PATH}/${bootstrap_name}" -v
+done
 
 echo "--- Prepare Elastic stack"
-stack_args=$(set +x; stack_version_args)
+# TODO: Remove --version below when composable tests pass on elastic-package's default stack
+# (Fleet/Kibana no longer require this snapshot).
+stack_args="--version 9.5.0-SNAPSHOT"
```
mrodm: This should be tried with 9.4.0-SNAPSHOT before merging if possible.
I haven't checked if the required changes are available in the latest Snapshot of 9.4.0

test/packages/parallel/nginx_composable.stack_version
mrodm, line 1: With the latest changes, I think this file is not used by the CI scripts.
If so, this could be deleted.

internal/testrunner/script/data_stream.go
```diff
@@ -121,7 +122,25 @@ func addPackagePolicy(ts *testscript.TestScript, neg bool, args []string) {
...
+	// so templ.Input is empty. Resolve the effective input type from the built tree.
+	if templ.Input == "" {
+		if builtRoot, builtPkg, berr := builder.ReadBuiltPackageManifest(manifestRoot); berr == nil {
+			if builtPT, berr := packages.SelectPolicyTemplateByName(builtPkg.PolicyTemplates, *polName); berr == nil {
+				inputName := config.Input
+				if inputName == "" {
+					if builtDS, berr := packages.ReadDataStreamManifestFromPackageRoot(builtRoot, dsMan.Name); berr == nil && len(builtDS.Streams) > 0 {
+						inputName = builtDS.Streams[0].Input
+					}
+				}
```
mrodm: Errors from these functions are not validated. Should they be raised if any?

scripts/test-composable-packages.sh
```diff
@@ -52,13 +72,17 @@ elastic-package profiles use composable
...
-echo "--- Building input package: ${COMPOSABLE_PACKAGES_PATH}/${COMPOSABLE_INPUT_PKG}"
-# The input package has no requires.input so no registry is needed at this stage.
-# After build, the package lands in build/packages/ where the local registry serves it.
-elastic-package build -C "${COMPOSABLE_PACKAGES_PATH}/${COMPOSABLE_INPUT_PKG}" -v
+for bootstrap_name in "${COMPOSABLE_BOOTSTRAP_PKGS[@]}"; do
+  echo "--- Building bootstrap input package: ${COMPOSABLE_PACKAGES_PATH}/${bootstrap_name}"
+  # Bootstrap packages have no (or satisfied) requires.input; no registry yet.
+  # After build, artifacts land in build/packages/ for the stack local registry.
+  elastic-package build -C "${COMPOSABLE_PACKAGES_PATH}/${bootstrap_name}" -v
+done
```
mrodm, line 74: Not required for this PR, but just thought that there could be added support for some kind of hooks for the packages tested under parallel or false positives could take advantage in case they need to run specific commands.

There could be some hooks:
- before and after building the package
- before and after installing the package
- before and after testing the package

AFAIK there is no need for that now, so just commenting here this idea.

teresaromero:
test integrations

internal/testrunner/script/data_stream.go
teresaromero: updated

scripts/test-composable-packages.sh
teresaromero: i am not sure i understand what you mean with hook here. you mean that before building a packagae we could have defined commands running? can you extend with an example?

test/packages/parallel/nginx_composable.stack_version
teresaromero: parallel tests still use this. i removed it and the version used at the test was 9.3.3

internal/kibana/packagepolicy.go
```diff
@@ -48,17 +48,20 @@ func BuildIntegrationPackagePolicy(
 	inputs := make(map[string]PackagePolicyInput)
 	for _, pt := range manifest.PolicyTemplates {
 		for _, input := range pt.Inputs {
-			inputKey := fmt.Sprintf("%s-%s", pt.Name, input.Type)
-			if input.Type == streamInput && pt.Name == policyTemplate.Name {
+			// ref is the identifier used in data stream manifests: the Name qualifier
+			// when set (disambiguated same-type inputs), otherwise the Type.
+			ref := inputRef(input)
```
jsoriano: Is this what we call effective input names in Kibana? Consider to use this name here.
```suggestion
			effectiveName := effectiveInputName(input)
```

internal/kibana/packagepolicy.go
```diff
@@ -310,3 +313,13 @@ func setVarFromUser(vars Vars, name, varType string, val packages.VarValue) {
 	}
 	vars[name] = Var{Type: varType, Value: val, fromUser: true}
 }
+
+// inputRef returns the identifier used to reference an input from data stream
+// manifests: the Name qualifier when set (for inputs disambiguated by name), or
+// the Type when no qualifier is present.
+func inputRef(input packages.Input) string {
```
jsoriano:
```suggestion
// effectiveInputName returns the identifier used to reference an input from data stream
// manifests: the Name qualifier when set (for inputs disambiguated by name), or
// the Type when no qualifier is present.
func effectiveInputName(input packages.Input) string {
```

internal/packages/packages.go
```diff
@@ -841,11 +842,15 @@ func (dsm *DataStreamManifest) indexTemplateNamePrefix() string {
...
-// FindInputByType returns the input for the provided type.
-func (pt *PolicyTemplate) FindInputByType(inputType string) *Input {
-	for _, input := range pt.Inputs {
-		if input.Type == inputType {
-			return &input
+// FindInput returns the first input matching identifier, checked against the
+// input's Name qualifier first (when set) and then its Type. Use this when the
+// identifier may be a name qualifier (set on inputs that share a type) rather
+// than a bare type string.
+func (pt *PolicyTemplate) FindInput(identifier string) *Input {
```
jsoriano:
```suggestion
// FindInput returns the first input matching the effective name, checked against the
// input's Name qualifier first (when set) and then its Type. Use this when the
// identifier may be a name qualifier (set on inputs that share a type) rather
// than a bare type string.
func (pt *PolicyTemplate) FindInput(effectiveName string) *Input {
```

internal/requiredinputs/streamdefs.go
```diff
@@ -41,11 +48,44 @@ func (r *RequiredInputsResolver) resolveStreamInputTypes(
 		return err
 	}
 
-	if err := applyInputTypesToComposableManifest(manifest, buildRoot, infoByPkg); err != nil {
+	streamInputRefs := buildStreamInputRefs(manifest, infoByPkg)
```
jsoriano: We should probably also use "effective name" instead of "ref" here.

internal/testrunner/runners/static/tester.go
```diff
@@ -248,28 +249,51 @@ func (r tester) getExpectedDatasets(pkgManifest *packages.PackageManifest) ([]st
...
+		return true
 	}
 
-	return true
+	// For composable integration packages, check the built manifest where "package:" references
+	// are materialised into concrete input types.
+	if r.testFolder.DataStream == "" {
+		return false
+	}
+	builtRoot, builtPkg, err := builder.ReadBuiltPackageManifest(r.packageRoot)
```
jsoriano, line 283: This method already receives a manifest. Do callers need to be able to work both with source and built manifests?
If this is always working with built packages callers could use always built manifests so we don't need to read it again here.

scripts/test-composable-packages.sh
mrodm: In the main script (`scripts/test-check-packages.sh`), it could be added some code like:

```sh
for d in test/packages/${PACKAGE_TEST_TYPE}/${PACKAGE_UNDER_TEST}/; do
  echo "--- Checking package ${d}"
  if [ -f "$d/pre-check.sh" ]; then
    bash "$d/pre-check.sh"
  fi
  elastic-package check -C "$d" -v
  if [ -f "$d/post-check.sh" ]; then
    bash "$d/post-check.sh"
  fi
done
```


```sh
  echo "--- Run tests for package ${d}"
  if [ -f "$d/pre-test.sh" ]; then
    bash "$d/pre-test.sh"
  fi
  # Run all tests
  # defer-cleanup is set to a short period to verify that the option is available
  elastic-package test -C "$d" -v \
    --report-format xUnit \
    --report-output file \
    --defer-cleanup 1s \
    --test-coverage \
    --coverage-format=generic
  if [ -f "$d/post-test.sh" ]; then
    bash "$d/post-test.sh"
  fi
```

Not sure if this could be helpful in the future, to try to allow customize even further the scenario for each package.

As I mentioned, just an idea that it came to my mind, no need to do it now.

test/packages/parallel/nginx_composable.stack_version
mrodm: Ah right! 👍 
I misread this and I thought this were related to the packages in `test/packages/composable` 😅

internal/testrunner/runners/system/tester.go
```diff
@@ -2210,57 +2250,96 @@ func (r *tester) checkEnrolledAgents(ctx context.Context, agentInfo agentdeploye
...
 	packageRoot string,
 ) (policy kibana.PackagePolicy, dsType string, dsDataset string, err error) {
-	if pkg.Type == "input" {
+	if packageRoot == "" {
+		return kibana.PackagePolicy{}, "", "", fmt.Errorf("package root is required")
+	}
+
+	// Always resolve against the built tree so that RequiredInputsResolver has already
+	// materialized package: references into concrete input types.
+	builtRoot, builtPkg, err := builder.ReadBuiltPackageManifest(packageRoot)
```
jsoriano, line 2272: We are reading and using the built package manifest in several places. I would prefer to read it only once and then pass it.
My main concern is not about reusing the read manifest, but about having to decide every time what manifest should be used.

As this is a system tester, we can probably use always the built manifest, as a built package should be available.

internal/testrunner/runners/system/tester_test.go
```diff
@@ -520,16 +519,14 @@ func TestBuildDataStreamName(t *testing.T) {
...
 			expected:       "logs-nginx.access-default",
 		},
 		{
-			title:          "otelcol input: .otel suffix appended",
+			title:          "otelcol input package: .otel suffix appended",
 			dsType:         "logs",
 			dsDataset:      "httpcheck",
 			namespace:      "default",
 			policyTemplate: packages.PolicyTemplate{Input: otelCollectorInputName},
-			packageType:    "input",
```
jsoriano, line 532: Don't we need to declare if this is an input package to be sure that this test covers this type?

internal/testrunner/runners/system/tester_test.go
```diff
@@ -905,6 +909,58 @@ func TestPipelineErrorMessage(t *testing.T) {
...
+	_, hasLogfile := pp.Inputs["nginx-logfile"]
+	assert.True(t, hasLogfile, "expected input key nginx-logfile, got keys: %v", keysOf(pp.Inputs))
+}
+
+func keysOf(m map[string]kibana.PackagePolicyInput) []string {
+	out := make([]string, 0, len(m))
+	for k := range m {
+		out = append(out, k)
+	}
+	return out
```
jsoriano:
```suggestion
	return slices.Sorted(maps.Keys(m))
```

internal/testrunner/script/data_stream.go
```diff
@@ -121,7 +122,38 @@ func addPackagePolicy(ts *testscript.TestScript, neg bool, args []string) {
 	templ, err := packages.SelectPolicyTemplateByName(pkgMan.PolicyTemplates, *polName)
 	ts.Check(decoratedWith("finding policy template", err))
 
-	policy, dsType, dsDataset, err := system.CreatePackagePolicy(installed.testingPolicy, pkgMan, templ, dsMan, config.Input, config.Vars, config.DataStream.Vars, installed.testingPolicy.Namespace, manifestRoot)
+	// For composable integration packages the source manifest has unresolved "package:" references,
```
jsoriano, line 125: We should probably test always with built packages to avoid this complexity.

jsoriano (commented):
Tested and seems to work great, thanks!

I have a couple of concerns on the code:
* There is some complexity on deciding when the source or the built manifests need to be used. In most testers I think we should always use built packages, and thus we can assume that we have the build manifests available.
* [Nit] For the qualifier name/type we use "effective input name" in Kibana, and "input ref" here. If possible it would be great to use the same terms to avoid confusion.

internal/requiredinputs/streamdefs.go
```diff
@@ -17,6 +17,7 @@ import (
 // inputPkgInfo holds the resolved metadata from an input package needed to
 // replace package: references in composable package manifests.
 type inputPkgInfo struct {
+	pkgName        string // manifest.name; used as the input name qualifier when disambiguation is needed
```
mrodm: It looks like that this variable is never used/read. Could this be removed ?

internal/testrunner/runners/system/tester.go
```diff
@@ -2210,57 +2250,96 @@ func (r *tester) checkEnrolledAgents(ctx context.Context, agentInfo agentdeploye
...
+	policyTemplateName string,
+	dataStreamName string,
+	cfgName string,
+	cfgVars, cfgDSVars common.MapStr,
+	suffix string,
+) (policy kibana.PackagePolicy, dsType string, dsDataset string, err error) {
+	builtPkg, err := packages.ReadPackageManifestFromPackageRoot(builtRoot)
+	if err != nil {
+		return kibana.PackagePolicy{}, "", "", fmt.Errorf("reading built package manifest at %s: %w", builtRoot, err)
+	}
```
mrodm: When the package is of type "integration" this call would produced the same builtPkg value as in https://github.com/elastic/elastic-package/pull/3480/changes#diff-0bdd696fb9abd5e4aa12a3bab634cbac31ec1a1349b3a61e8a784a6bdeb6e7d8R2272-R2275

Would it be valid to set as parameter `builtPkg` instead of `builtRoot`? Would it be equivalent ?

internal/requiredinputs/streamdefs.go
```diff
@@ -153,7 +202,7 @@ func applyInputTypesToDataStreamManifests(buildRoot *os.Root, infoByPkg map[stri
 				return fmt.Errorf("getting stream node at index %d in %q: %w", streamIdx, manifestPath, err)
 			}
 
-			upsertKey(streamNode, "input", strVal(info.identifier))
+			upsertKey(streamNode, "input", strVal(streamInputRefs[stream.Package]))
```
mrodm: Should it be ensured that `stream.Package` is present in `streamInputRefs` to return some error before doing this call ? As it has been done for `infoByPkg` at line 195.

internal/requiredinputs/streamdefs_test.go
```diff
@@ -496,6 +496,275 @@ streams:
...
+	manifest := &packages.PackageManifest{
+		PolicyTemplates: []packages.PolicyTemplate{
+			{
+				Inputs: []packages.Input{
+					{Package: "pkg_a"},
+					{Package: "pkg_b"},
+				},
+			},
+		},
+	}
```
mrodm, line 550: These tests just have one policy template, could you add one test that includes two policy templates to validate that scenario too?

mrodm (commented):
LGTM!
I've also tested locally and the `nginx_composable` built package looks good too 💪

internal/requiredinputs/streamdefs_test.go
teresaromero: sure, i've added the case and improved godoc description as it was not clear what the outcome is in that case. duplicate input types are counted by policy template.

internal/testrunner/runners/system/tester_test.go
teresaromero: the type was removed because we are now checking both integration and input packages. the function does not need the type to check if the policy template input is otelcol or not, and construct the dataset name with the suffix.

internal/testrunner/runners/static/tester.go
teresaromero: this is not trivial change, for static tests we need both package source and built.

internal/testrunner/runners/system/tester.go
teresaromero: i think this change is related to a test structure refactor. we needed to introduced the built bundle to work with composables, can we aford this debt and address the refactor as a followup? i can open an issue for it

internal/testrunner/script/data_stream.go
teresaromero: same as here https://github.com/elastic/elastic-package/pull/3480#discussion_r3195581141

internal/testrunner/runners/static/tester.go
jsoriano: It would be good to check why static tests need both, we can leave this for a future refactor.

internal/testrunner/runners/system/tester.go
jsoriano: Definitely, we can do this in a follow up.

internal/testrunner/script/data_stream.go
jsoriano: Ok, let's do it in a follow up.

jsoriano (approved):
👍 ready to go, and we can address the requested refactors in follow ups.

teresaromero:
Opening followup issue to address test refactor https://github.com/elastic/elastic-package/issues/3530 ;

teresaromero:
test integrations

internal/requiredinputs/streamdefs_test.go
mrodm: Thanks!

teresaromero:
test integrations

mrodm (approved):
🚀

PR #3508, opened 2026-04-28
[sql_server_input_otel][mysql_input_otel] Add signal-specific system and policy tests for logs and metrics

mrodm:
Update the `sql_server_input_otel` and `mysql_input_otel` test packages to add signal-specific system and policy tests, aligned with the changes in https://github.com/elastic/integrations/pull/18658.

### `sql_server_input_otel`

- Enable `test-logs-config` system test with `signal_types: logs`, covering `db.server.top_query` and `db.server.query_sample` events with `fields_present` assertions aligned with upstream receiver field names (`user.name` instead of `sqlserver.username`; `sqlserver.query_plan` removed due to ES `ignore_above` limit). Uses `top_query_lookback_time: 20s` to match the workload's lock window.
- Add `test-metrics-config` system test (skipped pending multisignal input support in elastic-package).
- Enable log event collection in default and datasource configs so the default system test also exercises log collection alongside metrics.
- Remove blanket `fields_present` from default and datasource configs — assertions are now handled per-signal config file.
- Fix Docker workload to hold an exclusive table lock (`TABLOCKX`) so foreground `SELECT`s appear as suspended sessions in `sys.dm_exec_requests`, making them reliably visible to the `query_sample` collector.
- Remove `skip` directive from all 5 existing policy tests.
- Add `test-resource-attributes` policy test for the `resource_attributes` var.
- Add `test-logs-custom-collection` policy test for non-default query-sample and top-query settings.

### `mysql_input_otel`

- Add `test-logs-config` and `test-metrics-config` signal-specific system tests with `fields_present` assertions using correct `mysqlreceiver` field names.
- Add policy tests for logs and metrics enabling.
- Restrict the default system test to a multisignal run (no `signal_types` filter), dropping `fields_present` in favour of the signal-specific configs.

## Why

These tests ensure that both OTel SQL integrations collect the expected fields for each signal type, catch regressions in field naming conventions, and validate that policy rendering produces correct agent configurations for all supported configuration options.

## Related

- Relates https://github.com/elastic/integrations/pull/18658
- Part of https://github.com/elastic/integrations/issues/17876

---


teresaromero (approved):
thanks

PR #3509, opened 2026-04-28
Support custom data_stream.type in system and policy tests

jsoriano:
https://github.com/elastic/package-spec/issues/849 introduced a way to override the data stream type for input packages. Users can use this toggle to decide the data stream to use for data collected with generic input packages.

Allow to override the data stream type using the `data_stream.type` variable in system and policy tests.

### DIFF
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -554,7 +554,7 @@ override the data stream routing set by Fleet:
   Requires Kibana 8.18.3 / 8.19.0 / 9.1.0 or later (Fleet PR [#214216](https://github.com/elastic/kibana/pull/214216)).
   Because the Kibana simplified Fleet API does not yet accept `data_stream.type` as a
   stream-level variable (it rejects unknown vars), tests that use this override must also
-  set `policy_api_format: legacy` until a Kibana fix is available.
+  set `policy_api_format: legacy` until a Kibana fix is available ([#266321](https://github.com/elastic/kibana/issues/266321)).
 
 ```yaml
 policy_api_format: legacy
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -543,6 +543,30 @@ for system tests.
 | wait_for_data_timeout | duration |  | Amount of time to wait for data to be present in Elasticsearch. Defaults to 10m. |
 | wait_for_dynamic_streams_stable | duration |  | For packages with dynamic data streams (e.g. `dynamic_signal_types`), minimum time the discovered data stream count must stay unchanged after the first stream appears before discovery completes. Defaults to 10s. |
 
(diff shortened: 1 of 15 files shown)


docs/howto/system_testing.md
```diff
@@ -543,6 +543,34 @@ for system tests.
...
+For **input packages**, the top-level `vars` field also accepts two special variables that
+override the data stream routing set by Fleet:
+
+- `data_stream.dataset` — overrides the dataset portion of the data stream name
+  (e.g. `sql_input.test` instead of the default `sql_input.sql_query`).
+- `data_stream.type` — overrides the signal type (`logs`, `metrics`, or `traces`).
+  Requires Kibana 8.18.3 / 8.19.0 / 9.1.0 or later (Fleet PR [#214216](https://github.com/elastic/kibana/pull/214216)).
+  Because the Kibana simplified Fleet API does not yet accept `data_stream.type` as a
+  stream-level variable (it rejects unknown vars), tests that use this override must also
+  set `policy_api_format: legacy` until a Kibana fix is available.
```
jsoriano: Follow up issue created for that: https://github.com/elastic/kibana/issues/266321

test/packages/parallel/sql_input/_dev/test/policy/test-mysql-logs-type.yml
```diff
@@ -0,0 +1,9 @@
+# The simplified Fleet API rejects data_stream.type because it is not declared
+# as a variable in the package manifest. Use the legacy API until Kibana's
+# simplified_package_policy_helper handles data_stream.type like data_stream.dataset.
```
mrodm, line 3: It could be added the link to the Kibana issue here or maybe in the docs. That would help us to know if `policy_api_format: legacy` could be removed eventually.

mrodm (approved):

test/packages/parallel/sql_input/_dev/test/policy/test-mysql-logs-type.yml
jsoriano: Added in https://github.com/elastic/elastic-package/pull/3509/commits/cc0c037488c31ffd3e7beb395ee9d99a28b90e26

mrodm (approved):

teresaromero (approved):
why we would want to change the type a datastream gets?

jsoriano:
> why we would want to change the type a datastream gets?

With input packages a user may want to chose the signal type they are collecting. Imagine for example querying HTTP endpoints with httpjson. This input package is defined with the "logs" type, but you could also collect metrics using HTTP JSON endpoints. With this feature we allow users to store the data they are collecting with the type they consider to be better suited.

See https://github.com/elastic/package-spec/issues/849.

Here we are allowing package developers to test their packages with this feature.

Issue #3443, opened 2026-04-08
Gradually stricter configuration for golangci-lint


jsoriano:
This issue tracks the TODOs introduced when setting up `golangci-lint` in https://github.com/elastic/elastic-package/pull/3440.

Each item represents a known gap that was suppressed for now to keep `make lint` green while the codebase is incrementally improved.

## Complexity thresholds

These thresholds were raised above the desired targets to make lint pass. They should be lowered as the flagged functions are refactored.

- [ ] **`gocyclo`**: current threshold is 51, target is 25. Worst offenders: `(*tester).prepareScenario` and `Run` in `internal/testrunner/script/script.go` (complexity 50).
- [ ] **`gocognit`**: current threshold is 70, target is 40. Two functions exceed 70 and have inline `//nolint` comments:
  - `cleanPolicyMap` in `internal/testrunner/runners/policy/policy.go` (complexity 103)
  - `Run` in `internal/testrunner/script/script.go` (complexity 89)

## Function argument limits (`revive`)

The `argument-limit` rule is set to 11 instead of the default 8. The following functions should be refactored to use option structs:

- [ ] `BuildIntegrationPackagePolicy` in `internal/kibana/packagepolicy.go` (11 args)
- [ ] `CreatePackagePolicy` in `internal/testrunner/runners/system/tester.go` (9 args)
- [ ] `processResults` in `cmd/testrunner.go` (9 args)

## MapStr typed API (`errcheck`)

- [ ] `common.MapStr.Put` and `common.MapStr.Delete` return errors only when dot-notation path traversal fails on a non-map intermediate value. All current callsites use simple string literal keys and could use direct map operations (`m[key] = value`, `delete(m, key)`), but we want to keep the possibility of using dot-notation.

Affected files:
- `internal/builder/dashboards.go`
- `internal/export/transform_decode.go`
- `internal/fields/dependency_manager.go`
- `internal/testrunner/runners/policy/policy.go`

## testscript interface signature (`unparam`)

- [ ] Testscript command handlers in `internal/testrunner/script/` must match the `testscript.Cmd` signature `func(ts *testscript.TestScript, neg bool, args []string)`. The `neg bool` parameter is required by the interface even when unused. Consider upstreaming a type alias or interface annotation to make this explicit without suppression.

PR #3510, opened 2026-04-28
Reduce complexity on top complex functions

jsoriano:
Reduce complexity on more complex functions according to gocyclo and gocognit.

Part of https://github.com/elastic/elastic-package/issues/3443.


internal/testrunner/script/script.go
```diff
@@ -131,75 +129,289 @@ func Run(dst *[]testrunner.TestResult, w io.Writer, opt Options) error {
...
 	}
-	var (
-		pkgRoot string
+	if len(pkgInfo.dirs) == 0 {
+		return nil
+	}
+
+	t, err := newT(opt, dst, w)
+	if err != nil {
+		return nil
```
mrodm: Should this be returning the error ?


```suggestion
		return err
```

internal/testrunner/script/script.go
```diff
@@ -131,75 +129,289 @@ func Run(dst *[]testrunner.TestResult, w io.Writer, opt Options) error {
...
+	}
+	revisions, err := revisionsFromRegistry(eprBaseURL, prof, manifest.Name)
+	if err != nil {
+		return nil, "", false, err
+	}
+	if len(revisions) > 0 && pkgInfo.currSemver != nil {
+		latestEPRVersion = revisions[len(revisions)-1].Version
+		latestSemver, err := semver.NewVersion(latestEPRVersion)
+		if err != nil {
+			return nil, "", false, fmt.Errorf("failed to parse latest epr version: %w", err)
```
mrodm: Would help debugging the error if the actual version string is added to the error message?


```suggestion
			return nil, "", false, fmt.Errorf("failed to parse latest epr version %q: %w", latestEPRVersion, err)
```

internal/testrunner/script/script.go
```diff
@@ -131,75 +129,289 @@ func Run(dst *[]testrunner.TestResult, w io.Writer, opt Options) error {
...
+	return manifest, latestEPRVersion, isLatestVersion, nil
+}
 
-	var stdinTempFile string
+func newT(opt Options, dst *[]testrunner.TestResult, w io.Writer) (*T, error) {
 	t := &T{
-		pkg: opt.Package,
-
-		verbose:       opt.Verbose,
-		stdinTempFile: stdinTempFile,
```
mrodm, line 199: This variable `stdinTempFile` is not present in the `T` struct from newT function. Is that ok ?

internal/testrunner/script/script.go
```diff
@@ -131,75 +129,289 @@ func Run(dst *[]testrunner.TestResult, w io.Writer, opt Options) error {
...
 
-		latestEPRVersion string
-		currVersion      string
-		prevVersion      string
-		currSemver       *semver.Version
-		breakingChange   bool
-	)
+	pkgRoot, err := packages.FindPackageRoot()
+	if err != nil {
+		if err == packages.ErrPackageRootNotFound {
```
mrodm: This remains as it was before. Just wondering if it should be changed to use `errors.Is()`.

internal/testrunner/script/script.go
jsoriano: Yes, fixed in https://github.com/elastic/elastic-package/pull/3510/changes/ea709c9d9120fdb69ee7d8bcbaf3a6cdedbbc3ab, and added a linter rule to catch these ones.

internal/testrunner/script/script.go
jsoriano: It was never used :shrug: Should I completely remove this?

mrodm: The only usage is for replacing files in the log messages

https://github.com/elastic/elastic-package/blob/7a37cb2c723978de94a7545fc227d08c32aa5393/internal/testrunner/script/script.go#L611

but it looks like it was never set any stdin file.

I'd say this could be removed and if needed later it could be restored. WDYT ?

jsoriano: Removing in https://github.com/elastic/elastic-package/pull/3510/commits/a7bcf5e632a7b1a59c159969dae27668689d12e5.

scripts/ruleguard/rules/rules.go
```diff
@@ -0,0 +1,25 @@
...
+
+// Package gorules contains custom go-ruleguard rules consumed by gocritic.
+//
+// These rules are not compiled as part of the program; they are loaded by
+// the ruleguard checker. The build tag prevents `go build ./...` from
+// trying to compile them (the dsl import is not in go.mod).
+//
+//go:build ruleguard
+
+package gorules
```
mrodm, line 13: Since the folder is named `ruleguard`, should this package name be the same as the folder name ?

Or viceversa name the folder as the package ?

mrodm (approved):
LGTM
Just a question about package name vs folder name.

scripts/ruleguard/rules/rules.go
jsoriano: The package must be called `gorules`.

Renaming directory to `rules`, as in the ruleguard repository https://github.com/quasilyte/go-ruleguard/tree/master/rules.

jsoriano: Done in https://github.com/elastic/elastic-package/pull/3510/commits/a285f79f737a6d71f708ce0d806eb1b9f69418c5

PR #3522, opened 2026-05-05
Enable errorlint and some gocritic checks

jsoriano:
Enable some additional linter checks that can prevent potential bugs:
* `errorlint`:
  * Ensure `errors.Is`/`As` is used for code handling specific types of errors.  
  * Ensure wrapping is used when formatting errors.
* `gocritic`:
  * `appendAssign` ensures `append` result is not assigned to a different variable, to avoid issues depending on the capacity of the modified slice.
  * `offBy1` avoids unsafe uses of indexes in slices.

Part of https://github.com/elastic/elastic-package/issues/3443.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,8 +3,7 @@ version: "2"
 linters:
   enable:
     - contextcheck
-    # TODO: enable errorlint to detect issues with error comparisons, wrapping and so on.
-    #- errorlint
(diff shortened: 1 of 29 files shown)


mrodm (approved):
thanks!

teresaromero (approved):
🧹  housekeeping win!

PR #3523, opened 2026-05-05
fix: preserve pipeline format when adding reroute processors from routing_rules.yml

### LINKED ISSUE #2861: JSON format pipeline and having a routing_rules.yml file causing error while running pipeline tests
## Description
When implementing the pipeline for Corelight integration, we need a routing_rules.yml to route data to different indices. After adding routing_rules.yml, pipeline tests passes if the ingest pipeline is defined in YAML, but fail if the ingest pipeline is defined in JSON. 

## Affected package/files
Package: `corelight`
Data stream: `log`
Ingest pipeline (JSON): `packages/corelight/data_stream/log/elasticsearch/ingest_pipeline/default.json`
Routing rules: `packages/corelight/data_stream/log/routing_rules.yml`

## Steps to reproduce
- Clone the repo: `https://github.com/sharadcrest/integrations`
- Checkout to `corelight_json_pipeline` branch using `git checkout corelight_json_pipeline`
- Go to corelight's package directory
- Install corelight package using `elastic-package install`
- Run pipeline tests via elastic-package: `elastic-package test pipeline -v`

## Expected result

elastic-package installs the ingest pipeline successfully.
Pipeline tests complete without errors, the same as when the pipeline is authored in YAML.

## Actual result
Test run fails during pipeline installation when the ingest pipeline is in JSON format:
```
Error: error running package pipeline tests: could not complete test run: installing ingest pipelines failed: unexpected response status for PutPipeline (500): 500 Internal Server Error (pipelineName: default-1755089181175456561, path: /home/devuser/bitbucket/integrations/packages/corelight/data_stream/log/elasticsearch/ingest_pipeline/default.json): elasticsearch error (type=not_x_content_exception): Compressor detection can only be called on some xcontent bytes or compressed xcontent bytes
Root cause:
[
  {
    "type": "not_x_content_exception",
    "reason": "Compressor detection can only be called on some xcontent bytes or compressed xcontent bytes",
    "position": {
      "offset": 0,
      "start": 0,
      "end": 0
    }
  }
]
```
## Notes & observations

- Using a YAML ingest pipeline (equivalent logic) works: tests pass and the pipeline installs.
- The JSON file appears valid when opened locally, but Elasticsearch responds as if the body is not recognized as JSON/xcontent.
- This looks similar to issues caused by:
    -   UTF-8 BOM at the start of the file.
    -   Incorrect Content-Type or compression flags when uploading.
    -   Hidden/invalid bytes introduced by editors or CI tooling.

### Diagnostics performed

- Validate JSON syntax using:
`jq . packages/corelight/data_stream/log/elasticsearch/ingest_pipeline/default.json
`
Result: Parses successfully.

- Check for BOM or hidden bytes:
   ```
   HEX=$(xxd -p -l 3 "$file")
      if [ "$HEX" = "efbbbf" ]; then
          echo "  ⚠️  File has BOM (Byte Order Mark)"
          STATUS=1
      else
          echo "  ✅ No BOM"
      fi
   ```
   (If “BOM present”, It prints warning message.)
   Result: got `✅ No BOM` for all json pipelines.

- Try uploading ingest pipeline using dev tools:
  Endpoint:  `PUT _ingest/pipeline/my-default-pipeline` Body: ${content of default.json}
  Result:  `{ "acknowledged": true}`
  NOTE: If this yields the same not_x_content_exception, it confirms the payload bytes are the problem, not elastic-package.
## Environment
Elastic Stack: 8.18.0
elastic-package: v0.112.0
OS: Linux (dev VM)

Reproducibility: Always, when ingest pipeline is JSON.

## Workarounds tried
- Converting the pipeline to YAML format makes the tests pass. However, switching to YAML isn’t an option since Corelight already maintains 120+ pipelines in JSON, and elastic-package does support JSON pipelines, so we need to keep the existing format.

  **Note**: If the pipelines remain in JSON and routing_rules.yml is removed, the pipeline tests also pass—but since routing_rules.yml is required, this workaround cannot be applied.

mrodm:
Fixes #2861

## Problem

When a data stream uses a JSON-format ingest pipeline (`default.json`) combined with `routing_rules.yml`, `addRerouteProcessors` was always marshaling the modified pipeline back to YAML regardless of the original format. This caused Elasticsearch to reject the pipeline with a `not_x_content_exception` because it received YAML content where JSON was expected.

Additionally, `RerouteProcessor` lacked `json` struct tags, causing field names to be serialized with uppercase initials (e.g., `Tag`, `If`) when converted to JSON, producing invalid processor definitions.

## Changes

- Propagate the pipeline format through `addRerouteProcessors` and use format-appropriate unmarshal/marshal (JSON or YAML).
- Add `json` struct tags to `RerouteProcessor` to ensure correct lowercase field names when serialized to JSON.
- Extract the `format` variable in `LoadIngestPipelineFiles` to avoid computing it twice.
- Add a unit test (`TestAddRerouteProcessorsJSONPipeline`) covering the JSON pipeline + routing rules code path.
- Add a new `log_json_pipeline` data stream to the `awsfirehose` test package with pipeline tests and routing rules to exercise the fix end-to-end.

## Testing

- New unit test verifies the output is valid JSON and contains the expected processors.
- New integration test data (`test-cloudtrail.json`, `test-vpcflow-log.json`) covers the rerouting logic for JSON pipelines.

## Author's checklist

- [x] Validate that keeps working with YAML pipelines and `routing_rules.yml` file.
- [x] Validate that works with JSON pipeline and `routing_rules.yml` file.
- [x] Validate that works with YAML pipelines without `routing_rules.yml` file.
    - Tested with `elastic_package_registry` from integrations repository 
- [x] Validate that works with JSON pipelines without `routing_rules.yml` file.
    - Tested with `mysql` package (data stream `slowlog`) from integrations repository

This PR can be tested via:
```sh
elastic-package -C test/packages/parallel/awsfirehose test pipeline -v
elastic-package stack down -v
```

When this is tested with elastic-package version `v0.123.0`, this gives the following error:
```
Error: error running package pipeline tests: could not complete test run: installing ingest pipelines failed: unexpected response status for PutPipeline (400): 400 Bad Request (pipelineName: default-1777987633876933265, path: /home/mariorodriguez/Coding/work/elastic-package/test/packages/parallel/awsfirehose/data_stream/log_json_pipeline/elasticsearch/ingest_pipeline/default.json): elasticsearch error (type=x_content_parse_exception): [1:13] Unrecognized token 'description': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (byte[])"description: Pipeline for rerouting logs streams from Amazon Kinesis Data Firehose.
on_failure:
    - set:
        field: error.message
        value: '{{ _ingest.on_failure_message }}'
processors:
    - set:
        field: ecs.version
        value: 8.0.0
    - set:
        field: cloud.provider
        value: aws
    - reroute:
        tag: awsfirehose.log_json_pipeline
        if: ctx.aws?.labels != null
        dataset:
            - '{{aws.labels.elastic_co/dataset}}'
        namespace:
   "[truncated 270 bytes]; line: 1, column: 13]
Root cause:
[
  {
    "type": "x_content_parse_exception",
    "reason": "[1:13] Unrecognized token 'description': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')\n at [Source: (byte[])\"description: Pipeline for rerouting logs streams from Amazon Kinesis Data Firehose.\non_failure:\n    - set:\n        field: error.message\n        value: '{{ _ingest.on_failure_message }}'\nprocessors:\n    - set:\n        field: ecs.version\n        value: 8.0.0\n    - set:\n        field: cloud.provider\n        value: aws\n    - reroute:\n        tag: awsfirehose.log_json_pipeline\n        if: ctx.aws?.labels != null\n        dataset:\n            - '{{aws.labels.elastic_co/dataset}}'\n        namespace:\n   \"[truncated 270 bytes]; line: 1, column: 13]",
    "position": {
      "offset": 0,
      "start": 0,
      "end": 0
    }
  }
]
```

When running the code in this PR, it runs pipeline tests successfully:
```sh
--- Test results for package: awsfirehose - START ---
╭─────────────┬───────────────────┬───────────┬──────────────────────────────────────────────────┬────────┬──────────────╮
│ PACKAGE     │ DATA STREAM       │ TEST TYPE │ TEST NAME                                        │ RESULT │ TIME ELAPSED │
├─────────────┼───────────────────┼───────────┼──────────────────────────────────────────────────┼────────┼──────────────┤
│ awsfirehose │ log               │ pipeline  │ (ingest pipeline warnings test-cloudtrail.json)  │ PASS   │ 177.075413ms │
│ awsfirehose │ log               │ pipeline  │ (ingest pipeline warnings test-vpcflow-log.json) │ PASS   │ 181.771697ms │
│ awsfirehose │ log               │ pipeline  │ test-cloudtrail.json                             │ PASS   │  30.685066ms │
│ awsfirehose │ log               │ pipeline  │ test-vpcflow-log.json                            │ PASS   │  28.595604ms │
│ awsfirehose │ log_json_pipeline │ pipeline  │ (ingest pipeline warnings test-cloudtrail.json)  │ PASS   │ 178.171931ms │
│ awsfirehose │ log_json_pipeline │ pipeline  │ (ingest pipeline warnings test-vpcflow-log.json) │ PASS   │ 179.515222ms │
│ awsfirehose │ log_json_pipeline │ pipeline  │ test-cloudtrail.json                             │ PASS   │  28.200119ms │
│ awsfirehose │ log_json_pipeline │ pipeline  │ test-vpcflow-log.json                            │ PASS   │  29.053974ms │
╰─────────────┴───────────────────┴───────────┴──────────────────────────────────────────────────┴────────┴──────────────╯
--- Test results for package: awsfirehose - END   ---
Done

```

---

### DIFF
--- a/internal/elasticsearch/ingest/datastream.go
+++ b/internal/elasticsearch/ingest/datastream.go
@@ -7,6 +7,7 @@ package ingest
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -20,6 +21,7 @@ import (
 
 	"github.com/elastic/elastic-package/internal/elasticsearch"
 	"github.com/elastic/elastic-package/internal/files"
+	"github.com/elastic/elastic-package/internal/formatter"
 	"github.com/elastic/elastic-package/internal/packages"
 )
 
@@ -43,10 +45,10 @@ type RoutingRule struct {
 }
 
 type RerouteProcessor struct {
-	Tag       string   `yaml:"tag"`
-	If        string   `yaml:"if"`
-	Dataset   []string `yaml:"dataset"`
-	Namespace []string `yaml:"namespace"`
+	Tag       string   `yaml:"tag" json:"tag"`
+	If        string   `yaml:"if" json:"if"`
+	Dataset   []string `yaml:"dataset" json:"dataset"`
+	Namespace []string `yaml:"namespace" json:"namespace"`
 }
 
 func InstallDataStreamPipelines(ctx context.Context, api *elasticsearch.API, dataStreamRoot string, repositoryRoot *os.Root) (string, []Pipeline, error) {
@@ -109,7 +111,8 @@ func LoadIngestPipelineFiles(dataStreamRoot string, nonce int64, repositoryRoot 
 			return nil, err
 		}
 
-		cWithRerouteProcessors, err := addRerouteProcessors(c, dataStreamRoot, path)
+		format := filepath.Ext(strings.TrimSuffix(path, ".link"))[1:]
+		cWithRerouteProcessors, err := addRerouteProcessors(c, dataStreamRoot, path, format)
 		if err != nil {
 			return nil, err
 		}
@@ -118,7 +121,7 @@ func LoadIngestPipelineFiles(dataStreamRoot string, nonce int64, repositoryRoot 
 		pipelines = append(pipelines, Pipeline{
 			Path:            path,
 			Name:            GetPipelineNameWithNonce(name[:strings.Index(name, ".")], nonce),
-			Format:          filepath.Ext(strings.TrimSuffix(path, ".link"))[1:],
+			Format:          format,
 			Content:         cWithRerouteProcessors,
 			ContentOriginal: c,
 		})
@@ -126,7 +129,7 @@ func LoadIngestPipelineFiles(dataStreamRoot string, nonce int64, repositoryRoot 
 	return pipelines, nil
 }
 
-func addRerouteProcessors(pipeline []byte, dataStreamRoot, pipelinePath string) ([]byte, error) {
+func addRerouteProcessors(pipeline []byte, dataStreamRoot, pipelinePath, format string) ([]byte, error) {
 	// Only attach routing_rules.yml reroute processors after the default pipeline
 	filename := filepath.Base(pipelinePath)
 	if filename != defaultPipelineJSON && filename != defaultPipelineYML &&
@@ -137,20 +140,26 @@ func addRerouteProcessors(pipeline []byte, dataStreamRoot, pipelinePath string) 
 	// Read routing_rules.yml and convert it into reroute processors in ingest pipeline
 	rerouteProcessors, err := loadRoutingRuleFile(dataStreamRoot)
 	if err != nil {
-		return nil, fmt.Errorf("failed loading routing rules: %v", err)
+		return nil, fmt.Errorf("failed loading routing rules: %w", err)
 	}
 	if len(rerouteProcessors) == 0 {
 		return pipeline, nil
 	}
 
-	var yamlPipeline map[string]any
-	err = yaml.Unmarshal(pipeline, &yamlPipeline)
-	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal ingest pipeline YAML data (path: %s): %w", pipelinePath, err)
+	var pipelineMap map[string]any
+	switch format {
+	case "json":
+		if err = formatter.JSONUnmarshalUsingNumber(pipeline, &pipelineMap); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal ingest pipeline JSON data (path: %s): %w", pipelinePath, err)
+		}
+	default:
+		if err = yaml.Unmarshal(pipeline, &pipelineMap); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal ingest pipeline YAML data (path: %s): %w", pipelinePath, err)
+		}
 	}
 
 	var processors []any
-	v, found := yamlPipeline["processors"]
+	v, found := pipelineMap["processors"]
 	if found {
 		list, ok := v.([]any)
 		if !ok {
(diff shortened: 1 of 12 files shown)


internal/elasticsearch/ingest/datastream.go
```diff
@@ -109,7 +111,8 @@ func LoadIngestPipelineFiles(dataStreamRoot string, nonce int64, repositoryRoot
 			return nil, err
 		}
 
-		cWithRerouteProcessors, err := addRerouteProcessors(c, dataStreamRoot, path)
+		format := filepath.Ext(strings.TrimSuffix(path, ".link"))[1:]
```
teresaromero: this might not happen because of the context around the function, but if path is ".link" or does not have extension after the [trim](https://pkg.go.dev/path/filepath#Ext) , getting `format` could trigger a panic because it will be an empty string[1:]. Can we add a safe check that path is not empty?

mrodm: This was already used previously:
```go
		pipelines = append(pipelines, Pipeline{
			Path:            path,
			Name:            GetPipelineNameWithNonce(baseName, nonce),
			Format:          filepath.Ext(strings.TrimSuffix(path, ".link"))[1:],
			Content:         cWithRerouteProcessors,
			ContentOriginal: c,
		})
```

According to the spec, these files could be just `json, yml and link`:
https://github.com/elastic/package-spec/blob/6d1dec05042b45b595230450e9d4c66fb9e30caf/spec/integration/elasticsearch/spec.yml#L13-L34


And the list of `pipelineFiles` is built the previous for loop
```go
	for _, pattern := range []string{"*.json", "*.yml", "*.link"}
```

I'll check in any case if it could be added a safeguard in case something changes in the spec to be safe.

mrodm: Updated in https://github.com/elastic/elastic-package/pull/3523/commits/2b5531b699dabc4db23fe9951a43473c5600b2b1

Thanks!

teresaromero: >This was already used previously:

i know.. but we might as well leave the room a bit cleaner :D thanks!

teresaromero (approved):

internal/elasticsearch/ingest/datastream.go
mrodm: > i know.. but we might as well leave the room a bit cleaner :D thanks!

Totally agreed! :+1:

PR #3531, opened 2026-05-07
support for 'compilation' of new dashboards-as-code format via `elastic-package build`

tommyers-elastic:
Lets package authors write dashboards in the new dashboards-as-code JSON format under `_dev/shared/*.json`. 

During build, each dashboard is installed to Kibana via the new API, exported back through the standard dashboards pipeline, and written to `kibana/dashboard/<package>-<id>.json`. The imported dashboard is cleaned up in Kibana. 

`<id>` is derived from the source filename, so output is stable across rebuilds. Requires Kibana ≥ 9.4.0; build fails with the usual `ErrUnavailableStack` when the source dir is present and no stack is reachable. Packages without `_dev/shared/*.json` still build offline.

End-to-end test: `test/packages/dashboards_as_code/sample/` plus `scripts/test-check-packages-dashboards-as-code.sh`, which brings up a stack and asserts the compiled SO is produced (the SO is deliberately not committed).


cmd/build.go
```diff
@@ -122,3 +134,22 @@ func buildCommandAction(cmd *cobra.Command, args []string) error {
 	cmd.Println("Done")
 	return nil
 }
+
+// hasDashboardsAsCode reports whether the package source contains any
+// dashboards-as-code JSON files that would require Kibana to compile.
+func hasDashboardsAsCode(packageRoot string) bool {
```
teresaromero: i would prefer to have a flag in technical preview to enable this on the build. the flag will enable a kibana client; if packages have dasboards is ok, if not is just skipped.

internal/builder/dashboards_as_code.go
```diff
@@ -0,0 +1,125 @@
...
+	if kibanaClient == nil {
+		return fmt.Errorf("package contains %s but no Kibana client is configured; "+
+			"set ELASTIC_PACKAGE_KIBANA_HOST or run 'elastic-package stack up' first", dashboardsAsCodeDir)
+	}
+
+	versionInfo, err := kibanaClient.Version()
+	if err != nil {
+		return fmt.Errorf("getting Kibana version information: %w", err)
+	}
+	if err := checkDashboardsAsCodeKibanaVersion(versionInfo); err != nil {
```
teresaromero, line 58: is this not checked by the package validation? when the version of a package is set with its kibana version?

internal/builder/dashboards_as_code.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return fmt.Errorf("importing dashboards-as-code failed: %w", err)
+	}
+
+	// Best-effort cleanup of the imported dashboard, regardless of how the
+	// rest of this function completes. Use a fresh context so cleanup runs
+	// even if ctx has been cancelled by the time we reach this point.
+	defer func() {
+		if cleanupErr := kibanaClient.DeleteDashboard(context.Background(), id); cleanupErr != nil {
+			if errors.Is(cleanupErr, context.Canceled) {
```
teresaromero: why is context canceled returned without log? why we are using a different context for this task, perhaps we want to add a timeout ?

internal/kibana/dashboards_as_code.go
```diff
@@ -0,0 +1,65 @@
...
+
+// DashboardsAPI is the prefix for the Kibana dashboards-as-code import API.
+const DashboardsAPI = "/api/dashboards"
+
+// ImportDashboardAsCode imports a dashboards-as-code JSON document via the
+// /api/dashboards Kibana endpoint, storing it under the supplied id (the PUT
+// form of the API guarantees the resulting saved object uses the id we choose,
+// rather than a freshly generated one). Returns the id reported by Kibana,
+// which should match the supplied id.
+func (c *Client) ImportDashboardAsCode(ctx context.Context, id string, body []byte) (string, error) {
```
teresaromero, line 25: I am not following the function, sorry. So ImportDashboardAsCode means: 
- attach a body and update whatever dashboard is at kibana? with the given id, if not exists, creates it? from the elastic-package pov, isnt this exporting it to kibana? where is the object saved?
OR
- is this sending the body to kibana and responding it with a good-to-go dashboard instead? in that case... can we rename the function, i feel Import semantics feels is a one way road, but is actually 2 ways? although i dont see where it saves it :D

tommyers-elastic: this creates or updates the dashboard in kibana. 

the new dashboard API handles the underlying saved object management. you can create, update, and delete dashboards via the new API without touching saved objects.

tommyers-elastic: the semantics of import/export in elastic-package follow this naming convention. 'export' is exporting objects _from_ kibana.

internal/builder/dashboards_as_code.go
tommyers-elastic: i'm not sure when this would fire. it looks just over the top defensive. the result of removing this is just that you'd see the debug log if the context was canceled for whatever reason. which is arguable the right thing anyway.

cmd/build.go
tommyers-elastic: good idea. will add.

internal/builder/dashboards_as_code.go
tommyers-elastic: the intention here is just to check if the stack being used to _build_ the package supports the API. it won't check vs the manifest version because existing check/build doesn't require a stack at all.

internal/builder/dashboards_as_code.go
tommyers-elastic: removed

cmd/build.go
tommyers-elastic: added `--compile-dashboards-as-code` flag to `build`

Issue #3524, opened 2026-05-05
validateIgnoredFields silently passes: runtime script never reads `_ignored`


andrewkroh:
## Summary

System tests are supposed to fail when indexed documents contain values in the `_ignored` metadata field, but they don't. The Painless runtime script that feeds the `ignored_fields` terms aggregation reads `_ignored` through the *stored-fields* lookup (`params['_fields']._ignored.values`), which returns nothing on at least Elasticsearch 9.3.4. The terms aggregation produces zero buckets, `validateIgnoredFields` sees an empty list, and the test passes — even when every document in the data stream has `_ignored` populated.

This defeats the entire `_ignored`-fields safety net that was added in #1738 ("Extend system test to validate absence of `_ignored`", b33490b6).

## How it was detected

While triaging an unrelated bug in the `netflow` integration, `elastic-package test system --data-streams log` against `packages/netflow` v2.25.0 **passed** on a 9.3.4 stack despite the indexed documents containing:

```json
{
  "_index": ".ds-logs-netflow.log-79901-2026.05.05-000001",
  "_ignored": ["event.created", "netflow.exporter.timestamp"],
  "_source": {
    "event": { "created": {} },
    "netflow": { "exporter": { "timestamp": {} } }
  },
  "ignored_field_values": {
    "event.created":             [{}],
    "netflow.exporter.timestamp": [{}]
  }
}
```

All 29 documents in the data stream had `_ignored` set on both fields. The package does **not** configure `skip_ignored_fields`, both fields are declared as `date` (`event.created` via ECS import, `netflow.exporter.timestamp` in `package-fields.yml`), and the test config does not set `skip_reason`/`skip_link`. The validation should have fired but didn't.

## Why it's failing

The `_ignored` metadata field is exposed via doc-values, **not** stored fields. On Elasticsearch 9.3.4, `params['_fields']['_ignored']` resolves to a `FieldLookup` whose `getValues()` returns an empty list for `_ignored`, even on documents whose search response clearly includes `"_ignored": [...]`.

Reproduction against the same cluster, same index, same query body as elastic-package:

```bash
# 1. Confirm 29 docs have _ignored populated
$ curl -sk -u elastic:changeme \
    'https://localhost:9200/logs-netflow.log-*/_search?size=0' \
    -H 'Content-Type: application/json' \
    -d '{"query":{"exists":{"field":"_ignored"}}}'
# → "hits": { "total": { "value": 29, "relation": "eq" } }

[...]

PR #3571, opened 2026-05-15
Add opt-in ignored fields doc-values query

mturac:
Refs #3524

## What changed
- Keep the default ignored-fields query on the existing `_fields` lookup to avoid changing every system test at once.
- Add `ELASTIC_PACKAGE_IGNORED_FIELDS_USE_DOC_VALUES=true` as a targeted opt-in for stricter doc-values based `_ignored` detection.
- Cover both query modes, the default exported query body, and the env switch with regression tests.

## Tests
- `CGO_ENABLED=0 go test ./internal/testrunner/runners/system -run 'Test(FieldsQueryIgnoredFieldsScript|IgnoredFieldsUseDocValues|NewConfig)$'`
- `CGO_ENABLED=0 go test ./internal/testrunner/runners/system ./internal/testrunner/script`
- `CGO_ENABLED=0 go test ./internal/testrunner/...`
- `make format`
- `git diff --check`

### DIFF
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -1007,6 +1007,9 @@ In this case, `elastic-package test system` will fail with an error and print a 
 skip_ignored_fields:
   - field.to.ignore
 ```
+
+For targeted verification runs, set `ELASTIC_PACKAGE_IGNORED_FIELDS_USE_DOC_VALUES=true` to read `_ignored` from doc values. This stricter lookup can surface ignored fields that the default `_fields` lookup misses.
+
 ### Kibana policy overrides
 
 If you need to test a system test with a Kibana policy override you can do that by setting the environment variable `ELASTIC_PACKAGE_KIBANA_POLICY_OVERRIDES` to be a path to a yaml file that contains the policy override.  For example:
--- a/internal/testrunner/runners/system/tester.go
+++ b/internal/testrunner/runners/system/tester.go
@@ -43,7 +43,13 @@ import (
 	"github.com/elastic/elastic-package/internal/wait"
 )
 
-const FieldsQuery = `{
+const (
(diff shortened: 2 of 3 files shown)


jsoriano:
test integrations

jsoriano (commented):
Change LGTM, let's wait to tests.

jsoriano:
This change produces failures in many packages, see https://github.com/elastic/integrations/pull/19022

They may be legit, but we need to think on an strategy to enable this in a controlled way, so we don't need to review and fix all packages at once.

mturac:
Thanks for checking this. That makes sense, I had only scoped the change to the query behavior, but the package-level fallout means the rollout needs to be narrower.

I’ll rework this so the stricter ignored-fields check can be enabled in a controlled way instead of forcing all affected packages to be fixed in one pass.

mturac:
Thanks again for the earlier package-matrix check. I reworked this to keep the default ignored-fields query on `params['_fields']._ignored.values`, so existing package tests should stay on the current behavior.

The stricter `doc['_ignored']` lookup is now gated behind `ELASTIC_PACKAGE_IGNORED_FIELDS_USE_DOC_VALUES=true`, and that option is read once before the `waitForDocs` retry loop so retries use a stable query. I also added docs and unit coverage for both query modes.

Local validation:
- `CGO_ENABLED=0 go test ./internal/testrunner/runners/system -run 'TestFieldsQueryIgnoredFieldsScript|TestIgnoredFieldsUseDocValues' -count=1`
- `CGO_ENABLED=0 go test ./internal/testrunner/runners/system -count=1`
- `CGO_ENABLED=0 go test ./internal/testrunner/... -count=1`
- `gofmt -l internal/testrunner/runners/system/tester.go internal/testrunner/runners/system/tester_test.go`
- `git diff --check`

Could you trigger Buildkite again so this can be checked against the full package matrix on the current SHA?

PR #3572, opened 2026-05-15
Review readme template to avoid vale linting warnings on comments

teresaromero:
## Summary

- ~Adds a `validate-integration-docs` skill under `.skills/skills/` that runs `vale` against an integration package's markdown, filters false positives inside Go-template comment blocks, and re-lints the rendered output after `elastic-package build`.~
- ~Adds `.skills/README.md` explaining how to install the skills via `npx skills@latest add`.~
- Updates comments under the base docs readme template so vale errors for the template comments are fixed on all new integrations

~`elastic-package` is the reference tool for building packages, so it's a natural place to keep skills that assist with the development workflow. The immediate goal was to have a skill that helps fix vale warnings while working on integration docs, without having to remember the filtering logic for template comments every time.~

EDIT:

skill was moved to integrations repository so is available from the repo where the packages are developed. This change is only to align the template comments so we prevent vale to run the same warnings on every pr

🤖 Generated with [Claude Code](https://claude.com/claude-code)


criamico:
Just an observation: in other repos I've seen the skills under `.agents/skills`, probably it would be better to move it there so it's also automatically picked up by any agent.

internal/packages/archetype/_static/package-docs-readme.md.tmpl
```diff
@@ -8,15 +8,15 @@ Find more detailed documentation guidelines in https://www.elastic.co/docs/exten
 # {[.Manifest.Title]} Integration for Elastic
 
 ## Overview
-{{/* Complete this section with a short summary of what data this integration collects and the service it integrates with.}}
+{{/* Complete this section with a short summary of what data this integration collects and the service it integrates with. */}}
```
mrodm, line 11: Would these changes affect the current docs written in the packages? I guess not, but it would be good to confirm.

.skills/skills/validate-integration-docs/SKILL.md
```diff
@@ -0,0 +1,70 @@
+---
+name: validate-integration-docs
+description: Lint an Elastic integration package's markdown with vale and propose fixes. Use when the user says "vale check," "lint docs," "validate the README," "check integration markdown," or wants to polish an integration's docs before opening a PR. Filters out false positives inside Go-template comment blocks in _dev/build/docs/, then re-runs vale against the rendered docs/ output after `elastic-package build` as a final pass.
+compatibility: Requires `elastic-package` CLI and `vale` on PATH. Designed for packages in elastic/integrations (or any repo using the _dev/build/docs/README.md → docs/README.md layout produced by elastic-package build).
+license: Apache-2.0
+metadata:
+  origin: elastic/elastic-package
```
mrodm: This could be applied (copied?) to the elastic/integrations. It looks like it could help developers to solve the Vale lint warnigs or errors posted in their Pull Requests.

mrodm (approved):
LGTM

Just minor comments.

.skills/skills/validate-integration-docs/SKILL.md
teresaromero: yes, i was thinking on this repo or the integrations one 🤔  i will check if there is a mechanism to link repos or perpahs i will just move the pr into the integrations repository

internal/packages/archetype/_static/package-docs-readme.md.tmpl
teresaromero: i will have to test this, but my guess is this template is used when creating the integration, so existent wont be affected as the changes dont appear after running build... but i have to double check

teresaromero:
> Just an observation: in other repos I've seen the skills under `.agents/skills`, probably it would be better to move it there so it's also automatically picked up by any agent.

ok! the aim was to have a general folder and then `npx skills add `to include the skill on project or global  or with the structure of the prefered agent (claude does not use that convention), as prefered by the user. this seems to fit into a single repository skill, and as talked on other comment i will be moving it to integrations repository as it makes there more sense. thanks!

.skills/skills/validate-integration-docs/SKILL.md
teresaromero: i've moved the skills directory under .agent to the integrations repo 👍🏻

internal/packages/archetype/_static/package-docs-readme.md.tmpl
teresaromero: tested with current packages to lint, check and build and there is no changes. this will only affect new packages using the tool when this change is released.

PR #3573, opened 2026-05-15
testrunner/runners/system: fall back to packageRoot in CreatePackagePolicy when built tree is unavailable

### LINKED ISSUE #3552: `add_package_policy -version` in script tests fails: `CreatePackagePolicy` resolves to build directory instead of EPR-extracted root
## Description

Since v0.121.0, script tests that use `add_package_policy -version <epr_version>` to create a policy against an older EPR package version fail with Fleet validation errors for required variables:

```
Package policy is invalid:
  inputs.cel.streams.o365.audit.vars.azure_tenant_id: Directory (tenant) ID is required
  inputs.cel.streams.o365.audit.vars.client_id: Application (client) ID is required
  inputs.cel.streams.o365.audit.vars.client_secret: Client Secret is required
```

The request body sent to Fleet contains no `streams` in the enabled input — only the enable/disable flags — so Fleet auto-creates stream entries with no user-provided vars, then rejects the policy because required fields have no defaults.

## Root cause

PR [#3474](https://github.com/elastic/elastic-package/pull/3474) (v0.123.0) changed `CreatePackagePolicy` in [`internal/testrunner/runners/system/tester.go`](https://github.com/elastic/elastic-package/blob/8c461517c3e2ec8af91a026ae48f526e14729430/internal/testrunner/runners/system/tester.go) to support composable integrations:

```go
// Match Fleet's view of the package: composable integrations use the built tree (requires.input).
root, err := builder.BuildPackagesDirectory(packageRoot, "")
// ...
allDatastreams, err := packages.ReadAllDataStreamManifests(root)
```

`BuildPackagesDirectory` reads the manifest at `packageRoot` to get the package name and version, then constructs a path under the repo's `build/packages/` directory:

```
build/packages/<name>/<version>
```

When `packageRoot` is the EPR-extracted directory (e.g. for `add_package_policy -version 3.8.1`), this resolves to `build/packages/o365/3.8.1`. That directory does not exist — only the dev version (`build/packages/o365/3.8.2`) is present. `ReadAllDataStreamManifests` silently returns `nil` for a non-existent path ([`packages.go` L731-732](https://github.com/elastic/elastic-package/blob/8ae01073f0bbd21ef5ff7f1bd208eaa731cd5e7e/internal/packages/packages.go#L731-L732)), so `buildStreamsForInput` iterates an empty slice and produces an empty streams map. The `omitempty` JSON tag on `Streams` drops it from the request.

Before #3474, `CreatePackagePolicy` passed `packageRoot` directly to `ReadAllDataStreamManifests`, which read the EPR-extracted manifests correctly.

## Reproduction

Any script test using the `install_package_from_registry` + `add_package_policy -version` pattern will fail when the EPR version differs from the dev version. Minimal reproduction:

```
install_package_from_registry -profile ... ${PACKAGE_NAME} ${LATEST_EPR_VERSION}
add_package_policy -profile ... -version ${LATEST_EPR_VERSION} test_config.yaml DATA_STREAM_NAME
```

Observed in the [o365 audit upgrade script test](https://github.com/elastic/integrations/blob/main/packages/o365/data_stream/audit/_dev/test/scripts/upgrade.txt).

## Affected versions

- v0.123.0 (first release containing #3474)
- v0.122.0 and earlier are not affected

## Suggested fix

`CreatePackagePolicy` should use `packageRoot` directly (as it did before) when the package root is not under the repository's source tree — i.e. when it is an EPR-extracted directory. The `BuildPackagesDirectory` indirection is only meaningful for composable integrations that need a built tree with bundled input packages.

One approach: fall back to `packageRoot` when `builder.BuildPackagesDirectory` resolves to a path that does not exist.

```go
root, err := builder.BuildPackagesDirectory(packageRoot, "")
if err != nil {
    return ..., fmt.Errorf("error locating built package directory: %w", err)
}
if _, err := os.Stat(filepath.Join(root, "data_stream")); errors.Is(err, os.ErrNotExist) {
    root = packageRoot
}
allDatastreams, err := packages.ReadAllDataStreamManifests(root)
```

kcreddy:
```
testrunner/runners/system: fall back to packageRoot in CreatePackagePolicy when built tree is unavailable

CreatePackagePolicy called ReadBuiltPackageManifest unconditionally,
which assumed the built package directory always exists and is reachable
from a Git repository. This fails when packageRoot is an EPR-extracted
package (add_package_policy -version) because the built tree contains a
different version, or when running outside a Git repository where no
build/ ancestor exists.

ReadAllDataStreamManifests silently returned nil for the non-existent
built path, producing an empty Streams map that Fleet rejected because
required variables had no defaults.

The fix is in CreatePackagePolicy: when ReadBuiltPackageManifest fails,
fall back to reading manifests directly from packageRoot. This keeps
ReadBuiltPackageManifest strict — callers that genuinely need the built
tree (composable integrations) still get a hard error if the build is
missing. An EPR package is already fully resolved (no package:
references), so the source manifest is accurate enough.

The fallback is logged at debug level so it appears in verbose output
without masking real build failures.

Fixes #3552

Assisted-by Cursor (claude-opus-4.6)
```

### DIFF
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -112,27 +112,16 @@ func BuildPackagesDirectory(packageRoot string, buildDir string) (string, error)
 
 // ReadBuiltPackageManifest locates the built package directory for packageRoot
 // and reads its manifest. Returns the built root path and parsed manifest.
-//
-// When the built directory does not exist or does not contain a valid manifest,
-// packageRoot is used directly. This handles two cases:
-//   - EPR-extracted packages passed via add_package_policy -version, where the
-//     built tree contains a different version than the one being installed.
-//   - Execution outside a Git repository (e.g. temp directories) where no
-//     build/ directory can be found.
 func ReadBuiltPackageManifest(packageRoot string) (string, *packages.PackageManifest, error) {
 	builtRoot, err := BuildPackagesDirectory(packageRoot, "")
-	if err == nil {
-		builtPkg, readErr := packages.ReadPackageManifestFromPackageRoot(builtRoot)
-		if readErr == nil {
-			return builtRoot, builtPkg, nil
-		}
-	}
-	// Built tree unavailable or incomplete; fall back to packageRoot.
-	pkg, err := packages.ReadPackageManifestFromPackageRoot(packageRoot)
 	if err != nil {
-		return "", nil, fmt.Errorf("reading package manifest from %s: %w", packageRoot, err)
+		return "", nil, err
 	}
-	return packageRoot, pkg, nil
+	builtPkg, err := packages.ReadPackageManifestFromPackageRoot(builtRoot)
+	if err != nil {
+		return "", nil, err
+	}
+	return builtRoot, builtPkg, nil
 }
 
 // buildPackagesZipPath function returns the path to zipped built package.
--- a/internal/builder/packages.go
+++ b/internal/builder/packages.go
@@ -112,16 +112,27 @@ func BuildPackagesDirectory(packageRoot string, buildDir string) (string, error)
 
 // ReadBuiltPackageManifest locates the built package directory for packageRoot
 // and reads its manifest. Returns the built root path and parsed manifest.
+//
+// When the built directory does not exist or does not contain a valid manifest,
+// packageRoot is used directly. This handles two cases:
+//   - EPR-extracted packages passed via add_package_policy -version, where the
+//     built tree contains a different version than the one being installed.
+//   - Execution outside a Git repository (e.g. temp directories) where no
+//     build/ directory can be found.
 func ReadBuiltPackageManifest(packageRoot string) (string, *packages.PackageManifest, error) {
 	builtRoot, err := BuildPackagesDirectory(packageRoot, "")
-	if err != nil {
-		return "", nil, err
+	if err == nil {
+		builtPkg, readErr := packages.ReadPackageManifestFromPackageRoot(builtRoot)
+		if readErr == nil {
+			return builtRoot, builtPkg, nil
+		}
 	}
-	builtPkg, err := packages.ReadPackageManifestFromPackageRoot(builtRoot)
+	// Built tree unavailable or incomplete; fall back to packageRoot.
+	pkg, err := packages.ReadPackageManifestFromPackageRoot(packageRoot)
 	if err != nil {
-		return "", nil, err
+		return "", nil, fmt.Errorf("reading package manifest from %s: %w", packageRoot, err)
 	}
-	return builtRoot, builtPkg, nil
+	return packageRoot, pkg, nil
 }
 
 // buildPackagesZipPath function returns the path to zipped built package.
--- a/internal/builder/packages_test.go
+++ b/internal/builder/packages_test.go
@@ -5,6 +5,7 @@
 package builder
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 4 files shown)


internal/builder/packages.go
```diff
@@ -112,16 +112,27 @@ func BuildPackagesDirectory(packageRoot string, buildDir string) (string, error)
...
-		return "", nil, err
+	if err == nil {
+		builtPkg, readErr := packages.ReadPackageManifestFromPackageRoot(builtRoot)
+		if readErr == nil {
+			return builtRoot, builtPkg, nil
+		}
 	}
-	builtPkg, err := packages.ReadPackageManifestFromPackageRoot(builtRoot)
+	// Built tree unavailable or incomplete; fall back to packageRoot.
+	pkg, err := packages.ReadPackageManifestFromPackageRoot(packageRoot)
```
jsoriano: This is error prone, a caller could silently fallback to the source manifest when they are interested on the built one.

The caller should take the responsibility of falling back to the source manifest if it thinks it is accurate enough.

If the caller is interested only in fields that appear on both kinds of manifests (such as the name and version), and is not concerned about other fields that could change in build time they they could directly use the package root.

But in general a caller that is interested on the built files, such as most testers, should ensure first that they have a fully built package.

kcreddy: Good point — the silent fallback in `ReadBuiltPackageManifest` is too broad and would hide real build problems for callers that genuinely need the built tree.

I've moved the fallback into `CreatePackagePolicy`, which is the only caller that receives EPR-extracted roots (via `addPackagePolicy -version`). `ReadBuiltPackageManifest` is now strict again — it fails if the built tree doesn't exist or contains a different version.

The reasoning: an EPR package downloaded from the registry is already fully resolved (no `package:` references), so reading manifests directly from `packageRoot` is accurate enough to build a correct policy. Callers that need composable resolution (the normal system test path, `fleetpolicy.go`, etc.) still get a hard error if the build is missing.

The fallback is logged at debug level so it's visible in `-v` output but doesn't mask problems silently.

Commit: [0901250](https://github.com/elastic/elastic-package/pull/3573/commits/0901250830d73f46a99da0996332f44df46f28c7)

jsoriano: `CreatePackagePolicy` must use a built manifest, we neither can silently fall back to the source manifest there.

Actually after re-reading #3552 I wonder if the problem is really related to use the built or the source manifest.

The failures are about required variables:
```
Package policy is invalid:
  inputs.cel.streams.o365.audit.vars.azure_tenant_id: Directory (tenant) ID is required
  inputs.cel.streams.o365.audit.vars.client_id: Application (client) ID is required
  inputs.cel.streams.o365.audit.vars.client_secret: Client Secret is required
```

In https://github.com/elastic/elastic-package/pull/3307 we introduced the use of the newer "simplified" format of the Fleet policy API. This API is stricter regarding validation of required fields.

If in this case these fields are really required and not set, the test should be fixed to set the required variables.
If not possible, fall back to use the legacy API. This can be done in system and policy sets setting the `policy_api_format: legacy` parameter. Not sure what would be the best way to do this in script tests.

teresaromero:
@kcreddy  i am still figuring out what is the error and why this changes. perhaps is obvious but i am not understanding the following: why the error is reported at the script test runner when using certain flag, and the changes are about system test runner?

ref https://github.com/elastic/elastic-package/issues/3552 is the flag also used for system? 

thanks

internal/builder/packages.go
kcreddy: The required variables are set in the test config ([`test_config.yaml`](https://github.com/elastic/integrations/blob/main/packages/o365/data_stream/audit/_dev/test/scripts/upgrade.txt#L65-L74) lines 65-74 has: `client_id`, `client_secret`, `azure_tenant_id` set). I believe the problem isn't missing vars or the simplified API being stricter — it's that the vars never reach Fleet at all.

The chain:
1. `ReadBuiltPackageManifest(eprRoot)` resolves to `build/packages/o365/3.8.1/` — which doesn't exist (only `3.8.2` is built).
2. `ReadAllDataStreamManifests(builtRoot)` returns `nil` because that directory has no `data_stream/`.
3. `buildStreamsForInput` iterates over empty `datastreams`, producing an empty `Streams` map.
4. `json:"streams,omitempty"` drops `Streams` from the request.
5. Fleet receives `enabled: true` with no streams, auto-creates them with no user vars, and rejects because the required fields have no defaults.

The legacy API would have the same issue — no streams means no vars get sent regardless of format.

Regarding "must use a built manifest": this seems true for composable integrations where `package:` references need resolution. But `add_package_policy -version` passes an EPR-extracted package downloaded from the registry. It's already fully resolved — there are no `package:` references. The EPR manifest and the built manifest are identical for non-composable packages. The built tree simply doesn't exist for the EPR version because only the dev version is built locally.

The alternative would be to also build (or extract) the EPR version into the `build/` tree, but that's a much larger change for a path that doesn't need it.

Let me know what you think.

kcreddy:
> @kcreddy i am still figuring out what is the error and why this changes. perhaps is obvious but i am not understanding the following: why the error is reported at the script test runner when using certain flag, and the changes are about system test runner?
> 
> ref #3552 is the flag also used for system?
> 
> thanks

@teresaromero
The script test runner's `addPackagePolicy` command in `data_stream.go` calls `system.CreatePackagePolicy()` (line 156) — the function lives in the system package because it's shared between both runners. The script test runner doesn't have its own policy-building code; it delegates to the same `CreatePackagePolicy` function that system tests use.

So the bug surfaces via the script test command `add_package_policy -version`, but the code that actually builds the Fleet policy request (and where the wrong `builtRoot` gets resolved) is `CreatePackagePolicy` in `internal/testrunner/runners/system/tester.go`. That's why the fix is there — it's the shared function both runners go through.

internal/builder/packages.go
jsoriano: I am refactoring `add_package_policy` in https://github.com/elastic/elastic-package/pull/3576 to work only with built packages, and thus removing the conditions added for composable packages. Once built, composable packages have the same format as integration packages.

This should also fix https://github.com/elastic/elastic-package/issues/3552, please check if it works for you.

kcreddy: Verified locally — built elastic-package from your PR and ran the `o365.audit` script tests (including `upgrade.txt`, which was the original failing case in #3552). All four tests pass. Your approach is cleaner than the fallback in this PR; closing mine in favour of yours.

kcreddy:
Closing as per discussion: https://github.com/elastic/elastic-package/pull/3573#discussion_r3266304670

PR #3576, opened 2026-05-18
Ensure testers create policies from built packages

### LINKED ISSUE #3552: `add_package_policy -version` in script tests fails: `CreatePackagePolicy` resolves to build directory instead of EPR-extracted root

jsoriano:
Most testers should work with built packages. This is specially relevant as we introduce more processing in build time, as we recently did with composable packages, but we also do with external and ECS fields, dashboards and so on. The source of a package is not reliable for functional testing.

This change address this issue when creating policies from packages. System tester already builds and uses built packages, but some steps in script tester need more flexibility, while still ensuring that they are using built packages when needed.

The helper used to create policies and shared by the system and script runner is moved to the kibana package. It is refactored to expect already read manifests, so each tester can decide their source for built manifests.

System tester remains mostly as is, only delegating creation of policy templates to the kibana package.

`add_package_policy` in the script tester stops assuming that it can receive a source manifest. Decision on where to read the manifest from is now based on explicit parameters and not on features specific to composable packages. If `--version` is provided, the package is obtained from EPR, and thus it is already built, if `--version` is not provided, the content is read from a local build.

Apart from some unit tests, a new integration test is added that performs an upgrade test for a composable package using the script test runner.

Fixes https://github.com/elastic/elastic-package/issues/3552.


scripts/test-script-upgrade.sh
```diff
@@ -0,0 +1,132 @@
...
+# Remove the registry override before starting the full stack.
+sed -i '/^stack\.epr\.base_url:/d' \
+  ~/.elastic-package/profiles/script-upgrade/config.yml
+
+echo "--- Start full stack"
+elastic-package stack down -v
+elastic-package stack up -d -v
+elastic-package stack status
+
+eval "$(elastic-package stack shellinit)"
```
mrodm: This command should not be needed.

internal/testrunner/runners/system/tester.go
```diff
@@ -2267,119 +2268,31 @@ func CreatePackagePolicy(
...
-			v, _ := common.MapStr(stream.Vars).GetValue("data_stream.dataset")
-			ds, _ := v.(string)
-			if ds == "" {
-				continue
-			}
-
-			return ds
+	var builtDS *packages.DataStreamManifest
+	var allDatastreams []packages.DataStreamManifest
+	if builtPkg.Type != "input" {
```
mrodm: As the error is explicitly for integration packages, would it be better to change the condition for that ?


```suggestion
	if builtPkg.Type == "integration" {
```

internal/kibana/packagepolicy.go
```diff
@@ -214,6 +215,113 @@ func BuildInputPackagePolicy(
...
+// index/data-stream names. It is a pure function: all disk I/O must be done by the
+// caller before invoking it.
+//
+// For input-type packages pass dsName="" and dsManifest=nil; allDatastreams is ignored.
+// For integration packages dsManifest must not be nil and allDatastreams must contain
+// the manifests for all data streams under the package root that was actually installed
+// (built tree for system tests, EPR-extracted directory for script tests). dsManifest's
+// Name must appear in allDatastreams; the function errors out otherwise to catch the
+// #3552 class of bug where allDatastreams is loaded from a different/non-existent root.
+func CreatePackagePolicy(
```
mrodm: Should the methods in `internal/resources/fleetpolicy.go` use this new method to create the policies too?

https://github.com/elastic/elastic-package/blob/05d10d03fce587e02c9fc72fc2e74cd01a5377b0/internal/resources/fleetpolicy.go#L203-L209

https://github.com/elastic/elastic-package/blob/05d10d03fce587e02c9fc72fc2e74cd01a5377b0/internal/resources/fleetpolicy.go#L236-L241

internal/builder/packages_test.go
```diff
@@ -162,3 +163,70 @@ func TestCopyLicenseTextFile_UsesExistingLicenseFile(t *testing.T) {
...
+// TestReadBuiltPackageManifest pins the strict behavior of ReadBuiltPackageManifest:
+// it must locate the built tree at build/packages/<name>/<version>/ matching the
+// version in the source manifest, and fail otherwise. The script tester relies on
+// this strictness — when -version is set, it deliberately bypasses
+// ReadBuiltPackageManifest because it would resolve to a non-existent path
+// (the EPR version is not built locally).
+func TestReadBuiltPackageManifest(t *testing.T) {
+	t.Run("ReturnsBuiltTree", func(t *testing.T) {
+		root := t.TempDir()
+		t.Chdir(root)
```
mrodm, line 188: I know it is unrelated to this PR, but it would be good to update this other test to use `t.Chdir` too. It could be done in a separate PR too.

https://github.com/elastic/elastic-package/blob/05d10d03fce587e02c9fc72fc2e74cd01a5377b0/internal/files/linkedfiles_test.go#L838-L842


```go
	t.Run("repository root name must be absolute", func(t *testing.T) {
		t.Chdir(tempDir)
```

kcreddy (approved):
I've tested the original issue from https://github.com/elastic/integrations/pull/18957 with the build from this PR, and it works fine now. Thanks!

scripts/test-script-upgrade.sh
jsoriano: Done in https://github.com/elastic/elastic-package/pull/3576/commits/61f4622f4450bdf8c0da9ceda303a37ce886468d.

internal/testrunner/runners/system/tester.go
jsoriano: Done in https://github.com/elastic/elastic-package/pull/3576/commits/61f4622f4450bdf8c0da9ceda303a37ce886468d.

internal/kibana/packagepolicy.go
jsoriano: I was thinking on doing this in a follow up, but is not such a big change. Doing it here in https://github.com/elastic/elastic-package/pull/3576/commits/42861c80e5a02cedb62dd0016058645d7873ce71.

jsoriano:
@mrodm applied your suggestions, thanks! Also renamed the new policy builder from `CreatePackagePolicy` to `BuildPackagePolicy`, so the `Build` method builds the struct for the request and the `Create` method in the kibana client actually creates the policy.

internal/builder/packages_test.go
jsoriano: Done in https://github.com/elastic/elastic-package/pull/3576/commits/61f4622f4450bdf8c0da9ceda303a37ce886468d.

mrodm (approved):
Thanks!

teresaromero (commented):
sorry i was reviewing now and got merged. thanks for this change. i am still confused a but with built-source manifest/policies. i wonder if there should be a refactor to have separated structs and not using the same type for both "entities", now that they are diverging (composables)

jsoriano:
> i wonder if there should be a refactor to have separated structs and not using the same type for both "entities", now that they are diverging (composables)

Yeah, maybe we should have completely different structs and associated functions for source and built packages. Created issue about this: https://github.com/elastic/elastic-package/issues/3585

Manifests are diverging more now with the resolution of `requires`, impacting policy creation, but there are other differences that have been relevant in the past. For example resolution of `external` fields or automatic inclusion of ECS fields has also caused problems while validating documents. In general testers should use built packages so we don't need to think much about these differences.

PR #3589, opened 2026-05-21
Allow constant_keyword to keyword field overrides

### LINKED ISSUE #3583: Allow override constant_keyword->keyword
There is an `allowedTypeOverride` allows for overriding `keyword` fields to `constant_keyword`.
https://github.com/elastic/elastic-package/blob/05d10d03fce587e02c9fc72fc2e74cd01a5377b0/internal/fields/dependency_manager.go#L311

I'd like to propose adding an override exception for going the other way. In context of https://github.com/elastic/integrations/pull/19055, there are some transforms that need custom mapping for `data_stream.namespace` within the transform destination index need to be `keyword` to allow multiple namespaces to be processed.

resolvicomai:
## Summary
- allow external field imports to override `constant_keyword` fields to `keyword`
- keep the override allowlist narrow instead of allowing arbitrary type changes
- add regression coverage for `data_stream.namespace`

Fixes #3583

## Validation
- go test ./internal/fields
- git diff --check

### DIFF
--- a/internal/fields/dependency_manager.go
+++ b/internal/fields/dependency_manager.go
@@ -309,6 +309,7 @@ func allowedTypeOverride(fromType, toType string) bool {
 	}{
 		// Support the case of setting the value already in the mappings.
 		{"keyword", "constant_keyword"},
+		{"constant_keyword", "keyword"},
 
 		// Support objects in ECS where the developer must decide if using
 		// a group or nested object.
--- a/internal/fields/dependency_manager_test.go
+++ b/internal/fields/dependency_manager_test.go
@@ -83,6 +83,25 @@ func TestDependencyManagerInjectExternalFields(t *testing.T) {
 			valid:   true,
 		},
 		{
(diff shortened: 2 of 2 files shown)


internal/fields/dependency_manager.go
```diff
@@ -309,6 +309,7 @@ func allowedTypeOverride(fromType, toType string) bool {
 	}{
 		// Support the case of setting the value already in the mappings.
 		{"keyword", "constant_keyword"},
+		{"constant_keyword", "keyword"},
```
jsoriano, line 312: As a general rule it could be counterintuitive to have different namespaces in the same data_stream.

Please add a comment here mentioning the use case that motivates supporting this.
```suggestion

		// Transforms can contain documents from multiple data streams, with different data_stream.* values.
		// See https://github.com/elastic/elastic-package/issues/3583.
		{"constant_keyword", "keyword"},
```

PR #3592, opened 2026-05-22
Wire --defer-cleanup and --metrics-collection-interval flags to rally benchmark runner

### LINKED ISSUE #3591: [rally benchmarking] `--defer-cleanup` flag is registered but not passed to runner
## Summary

The `--defer-cleanup` flag is exposed on `elastic-package benchmark rally` but has no effect. Rally benchmarks delete ingested data immediately after the run completes, even when `--defer-cleanup` is set (e.g. `30m`).

`benchmark system` wires this flag correctly; `benchmark rally` does not.

## Steps to reproduce

1. Start the elastic-package stack (`elastic-package stack up`).
2. From an integration package directory, run a rally benchmark with deferred cleanup:

```bash
elastic-package benchmark rally \
  --benchmark <benchmark-name> \
  -v \
  --defer-cleanup 30m
```

3. The command returns immediately after results are printed (no 30-minute wait).
With -v, there is no log line `waiting for 30m0s before tearing down....`.

## Expected behavior

Expected waiting time before cleanup.

kcreddy:
```
Wire --defer-cleanup and --metrics-collection-interval flags to rally benchmark runner

The two flags were registered on the rally subcommand but never
extracted in rallyCommandAction or passed to the runner options.
The rally runner already checked DeferCleanup and MetricsInterval
in TearDown and startMetricsColletion, so they silently did nothing.

Add WithDeferCleanup and WithMetricsInterval option helpers to the
rally package, read both flags in rallyCommandAction, and thread them
into the options slice — matching how benchmark system already wires
them.

Fixes #3591
```

Closes #3591

### DIFF
--- a/cmd/benchmark.go
+++ b/cmd/benchmark.go
@@ -267,6 +267,16 @@ func rallyCommandAction(cmd *cobra.Command, args []string) error {
 		return cobraext.FlagParsingError(err, cobraext.BenchNameFlagName)
 	}
 
+	deferCleanup, err := cmd.Flags().GetDuration(cobraext.DeferCleanupFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.DeferCleanupFlagName)
+	}
+
+	metricsInterval, err := cmd.Flags().GetDuration(cobraext.BenchMetricsIntervalFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.BenchMetricsIntervalFlagName)
+	}
+
 	dataReindex, err := cmd.Flags().GetBool(cobraext.BenchReindexToMetricstoreFlagName)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/benchmark.go
```diff
@@ -267,6 +267,16 @@ func rallyCommandAction(cmd *cobra.Command, args []string) error {
 		return cobraext.FlagParsingError(err, cobraext.BenchNameFlagName)
 	}
 
+	deferCleanup, err := cmd.Flags().GetDuration(cobraext.DeferCleanupFlagName)
```
kcreddy, line 270: The `--defer-cleanup` flag was registered on `benchmark rally` but never passed to the runner, making it a silent no-op. Wiring it through lets users keep ingested data alive after a rally run — useful for loading dashboards with realistic high-volume data, inspecting index state, or debugging unexpected benchmark results. The same gap existed for `--metrics-collection-interval`, which is now also connected.

teresaromero (approved):
🚀  thanks!

PR #3593, opened 2026-05-22
feat: add `elastic-package requires update` command to bump requires pins from the registry

### LINKED ISSUE #17404: Automatically update package dependencies
In the context of the composable packages project, integration packages are going to start having dependencies on other packages (see https://github.com/elastic/package-spec/issues/1083).

It would be good to have a way to automate updates, so when a new input package is published, the integration packages that depend on it are also updated automatically.

For convenience, it would be good if the changes on packages are grouped in PRs by owner.

teresaromero:
## What

Adds `elastic-package requires update` — a new subcommand that reads `requires.input` and `requires.content` entries from `manifest.yml`, queries the package registry for the latest published versions, and writes the updated pins back to the file.

Key behaviours:

- Only applies to integration packages with a `requires` block; other package types are skipped.
- Respects `conditions.kibana.version`: a newer dependency version is only proposed when its Kibana constraint overlaps with the integration's.
- When a newer version exists but requires a higher Kibana version, a warning is printed suggesting to bump `conditions.kibana.version`.
- `--dry-run` reports proposals without modifying the file.
- `--format json` emits machine-readable output suited for CI batch jobs.
- Normalises constraint-style pins (e.g. `^0.3.0`) to exact semver on write.

Output example:

```bash

➜  elastic-package requires update -C packages/nginx_integration_otel  --dry-run                 
WARN package nginx_otel 0.5.0 is available but requires kibana ^9.4.0; integration conditions.kibana.version is ^9.2.0 — consider bumping conditions.kibana.version
Package: nginx_otel_integration
Code owner: elastic/ecosystem
Requires updates:
╭─────────┬──────────────────┬─────────┬──────────┬────────┬──────────────────────────────╮
│  KIND   │    DEPENDENCY    │ CURRENT │ PROPOSED │ KIBANA │           WARNING            │
├─────────┼──────────────────┼─────────┼──────────┼────────┼──────────────────────────────┤
│ input   │ filelog_otel     │ 0.1.0   │ 0.2.0    │ ^9.2.0 │                              │
├─────────┼──────────────────┼─────────┼──────────┼────────┼──────────────────────────────┤
│ input   │ nginx_otel_input │ 0.1.0   │ 0.2.0    │ ^9.2.0 │                              │
├─────────┼──────────────────┼─────────┼──────────┼────────┼──────────────────────────────┤
│ content │ nginx_otel       │ 0.3.0   │ -        │        │ package nginx_otel 0.5.0     │
│         │                  │         │          │        │ is available but requires    │
│         │                  │         │          │        │ kibana ^9.4.0; integration   │
│         │                  │         │          │        │ conditions.kibana.version    │
│         │                  │         │          │        │ is ^9.2.0 — consider bumping │
│         │                  │         │          │        │ conditions.kibana.version    │
╰─────────┴──────────────────┴─────────┴──────────┴────────┴──────────────────────────────╯
Dry run: manifest.yml was not modified

➜  elastic-package requires update -C packages/nginx_integration_otel  --dry-run --format json
WARN package nginx_otel 0.5.0 is available but requires kibana ^9.4.0; integration conditions.kibana.version is ^9.2.0 — consider bumping conditions.kibana.version
{
  "package": "nginx_otel_integration",
  "codeowner": "elastic/ecosystem",
  "proposals": [
    {
      "kind": "input",
      "package": "filelog_otel",
      "current": "0.1.0",
      "proposed": "0.2.0",
      "kibana_constraint": "^9.2.0"
    },
    {
      "kind": "input",
      "package": "nginx_otel_input",
      "current": "0.1.0",
      "proposed": "0.2.0",
      "kibana_constraint": "^9.2.0"
    },
    {
      "kind": "content",
      "package": "nginx_otel",
      "current": "0.3.0",
      "proposed": "",
      "warning": "package nginx_otel 0.5.0 is available but requires kibana ^9.4.0; integration conditions.kibana.version is ^9.2.0 — consider bumping conditions.kibana.version"
    }
  ]
}



➜  elastic-package requires update -C packages/nginx_otel  --dry-run                             
Package: nginx_otel
Code owner: elastic/obs-infraobs-integrations
package type is "content"; requires update only applies to integration packages with requires
```

## Why

Integration packages pin exact versions of input/content packages in `manifest.yml`. Keeping those pins up to date today requires manually checking the registry and editing the file — error-prone at scale across the integrations repository.

`requires update` makes bumping pins a single command, safe to run in CI across all integration packages, with `--format json` output for grouping PRs by code owner.

## Kibana compatibility check

To determine whether a newer dependency version is compatible, the command checks whether the integration's and dependency's `conditions.kibana.version` constraints share any valid Kibana version.

- Resolves https://github.com/elastic/integrations/issues/17404 (partially)


cmd/foreach.go
```diff
@@ -34,6 +34,7 @@ func getAllowedSubCommands() []string {
 		"format",
 		"install",
 		"lint",
+		"requires",
```
teresaromero, line 37: not sure this will be used on the automated script but i believe is worth enabling it for local use

docs/howto/dependency_management.md
```diff
@@ -97,9 +97,9 @@ and use a following field definition:
   external: ecs
 ```
 
-## Composable packages and the package registry
+## Integrations with requires and the package registry
 
-Composable (integration) packages can also depend on input or content packages by declaring them under
+Integration packages with `requires` can depend on input or content packages by declaring them under
 `requires` in `manifest.yml`. Depending on the package type, dependencies are resolved
```
jsoriano: Redundant `requires`?
```suggestion
Integration packages can depend on input or content packages by declaring them under
`requires` in `manifest.yml`. Depending on the package type, dependencies are resolved
```

docs/howto/dependency_management.md
```diff
@@ -97,9 +97,9 @@ and use a following field definition:
   external: ecs
 ```
 
-## Composable packages and the package registry
+## Integrations with requires and the package registry
```
jsoriano: Thanks 👍

Maybe instead of "requires" we can say "required packages", or "dependencies".

```suggestion
## Integrations with dependencies and the package registry
```
```suggestion
## Integrations with required packages and the package registry
```

Or backquote `requires` so it refers to the setting in the manifest.
```suggestion
## Integrations with `requires` and the package registry
```

internal/requiresupdates/compatibility.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		key := v.Original()
+		if _, ok := seen[key]; ok {
+			return
+		}
+		seen[key] = struct{}{}
+		result = append(result, v)
+	}
+	for _, constraint := range constraints {
+		for _, branch := range strings.Split(constraint, "||") {
```
mrodm: This assumes that constraints are somehow ordered, isn't it?
```yaml
conditions:
  kibana:
    version: "^8.0.0 || ^9.0.0"
```
Would it work successfully with this other constraint ? 

```yaml
conditions:
  kibana:
    version: "^9.0.0 || ^8.0.0"
```

Based on the `minVersionFromConstraintBranch` function.

Not sure if this is happening now in the current packages.

internal/requiresupdates/compatibility.go
```diff
@@ -0,0 +1,118 @@
...
+		Conditions: packages.Conditions{
+			Kibana: packages.KibanaConditions{Version: dependencyKibana},
+		},
+	}
+	for _, v := range candidates {
+		vStr := fmt.Sprintf("kibana.version=%s", v.Original())
+		if packages.CheckConditions(integManifest, []string{vStr}) == nil &&
+			packages.CheckConditions(depManifest, []string{vStr}) == nil {
+			return true, nil
+		}
```
mrodm: I'm not sure if this is doing what we expect (or what I think I would expect).

For instance, an integration package with this kibana constraint:
```yaml
conditions:
  kibana:
    version: "~9.1.0 || ^9.4.0""
```
and a new input package dependency (set as required) update with kibana constraint:
```yaml
conditions:
  kibana:
    version: "^9.4.0""
```

Would return `true` in `package.CheckConditions` because the constraint is validated with the part of the constraint `^9.4.0`. However, that integration package supports some Elastic stacks that this new input package would not be valid (9.1.x).

And this could be even more complicated to calculated if both packages contain constraints with `||` or `~` instead of `^`.

Example of a package with several constraints with OR:
https://github.com/elastic/integrations/blob/cc8af15ae202a23ee52f5099793c40cce3ce50f2/packages/microsoft_sentinel/manifest.yml#L13

There are also opened constrains (that I guess they would be less problematic here):
https://github.com/elastic/integrations/blob/cc8af15ae202a23ee52f5099793c40cce3ce50f2/packages/cloud_asset_inventory/manifest.yml#L17


I'm not sure about the approach to take here. @jsoriano could you please chime in here too?

cmd/requires.go
```diff
@@ -0,0 +1,195 @@
...
+func printRequiresUpdateResult(result *requiresupdates.Result, w io.Writer, format string) error {
+	if result == nil {
+		return nil
+	}
+	switch format {
+	case requiresFormatJSON:
+		enc := json.NewEncoder(w)
+		enc.SetEscapeHTML(false)
+		enc.SetIndent("", "  ")
+		return enc.Encode(result)
```
mrodm: Maybe it would complicate or couple with the package-spec version.

Just remembered that there is a JSON formatter used in the builder step that sets those same settings:

https://github.com/elastic/elastic-package/blob/13785cc9379a8f5a5f4fd8cdc2234e0225ca4084/internal/formatter/json.go#L65-L78

internal/requiresupdates/updates.go
```diff
@@ -0,0 +1,374 @@
...
+	var filtered []packages.PackageManifest
+	for _, rev := range revisions {
+		ok, err := kibanaConstraintsOverlap(integrationKibana, rev.Conditions.Kibana.Version)
+		if err != nil {
+			return nil, err
+		}
+		if ok {
+			filtered = append(filtered, rev)
+		}
+	}
```
mrodm, line 254: Currently the output when using `-v` (debug mode), it does not add much information to know what it is checking:

```
 $ elastic-package requires update -v
2026/05/28 12:36:15 DEBUG Enable verbose logging
2026/05/28 12:36:15 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
2026/05/28 12:36:16 DEBUG Verify Kibana version (constraint: ^9.4.0, requirement: 9.4.0)
2026/05/28 12:36:16 DEBUG Constraint kibana.version = ^9.4.0 is satisfied
2026/05/28 12:36:16 DEBUG Verify Kibana version (constraint: ^9.2.0, requirement: 9.4.0)
2026/05/28 12:36:16 DEBUG Constraint kibana.version = ^9.2.0 is satisfied
2026/05/28 12:36:16 DEBUG Verify Kibana version (constraint: ^9.4.0, requirement: 9.4.0)
2026/05/28 12:36:16 DEBUG Constraint kibana.version = ^9.4.0 is satisfied
2026/05/28 12:36:16 DEBUG Verify Kibana version (constraint: ^9.2.0, requirement: 9.4.0)
2026/05/28 12:36:16 DEBUG Constraint kibana.version = ^9.2.0 is satisfied
2026/05/28 12:36:16 DEBUG Verify Kibana version (constraint: ^9.4.0, requirement: 9.4.0)
2026/05/28 12:36:16 DEBUG Constraint kibana.version = ^9.4.0 is satisfied
2026/05/28 12:36:16 DEBUG Verify Kibana version (constraint: ^9.4.0, requirement: 9.4.0)
2026/05/28 12:36:16 DEBUG Constraint kibana.version = ^9.4.0 is satisfied
Package: nginx_otel_integration
Code owner: elastic/ecosystem
No dependencies to update

```

Could it be added at least a message here with the details of the comparisons ? 

```go
	for _, rev := range revisions {
		logger.Debugf("Checking compatibility with integration kibana.version=%s and revision kibana.version=%s for package %s %s", integrationKibana, rev.Conditions.Kibana.Version, rev.Name, rev.Version)
		ok, err := kibanaConstraintsOverlap(integrationKibana, rev.Conditions.Kibana.Version)
		if err != nil {
			return nil, err
		}
		if ok {
			filtered = append(filtered, rev)
		}
	}
```

I'm not sure if it would be helpful to add another debug log message just before each:
```
2026/05/28 12:36:16 DEBUG Verify Kibana version (constraint: ^9.4.0, requirement: 9.4.0)
2026/05/28 12:36:16 DEBUG Constraint kibana.version = ^9.4.0 is satisfied
```

in `kibanaConstraintOverlap` function:

```go
	for _, v := range candidates {
		logger.Debugf("Checking compatibility with integration kibana.version=%s and dependency kibana.version=%s", integrationKibana, v.Original())
		vStr := fmt.Sprintf("kibana.version=%s", v.Original())
		if packages.CheckConditions(integManifest, []string{vStr}) == nil &&
			packages.CheckConditions(depManifest, []string{vStr}) == nil {
			return true, nil
		}
	}
```
Maybe the latter is too much, WDYT ?

internal/requiresupdates/compatibility_test.go
```diff
@@ -0,0 +1,88 @@
...
+			integration: ">9.5.0,<9.6.0",
+			dependency:  ">=9.5.1",
+			want:        true,
+		},
+		{
+			name:        "strict-greater range does not overlap higher constraint",
+			integration: ">9.5.0,<9.6.0",
+			dependency:  "^9.6.0",
+			want:        false,
+		},
```
jsoriano, line 63: Could we add a test case like this one? This would be a common case when integrations start supporting new majors.
```suggestion
		{
			name:        "strict-greater range does not overlap higher constraint",
			integration: "^9.5.0 || ^10.0.0",
			dependency:  "^9.6.0",
			want:        false,
		},
```

internal/requiresupdates/compatibility_test.go
```diff
@@ -0,0 +1,88 @@
...
+		{
+			name:        "identical constraints overlap",
+			integration: "^9.4.0",
+			dependency:  "^9.4.0",
+			want:        true,
+		},
+		{
+			name:        "non-overlapping ranges",
+			integration: ">=9.4.0,<9.6.0",
+			dependency:  "^9.6.0",
```
jsoriano, line 29: A more realistic case that should be also a non-overlapping range, would be this one:

```suggestion
			integration: "^9.4.0",
			dependency:  "^9.6.0",
```

internal/requiresupdates/compatibility_test.go
```diff
@@ -0,0 +1,88 @@
...
+			name:        "dependency range contained in integration range",
+			integration: ">=9.4.0,<9.6.0",
+			dependency:  "^9.4.0",
+			want:        true,
+		},
+		{
+			name:        "empty integration constraint always overlaps",
+			integration: "",
+			dependency:  "^9.6.0",
+			want:        true,
```
jsoriano: This should fail. This means that the integration doesn't have any kibana restriction, but the input has. The resulting built integration should also have `^9.6.0`, not an empty constraint.
```suggestion
			name:        "empty integration constraint always overlaps",
			integration: "",
			dependency:  "^9.6.0",
			want:        false,
```

internal/requiresupdates/compatibility_test.go
```diff
@@ -0,0 +1,88 @@
...
+		{
+			name:        "strict-greater range does not overlap higher constraint",
+			integration: ">9.5.0,<9.6.0",
+			dependency:  "^9.6.0",
+			want:        false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ok, err := kibanaConstraintsOverlap(tt.integration, tt.dependency)
```
jsoriano: We should really check if the constraint of the integration is a subset of the constraint of the dependency.

Given an integration and a dependency, for any version of kibana that matches the constraint in the integration, it must also match the constraint of the dependency.

Checking the overlap is a good approximation for realistic use cases, but won't catch all of them. Probably any algorithm for this is going to be an approximation, so this one can be fine if we don't find another one more accurate.

Maybe an option is to get the approximated maximum and minimum supported versions in the integration constraint, and check that both match with the dependency constraint.

cmd/requires.go
```diff
@@ -0,0 +1,195 @@
...
+	updateCmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update requires.input and requires.content versions from the registry",
+		Long:  requiresUpdateLongDescription,
+		Args:  cobra.NoArgs,
+		RunE:  requiresUpdateCommandAction,
+	}
+	updateCmd.Flags().Bool(cobraext.RequiresDryRunFlagName, false, cobraext.RequiresDryRunFlagDescription)
+	updateCmd.Flags().String(cobraext.RequiresFormatFlagName, requiresFormatTable, fmt.Sprintf(cobraext.RequiresFormatFlagDescription, strings.Join(requiresFormatChoices, "|")))
+
```
jsoriano, line 54: Have you considered adding a flag to optionally add the changelog automatically?

internal/requiresupdates/compatibility.go
jsoriano: Yes, I think we need to reconsider the algorithm here. What we really would need to check is that the integration constraint is a subset of the dependency constraint. Added some comments about this in the tests.

internal/requiresupdates/manifest.go
jsoriano, line 1: Try to reuse `internal/yamledit`, it already has helpers for the kind of functionality needed here.

internal/requiresupdates/updates.go
```diff
@@ -0,0 +1,374 @@
...
+
+func resultFromManifest(manifest *packages.PackageManifest) Result {
+	return Result{
+		Package:   manifest.Name,
+		CodeOwner: manifest.Owner.Github,
+	}
+}
+
+// Update resolves and optionally applies dependency version bumps for an integration package with requires.
+func Update(opts Options) (*Result, error) {
```
jsoriano: Instead of having a single method that resolves and optionally applies version dumps, and given that we already have the type for proposals, maybe we should have two methods, one to resolve, and another one to apply.
```
func Resolve(opts Options) ([]UpdateProposal, error)
func Apply(manifest []bytes, proposals []UpdateProposal) ([]bytes, error)
```

internal/requiresupdates/updates.go
```diff
@@ -0,0 +1,374 @@
...
+		Kind:             kind,
+		Package:          dep.Package,
+		Current:          dep.Version,
+		Proposed:         latestCompatible.Version,
+		KibanaConstraint: latestCompatible.Conditions.Kibana.Version,
+		Warning:          kibanaBumpWarning(dep, latestUnfiltered, integrationKibana, latestCompatibleVer, nil),
+	}, nil
+}
+
+func fetchCompatibleRevisions(opts Options, integrationKibana, packageName string) ([]packages.PackageManifest, error) {
```
jsoriano: This method gets the versions by querying from all the derived kibana versions. This makes several queries to the registry, not sure if this will be fully accurate for what we need.

Maybe an initial approach while we don't have another way to get packages from a given constraint from the registry is to query all versions of a package with a single request (with `all=true`), and then do the filtering locally.

internal/requiresupdates/updates.go
```diff
@@ -0,0 +1,374 @@
...
+			proposals = append(proposals, *proposal)
+		}
+	}
+	return proposals, nil
+}
+
+func resolveDependency(opts Options, integrationKibana string, kind DependencyKind, dep packages.PackageDependency) (*UpdateProposal, error) {
+	unfiltered, err := opts.RegistryClient.Revisions(dep.Package, registry.SearchOptions{
+		Prerelease:   true,
+		Experimental: true,
```
jsoriano: Umm, we probably don't want to update to prerelease versions, at least by default. This can be another reason to just query all versions of the package and then filter locally. An exception could be a dependency that only has prerelease versions.

There could be a flag to decide if we want or not to update to prerelease versions, but getting them should not be the default.

internal/requiresupdates/updates.go
```diff
@@ -0,0 +1,374 @@
...
+			return nil, err
+		}
+		if proposal != nil {
+			proposals = append(proposals, *proposal)
+		}
+	}
+	return proposals, nil
+}
+
+func resolveDependency(opts Options, integrationKibana string, kind DependencyKind, dep packages.PackageDependency) (*UpdateProposal, error) {
```
jsoriano, line 136: Nit. We could have some unit tests for this using `httptest` and/or with `go-vcr` to record real interactions.

jsoriano (commented):
The high-level approach looks good, but some details may need another look.

internal/requiresupdates/updates.go
teresaromero: as all the otel input packages are still a prerelease, that is why i configured as so, i will check for the flag so we can use it instead

jsoriano: Yeah, we can keep an exception for packages that only have prerelease versions. But for GA packages that use input packages that have GA versions we should definitely default to GA versions.

Btw, in principle `requires` can be used too with non-OTel packages, so we could create now an integration package that requires input packages with GA versions.

docs/howto/dependency_management.md
teresaromero: I've rephrase it. thnks

internal/requiresupdates/compatibility.go
teresaromero: Replaced the overlap check with a subset check. The algorithm splits the integration constraint on `||` and probes each branch with three representative versions (floor, floor+1 patch, major ceiling) against the dependency constraint. If any probe satisfies the branch but not the dependency, it returns false.

internal/requiresupdates/compatibility.go
teresaromero: The new subset check evaluates each `||` branch independently, so order does not affect the result. Both orderings are covered by the `OR ordering invariant` test cases in `compatibility_test.go`.

internal/requiresupdates/compatibility_test.go
teresaromero: Replaced overlap with subset check as described above.

cmd/requires.go
teresaromero: Added `NewJSONFormatter()` as a spec-version-agnostic constructor and reused it here. The existing `JSONFormatterBuilder(specVersion)` is kept for spec content where HTML-encoding behaviour matters.

internal/requiresupdates/manifest.go
teresaromero: Replaced hand-rolled YAML helpers with `yamledit.NewDocumentBytes`, `GetSequenceNode`, `SetKeyValue`, and `Write`.

internal/requiresupdates/updates.go
teresaromero: Done. `Resolve` handles resolution only; `Apply` rewrites manifest bytes in memory; the command layer owns the disk write. Dry-run is a natural consequence of skipping `Apply`.

internal/requiresupdates/updates.go
teresaromero: `fetchAllRevisions` now issues a single request with `All: true` and filters locally in `filterByKibanaSubset`.

internal/requiresupdates/updates.go
teresaromero: I've added couple logs to be on debug so we could keep beter tracing of what is happening. i dont think it will be too noisy as is an opt-in

cmd/requires.go
teresaromero: yes, still pending to implement it

jsoriano (approved):
LGTM, thanks for addressing the comments.

If you don't add the flag to add the changelog in this PR please create a follow up issue (or PR).

teresaromero:
> If you don't add the flag to add the changelog in this PR please create a follow up issue (or PR).

i've opened a followup here https://github.com/elastic/elastic-package/pull/3626

internal/requiresupdates/updates.go
mrodm: Thanks!

mrodm (approved):
Tested locally and it looks great!
Thanks!

PR #3597, opened 2026-05-25
Verify signatures of packages downloaded from package registry

### LINKED ISSUE #3466: Verify EPR package zip signatures on download (registry client)
This work is tracked under #3277.

## Context

Detached OpenPGP verification for package zips downloaded from the Package Registry was removed from the #3459 branch to land separately. See the review thread: https://github.com/elastic/elastic-package/pull/3459#discussion_r3080770095

## Summary

Reintroduce verification when downloading packages from EPR (for example when resolving required input packages). Before implementing, align with the team on defaults and UX:

- The earlier approach used **opt-in** environment variables (`ELASTIC_PACKAGE_VERIFY_PACKAGE_SIGNATURE`, `ELASTIC_PACKAGE_VERIFIER_PUBLIC_KEYFILE`).
- Review feedback pointed toward **opt-out** verification and/or **bundling** the Elastic public GPG key so the integrations publish pipeline does not require extra environment setup.

## Acceptance criteria

- [ ] Document the environment contract and the EPR `.zip.sig` contract in [`tools/readme/readme.md.tmpl`](https://github.com/elastic/elastic-package/blob/main/tools/readme/readme.md.tmpl) and regenerate `README.md`.
- [ ] Implement verification on the registry download path ([`internal/registry/client.go`](https://github.com/elastic/elastic-package/blob/main/internal/registry/client.go) or a small dedicated helper in the same package).
- [ ] Add unit tests in `internal/registry` for success, missing signature, and invalid/wrong signature (zip removed on verification failure as appropriate).

jsoriano:
Verify GPG signatures of packages downloaded from the Package Registry. This is specially relevant when building packages that require input packages, so it is ensured that the packages used as dependencies are signed by trusted entities.

Verification uses GPG keyrings with trusted public keys. A keyring with the Elastic key (https://packages.elasticsearch.org/GPG-KEY-elasticsearch) is embedded by default.
Custom keyrings can be created by concatenating GPG public keys in a single file.

Two environment variables are added to control this behavior:
* `ELASTIC_PACKAGE_VERIFIER_DISABLE` disables signature verification.
* `ELASTIC_PACKAGE_VERIFIER_GPG_KEYRING` to configure a path to a custom keyring.

A script and updatecli configuration are included to automatically rotate the public key if it were to change.

Fixes https://github.com/elastic/elastic-package/issues/3466.


teresaromero (approved):
🚀  thanks!

For the automatic rotation workflow, Copilot has flagged a concern as we are trusting the downloaded key from the url. I think the risk is lowered by the fact that the rotation is done via PR and this will be reviewed by a human prior to merge, so in that review we could check the key is the expected one from a trusted source. Could we include some disclaimer text at the PR so reviewers have this in mind?

.github/workflows/updatecli/updatecli.d/bump-elastic-gpg-key.yml
```diff
@@ -0,0 +1,86 @@
...
+  default:
+    kind: github
+    spec:
+      owner: '{{ .scm.owner }}'
+      repository: '{{ .scm.repository }}'
+      user: '{{ requiredEnv "GITHUB_ACTOR" }}'
+      username: '{{ requiredEnv "GITHUB_ACTOR" }}'
+      token: '{{ requiredEnv "GITHUB_TOKEN" }}'
+      commitusingapi: true
+      branch: main
```
mrodm, line 54: Could it be that `updatecli` execution is failing because this workflow is executed with the `main` branch ? 

```
The shell :shell: command "/bin/sh /tmp/updatecli/bin/cda55642ec763d37553a4588793dbd18a8005f5e465eb600ac39161eb9bc63d6.sh" exited on error (exit code 1) with the following output:
----
----

command stderr output was:
----
go: downloading go1.26.0 (linux/amd64)
stat /tmp/updatecli/github/elastic/elastic-package/internal/registry/gpgkey/fetch: directory not found

----
```

internal/registry/elastic-gpg-key.asc
mrodm, line 1: In order to assign reviewers of the Pull Requests created by updatecli.
I think it would be good to add `ecosystem` team as codeowners of this file. 

The same for the file that contains the fingerprint of the Elastic Public GPG key (`internal/registry/gpgkey_test.go`).

I'm not sure about this other file `internal/registry/gpgkey/fetch/main.go` if `ecosystem` should be set as the codeowner too.

jsoriano:
@teresaromero thanks for pointing this out.

Investigated a little bit about how to improve this situation and also found out that we should be using https://artifacts.elastic.co/GPG-KEY-elasticsearch. It is the same public key as the one in https://packages.elasticsearch.org, but more current, and re-armored with more modern cryptographic parameters.

Changes in the last commits:
* Add recommendations to the PR descriptions about how to review a change in the key.
* Use the key from artifacts.elastic.co.
* Fix some other issues in the updatecli configuration.

.github/workflows/updatecli/updatecli.d/bump-elastic-gpg-key.yml
jsoriano: Yes, it looks like this will get solved after merge.

internal/registry/elastic-gpg-key.asc
jsoriano: Added in https://github.com/elastic/elastic-package/pull/3597/commits/ab502ab016a856bc1427ee3cb2238588687c7033, thanks.

mrodm (approved):

PR #3617, opened 2026-05-28
Wire elastic-package to package-spec validation modes

### LINKED ISSUE #549: [Change Proposal] Add different validation modes for source and built packages
Now we are using the very same spec to validate source and built packages, but the content can be actually different. For example dev files or external fields can be found in source packages but are not expected to be on built packages. Adding two validation modes can allow us to be more strict with the content and format of built packages. It can also help on testing package build implementations like the one in `elastic-package`.

We would need to:
* Expose in the `validator` package an interface to select the validation mode.
* Define somehow in the spec the contents for each mode.

teresaromero:
## Problem

`elastic-package` calls `package-spec` validation at several points in the lifecycle of a package — during `lint` (source tree), `build` (built artifact, directory and zip), and `install --zip` (pre-built zip). All of these call sites were using the same deprecated `ValidateFrom*` / `ValidateAndFilter*` functions, which run in `ModeLegacy`. This means the validation rules applied were identical regardless of whether the package was a source checkout or a built artifact.

## Why

`package-spec` now exposes explicit validation modes (`ModeSource` and `ModeBuild`) that reflect what a package actually is at each phase:

- **`ModeSource`** allows source-only artifacts: `_dev/` directories, `.link` files, and `external: ecs` field references. These are normal in a checked-out package and are resolved or stripped during the build step.
- **`ModeBuild`** rejects those same artifacts. A built package that still contains `_dev/` or unresolved `external: ecs` references is incorrect and should fail validation.

Running `lint` with `ModeLegacy` (instead of `ModeSource`) could silently allow or reject things inconsistently with what the source phase actually permits. Running `build` and `install` with `ModeLegacy` meant that built artifacts carrying source-only remnants would not be caught.

## Changes

The `internal/validation` package is rewritten to expose three named functions, one per validation context: `ValidateSourceFromPath`, `ValidateBuiltFromPath`, and `ValidateBuiltFromZip`. The mode selection is encapsulated inside the validation package — call sites never reference `validator.Mode` directly. Dead code (`ValidateFromPath`, `ValidateFromZip`) with zero callers is removed.

All four call sites are updated to use the context-appropriate function: `lint` uses `ValidateSourceFromPath`; `build` uses `ValidateBuiltFromPath` and `ValidateBuiltFromZip`; `install --zip` uses `ValidateBuiltFromZip`.

Tests are added to `internal/validation` covering the happy path for each new function. `TestValidateBuiltFromPath` and `TestValidateBuiltFromZip` each include a rejection case confirming that a source package containing `_dev/` is rejected — guarding against silent reversion to `ModeLegacy` behaviour on both the directory and zip code paths. A `makeZipFromDir` test helper generates zip fixtures on-the-fly, so no binary test fixture is committed.

`ValidateBuiltFromZip` also gains `logger.Debugf` calls in the two previously silent error paths (zip open failure and subdirectory extraction failure), making them visible under `--verbose` and consistent with the logging in `validateFromPath`.

## Dependency

This PR depends on [elastic/package-spec#1178](https://github.com/elastic/package-spec/pull/1178) and package-spec being released

## Testing

Tested by running a local build against the integrations packages. After the change, all packages validate as expected with no new failures and no regressions.

- Resolves https://github.com/elastic/package-spec/issues/549


internal/validation/validation.go
```diff
@@ -77,6 +109,7 @@ func fsFromPackageZip(fsys fs.FS) (fs.FS, error) {
 	return subDir, nil
 }
 
+// TODO: follow-up issue — move this logic into specerrors as an exported function so consumers don't reimplement it.
```
teresaromero, line 115: filed issue https://github.com/elastic/elastic-package/issues/3618

internal/validation/validation.go
```diff
@@ -13,59 +13,94 @@ import (
...
+	allErrors := v.ValidateFromPath(packageRoot)
 	if allErrors == nil {
 		return nil, nil
 	}
-
 	fsys := os.DirFS(packageRoot)
 	result, err := filterErrors(allErrors, fsys)
 	if err != nil {
-		return err, nil
+		logger.Debugf("failed to filter validation errors: %v", err)
```
teresaromero, line 50: at this point allErrors is not nil; so in case any error happens during the filering, they were ignored... i've added a debug log with the error and return allErrors unfiltered; same un the zip case

mrodm (approved):
LGTM!
Tested locally to force failures in build and source modes and `elastic-package lint` and `elastic-package build` commands detected those.

Thanks!

PR #3623, opened 2026-05-29
internal/{servicedeployer,testrunner,validation},docs: generate on-disk TLS certs for mock services

efd6:
The compose deployer now creates a CA and per-service TLS certificates before docker compose up whenever a service declares a hostname. Cert files are written to {OutputDir}/tls/ and exposed to containers via the TLS_DIR environment variable.

Each certificate carries dual SANs (the compose hostname and the svc-{name} network alias) so both connection patterns work.

A new tls_ca Handlebars helper emits the matching CA PEM into test configs for certificate_authorities blocks. On --no-provision and --tear-down runs the deployer reads back the existing ca.crt so the helper continues to work without regenerating certificates.

The TLS indent lint check is extended to cover tls_ca (which takes no domain argument, unlike tls_cert/tls_key).

Ref #3604


internal/testrunner/runners/system/tls_helpers.go
```diff
@@ -95,6 +96,29 @@ func tlsHelpers() map[string]interface{} {
...
+// Template usage:
+//
+//	{{{tls_ca indent=10}}}
+func tlsCAHelper(svcInfo servicedeployer.ServiceInfo) map[string]interface{} {
+	return map[string]interface{}{
+		"tls_ca": func(options *raymond.Options) raymond.SafeString {
+			if svcInfo.CustomProperties == nil {
+				return ""
+			}
+			pem, ok := svcInfo.CustomProperties["TLS_CA_PEM"]
```
mrodm: Would it make sense to use the same variable (making it public) from `internal/servicedeployre/compose_tls.go` ?

https://github.com/elastic/elastic-package/pull/3623/changes#diff-5505fd7b89d2f029be4e37e1155b7d57a82872619080be8572814bbc0cde3cacR21

docs/howto/system_testing.md
```diff
@@ -750,6 +750,61 @@ YAML block scalars require consistent indentation.
...
+Reference `TLS_DIR` in `docker-compose.yml` to volume-mount the certs:
+
+```yaml
+services:
+  mock_server:
+    hostname: guardduty.xxxx.amazonaws.com
+    image: mock:latest
+    command: --tls-cert=/tls/mock_server.crt --tls-key=/tls/mock_server.key
+    volumes:
+      - ${TLS_DIR}:/tls:ro
```
mrodm, line 790: We should add an integration test (a new test package) under `test/packages/parallel` to prevent future regressions.

docs/howto/system_testing.md
```diff
@@ -750,6 +750,61 @@ YAML block scalars require consistent indentation.
...
+`hostname:`, the compose deployer creates a CA and per-service certificate
+before `docker compose up`.
+
+The generated files are written to a `tls/` subdirectory under the test
+output directory and exposed to compose via the `TLS_DIR` environment
+variable:
+
+| File | Description |
+|---|---|
+| `${TLS_DIR}/ca.crt` | CA certificate (PEM) |
```
mrodm, line 773: TLS_DIR should also be added as part of the documentation (table) at the Placeholders section. Similar to SERVICE_LOGS_DIR.

And I'd say that these helpers should be added also in that table:
- tls_cert
- tls_key
- tls_ca

This could help developers to have all the available placeholders in one location, and then they could read more about them in the sections you added.

docs/howto/system_testing.md
```diff
@@ -750,6 +750,61 @@ YAML block scalars require consistent indentation.
...
+
+`tls_ca` takes no domain argument — there is a single shared CA per test
+run. Like `tls_cert`/`tls_key`, it accepts an `indent=` parameter for
+YAML alignment.
+
+On `--no-provision` and `--tear-down` runs the deployer reads back the
+existing `ca.crt` so `tls_ca` continues to work without regenerating
+certificates.
+
 **NOTE**: Terraform variables in the form of environment variables (prefixed with `TF_VAR_`) are not injected and cannot be used as placeholder (their value will always be empty).
```
mrodm: With these changes, this paragraph looks like it belongs to this new section. But IMO it could be set at the position where it was previosly (at the end of Placeholders section, before the sub-sections). WDYT?

```
#### Placeholders

The `SERVICE_LOGS_DIR` placeholder is not the only one available for use in a data stream's `test-<test_name>-config.yml` file. The complete list of available placeholders is shown below.

| Placeholder name | Data type | Description |
| --- | --- | --- |
| `Hostname`| string | Addressable host name of the integration service. |
| `Ports` | []int | Array of addressable ports the integration service is listening on. |
| `Port` | int | Alias for `Ports[0]`. Provided as a convenience. |
| `Logs.Folder.Agent` | string | Path to integration service's logs folder, as addressable by the Agent. |
| `SERVICE_LOGS_DIR` | string | Alias for `Logs.Folder.Agent`. Provided as a convenience. |
| `TEST_RUN_ID` | string | Unique identifier for the test run, allows to distinguish each run. Provided as a convenience. |

Placeholders used in the `test-<test_name>-config.yml` must be enclosed in `{{{` and `}}}` delimiters, per Handlebars syntax.

**NOTE**: Terraform variables in the form of environment variables (prefixed with `TF_VAR_`) are not injected and cannot be used as placeholder (their value will always be empty).

#### TLS certificate helpers

```

mrodm (approved):
Thanks!

PR #3626, opened 2026-06-01
feat: add --changelog flag to requires update command

### LINKED ISSUE #17404: Automatically update package dependencies

teresaromero:
- Closes https://github.com/elastic/integrations/issues/17404 (partially)

## Summary

Extends `requires update` with two optional flags that automate changelog maintenance when bumping package dependencies:

```
elastic-package requires update [--changelog] [--changelog-type <type>]
```

- `--changelog` — for each bumped dependency, appends one changelog entry to `changelog.yml` and bumps the package's own version in both `changelog.yml` and `manifest.yml`.
- `--changelog-type` — overrides the inferred entry type for all generated entries (`bugfix`, `enhancement`, or `breaking-change`). Without this flag the type is inferred from the semver bump tier: major → `breaking-change`, minor/patch → `enhancement`.

When combined with `--dry-run`, nothing is written; the would-be new version and per-entry descriptions are printed instead. The dry-run path runs the same guards as the live path — a package that would be rejected by the live run (version divergence, pre-release top version) is also rejected in dry-run rather than printing a misleading clean preview.

---

## How it works

1. **Tier inference** — each proposal's `Current`→`Proposed` semver delta determines its tier (`TierMajor`, `TierMinor`, `TierPatch`). Content-dependency constraints that can't be parsed as an exact version default to `TierMinor`.
2. **Aggregate tier** — the largest tier across all bumped proposals drives the package version bump (same `major|minor|patch` mode as `changelog add --next <tier>`).
3. **Atomic write** — `PlanChangelog` reads `changelog.yml`, runs all version guards, and computes the `PatchYAML` input in memory before any file is touched. The same bytes are reused by `ApplyChangelog` for the write, so there is no second read and no TOCTOU window between validation and patching.
4. **Placeholder link** — every generated entry uses `https://github.com/elastic/integrations/pull/REPLACE_ME` as its link. The https://github.com/elastic/integrations/pull/19217 PR-opening workflow is expected to replace this token with the real PR URL. This is a contract — the literal value must be confirmed with that workflow before merge.
5. **`new_version` in JSON output** — `Result.NewVersion` is populated when `--changelog` bumps the version, making it available to automation via `--format json`.
6. **Pre-release guard** — the command errors if the changelog top version carries a pre-release suffix (e.g. `1.2.0-SNAPSHOT`). Experimental packages (`0.x.y`, no suffix) are unaffected.

---

## Reviewer notes

- **Placeholder link contract** — `ChangelogPlaceholderLink` (`https://github.com/elastic/integrations/pull/REPLACE_ME`) is a contract with the elastic/integrations#19217 PR-opening workflow, which must search-and-replace this exact token with the real PR URL. The literal value is still proposed — needs confirmation with that workflow before merge.
- **Tier inference for content constraints** — `Current` values that are semver constraints (e.g. `^0.3.0`) cannot be diffed against `Proposed`; they default to `TierMinor`. This is a deliberate approximation since Masterminds/semver exposes no constraint floor.
- **Atomic write ordering** — `PlanChangelog` validates and builds the revision in memory; `ApplyChangelog` reuses the already-read bytes for `PatchYAML`, avoiding a second read. `changelog.yml` is written first, then `manifest.yml`. A failure between the two writes would leave the files inconsistent; this was accepted as a low-probability edge case vs. the complexity of a two-phase commit.
- **Version divergence guard** — the command errors if `manifest.yml` version differs from the `changelog.yml` top revision before applying changes. package-spec requires they agree; this protects against already-dirty packages.
- **Dry-run consistency** — `--changelog --dry-run` calls `PlanChangelog` directly and surfaces the same guard errors (version divergence, pre-release top version) as the live path. Previously it recomputed version and entry types independently and skipped the guards, so a dry-run could print a clean preview for a package the live run would reject.
- **Warning-only proposals guard** — `PlanChangelog` errors immediately if all proposals are warning-only (no `Proposed` value), preventing a no-op revision with zero entries and an unchanged version from being written.


docs/howto/dependency_management.md
```diff
@@ -172,6 +172,62 @@ To refresh many integration packages in a repository (for example from a schedul
...
+file valid. Replace `REPLACE_ME` with the real PR number once the PR is opened.
+
+**Dry-run.** `--changelog --dry-run` writes nothing and prints the would-be new version and the
+per-dependency entry types.
+
+**JSON output.** `--format json` includes `new_version` in the response when `--changelog`
+bumped the package version.
+
+**Guards.** The command refuses to proceed when the changelog top version and `manifest.yml`
+version disagree, or when the changelog top version is a pre-release. Proposals that are
```
mrodm, line 225: According to this doc , the changelog is not added if the top version is a pre-release.

Just checked that there is this method in `ApplyChangelog`:
```go
	if err := assertManifestVersionMatchesChangelogTop(manifestBytes, revisions); err != nil {
		return nil, "", err
	}
```

Should this guard be added ? or is it managed already in another method?

internal/requiresupdates/changelog.go
```diff
@@ -0,0 +1,201 @@
...
+
+	tier := AggregateTier(proposals)
+	next, err := changelog.NextVersionFromRevisions(revisions, tier.NextMode())
+	if err != nil {
+		return nil, "", err
+	}
+	manifestBytes, err = changelog.SetManifestVersion(manifestBytes, next.String())
+	if err != nil {
+		return nil, "", err
+	}
```
mrodm: Taking as a example `elastic-package changelog` command, changelog and manifest updates are separated.

https://github.com/elastic/elastic-package/blob/6158fcdc6c5e0036aa29bb0538eef94050535452/cmd/changelog.go#L98-L106

What about moving the writing of the manifet out of the `ApplyChangelog`?  So applyChangelog just updates changelog and not manifest ? That write could be done by `requires.Apply`.

cmd/requires.go
```diff
@@ -155,6 +170,24 @@ func requiresUpdateCommandAction(cmd *cobra.Command, _ []string) error {
...
+				}
+				t := changelogType
+				if t == "" {
+					t = requiresupdates.DefaultChangelogType(p.Tier())
+				}
+				cmd.Printf("  - [%s] %s: %s -> %s\n", t, p.Package, p.Current, p.Proposed)
+			}
+		}
 	} else if applied {
 		cmd.Println("Updated manifest.yml")
```
mrodm: If `--changelog` is set, this message should also indicate that `changelog.yml` has been updated.

cmd/requires.go
```diff
@@ -51,6 +51,8 @@ func setupRequiresCommand() *cobraext.Command {
 	updateCmd.Flags().Bool(cobraext.RequiresDryRunFlagName, false, cobraext.RequiresDryRunFlagDescription)
 	updateCmd.Flags().String(cobraext.RequiresFormatFlagName, requiresFormatTable, fmt.Sprintf(cobraext.RequiresFormatFlagDescription, strings.Join(requiresFormatChoices, "|")))
 	updateCmd.Flags().Bool(cobraext.RequiresPrereleaseFlagName, false, cobraext.RequiresPrereleaseFlagDescription)
+	updateCmd.Flags().Bool(cobraext.RequiresChangelogFlagName, false, cobraext.RequiresChangelogFlagDescription)
+	updateCmd.Flags().String(cobraext.RequiresChangelogTypeFlagName, "", cobraext.RequiresChangelogTypeFlagDescription)
```
mrodm: There is no documentation added for these 2 new flags here: 
https://github.com/elastic/elastic-package/blob/6158fcdc6c5e0036aa29bb0538eef94050535452/cmd/requires.go#L35-L40

cmd/requires.go
```diff
@@ -70,6 +72,13 @@ const (
 
 var requiresFormatChoices = []string{requiresFormatTable, requiresFormatJSON}
 
+// changelogTypeChoices mirrors the changelog entry type enum defined in the package-spec
+// (spec/integration/changelog.spec.yml). The canonical validation happens when the
+// package is built and checked against the spec; this list enables early flag
+// validation. "deprecation" is intentionally excluded as it is not a valid type
+// for automated dependency-bump changelog entries.
+var changelogTypeChoices = []string{"bugfix", "enhancement", "breaking-change"}
```
mrodm: Taking a look to this list, there is another place where these values are king of defined.

https://github.com/elastic/elastic-package/blob/6158fcdc6c5e0036aa29bb0538eef94050535452/internal/cobraext/flags.go#L107

I know that this is unrelated to tihs PR, but could you add `deprecation` there in the flag description?

I wonder if we could set the list in the description programmatically.

docs/howto/dependency_management.md
teresaromero: i am being aware now that i was thinking on "prereleased" as "experimental" too.... and this might be an implementation gap. i think we should probably ignore changelog for prerelease (-suffix) but allow for experimental. 

`assertManifestVersionMatchesChangelogTop` just checks the manifest is aligned with the changelog latest entry

docs/howto/dependency_management.md
```diff
@@ -172,6 +181,62 @@ To refresh many integration packages in a repository (for example from a schedul
...
+
+Pass `--changelog` to write changelog entries in addition to rewriting the `manifest.yml` pins.
+With this flag the command:
+
+- adds **one changelog entry per bumped dependency** to `changelog.yml`, and
+- **bumps the package's own version** in both `changelog.yml` and `manifest.yml`.
+
+Neither a human nor automation has to hand-edit the changelog afterward.
+
+**Version bump tier.** The package version is bumped by the largest semver tier across all
```
mrodm: I think the `tier` term was not used in elastic-package previously. So, I think it could be interesting to add some brief definition and/or include the options that could be used there (major, minor or patch?)

Does it refer to the type in the changelog entry?


```suggestion
**Version bump tier.** The package version is bumped by the largest semver tier (major, minor or patch) across all
```

internal/requiresupdates/changelog.go
```diff
@@ -0,0 +1,229 @@
...
+		return ""
+	}
+	return ""
+}
+
+// nextVersion reads the changelog top version from packageRoot and returns it
+// incremented by tier. Returns 0.0.0 when the changelog is empty.
+func nextVersion(tier BumpTier, packageRoot string) (*semver.Version, error) {
+	return changelog.NextVersion(packageRoot, tier.NextMode())
+}
```
mrodm: This function is used just in testing.
Could it be moved to the `changelog_test.go` file ?

internal/requiresupdates/updates.go
```diff
@@ -116,6 +120,15 @@ func Apply(manifestBytes []byte, proposals []UpdateProposal) ([]byte, error) {
...
 }
 
+// ApplyManifestVersion sets the package version field in manifestBytes.
+func ApplyManifestVersion(manifestBytes []byte, version string) ([]byte, error) {
+	updated, err := changelog.SetManifestVersion(manifestBytes, version)
+	if err != nil {
+		return nil, fmt.Errorf("setting manifest version failed: %w", err)
+	}
+	return updated, nil
+}
```
mrodm: As it is `ApplyManifestVersion` looks just a wrapper of the `changelog.SetManifestVersion`. Could it be replaced the usages of ApplyManifestVersion` for the other method ?

cmd/requires.go
```diff
@@ -70,6 +80,11 @@ const (
 
 var requiresFormatChoices = []string{requiresFormatTable, requiresFormatJSON}
 
+// updateRequiresChangelogTypeChoices is the subset of changelog entry types valid for --changelog-type.
+// "deprecation" (valid for `changelog add --type`) is intentionally excluded: it is not a meaningful
+// type for an automated dependency-bump entry.
+var updateRequiresChangelogTypeChoices = []string{"bugfix", "enhancement", "breaking-change"}
```
mrodm, line 87: 👍

cmd/requires.go
```diff
@@ -155,15 +176,73 @@ func requiresUpdateCommandAction(cmd *cobra.Command, _ []string) error {
...
+			plan, err := requiresupdates.PlanChangelog(packageRoot, manifestBytes, result.Proposals, changelogType)
+			if err != nil {
+				return err
+			}
+			cmd.Printf("Dry run: would bump package version to %s and add changelog entries:\n", plan.NextVersion)
+			for _, e := range plan.Revision.Changes {
+				cmd.Printf("  - [%s] %s\n", e.Type, e.Description)
+			}
+		}
 	} else if applied {
```
mrodm: Would it be the same as `else if hasBumps`?

If it would be `dry_run==false` and `hasBumps==true` but there is some error, the execution would fail due to line  157. Is that right?

I'm trying to check whether or not there could be needed some another `else if` to add here.

cmd/requires.go
```diff
@@ -155,15 +176,73 @@ func requiresUpdateCommandAction(cmd *cobra.Command, _ []string) error {
...
 		cmd.Println("Updated manifest.yml")
+		if changelogEnabled {
+			cmd.Println("Updated changelog.yml")
+		}
 	} else if len(result.Proposals) == 0 && result.SkipReason == "" {
 		cmd.Println("No dependencies to update")
 	}
```
mrodm: What would be the output if there is any `result.SkipReason != ""` ?

mrodm (commented):
Tested locally and it looks good!
Added some comments and questions.

cmd/requires.go
teresaromero: Added a dedicated `else if result.SkipReason != ""` arm before the `len == 0` branch; also dropped the now-redundant `&& result.SkipReason == ""` guard from the last condition.

cmd/requires.go
teresaromero: `applied` variable removed; the two top-level `hasBumps` arms were merged into a single `if hasBumps { if dryRun { … } else { … } }` block, making the intent clear without a tracking variable.

internal/requiresupdates/updates.go
teresaromero: `ApplyManifestVersion` was removed; both callers (`cmd/requires.go:236` and `changelog_test.go:487`) now call `changelog.SetManifestVersion` directly.

internal/requiresupdates/changelog.go
teresaromero: The `nextVersion` wrapper was removed from `changelog.go`; `TestNextVersion` now calls `changelog.NextVersion(dir, tc.tier.NextMode())` directly.

cmd/requires.go
```diff
@@ -153,17 +173,79 @@ func requiresUpdateCommandAction(cmd *cobra.Command, _ []string) error {
...
+				}
+			}
+		} else {
+			cmd.Println("Updated manifest.yml")
+			if changelogEnabled {
+				cmd.Println("Updated changelog.yml")
+			}
+		}
+	} else if result.SkipReason != "" {
+		cmd.Println(result.SkipReason)
```
mrodm: Testing this with an otel input package in the integrations repository:

```sh
cd packages/nginx_input_otel
elastic-package requires update -v --dry-run --format table
```

The skip reason message appears twice

```
 $ elastic-package requires update -v --dry-run --format table
2026/06/17 11:00:28 DEBUG Enable verbose logging
2026/06/17 11:00:28 DEBUG Distribution built without a version tag, can't determine release chronology. Please consider using official releases at https://github.com/elastic/elastic-package/releases
2026/06/17 11:00:28 DEBUG using package registry: https://epr.elastic.co
Package: nginx_otel_input
Code owner: elastic/obs-infraobs-integrations
package type is "input"; requires update only applies to integration packages with requires
package type is "input"; requires update only applies to integration packages with requires

```

mrodm (commented):
LGTM
Added a comment about showing to the user twice the skip reason message in the case of otel input packages for instance.

mrodm (approved):
LGTM!

PR #3648, opened 2026-06-12
Bundle templates for required input package templates at the stream level

jsoriano:
Integration packages can include agent configuration templates at the stream and the input level.

When templates of required input are bundled they must be bundled as if they where at the stream level, so each stream can have its own configuration, but it is redundant to bundle them also at the input level.

Stop bundling templates also at the input level.

Why?
---

The input level provides the same configuration for all streams in the same input, using to resolve the template the variables defined at the input level.
Then the stream level templates use both the input and stream level variables to provide stream specific configuration.

For example in the usual case of having access and error logs for a web server, this separation allows to have some common settings at the input level, for both log inputs, and some specific settings at the stream level.

In the case of composable packages it doesn't really matter to define common settings at the input level, as they are going to be duplicated at the stream level using the stream specific configuration. So we stop doing this by default. All streams of the same input will have the same configuration template, that will be feeded with a combination of common input level variables and per-stream stream variables. Developers can still add input-level templates and configuration levels to extend the behavior of the required package.

Apart from that, Fleet puts configuration for data routing at the input level (`data_stream.*`  variables), placing additional configuration at this level is prone to conflicts, as input packages use to define `data_stream` blocks in their template configurations (see for example https://github.com/elastic/integrations/blob/9c51a22a0633904e3916f64bad9879ae316a2096/packages/log/agent/input/input.yml.hbs#L16). With this change we avoid issues at this point.
As a follow up we could review how Fleet compiles inputs and streams, as they seem to have different merge strategies.

For example when trying to bundle the log package in the ECS nginx package we found out that the documents were not correctly routed to the configured data stream, though it works with the changes in this branch.


teresaromero (approved):
👍🏻  thanks. it will be good to find a place where we document clearly this behavior. when working on the template validation rules on package-spec i understood fleet does not relate between the input and the stream files, we should check that not bundling the template for the input does not trigger a validation error not finding the default file

jsoriano:
> it will be good to find a place where we document clearly this behavior.

Yeah, this is a good idea. Though it can be complicated to do proactively because it depends on the behavior implemented in Fleet.

PR #3699, opened 2026-06-24
docs/_static/inputs: fix input name capitalisations

efd6:
See title.


teresaromero (approved):

mrodm:
@efd6 and @teresaromero just as a note, this will imply that in the next `elastic-package` release, some packages in the integrations repository should be updated. As it happened in this PR for the `apache` test package, their README files will be outdated.

PR #3713, opened 2026-06-26
fix(requiredinputs): rewrite {{data_stream.dataset}} to _meta path in bundled templates

teresaromero:
## What changed

- When bundling an input package template into a composable integration, rewrite `{{data_stream.dataset}}` → `{{_meta.stream.data_stream.dataset}}` in the copied template content.

## Context

For standalone `type: input` packages, Fleet auto-injects `data_stream.dataset` as a stream var, so `{{data_stream.dataset}}` resolves correctly in `.hbs` templates. For `type: integration` (composable) packages Fleet does **not** inject that var — `{{data_stream.dataset}}` renders empty, which silently breaks routing to the correct index template.

Fleet always injects `_meta.stream.data_stream.dataset` from package metadata for all package types, and that value is correct for composable packages. Rewriting the variable reference at bundle time eliminates the need for composable package authors to add a user-visible `data_stream.dataset` stream var override as a workaround.

## Tests

```
go test ./internal/requiredinputs/...
make format
```

## Reviewer notes

The rewrite is a byte-level `strings.ReplaceAll` on the copied template before it is written to the build root. It only touches `{{data_stream.dataset}}`; `{{data_stream.type}}` and `{{data_stream.namespace}}` are unaffected (covered by the partial-match test case).

Found while working on https://github.com/elastic/integrations/pull/19719 - tested manually with this package

### DIFF
--- a/internal/requiredinputs/copy.go
+++ b/internal/requiredinputs/copy.go
@@ -5,6 +5,7 @@
 package requiredinputs
 
 import (
+	"bytes"
 	"fmt"
 	"io/fs"
 	"os"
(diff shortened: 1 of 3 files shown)


teresaromero:
closing this fix in favor of a new pr with an other approach to the problem

Issue #3710, opened 2026-06-25
[policy-test] Add support for stack version conditions in policy test configuration


mrodm:
## Summary

Add support for a `condition` field in policy test configuration files (`_dev/test/policy/*.yml`) that allows a test to be skipped when the running stack version does not satisfy a specified semver constraint.

This allows package authors to express that a given policy test configuration is only valid or meaningful on certain stack versions, rather than relying on unconditional `skip` entries with manual version notes in the reason string.

## Motivation

Policy tests are currently either always run or unconditionally skipped. There is no way to express "run this test only on stack >= 9.5.0" without permanently skipping it for all versions.

A concrete example of this gap: OTel input packages (e.g. `docker_input_otel`) support a minimum stack of 9.2.0, but Kibana changes introduced in 9.5.0-SNAPSHOT altered how OTel extension IDs are suffixed in generated policies. Updating the policy test expected output for 9.5.0-SNAPSHOT behavior breaks the tests on 9.2.0 and 9.4.0. Without version-gating, there is no way to maintain correct expected output for both behaviours simultaneously.

Related: https://github.com/elastic/integrations/issues/18803#issuecomment-4744033420 and https://github.com/elastic/elastic-package/issues/3630.

## Proposed config syntax

Add a top-level `condition` key to the policy test YAML:

```yaml
# Run only on stack >= 9.5.0
condition:
  stack_versions: ">=9.5.0"
vars:
  enable_request_tracer: true
```

```yaml
# Run only on stack >= 9.0.0 and < 9.5.0
condition:
  stack_versions: ">=9.0.0 <9.5.0"
vars:
  some_legacy_option: true
```

The `stack_versions` value is a semver range string, consistent with the format already used in `manifest.yml` for `conditions.kibana.version`.

## Behaviour

- If the running stack version satisfies `stack_versions`, the test runs normally.
- If it does not satisfy the constraint, the test is skipped with a log message indicating the current stack version and the unsatisfied constraint, e.g.:
  `Skipping test-foo: stack version 9.4.0 does not satisfy >=9.5.0`
- If the stack version cannot be retrieved from Kibana's `/api/status`, the test fails with an error (policy tests require a running Kibana).
- `-SNAPSHOT` suffix is stripped before the semver comparison (e.g. `9.5.0-SNAPSHOT` is treated as `9.5.0`).
- If both `skip` and `condition` are present, `skip` takes precedence.

## Implementation notes

[...]

Issue #3630, opened 2026-06-02
Policy test normalization misses suffixed OTel extension IDs in `service.extensions` and `auth.authenticator`


mrodm:
## Summary

Policy tests for OTel-based integrations (e.g. `elasticapm_input_otel`) started failing after https://github.com/elastic/kibana/pull/270771 was merged (2026-05-22). That Kibana change fixed a real runtime bug where Fleet's per-stream OTel config generator suffixes extension keys for cross-stream uniqueness (e.g. `apikeyauth` → `apikeyauth/<input-id>-<stream-id>`), but the references to those extensions in `service.extensions` and `auth.authenticator` were left as bare names.

Now that Kibana correctly suffixes both reference sites, `normalizePolicyToCanonical` in elastic-package fails to normalize the dynamic UUIDs in those two locations, so test output can never match the static `.expected` files.

## Observed failure

From https://github.com/elastic/integrations/issues/19332:

```diff
unexpected content in policy: --- want
+++ got
@@ -80,3 +80,3 @@
         auth:
-            authenticator: apikeyauth
+            authenticator: apikeyauth/otelcol-elasticapmintakereceiver-2ad3f316-95ec-4749-955d-bb680ccb3a6f-otelcol-elasticapm_input_otel-elasticapmintakereceiver-2ad3f316-95ec-4749-955d-bb680ccb3a6f
         endpoint: localhost:8200
@@ -85,3 +85,3 @@
     extensions:
-        - apikeyauth
+        - apikeyauth/componentid-0
     pipelines:
```

Two normalization gaps:

1. **`service.extensions` array** — entries like `apikeyauth/otelcol-<uuid>` are not being normalized to `apikeyauth/componentid-N`, so the `.expected` file can never match.
2. **`auth.authenticator` value** inside receiver bodies — the full suffixed ID `apikeyauth/otelcol-<long-uuid>` is not normalized at all, while the corresponding extension key in the `extensions:` block is already correctly normalized to `apikeyauth/componentid-N`.

## Root cause

`normalizePolicyToCanonical` in [`internal/testrunner/runners/policy/policy.go#L243`](https://github.com/elastic/elastic-package/blob/b314e52d3b8da431f7941336cd8e3554f262d192/internal/testrunner/runners/policy/policy.go#L243) handles normalization of dynamic component IDs for extension **keys**, but does not normalize the suffixed IDs that now appear as **values** in:
- `service.extensions[]` items
[...]

PR #3719, opened 2026-06-29
requiredinputs: exclude data_stream.dataset from bundled stream vars

teresaromero:
## Why

Composable integrations (type `integration` with `requires.input`) import one or more input packages at build time. Input packages declare `data_stream.dataset` as a user-facing policy template variable because, for a standalone input package, users choose which dataset their data lands in.

When bundling a composable integration, the build merges non-promoted vars from the imported input package into the integration's data stream manifest. This caused `data_stream.dataset` to leak into the bundled stream-level `vars` section, which is wrong for two reasons:

1. **Semantic**: in an integration the dataset is fixed — either derived from the data stream name (`<package>.<data_stream>`) or set explicitly via the data stream's `dataset` field. It is not meant to be user-configurable.
2. **Fleet rejects the policy**: Fleet validates that every var sent in `stream.vars` is declared in the package's stream manifest. Because `data_stream.dataset` ended up in the bundled manifest without a meaningful `default`, Fleet responded with `400 Bad Request: Variable ... data_stream.dataset not found` when the integration was tested or installed.

## What

In `mergeStreamsInDSManifest` (`internal/requiredinputs/variables.go`), `data_stream.dataset` is stripped from the imported input package's own vars (`excludeVarByName`) before the merge runs, rather than being blocked via the `promotedNames` set as in the first version of this PR.

That distinction matters: `promotedNames` is checked before data-stream overrides are considered, so blocking it there silently dropped `data_stream.dataset` even when **the integration itself** declared the var (a real, supported pattern — see below). Stripping it from the input package's base vars instead means an integration-declared override still goes through the normal "novel var" merge path and is preserved.

An integration can declare `data_stream.dataset` itself in two independent places, and both are honored:
- Directly on the data stream (`data_stream/*/manifest.yml` → `streams[].vars`).
- Promoted to the input level (`policy_templates[].inputs[].vars`).

If both are declared, no dedup happens on the elastic-package side — Fleet's normal variable-scope precedence (stream overrides input overrides package) resolves which value applies when the policy is rendered. No value is injected into the built YAML by elastic-package itself; the actual injection for the "nothing declared" case happens in Fleet at template-compilation time (see Related).

## Testing

Added fixtures and unit tests in `internal/requiredinputs` covering the full matrix of how an integration can relate to `data_stream.dataset`:
- Input package declares it, integration declares nothing → excluded from the bundle, other input vars still bundled.
- Integration fixes the dataset via the data stream's `dataset` field only (no var) → field untouched, var still excluded.
- Integration also declares its own `data_stream.dataset` var on the data stream → integration's declaration wins over the input package's.
- Input package's *only* var is `data_stream.dataset` → exclusion leaves zero base vars, and the merge still correctly picks up the integration's own overrides instead of short-circuiting (this was a latent gap introduced while iterating on the fix — closed here with a regression test).
- `data_stream.dataset` promoted to the input level, with and without an additional data-stream-level override.

## Related

> The companion Kibana PR https://github.com/elastic/kibana/pull/275312 (inject `data_stream.dataset` into `templateVars` for composable integrations when no var is present) is **merged**. It's still relevant context for understanding why the var can be safely omitted from the bundle, but is no longer a blocking sequencing dependency for this PR.

## Reviewer notes

- Kept `data_stream.dataset` as a single hardcoded exception rather than a generalized "excluded vars" mechanism, per discussion on the previous revision — no other var currently needs this treatment; happy to refactor into a list if more exceptions come up.
- Didn't add a dedicated OTel-input-based fixture (e.g. mirroring `test/packages/parallel/nginx_composable`): its input packages (`filelog_otel`, `nginx_otel_input`) don't declare `data_stream.dataset` at all today, so that exclusion path isn't exercised by them either way. That package is still built/tested generically by existing CI, so there's no regression risk, but it doesn't add coverage specific to this change.


internal/requiredinputs/testdata/with_merging_dataset_excluded/manifest.yml
```diff
@@ -0,0 +1,33 @@
...
+  Composable package whose imported input package declares data_stream.dataset.
+  The bundler must exclude it from stream-level vars; it is fixed by the
+  integration and must not be user-configurable.
+version: 0.1.0
+type: integration
+categories:
+  - custom
+conditions:
+  kibana:
+    version: "^8.0.0"
```
mrodm: For completeness, this should be 9.4.0 or maybe even 9.5.0

internal/requiredinputs/testdata/with_merging_dataset_excluded/data_stream/test_logs/manifest.yml
```diff
@@ -0,0 +1,6 @@
+title: Test Logs
+type: logs
+streams:
+  - package: input_with_dataset_var
+    title: Test Logs
+    description: Collect test logs using the input package with dataset var.
```
mrodm, line 6: Could it be added another data stream in this test package (fixture), where the `data_stream.dataset` is set as part of the integration package (overriding the value)? 

There are some packages that define `data_stream.dataset` in the integrations repository, and this should still work.

Would it be needed to add another test package here as a fixture to test an OTel integration package? 
There is a `test/packages/parallel/nginx_composable` package, we would need to ensure that these changes are valid in those packages too (OTel ones).

internal/requiredinputs/variables.go
```diff
@@ -261,6 +261,13 @@ func mergeStreamsInDSManifest(
 
 		promotedNames := promotedVarNamesForStream(stream.Package, dsName, promotedVarOverridesByScope)
 
+		// Exclude data_stream.dataset from the input package import: integrations fix
+		// the dataset by data stream name and must not expose it as a user-configurable var.
+		const datasetVarName = "data_stream.dataset"
+		if _, hasDataset := baseVarByName[datasetVarName]; hasDataset {
+			promotedNames[datasetVarName] = true
+		}
```
mrodm: Should it be created a new variable for this list instead? This looks like some exceptions that need to be omitted.

mrodm (commented):
Added a few comments.

I've checked that there is test package for `nginx_composable`, that IIRC it was based on OTel input packages. It would be good to add another test package for an integration package with dependencies based on input packages (based on beats).

internal/requiredinputs/variables.go
teresaromero: i would like to keep it as dataset exception for now as is the one we know. if more exceptions come along we can then do a refactor

teresaromero: after the review this was changed, as it was removing it from the bundled package; for integrations IF the data_stream.dataset is declared at the datastream by the user, this should remain valid.

This keeps the behavour we currently have for some integrations that use this mechanism. The removal should be done on the input we are requiring. When no dataset or variable override, fleet (as now) is the one creating the default dataset based on the package name and stream type

internal/requiredinputs/testdata/with_merging_dataset_excluded/data_stream/test_logs/manifest.yml
teresaromero: i've added a battery of tests covering the possible dataset scenarios. this are kept as unit test fixtures for now. tested manually with the old nginx otel composable and it works aligned with the beats integration. 

still pending to test the manual ingestion instead of system tests.

mrodm:
test integrations

mrodm (approved):
LGTM!
It would be good to add a new test package in `test/packages/parallel` with an integration package with dependencies based on beats inputs. It could be based on https://github.com/elastic/integrations/pull/19719

If needed, it could be pinned the stack version to use for testing this package in that case. Similar to:
https://github.com/elastic/elastic-package/blob/898ce48c3828d7ef94995468a224056d4ba900ed/test/packages/parallel/mysql_input_otel.stack_version

It looks like that there are some other test packages that have pinned the version to be 9.4.0-SNAPSHOT, and those could be removed too (zipkin_input_otel, sql_server_input_otel, mysql_input_otel, nginx_composable). Not to be done as part of this PR. And those packages probably start failing due to the policy tests issue mentioned in: 
- https://github.com/elastic/elastic-package/issues/3710
- https://github.com/elastic/elastic-package/issues/3630
- https://github.com/elastic/integrations/issues/18803#issuecomment-4744033420

PR #3727, opened 2026-07-01
internal/testrunner/runners/pipeline: add OTel validation for otelcol inputs

### LINKED ISSUE #3711: Pipeline tests should use OTel-aware field validation for otelcol input packages
When a data stream uses `input: otelcol`, the system test runner correctly enables OTel-aware field validation via `WithOTelValidation(true)`, which skips strict field-path checking. This is necessary because the OTel component templates (`otel@mappings`, `logs-otel@mappings`) define `attributes` and `resource.attributes` as `passthrough` containers with dynamic sub-fields — a type that the package-spec does not support in field definitions.

The pipeline test runner does not do this. It validates every field in the simulated output against the package's field definitions. This creates an impossible situation for any otelcol package that has a custom ingest pipeline:

1. OTLP data arrives with fields under `attributes.*`. A custom ingest pipeline necessarily reads from and writes to these paths.
2. To pass pipeline test field validation, `attributes.*` fields must be defined in the package's `fields/*.yml`.
3. Defining any `attributes.*` field creates an implicit `object` mapping in the package's component template.
4. The OTel component templates define `attributes` as `type: passthrough`. Elasticsearch rejects the composed index template because `object` and `passthrough` are incompatible.

So the package author is stuck: defining `attributes.*` fields breaks system tests (template composition error), and not defining them breaks pipeline tests (fields reported as undefined).

The system test runner already handles this:

```go
// internal/testrunner/runners/system/tester.go
if dsManifest.Input == "otelcol" {
    validatorOptions = append(validatorOptions,
        fields.WithOTelValidation(true),
    )
}
```

The pipeline test runner should do the same: check the data stream manifest and pass `WithOTelValidation(true)` when the input is `otelcol`.

**Relevant code:**

- System test runner (has the fix): `internal/testrunner/runners/system/tester.go` — checks `dsManifest.Input == "otelcol"`
- Pipeline test runner (missing the fix): `internal/testrunner/runners/pipeline/tester.go` — validator options at ~line 217, no OTel check
- Validator: `internal/fields/validate.go` — `WithOTelValidation` / `enabledOTelValidation` skips `validateMapElement` when true

**Workaround:** Skip pipeline tests in the package's `_dev/test/config.yml`. The ingest pipeline warnings tests (which verify the pipeline runs without errors) still execute. The system test covers end-to-end validation with OTel-aware field checking.

efd6:
Enable OTel field validation mode and accept the .otel dataset suffix in pipeline tests for packages that declare an otelcol input stream. This mirrors the behaviour already present in the system test runner.

Extract defaultExpectedDatasets and isOTelCollectorInput as testable helpers and add table-driven unit tests for both.

> [!NOTE]
> Tested with the elastic/integrations#19765 addition with its pipeline tests unskipped: confirmed working.

Fixes #3711

### DIFF
--- a/internal/testrunner/runners/pipeline/tester.go
+++ b/internal/testrunner/runners/pipeline/tester.go
@@ -206,11 +206,7 @@ func (r *tester) run(ctx context.Context) ([]testrunner.TestResult, error) {
 	}
 
 	if len(expectedDatasets) == 0 {
-		expectedDataset := dsManifest.Dataset
-		if expectedDataset == "" {
-			expectedDataset = pkgManifest.Name + "." + r.testFolder.DataStream
-		}
-		expectedDatasets = []string{expectedDataset}
+		expectedDatasets = defaultExpectedDatasets(pkgManifest.Name, r.testFolder.DataStream, dsManifest)
 	}
 
 	results := make([]testrunner.TestResult, 0)
@@ -222,6 +218,7 @@ func (r *tester) run(ctx context.Context) ([]testrunner.TestResult, error) {
 		fields.WithExpectedDatasets(expectedDatasets),
 		fields.WithEnabledImportAllECSSChema(true),
 		fields.WithSchemaURLs(r.schemaURLs),
+		fields.WithOTelValidation(isOTelCollectorInput(dsManifest)),
 	}
 	result, err := r.runTestCase(ctx, r.testCaseFile, dataStreamRoot, dsManifest.Type, entryPipeline, validatorOptions)
 	if err != nil {
@@ -580,4 +577,29 @@ func checkErrorMessage(event json.RawMessage) error {
 	default:
 		return fmt.Errorf("unexpected pipeline error (unexpected error.message type %T): %[1]v", m)
 	}
+}
+
+// defaultExpectedDatasets returns the expected dataset values when no reroute
+// processors override them. For otelcol inputs, both the base dataset and the
+// agent-appended .otel variant are expected.
+func defaultExpectedDatasets(pkgName, dataStream string, dsManifest *packages.DataStreamManifest) []string {
+	ds := dsManifest.Dataset
+	if ds == "" {
+		ds = pkgName + "." + dataStream
+	}
+	datasets := []string{ds}
+	if isOTelCollectorInput(dsManifest) {
+		datasets = append(datasets, ds+".otel")
+	}
+	return datasets
+}
+
+// isOTelCollectorInput returns true if the data stream manifest declares an otelcol input.
+func isOTelCollectorInput(dsManifest *packages.DataStreamManifest) bool {
+	for _, s := range dsManifest.Streams {
+		if s.Input == "otelcol" {
+			return true
+		}
+	}
+	return false
 }
--- a/internal/testrunner/runners/pipeline/tester.go
+++ b/internal/testrunner/runners/pipeline/tester.go
@@ -596,10 +596,7 @@ func defaultExpectedDatasets(pkgName, dataStream string, dsManifest *packages.Da
 
 // isOTelCollectorInput returns true if the data stream manifest declares an otelcol input.
 func isOTelCollectorInput(dsManifest *packages.DataStreamManifest) bool {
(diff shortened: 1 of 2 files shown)


internal/testrunner/runners/pipeline/tester_test.go
```diff
@@ -0,0 +1,143 @@
...
+		{
+			title: "otelcol among other inputs",
+			manifest: &packages.DataStreamManifest{
+				Streams: []packages.Stream{
+					{Input: "logfile"},
+					{Input: "otelcol"},
+				},
+			},
+			expected: true,
+		},
```
mrodm, line 134: Not sure this could be problematic. 
If the package contains 2 or more streams, being one of them `otelcol` input, it will turn on the OTel validation. 

That validation implies that no field mappings are validated. And therefore, for the case of pipeline tests will be only comparing the expected and the got results.

Do you think this would be ok @efd6 ?

cc @teresaromero

internal/testrunner/runners/pipeline/tester.go
```diff
@@ -581,3 +578,28 @@ func checkErrorMessage(event json.RawMessage) error {
...
+}
+
+// isOTelCollectorInput returns true if the data stream manifest declares an otelcol input.
+func isOTelCollectorInput(dsManifest *packages.DataStreamManifest) bool {
+	for _, s := range dsManifest.Streams {
+		if s.Input == "otelcol" {
+			return true
+		}
+	}
+	return false
```
mrodm: Nit. Just a suggestion that it could be used `slices.ContainsFunc` here.


```suggestion
	return slices.ContainsFunc(dsManifest.Streams, func(s packages.Stream) bool {
		return s.Input == "otelcol"
	})
}
```

internal/testrunner/runners/pipeline/tester.go
```diff
@@ -581,3 +578,28 @@ func checkErrorMessage(event json.RawMessage) error {
...
+	if isOTelCollectorInput(dsManifest) {
+		datasets = append(datasets, ds+".otel")
+	}
+	return datasets
+}
+
+// isOTelCollectorInput returns true if the data stream manifest declares an otelcol input.
+func isOTelCollectorInput(dsManifest *packages.DataStreamManifest) bool {
+	for _, s := range dsManifest.Streams {
+		if s.Input == "otelcol" {
```
teresaromero: this can be an issue when we have composable packages. pipeline test read the source of the package, so instead of `input` we have `package`. the `input` field will be on the built artifact. we need to address this as part of the composable initiative cc @mrodm

internal/testrunner/runners/pipeline/tester.go
```diff
@@ -581,3 +578,28 @@ func checkErrorMessage(event json.RawMessage) error {
...
+// processors override them. For otelcol inputs, both the base dataset and the
+// agent-appended .otel variant are expected.
+func defaultExpectedDatasets(pkgName, dataStream string, dsManifest *packages.DataStreamManifest) []string {
+	ds := dsManifest.Dataset
+	if ds == "" {
+		ds = pkgName + "." + dataStream
+	}
+	datasets := []string{ds}
+	if isOTelCollectorInput(dsManifest) {
+		datasets = append(datasets, ds+".otel")
```
teresaromero, line 592: related to https://github.com/elastic/elastic-package/pull/3727/changes#r3511706528 

i dont think we want the suffix added to any stream that is not otelcol. if i am not mistaken this will be true when ANY stream input is otelcol; from a list of streams possible... 

i dont know if having inputs suffixed with `.otel` is something we could accept. can we test it?

internal/testrunner/runners/pipeline/tester_test.go
efd6: At the moment we have that behaviour but worse; it is necessary to disable otelcol data stream pipeline tests completely.

internal/testrunner/runners/pipeline/tester.go
efd6: This is an acceptance list, we're not mutating anything.

teresaromero (approved):

PR #3760, opened 2026-07-15
fix(policy): normalize bare OTel extension refs across Fleet versions

mrodm:
## Summary

Policy tests for OTel-based packages (e.g. `otlp_input_otel`) were failing on 9.5.0-SNAPSHOT because Fleet started suffixing bare extension IDs (`basicauth`, `bearertokenauth`, `oidc`, etc.) with component ID suffixes (`basicauth/componentid-0`). Existing expected files were partially updated — the extension map keys gained the suffix, but string references in `service.extensions` and `auth.authenticator` still used the bare name, causing spurious diffs.

This PR adds pre-normalization logic so that expected files in any of the three possible states all normalize to the same canonical form:

| State | Extension map key | References |
|---|---|---|
| Fully bare | `basicauth:` | `authenticator: basicauth` |
| Mixed | `basicauth/componentid-0:` | `authenticator: basicauth` |
| Fully suffixed | `basicauth/componentid-0:` | `authenticator: basicauth/componentid-0` |

In the fully-suffixed state the new logic is a no-op: `resolveExtensionRefs` builds the type→canonical map but finds no bare type names at reference positions, so nothing changes.

## Related

- Relates to https://github.com/elastic/elastic-package/issues/3710
- Relates to https://github.com/elastic/integrations/issues/18803
- Relates to https://github.com/elastic/integrations/issues/20180
- https://github.com/elastic/integrations/pull/20141 — integrations PR testing OTel packages against the current stack and 9.5.0-SNAPSHOT

### `preNormalizePolicy`

- **Bare extension map keys** (e.g. `extensions: { basicauth: {...} }`): renamed to `<name>/_bare` so they are treated as variable keys and participate in component-ID normalization.
- **Bare pipeline keys** (`logs`, `metrics`) and bare `forward` connector: renamed to `<name>/_bare` (existing behaviour, now also covers extension keys).
- **`/_bare` sentinel collision guard**: before any rename, checks whether the target `<name>/_bare` key already exists. If it does, the rename is skipped so the existing entry is not silently overwritten. Applied to all three rename sites.
- **OTTL `where … == nil` suffixes**: stripped from `set()` statements added by Fleet since kibana#274993 (existing behaviour, unchanged).

### `preNormalizeNode`

Simplified: now only renames `"forward"` → `"forward/_bare"` inside list elements and strips OTTL `where` clauses. Extension string references are no longer touched here — they are resolved later by `resolveExtensionRefs` at known structural positions only.

### `resolveExtensionRefs`

Runs after `buildSectionMapping` (so canonical IDs are known) but before `applyNormalization`. Rewrites bare extension type-name strings at the three known OTel reference positions:

1. `service.extensions[]` — list items
2. `*.auth.authenticator` — scalar value at any nesting depth (gRPC + HTTP, client and server, per `config/configauth`)
3. `*.middlewares[].id` — `id` value inside list elements at any nesting depth (per `config/configmiddleware`, not yet used in current elastic packages)

`buildExtensionTypeMapping` builds the `bare-type → canonical` map from extension keys. Types with more than one extension are excluded to prevent non-deterministic resolution — the expected file must use the full canonical ID in that case.

## Test coverage

New unit tests in `TestNormalizePolicyToCanonical`:

- Bare extension key referenced from `service.extensions`
- Bare extension key referenced from `auth.authenticator`
- Bare and suffixed extension normalizing to the same result (equivalence)
- Mixed-state: suffixed map key with bare references (the primary failing case)
- Two bare extensions with correct cross-references
- Extension type name used as arbitrary config value is **not** renamed (regression for the over-broad renaming that was replaced by targeted walking)
- Two extensions sharing the same type prefix: bare-type mapping is skipped, full IDs must be used (non-determinism guard)
- `/_bare` sentinel collision: bare rename skipped when `_bare` variant already exists (covers both `forward/_bare` connector and `basicauth/_bare` extension)
- Bare `forward` connector normalization (existing)
- Bare and suffixed `forward` equivalence (existing)
- Bare and suffixed pipeline keys normalize to the same result (kibana#270487)
- OTTL `where` clause stripping (existing)

New end-to-end cases in `TestComparePolicies`:

- Mixed-state expected equals fully-suffixed found with same body (positive)
- Multi-extension mixed-state expected equals fully-suffixed found (positive)
- Different extension body detected after bare-ref normalization (negative)
- Different extension type detected (negative)
- Extra extension in found detected (negative)
- Swapped authenticator assignments between receivers detected (negative)

## Merge commit message

```
fix(policy): normalize bare OTel extension refs across Fleet versions

Fleet 9.5.0 started suffixing bare OTel extension IDs (e.g. basicauth →
basicauth/componentid-0). Add pre-normalization so policy expected files
in any state (fully bare, mixed, fully suffixed) compare cleanly against
any Fleet version.
```

---

> This PR was developed with the assistance of [Claude](https://claude.ai) (claude-sonnet-4-6).


mrodm:
test integrations

mrodm:
test integrations

mrodm:
test integrations

mrodm:
test serverless

internal/testrunner/runners/policy/policy.go
teresaromero, line 1: can we move the normalization helpers into its own file? i feel this file is getting to large and this could use its own house

internal/testrunner/runners/policy/policy.go
```diff
@@ -380,6 +476,114 @@ func replaceOrRecurse(v any, idMapping map[string]string) any {
...
+
+	// Resolve auth.authenticator and middlewares[].id throughout the rest of the tree.
+	resolveExtensionRefNode(root, typeToCanonical)
+}
+
+// buildExtensionTypeMapping returns a map from bare extension type name (e.g. "basicauth") to
+// its canonical component ID (e.g. "basicauth/componentid-0"), derived from the extension keys
+// and their idMapping entries. Types with more than one extension are excluded to prevent
+// non-deterministic resolution.
+func buildExtensionTypeMapping(root map[string]any, idMapping map[string]string) map[string]string {
```
teresaromero:
```suggestion
func buildExtensionTypeMapping(extensions map[string]any, idMapping map[string]string) map[string]string {
```

can we scope the param to just the extensions required?

internal/testrunner/runners/policy/policy.go
```diff
@@ -236,6 +236,93 @@ func isOTelVariableKey(key string) bool {
 	return strings.Contains(key, "/")
 }
 
+// ottlConditionalDataStreamAttr matches " where attributes["data_stream.<field>"] == nil"
+// suffixes appended by Fleet since kibana#274993. Stripping them makes policy comparison
+// stable across Fleet versions that do and don't add the guard.
+var ottlConditionalDataStreamAttr = regexp.MustCompile(` where attributes\["data_stream\.\w+"\] == nil$`)
```
teresaromero:
```suggestion
var otelConditionalDataStreamAttr = regexp.MustCompile(` where attributes\["data_stream\.\w+"\] == nil$`)
```

??

internal/testrunner/runners/policy/policy.go
```diff
@@ -236,6 +236,93 @@ func isOTelVariableKey(key string) bool {
...
+	// Walk string elements in arrays to:
+	//   - replace bare "forward" connector refs with "forward/_bare" (pipeline arrays)
+	//   - strip conditional "where ... == nil" OTTL suffixes from set() statements
+	preNormalizeNode(root)
+}
+
+// preNormalizeNode recursively walks the tree. String elements inside slices are rewritten;
+// map keys are left to the later normalization pass. Extension string references are NOT
+// touched here — they are resolved at known structural positions by resolveExtensionRefs.
+func preNormalizeNode(node any) {
```
teresaromero: will this replace "forward" in any place it appears across the policy? do we want to scope this specifically to a given section?

internal/testrunner/runners/policy/policy.go
```diff
@@ -380,6 +476,114 @@ func replaceOrRecurse(v any, idMapping map[string]string) any {
...
+	case map[string]any:
+		for k, v := range n {
+			switch k {
+			case "authenticator":
+				if s, ok := v.(string); ok {
+					if canonical, found := typeToCanonical[s]; found {
+						n[k] = canonical
+					}
+				}
+			case "middlewares":
```
teresaromero: can we add some unit testing for this case?

internal/testrunner/runners/policy/policy.go
mrodm: I'd keep here ottl since it refers to the Opentelemetry Transformation Language (OTTL). `otel` could be seen as a broader concept here.

teresaromero: meme

internal/testrunner/runners/policy/policy.go
mrodm: Add it here https://github.com/elastic/elastic-package/pull/3760/commits/69ab2182006f36a3f0f57ed35e05642174adff7a

internal/testrunner/runners/policy/policy.go
mrodm: Updated the code to use `extensions` here https://github.com/elastic/elastic-package/pull/3760/commits/de6bbfe9c7fe1edc30cbb506e95252c539232e2f

internal/testrunner/runners/policy/policy.go
mrodm: Good catch!
I added a test to avoid regresssions and updated the code to avoid that https://github.com/elastic/elastic-package/pull/3760/commits/9a63a1dd1b37c0fbabaa151c0f03c7a79f049c36

internal/testrunner/runners/policy/policy.go
mrodm: You're right. I added two new files clean.go and normalize.go to split the functions in those two files. Tests have been moved accordingly too.

Changes performed in https://github.com/elastic/elastic-package/pull/3760/commits/f10636feb7c8e620b141ae71398ceb48f69a7358

mrodm:
test serverless

mrodm:
test integrations

mrodm:
test integrations

mrodm:
test serverless

mrodm:
Tests run after the addressing review comments:
- Serverless https://buildkite.com/elastic/elastic-package-test-serverless/builds/898
- Integrations local minimum stacks supported https://buildkite.com/elastic/integrations/builds/46746
- Integrations 9.5.0-SNAPSHOT stack: https://buildkite.com/elastic/integrations-test-stack/builds/71

teresaromero (approved):
🚀

PR #3768, opened 2026-07-20
internal/testrunner/script: add stack_constraint condition for version-gated tests

efd6:
Add a testscript condition that checks whether the running stack version satisfies a semver constraint. This allows script tests to skip when the stack is too old (or too new) for the feature under test.

Usage: [!stack_constraint:>=9.5.0] skip 'requires stack >= 9.5.0'

Pre-release suffixes (e.g. -SNAPSHOT) are stripped before checking so that snapshots satisfy release constraints. The stack version is also exposed as the STACK_VERSION environment variable.


mrodm (approved):
👍

PR #3771, opened 2026-07-21
testrunner/runners/policy: add configurable field ignoring

efd6:
Please take a look.

ref https://github.com/elastic/integrations/issues/14136#issuecomment-5029715284

### DIFF
--- a/internal/testrunner/globaltestconfig.go
+++ b/internal/testrunner/globaltestconfig.go
@@ -82,6 +82,18 @@ type GlobalRunnerTestConfig struct {
 	Parallel        bool                     `config:"parallel"`
 	Requires        []PackageTestRequirement `config:"requires"`
 	SkippableConfig `config:",inline"`
+
+	// PackageVersion overrides the package version during test setup.
+	// When set, the source manifest is temporarily patched so that
+	// _meta.package.version resolves to a stable, known value instead
+	// of the real (changing) version. Only meaningful for policy tests.
+	PackageVersion string `config:"package_version"`
+
+	// IgnoreFields lists rendered stream fields to strip before comparing
+	// the actual policy against the expected file. Paths use dot notation
+	// scoped to the stream level (e.g. "state.user_agent" strips
+	// inputs[].streams[].state.user_agent). Only meaningful for policy tests.
+	IgnoreFields []string `config:"ignore_fields"`
 }
 
 func ReadGlobalTestConfig(packageRoot string) (*globalTestConfig, error) {
--- a/internal/testrunner/globaltestconfig.go
+++ b/internal/testrunner/globaltestconfig.go
@@ -20,7 +20,7 @@ import (
 type globalTestConfig struct {
 	Asset    GlobalRunnerTestConfig `config:"asset"`
 	Pipeline GlobalRunnerTestConfig `config:"pipeline"`
-	Policy   GlobalRunnerTestConfig `config:"policy"`
+	Policy   PolicyRunnerTestConfig `config:"policy"`
 	Static   GlobalRunnerTestConfig `config:"static"`
 	System   GlobalRunnerTestConfig `config:"system"`
 }
@@ -49,7 +49,7 @@ func (c *globalTestConfig) MergedRequiresSourceOverrides(packageRoot string) (ma
 		cfg   GlobalRunnerTestConfig
 	}{
 		{"system", c.System},
-		{"policy", c.Policy},
+		{"policy", c.Policy.GlobalRunnerTestConfig},
 		{"pipeline", c.Pipeline},
 		{"static", c.Static},
 		{"asset", c.Asset},
@@ -82,17 +82,17 @@ type GlobalRunnerTestConfig struct {
 	Parallel        bool                     `config:"parallel"`
 	Requires        []PackageTestRequirement `config:"requires"`
 	SkippableConfig `config:",inline"`
+}
 
-	// PackageVersion overrides the package version during test setup.
-	// When set, the source manifest is temporarily patched so that
-	// _meta.package.version resolves to a stable, known value instead
-	// of the real (changing) version. Only meaningful for policy tests.
-	PackageVersion string `config:"package_version"`
+// PolicyRunnerTestConfig extends GlobalRunnerTestConfig with fields specific
+// to the policy test runner.
+type PolicyRunnerTestConfig struct {
+	GlobalRunnerTestConfig `config:",inline"`
 
 	// IgnoreFields lists rendered stream fields to strip before comparing
 	// the actual policy against the expected file. Paths use dot notation
 	// scoped to the stream level (e.g. "state.user_agent" strips
-	// inputs[].streams[].state.user_agent). Only meaningful for policy tests.
+	// inputs[].streams[].state.user_agent).
 	IgnoreFields []string `config:"ignore_fields"`
 }
 
--- a/internal/testrunner/runners/policy/runner.go
+++ b/internal/testrunner/runners/policy/runner.go
@@ -6,6 +6,7 @@ package policy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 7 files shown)


efd6:
The linter is wrong. It claims:
```
mv /buildkite/builds/bk-agent-prod-k8s-1784605523478880110/elastic/elastic-package/build/tools/golangci-lint /buildkite/builds/bk-agent-prod-k8s-1784605523478880110/elastic/elastic-package/build/tools/golangci-lint-v2.12.2
/buildkite/builds/bk-agent-prod-k8s-1784605523478880110/elastic/elastic-package/build/tools/golangci-lint-v2.12.2 run
internal/testrunner/runners/policy/runner.go:223:40: patchManifestVersion - newVersion always receives "0.0.0" (unparam)
func patchManifestVersion(packageRoot, newVersion string) (restore func() error, err error) {
```

This is not the case since it's called with `r.globalTestConfig.PackageVersion` in the second position [here](https://github.com/elastic/elastic-package/blob/073db4d7d0d9626737a8f041193a5905a2d5e99d/internal/testrunner/runners/policy/runner.go#L178), this value is configurable from YAML files.

This is a bug in golangci-lint; the upstream linter that is doing the work does not find an issue.
```
..package [dynamic_policy_tests|✔ ] ❯❯❯ golangci-lint run                 
internal/testrunner/runners/policy/runner.go:223:40: patchManifestVersion - newVersion always receives "0.0.0" (unparam)
func patchManifestVersion(packageRoot, newVersion string) (restore func() error, err error) {
                                       ^
1 issues:
* unparam: 1
..package [dynamic_policy_tests|✔ ] ❯❯❯ unparam ./internal/testrunner/runners/policy/... && echo no problems here
no problems here
```
however, it does fail if I call with `-test` so maybe it's a bug in upstream `unparam`. It looks like the packages are not built correctly when test packages and non-test packages are included together.

Confirmed bug in `unparam` with a minimal repro. We should not be using this linter.

internal/testrunner/globaltestconfig.go
```diff
@@ -82,6 +82,18 @@ type GlobalRunnerTestConfig struct {
 	Parallel        bool                     `config:"parallel"`
 	Requires        []PackageTestRequirement `config:"requires"`
 	SkippableConfig `config:",inline"`
+
+	// PackageVersion overrides the package version during test setup.
+	// When set, the source manifest is temporarily patched so that
+	// _meta.package.version resolves to a stable, known value instead
+	// of the real (changing) version. Only meaningful for policy tests.
```
teresaromero: > Only meaningful for policy tests.

If this configs are only applicable to policy testing, can we move them to a specific PolicyRunnerTestConfig that inherits de global and adds this two fields?

We also need to enable this fields under the spec. So either an issue or an additional PR should be filed for the spec to support this new fields, ideally also for the policy tests config spec

teresaromero (commented):
Thanks for the PR. I'm uneasy about package_version patching manifest.yml for install. Even with validation skipped, mutating the source tree for a stable _meta.package.version feels brittle (restore on failure/crash, line-based YAML edits, install/uninstall coupling). From your note on https://github.com/elastic/integrations/issues/14136#issuecomment-5029715284, the hard part is that we can't inject a constant into _meta — but I don't think rewriting the package version is the right workaround.

  For the CrowdStrike-style case, I'd prefer ignore_fields alone as the first step: strip dynamic stream fields (e.g. state.user_agent / user_agent) during compare so expecteds don't churn across package bumps or stack versions (unknown vs real version). That matches how we already drop meta.package.version in policyEntryFilters, and the broader direction in https://github.com/elastic/elastic-package/issues/3710 / https://github.com/elastic/elastic-package/pull/3760 and your normalize suggestion (https://github.com/elastic/integrations/issues/18803#issuecomment-4888078609): absorb non-deterministic policy output in elastic-package rather than changing what we install or version-gating every case.

  If we later need to assert the field exists (or matches a shape) without pinning a concrete version, we could extend the policy config with declared dynamic fields and optional regex normalization — still without rewriting the package version.

  Also: these keys should be policy-only (not on the shared runner config type), and need a package-spec update — _dev/test/config.yml's config_tests has additionalProperties: false today, so packages can't use ignore_fields under policy: without failing validation.

  I'd drop package_version for now and land ignore_fields (+ package-spec).
  
  EDIT: also, could we add some docs on the how-to guides to reflect this update for policy tests?

internal/testrunner/globaltestconfig.go
```diff
@@ -82,6 +82,18 @@ type GlobalRunnerTestConfig struct {
...
+	PackageVersion string `config:"package_version"`
+
+	// IgnoreFields lists rendered stream fields to strip before comparing
+	// the actual policy against the expected file. Paths use dot notation
+	// scoped to the stream level (e.g. "state.user_agent" strips
+	// inputs[].streams[].state.user_agent). Only meaningful for policy tests.
+	IgnoreFields []string `config:"ignore_fields"`
```
mrodm, line 96: I was wondering if this could be extend to be able to define in the policy tests config a key to be able to add more normalize fields a part form the ones hardcoded in the elastic-package code

```yaml
normalize_fields:                       # single new config key
  # replace: value differs (e.g. embedded version) -> stable token
  - path: "inputs.streams.state.user_agent"
    pattern: "Elastic-[^/]+/[0-9.]+"    # optional; regex on the value
    value: "Elastic-Agent/VERSION"
  # ignore: field absent on old stacks -> drop it (no value => delete)
  - path: "inputs.streams.state.batch_size"

```

Not tested , but the idea here would be to allow defining how to replace the value of given keys that match some regexes.

The second entry in that example, it would allow to delete the keys (like ignoring them).

The benefit of replacing via regexes, is that values that do not match regex will be kept and make the tests failing.


As mentioned by Tere, all these keys require changes in package-spec in order to allow developers to define them in the corresponding files.

internal/testrunner/runners/policy/runner.go
```diff
@@ -172,6 +173,14 @@ func (r *runner) Type() testrunner.TestType {
 }
 
 func (r *runner) setupSuite(ctx context.Context, manager *resources.Manager) (cleanup func(ctx context.Context) error, err error) {
+	var restoreVersion func() error
+	if r.globalTestConfig.PackageVersion != "" {
+		restoreVersion, err = patchManifestVersion(r.packageRoot, r.globalTestConfig.PackageVersion)
+		if err != nil {
+			return nil, fmt.Errorf("patching manifest version for policy test: %w", err)
+		}
+	}
```
mrodm: I'd prefer to find some other approach if possible for this use case. Even if there is a cleanup function returned, I guess there could be scenarios where the manifest file could be left with the changes (some panic?).

mrodm (commented):
Agree with @teresaromero's review here https://github.com/elastic/elastic-package/pull/3771/#pullrequestreview-4786767105

I'd prefer to find a solution for this use case that do not require updating the package manifest.

I added another suggestion to use something like `ignore_fields` but with regexes. Thereby, just the keys that contain values that matches regexes are replaced or dropped.

teresaromero (dismissed):
are you taking into consideration the normalization comment from @mrodm or perhaps that can be moved to other issue? if for now we are not needing to check dynamic values i guess this looks ok. 

i will rather wait for mario to review again too. thanks

efd6:
@teresaromero elastic/package-spec#1214

mrodm (commented):
There are some conflicts after merging this PR https://github.com/elastic/elastic-package/pull/3760

That PR performed some refactors to split the functions in different files, that would probably caused those conflicts.

For this PR, this new setting is just added to the global settings. That would make not possible to define specific `ignore_fields` per policy test. Could this be an issue ? Those fields would be ignored in all policy tests.

efd6:
> That would make not possible to define specific `ignore_fields` per policy test. Could this be an issue ? Those fields would be ignored in all policy tests.

I can't see a good reason to ignore fields only sometimes.

teresaromero (approved):
This depends on https://github.com/elastic/package-spec/pull/1214 to be merged and released. - Do not merge until package-spec is updated.

It will be good to have a test package where this is being used in action. Will you be updating https://github.com/elastic/integrations/pull/20241 with this approach? this change can be included on the next release so you can use it there. 

The proposal of validating regex patterns for dynamic fields i think is still relevant. We should perhaps file an issue to include this on upcoming releases, as this pr opens the door to ignore and reducing the test scope.

efd6:
I was planning to merge elastic/integrations#20241 as is and adding this when it becomes available; this will result in a low level of flaky tests when running serverless tests, but I think that's OK.

mrodm (approved):
> > That would make not possible to define specific `ignore_fields` per policy test. Could this be an issue ? Those fields would be ignored in all policy tests.
> 
> I can't see a good reason to ignore fields only sometimes.

👍 

> The proposal of validating regex patterns for dynamic fields i think is still relevant. We should perhaps file an issue to include this on upcoming releases, as this pr opens the door to ignore and reducing the test scope.

I guess this could be interesting to do at some point too 👍

PR #3792, opened 2026-07-29
fix: handle non-string title in export dashboard standardization

mrodm:
## Summary

`elastic-package export dashboards` panicked when processing Kibana dashboards
that contain XY chart panels. The root cause was an unsafe type assertion in
`standardizeObjectProperties`: the code assumed the `title` key was always a
`string`, but in axis config objects (`embeddableConfig.axis.x/y/y2`) it is a
nested map like `{"visible": false}`.

Example (snippet from dashboard export):
```json
{
  "x": {
    "domain": {
      "rounding": false,
      "type": "fit"
    },
    "grid": {
      "visible": true
    },
    "labels": {
      "orientation": "horizontal"
    },
    "scale": "temporal",
    "ticks": {
      "visible": true
    },
    "title": {
      "visible": false
    }
  },
  "y": {
    "domain": {
      "rounding": true,
      "type": "full"
    },
    "grid": {
      "visible": true
    },
    "labels": {
      "orientation": "horizontal"
    },
    "scale": "linear",
    "ticks": {
      "visible": true
    },
    "title": {
      "visible": false
    }
  }
}
```

## Changes

- **`internal/export/transform_standardize.go`**: Replace the direct type
  assertion `value.(string)` for the `title` key with a safe assertion. When
  `title` is not a string, the transformation is skipped and the value falls
  through to the general map/array recursion unchanged.
- **`internal/export/transform_standardize_test.go`** *(new)*: Unit tests for
  `standardizeObjectProperties` covering:
  - String title with/without ` ECS` suffix
  - Map-typed title (the axis config case) — no panic, value preserved
  - Nested map and array traversal
  - Markdown link rewriting via `adjustObjectID`

Tested with dashboards defined in https://github.com/elastic/integrations/pull/20265

## Proposed commit

```
fix: handle non-string title values in standardizeObjectProperties

In Kibana dashboard JSON, axis config objects (embeddableConfig.axis.x/y/y2)
use "title" as a nested map (e.g. {"visible": false}) rather than a string.
The previous direct type assertion panicked in those cases. Switch to a safe
assertion so non-string title values fall through to the general map/array
recursion unchanged.

Add unit tests covering: string title with/without ECS suffix, map-typed
title (axis case), nested and array traversal, and markdown link ID
adjustment.
```

---

### DIFF
--- a/internal/export/transform_standardize.go
+++ b/internal/export/transform_standardize.go
@@ -57,11 +57,16 @@ func adjustObjectReferences(ctx *transformationContext, references []interface{}
 func standardizeObjectProperties(ctx *transformationContext, object common.MapStr) (common.MapStr, error) {
 	for key, value := range object {
 		if key == "title" {
-			_, err := object.Put(key, adjustTitleProperty(value.(string)))
-			if err != nil {
-				return nil, fmt.Errorf("can't update field (key: %s): %w", key, err)
+			// "title" is not always a string — e.g. in axis config objects it is a map
+			// like {"visible": false}. Skip the string transformation in those cases and
+			// fall through to the general map/array handling below.
+			if strVal, ok := value.(string); ok {
+				_, err := object.Put(key, adjustTitleProperty(strVal))
+				if err != nil {
+					return nil, fmt.Errorf("can't update field (key: %s): %w", key, err)
+				}
+				continue
 			}
-			continue
 		}
 
 		if key == "markdown" {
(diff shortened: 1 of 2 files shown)


teresaromero (approved):
🐛 thanks!

mrodm:
Tested again with the changes in this PR https://github.com/elastic/integrations/pull/20265
before they were updated to 9.4.3 (using changeset https://github.com/elastic/integrations/pull/20265/commits/8d6cd215d0b04343c252ceef166101577e22904c)

PR #3808, opened 2026-08-04
Add install scripts for macOS, Linux, and Windows

teresaromero:
## Problem

Installing or updating `elastic-package` requires manually navigating to the Releases page, picking the right binary for your OS and architecture, downloading, extracting, removing the macOS quarantine attribute, and moving the binary into `$PATH`. This is tedious to repeat on every release.

## Solution

Two scripts that handle the full install/update flow in one command:

**macOS / Linux**
```bash
curl -fsSL https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/install.sh | bash
```

**Windows (PowerShell)**
```powershell
irm https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/install.ps1 | iex
```

Both scripts:
- Auto-detect OS and architecture
- Fetch the latest release (or a pinned version via `--version` / `-Version`)
- Verify the SHA-256 checksum against the release checksums file before extracting
- Skip the download if the installed version already matches the target
- Show an upgrade message when replacing an older version
- Create the install directory if it doesn't exist

Platform-specific behaviour:
- **macOS**: removes the quarantine attribute automatically
- **Windows**: adds the install directory to the user `PATH` permanently; defaults to `%LOCALAPPDATA%\Programs\elastic-package` (no admin required)

The README Getting started section is updated to lead with the script install method.

## Reviewer notes

- The bash script saves the archive to disk before extraction (rather than piping `curl | tar`) so the checksum can be verified first; the temp directory is `0700` and cleaned on exit.
- `sudo` is only invoked for `mkdir` and `mv` when the unprivileged attempt fails — not proactively.
- The PowerShell PATH dedup uses `Split(';') -notcontains` rather than `-notlike` to avoid bracket characters in the path being interpreted as wildcards.
- Windows is deliberately out of scope for `install.sh`; the error message points to `install.ps1`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,8 @@ irm https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/insta
 ```
 
 The PowerShell script defaults to `%LOCALAPPDATA%\Programs\elastic-package` and adds that directory to
-your user `PATH` automatically. It supports `amd64`, `arm64`, and `386` architectures.
+your user `PATH` automatically. It supports `amd64`, `arm64`, and `386` architectures, and requires
+PowerShell 5.1 or later (included with Windows 10 and later).
 Both scripts skip the download if the installed version is already up to date.
 
 ### Manual download
--- a/README.md
+++ b/README.md
@@ -33,10 +33,22 @@ curl -fsSL https://raw.githubusercontent.com/elastic/elastic-package/main/script
 The script supports macOS (darwin) and Linux on `amd64`, `arm64`, and `386` architectures.
 On macOS it automatically removes the quarantine attribute so the binary can run without Gatekeeper warnings.
 
+To install a specific version, pass `--version`:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/install.sh | bash -s -- --version 0.125.0
+```
+
 On **Windows**, use the PowerShell script instead:
 
 ```powershell
 irm https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/install.ps1 | iex
+```
+
+To install a specific version on Windows:
+
+```powershell
+& ([scriptblock]::Create((irm 'https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/install.ps1'))) -Version 0.125.0
 ```
 
 To install to a custom directory on Windows:
--- a/README.md
+++ b/README.md
@@ -14,11 +14,34 @@ explore the builder tools.
 
 ## Getting started
 
-Download latest release from the [Releases](https://github.com/elastic/elastic-package/releases/latest) page.
+### Install with a script (recommended)
+
+The quickest way to install or update `elastic-package` is with the install script. It automatically detects
+your operating system and architecture, downloads the latest release, and installs the binary to `/usr/local/bin`
+(override with the `INSTALL_DIR` environment variable):
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/install.sh | bash
+```
+
+To install to a custom directory (no `sudo` required):
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/install.sh | INSTALL_DIR=$HOME/.local/bin bash
+```
+
+The script supports macOS (darwin) and Linux on `amd64`, `arm64`, and `386` architectures.
+On macOS it automatically removes the quarantine attribute so the binary can run without Gatekeeper warnings.
+
+### Manual download
+
+Download the latest release directly from the [Releases](https://github.com/elastic/elastic-package/releases/latest) page.
 
 On macOS, use `xattr -r -d com.apple.quarantine elastic-package` after downloading to allow the binary to run.
 
-Alternatively, you may use `go install` but you will not be able to use the `elastic-package version` command or check updates.
+### Install with Go
+
+You may also use `go install`, but you will not be able to use the `elastic-package version` command or check for updates.
 
 ```bash
 go install github.com/elastic/elastic-package@latest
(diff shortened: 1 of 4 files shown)


scripts/install.sh
```diff
@@ -0,0 +1,168 @@
...
+fi
+
+if command -v sha256sum >/dev/null 2>&1; then
+  echo "${EXPECTED_HASH}  ${ARCHIVE}" | sha256sum -c - >/dev/null
+elif command -v shasum >/dev/null 2>&1; then
+  echo "${EXPECTED_HASH}  ${ARCHIVE}" | shasum -a 256 -c - >/dev/null
+else
+  echo "Warning: cannot verify checksum (neither sha256sum nor shasum found)." >&2
+fi
+echo "Checksum verified."
```
mrodm: This message could be misleading for those scenarios where the validation has been skipped (e.g. no commands available).

scripts/install.sh
```diff
@@ -0,0 +1,168 @@
...
+    exit 1
+    ;;
+esac
+
+# Detect architecture
+ARCH="$(uname -m)"
+case "$ARCH" in
+  x86_64)        ARCH="amd64" ;;
+  aarch64|arm64) ARCH="arm64" ;;
+  i386|i686) ARCH="386" ;;
```
mrodm: Nit.


```suggestion
  x86_64)        ARCH="amd64" ;;
  aarch64|arm64) ARCH="arm64" ;;
  i386|i686)     ARCH="386" ;;
```

scripts/install.sh
```diff
@@ -0,0 +1,168 @@
...
+    echo "Failed to determine the latest release version." >&2
+    exit 1
+  fi
+fi
+
+# Check currently installed version, skip if already at the target version
+CURRENT_VERSION=""
+if command -v elastic-package >/dev/null 2>&1; then
+  CURRENT_VERSION=$(elastic-package version 2>/dev/null | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 || true)
+fi
```
mrodm, line 92: Should it be checking also here if `elastic-package` is present in the INSTALL_DIR path ? 

`install.ps1` does something like that since Install Dir variable is defined with the custom value or the default one.

Something like  

```sh
CURRENT_VERSION=""                                        
EXISTING_BINARY="${INSTALL_DIR}/elastic-package"          
if [ -x "$EXISTING_BINARY" ]; then                        
  CURRENT_VERSION=$("$EXISTING_BINARY" version 2>/dev/null | grep -oE '[0-9]+.[0-9]+.[0-9]+' | head -1 || true)  
elif command -v elastic-package >/dev/null 2>&1; then     
  CURRENT_VERSION=$(elastic-package version 2>/dev/null | grep -oE '[0-9]+.[0-9]+.[0-9]+' | head -1 || true)  
fi                                                        
```

scripts/install.sh
```diff
@@ -0,0 +1,168 @@
...
+if [ -w "$INSTALL_DIR" ]; then
+  mv "$BINARY" "${INSTALL_DIR}/elastic-package"
+else
+  echo "Installing to ${INSTALL_DIR} requires elevated privileges..."
+  sudo mv "$BINARY" "${INSTALL_DIR}/elastic-package"
+fi
+
+echo ""
+echo "elastic-package v${VERSION} installed to ${INSTALL_DIR}/elastic-package"
+echo ""
```
mrodm, line 170: Is `elastic-package` inside the tar.gz with execute permission ? 

Should we add `chmod +x "$BINARY"` ?

scripts/install.sh
```diff
@@ -0,0 +1,168 @@
...
+esac
+
+# Resolve target version: use requested version or fetch latest
+if [ -n "$REQUESTED_VERSION" ]; then
+  VERSION="$REQUESTED_VERSION"
+else
+  echo "Fetching latest elastic-package release..."
+  VERSION=$(curl -fsSL "https://api.github.com/repos/${REPO}/releases/latest" \
+    | grep '"tag_name"' \
+    | sed 's/.*"tag_name": *"v\([^"]*\)".*/\1/')
```
mrodm: I guess here it is not used `jq` to avoid any extra dependencies. Is that the case?

teresaromero: yes, the idea here was to make it as "universal" as possible, and jq needs installation. However i found an other way to parse the version following the redirect that github does when fetching the latest url.

teresaromero:
Tested the windows script locally and working as expected 👍🏻 
Screenshot 2026-08-07 at 09 21 22

scripts/install.sh
```diff
@@ -56,32 +56,38 @@ ARCH="$(uname -m)"
...
   VERSION="$REQUESTED_VERSION"
 else
   echo "Fetching latest elastic-package release..."
-  VERSION=$(curl -fsSL "https://api.github.com/repos/${REPO}/releases/latest" \
-    | grep '"tag_name"' \
-    | sed 's/.*"tag_name": *"v\([^"]*\)".*/\1/')
+  LATEST_URL=$(curl -fsSL -o /dev/null -w '%{url_effective}' \
+    "https://github.com/${REPO}/releases/latest")
+  VERSION="${LATEST_URL##*/}"
+  VERSION="${VERSION#v}"
```
mrodm, line 76: Install.ps1 does not use this method. Would it be easy to apply this change there too?

teresaromero: i've done the equivalent. also added a note on the readme for windows compatibly that was missing

README.md
```diff
@@ -59,7 +59,8 @@ irm https://raw.githubusercontent.com/elastic/elastic-package/main/scripts/insta
 ```
 
 The PowerShell script defaults to `%LOCALAPPDATA%\Programs\elastic-package` and adds that directory to
-your user `PATH` automatically. It supports `amd64`, `arm64`, and `386` architectures.
+your user `PATH` automatically. It supports `amd64`, `arm64`, and `386` architectures, and requires
+PowerShell 5.1 or later (included with Windows 10 and later).
```
mrodm, line 63: It is needed to update the template in order to change this README file. This should be added to the `tools/readme/readme.md.tmpl` file too:

https://github.com/elastic/elastic-package/pull/3808/changes#diff-2cbbc2f5de0fcbe7a701cf3347141019b12b8d8ae570644a088dd2087d2b6af6R61-R63

mrodm (approved):

PR #3839, opened 2026-08-13
fix: avoid false positive in mapping validation for dynamic children of a nested ECS parent

### LINKED ISSUE #3639: Mapping validation fails for dynamic child fields under explicit nested ECS parent
`elastic-package test system` will fail mapping validation when a package explicitly defines an ECS nested parent field and Elasticsearch dynamically maps child fields under it during ingest.

Example package field definition:
```
- name: email.attachments
  external: ecs
```
This builds correctly to:
```
- name: email.attachments
  type: nested
```
During system tests, the simulated index template mapping contains the explicit parent mapping:
```
"email": {
  "properties": {
    "attachments": {
      "type": "nested"
    }
  }
}
```
After indexing test documents, Elasticsearch dynamically adds child mappings under the nested field:
```
"email": {
  "properties": {
    "attachments": {
      "type": "nested",
      "properties": {
        "file": {
          "properties": {
            "size": { "type": "long" }
          }
        }
      }
    }
  }
}
```
`elastic-package` then reports:
```
undefined field mappings found in path: "email.attachments"
```
This is a false positive. The explicit parent mapping is valid and the child fields are dynamically mapped by Elasticsearch. Dynamic fields are already accepted when they are absent from the simulated preview and match dynamic templates or ECS schema definitions, but this case fails because the dynamically mapped fields are children of a field that exists in the preview.

### Expected behavior

`elastic-package` should allow dynamically created child mappings under an explicitly defined nested parent, provided those child mappings are valid according to the fallback ECS schema and/or dynamic templates, matching the validation for other dynamically mapped fields.

### Possible fix

When preview and actual mappings both define a field as type: nested, but actual has additional properties, validate those properties with the same logic used for mappings not present in the preview, instead of failing at the parent path.

mrodm:
## Summary

When a package declares an ECS `nested` field (e.g. `email.attachments`) and Elasticsearch dynamically creates child mappings during document ingest, the mapping validator was incorrectly raising a false positive:

```
undefined field mappings found in path: "email.attachments"
```

The preview (simulated index template) carries `{"type":"nested"}` with no `properties`, while the actual data stream mappings gain `properties` from dynamically ingested documents. The validator treated this as an undefined field at the parent level and never inspected the children individually.

This fix detects when a preview mapping has a type that Elasticsearch can populate with dynamically created children (currently `nested`) and the actual mapping preserves the same type. In that case the children are validated via the existing `validateMappingsNotInPreview` path — accepted if they match a dynamic template or the ECS schema, reported as errors otherwise.

Closes #3639

## Changes

- **`internal/fields/mappings.go`** — add `allowsDynamicChildMappings` helper and wire it into `compareMappings` inside the `else if !isObject(preview)` branch.
- **`internal/fields/mappings_test.go`** — add / update 6 test cases covering: ECS children pass, mixed ECS + non-ECS children (error on non-ECS only), dynamic-template match, type drift (error preserved), `object` parent out of scope.
- **`test/packages/other/ecs_nested_fields/`** — new e2e test package that reproduces the issue end-to-end: declares only `email.attachments` in `fields/ecs.yml`, ingest pipeline populates `file.{name,extension,mime_type,size,hash.sha256}` dynamically.

## Proposed commit message

```
fix: avoid false positive in mapping validation for dynamic children of a nested ECS parent

When a package declares an ECS `nested` field (e.g. `email.attachments`) without
declaring its children, and Elasticsearch creates those children dynamically during
ingest, the validator was reporting "undefined field mappings found in path" at the
parent level instead of inspecting the children individually.

Add `allowsDynamicChildMappings` to detect when the preview has a type that allows
dynamic children (`nested`) and the actual mapping preserves that type. Wire it into
`compareMappings` so those children are validated through `validateMappingsNotInPreview`
— accepted if they match a dynamic template or the ECS schema, errors otherwise.

```

## Testing

Verified on two stacks with both the released `v0.125.1` binary (reproduces the false positive) and the patched binary (passes):

| Stack | Binary | Result |
|-------|--------|--------|
| 9.5.0 | v0.125.1 | FAIL — `undefined field mappings found in path: "email.attachments"` |
| 9.5.0 | patched | PASS |
| 8.19.20-SNAPSHOT | v0.125.1 | FAIL |
| 8.19.20-SNAPSHOT | patched | PASS |

Also tested against `microsoft_exchange_online_message_trace` from the integrations repo with the child-field workaround removed — both `cel` and `logfile` test cases pass on both stacks.

---

### DIFF
--- a/internal/fields/mappings.go
+++ b/internal/fields/mappings.go
@@ -300,6 +300,24 @@ func isNumberTypeField(previewType, actualType string) bool {
 	return false
 }
 
+// mappingTypesWithDynamicChildren lists mapping types whose child mappings can be created
+// dynamically by Elasticsearch during document ingest, even when no child mapping is defined
+// in the index template (e.g. ECS nested fields like "email.attachments").
+// Add "object" here once the scope is extended to cover plain object parents.
+var mappingTypesWithDynamicChildren = []string{"nested"}
+
+// allowsDynamicChildMappings reports whether the preview mapping is a type that Elasticsearch
+// can populate with dynamically created child mappings, and that the actual mapping preserves
+// the same type.  When both conditions hold, the extra properties in the actual mapping should
+// be validated as dynamic fields rather than treated as an error.
+func allowsDynamicChildMappings(preview, actual map[string]any) bool {
+	previewType := mappingParameter("type", preview)
+	if !slices.Contains(mappingTypesWithDynamicChildren, previewType) {
+		return false
+	}
+	return previewType == mappingParameter("type", actual)
+}
+
 func (v *MappingValidator) validateMappingInECSSchema(currentPath string, definition map[string]any) multierror.Error {
 	found := FindElementDefinition(currentPath, v.Schema)
 	if found == nil {
@@ -457,8 +475,18 @@ func (v *MappingValidator) compareMappings(path string, couldBeParametersDefinit
 			}
 			return nil
 		} else if !isObject(preview) {
-			errs = append(errs, fmt.Errorf("undefined field mappings found in path: %q", path))
-			return errs.Unique()
+			if !allowsDynamicChildMappings(preview, actual) {
+				errs = append(errs, fmt.Errorf("undefined field mappings found in path: %q", path))
+				return errs.Unique()
+			}
+			// Elasticsearch created these child mappings dynamically while ingesting documents.
+			// Validate them the same way as any mapping not present in the preview.
+			dynamicErrors := v.validateMappingsNotInPreview(path, actual, dynamicTemplates)
+			errs = append(errs, dynamicErrors...)
+			if len(errs) > 0 {
+				return errs.Unique()
+			}
+			return nil
 		}
 		previewProperties, err := getMappingDefinitionsField("properties", preview)
 		if err != nil {
--- a/test/packages/other/ecs_nested_fields/data_stream/logs/sample_event.json
+++ b/test/packages/other/ecs_nested_fields/data_stream/logs/sample_event.json
@@ -1,0 +1,84 @@
+{
+    "@timestamp": "2026-08-13T09:22:33.634Z",
+    "agent": {
+        "ephemeral_id": "714157b9-dbd7-42e3-b542-8bf12d54dfcf",
+        "id": "86c418aa-8eb9-4976-aaaf-40c30b6e3a3d",
+        "name": "elastic-agent-52093",
+        "type": "filebeat",
+        "version": "9.5.0"
+    },
+    "data_stream": {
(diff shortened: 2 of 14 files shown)


mrodm:
test integrations

mrodm:
Tested in https://github.com/elastic/integrations/pull/20700 , the failures reported there are not related to this change. Those errors are related to not being able to create resources in AWS required for testing.

test/packages/other/ecs_nested_fields/data_stream/logs/sample_event.json
```diff
@@ -0,0 +1,84 @@
...
+                ],
+                "file.name.text": [
+                    "report.pdf"
+                ],
+                "file.size": [
+                    52428
+                ]
+            }
+        ]
+    },
```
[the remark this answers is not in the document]
mrodm: This file is obtained by `elastic-package` from the docs validated in system tests. I prefer to keep it as it is without changing it.

internal/fields/mappings.go
```diff
@@ -457,8 +475,18 @@ func (v *MappingValidator) compareMappings(path string, couldBeParametersDefinit
...
 		} else if !isObject(preview) {
-			errs = append(errs, fmt.Errorf("undefined field mappings found in path: %q", path))
-			return errs.Unique()
+			if !allowsDynamicChildMappings(preview, actual) {
+				errs = append(errs, fmt.Errorf("undefined field mappings found in path: %q", path))
+				return errs.Unique()
+			}
+			// Elasticsearch created these child mappings dynamically while ingesting documents.
+			// Validate them the same way as any mapping not present in the preview.
+			dynamicErrors := v.validateMappingsNotInPreview(path, actual, dynamicTemplates)
```
teresaromero, line 484: While reviewing this PR with Claude, it flagged a pre-existing bug in `validateMappingsNotInPreview` that this change makes more reachable.

At line 605, when an exception field is encountered inside the for fieldPath, object := range flattenFields loop, the function does return nil instead of continue. Since map iteration order in Go is non-deterministic, if an exception field happens to be visited first, all remaining fields in the loop are silently skipped — including genuinely undefined ones. This PR wires the new nested-children path through this function, so the bug now applies to dynamic children of nested ECS parents as well.

Should this continue instead of return?

mrodm: Testing this in a this other PR https://github.com/elastic/elastic-package/pull/3842

mrodm: @teresaromero we could deal with that bug in this other PR #3842. 
Could you do another review for this PR ?

teresaromero (approved):

PR #3840, opened 2026-08-13
fix: prevent empty ILM heading and discover lifecycle.yml data streams

### LINKED ISSUE #3709: README renders empty sections when no ILM policies are defined
## README renders empty sections

`elastic-package build` can produce README sections with a heading but no content.

### Confirmed bug

`renderILMPaths` (`internal/docs/ilm.go`) always writes the `### Data streams using ILM policies` heading before checking whether any ILM policies exist. When none are found the heading is emitted with no body.

Example output in `elastic_package_registry` (https://github.com/elastic/integrations/pull/19719/changes#r3472932654):

```markdown
### Data streams using ILM policies


```

Fix: return `""` before writing the heading when `dataStreamNames` is empty.

### Related: `findILMPaths` misses `lifecycle.yml`

The discovery glob only matches `data_stream/*/elasticsearch/ilm/` directories. Packages that define ILM via `data_stream/*/lifecycle.yml` are not found, so `dataStreamNames` ends up empty and the heading-only output is triggered even when ILM is configured.

Fix: extend `findILMPaths` to also detect `lifecycle.yml` files.

### Audit of other template functions

Only `ilm` has this pattern. `transform` already early-returns before writing its heading. `fields`, `inputDocs`, `alertRuleTemplates`, and `sloTemplates` do not embed headings — any empty section there would be an authoring issue in the package's README template.

### Tasks

- [ ] Fix `renderILMPaths` to guard the heading behind a non-empty check
- [ ] Fix `findILMPaths` to include `lifecycle.yml`-based data streams
- [ ] Scan `elastic/integrations` for READMEs with empty sections to measure impact
- [ ] Consider a lint check in `elastic-package check` to flag empty sections

teresaromero:
Resolve #3709

## Why

The `{{ ilm }}` README template helper had three bugs:

1. With no matching data streams it rendered an empty heading instead of an empty string.
2. `findILMPaths` only globbed `elasticsearch/ilm/` directories, so data streams using only `lifecycle.yml` (DLM) were never discovered by bare `{{ ilm }}`.
3. Input packages with a root `lifecycle.yml` were not supported at all.

## What

**Empty heading guard** — added an early-return when the resolved data stream list is empty, consistent with how `{{ transform }}` behaves.

**`lifecycle.yml` discovery** — `findILMPaths` now also globs `data_stream/*/lifecycle.yml`; a `seen` map deduplicates streams that have both files. Returns a sorted list for deterministic output.

**Input package support** — bare `{{ ilm }}` now checks for a root `lifecycle.yml` first (input packages), falls back to per-data-stream discovery (integration packages). Named args still bypass this and go straight to data-stream rendering.

Also adds documentation in the how-to guide and test coverage (unit tests + two new test packages: `readme_ilm` for integration packages, `readme_ilm_input` for input packages).

### DIFF
--- a/docs/howto/add_package_readme.md
+++ b/docs/howto/add_package_readme.md
@@ -124,9 +124,12 @@ List of placeholders that can be used in the Markdown templates:
       ```
     - Requirements:
         - It is needed to define a file with the links definitions. More information [in this section](#requirements)
-- `ilm [data_stream...]`: this placeholder renders ILM policies defined under
-  `data_stream/<name>/elasticsearch/ilm/`.
-    - Example of usage (all data streams with ILM policies):
+- `ilm [data_stream...]`: this placeholder renders ILM policies or lifecycle
+  configuration for each data stream. Two file locations are supported:
+    - `data_stream/<name>/elasticsearch/ilm/<policy>.json` — ILM policy file.
+    - `data_stream/<name>/lifecycle.yml` — DLM (Data Lifecycle Management)
+      configuration (takes precedence when both exist for the same data stream).
+    - Example of usage (all data streams):
       ```
       {{ ilm }}
       ```
@@ -134,10 +137,11 @@ List of placeholders that can be used in the Markdown templates:
       ```
       {{ ilm "metrics" }}
       ```
-    - When no data stream arguments are given, `elastic-package` discovers data streams
-      that have an `elasticsearch/ilm/` directory and documents each policy as a
+    - When no data stream arguments are given, `elastic-package` discovers all
+      data streams that have either an `elasticsearch/ilm/` directory or a
+      `lifecycle.yml` file, and documents one policy per data stream as a
       flattened key/value table.
-    - If there are no ILM policies to document (or discovery finds none), the placeholder
+    - If there are no policies to document (or discovery finds none), the placeholder
       renders as an empty string: no heading and no empty section are written.
     - Example of the rendered output when policies exist:
       ```
--- a/docs/howto/add_package_readme.md
+++ b/docs/howto/add_package_readme.md
@@ -143,7 +143,7 @@ List of placeholders that can be used in the Markdown templates:
       flattened key/value table.
     - If there are no policies to document (or discovery finds none), the placeholder
       renders as an empty string: no heading and no empty section are written.
-    - Example of the rendered output when policies exist:
+    - Example of the rendered output for an ILM policy (`elasticsearch/ilm/<policy>.json`):
       ```
       ### Data streams using ILM policies
 
@@ -153,6 +153,16 @@ List of placeholders that can be used in the Markdown templates:
       | policy.phases.delete.min_age | 30d |
       | policy.phases.hot.actions.rollover.max_age | 30d |
       ...
+      ```
+    - Example of the rendered output for a DLM `lifecycle.yml` (for example
+      `data_retention: "30d"`):
+      ```
+      ### Data streams using ILM policies
+
+      #### metrics Policy
+      | Key | Value |
+      |---|---|
+      | data_retention | 30d |
       ```
 - `transform`: this placeholder renders transforms defined under `elasticsearch/transform/`.
     - Example of usage:
--- a/docs/howto/add_package_readme.md
+++ b/docs/howto/add_package_readme.md
@@ -124,6 +124,39 @@ List of placeholders that can be used in the Markdown templates:
       ```
     - Requirements:
         - It is needed to define a file with the links definitions. More information [in this section](#requirements)
+- `ilm [data_stream...]`: this placeholder renders ILM policies defined under
+  `data_stream/<name>/elasticsearch/ilm/`.
+    - Example of usage (all data streams with ILM policies):
+      ```
+      {{ ilm }}
+      ```
+    - Example of usage (specific data streams):
+      ```
+      {{ ilm "metrics" }}
+      ```
+    - When no data stream arguments are given, `elastic-package` discovers data streams
+      that have an `elasticsearch/ilm/` directory and documents each policy as a
+      flattened key/value table.
+    - If there are no ILM policies to document (or discovery finds none), the placeholder
+      renders as an empty string: no heading and no empty section are written.
+    - Example of the rendered output when policies exist:
+      ```
+      ### Data streams using ILM policies
+
+      #### metrics Policy
+      | Key | Value |
+      |---|---|
+      | policy.phases.delete.min_age | 30d |
+      | policy.phases.hot.actions.rollover.max_age | 30d |
+      ...
+      ```
+- `transform`: this placeholder renders transforms defined under `elasticsearch/transform/`.
+    - Example of usage:
+      ```
+      {{ transform }}
+      ```
+    - If there are no transforms defined, the placeholder renders as an empty string
+      (no heading is written), same as `ilm`.
 
 ## Requirements
 
--- a/docs/howto/add_package_readme.md
+++ b/docs/howto/add_package_readme.md
@@ -125,22 +125,27 @@ List of placeholders that can be used in the Markdown templates:
     - Requirements:
         - It is needed to define a file with the links definitions. More information [in this section](#requirements)
 - `ilm [data_stream...]`: this placeholder renders ILM policies or lifecycle
-  configuration for each data stream. Two file locations are supported:
-    - `data_stream/<name>/elasticsearch/ilm/<policy>.json` — ILM policy file.
-    - `data_stream/<name>/lifecycle.yml` — DLM (Data Lifecycle Management)
-      configuration (takes precedence when both exist for the same data stream).
-    - Example of usage (all data streams):
+  configuration. Behavior depends on the package type:
+    - **Integration packages** — reads per-data-stream policy files:
+        - `data_stream/<name>/elasticsearch/ilm/<policy>.json` — ILM policy file.
+        - `data_stream/<name>/lifecycle.yml` — DLM (Data Lifecycle Management)
+          configuration (takes precedence when both exist for the same data stream).
+    - **Input packages** — reads a single package-root lifecycle file:
+        - `lifecycle.yml` at the package root (bare `{{ ilm }}` only; data stream
+          name arguments are not supported for input packages and skip this path).
+    - Example of usage (all data streams / input package root):
       ```
       {{ ilm }}
       ```
-    - Example of usage (specific data streams):
+    - Example of usage (specific data streams, integration packages only):
       ```
       {{ ilm "metrics" }}
       ```
-    - When no data stream arguments are given, `elastic-package` discovers all
-      data streams that have either an `elasticsearch/ilm/` directory or a
-      `lifecycle.yml` file, and documents one policy per data stream as a
-      flattened key/value table.
+    - When no data stream arguments are given, `elastic-package` reads the package
+      manifest type. For integration packages it discovers all data streams that
+      have either an `elasticsearch/ilm/` directory or a `lifecycle.yml` file, and
+      documents one policy per data stream as a flattened key/value table. For
+      input packages it renders the single root `lifecycle.yml`.
     - If there are no policies to document (or discovery finds none), the placeholder
       renders as an empty string: no heading and no empty section are written.
     - Example of the rendered output for an ILM policy (`elasticsearch/ilm/<policy>.json`):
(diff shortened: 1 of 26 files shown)


internal/docs/ilm.go
teresaromero: @kcreddy — quick follow-up from your review comment on #3221 where you asked to document `lifecycle.yml` / DLM alongside ILM.

The named form `{{ ilm "ds" }}` does resolve `lifecycle.yml` (added in #3221), but bare `{{ ilm }}` discovery still only globs `elasticsearch/ilm/` — so a data stream with only a `lifecycle.yml` and no ILM dir won't appear in the auto-discovered section at all.

Was that intentional (keep `{{ ilm }}` scoped to ILM policies only), or should discovery be extended to also pick up `lifecycle.yml`? cc @jrmolin

kcreddy: Thanks for catching this. From my understanding, it looks like an oversight and not intentional. Bare `{{ ilm }}` should also discover data streams that only have `lifecycle.yml` — auto-discovery shouldn't be limited to `elasticsearch/ilm/`.

teresaromero: is it ok if i include the fix as scope of this pr? or do you prefer i file an issue?

kcreddy: @teresaromero, I'm okay adding it to this PR unless @jrmolin thinks otherwise.

internal/docs/readme_test.go
```diff
@@ -470,6 +470,65 @@ Introduction to the package
 	},
 }
 
+func TestRenderILMPathsLifecycleYMLDiscovered(t *testing.T) {
+	packageRoot := t.TempDir()
+
+	createILMFileYML(t, packageRoot, "mystream")
```
mrodm, line 478: Looking at the code of this funciton, this does not generate the expected content for the `lifecycle.yml` file.

I see that this function was added here : https://github.com/elastic/elastic-package/pull/3221


https://github.com/elastic/elastic-package/blob/ee68f5c920e018a558bee1727e5b92aef35e8880/internal/docs/readme_test.go#L686-L711


But this files should follow this spec:
https://github.com/elastic/package-spec/blob/4a1fe606db3b60978df8174fae478b44afc651ec/spec/integration/data_stream/lifecycle.spec.yml


Example: 
https://github.com/elastic/integrations/blob/7860b3c9fde8757e4e0c2430543ba94189423ede/packages/ti_eset/data_stream/url/lifecycle.yml


I'm not sure this is doing what it is expected for the `lifecycle.yml` files.

Could you add a new test package under `test/packages/other` that one data stream contains the ILM as json (`data_stream/<name>/elasticsearch/ilm/default_policy.json`) and other data streaam contins the `lifecycle.yml` (`data_stream/<other>/lifecycle.yml`) ?

internal/docs/ilm.go
```diff
@@ -135,22 +135,30 @@ func renderILMPaths(packageRoot string, args []string) (string, error) {
...
+	}
 
-	result := make([]string, 0, len(ilmPaths))
+	lifecyclePaths, err := filepath.Glob(filepath.Join(packageRoot, "data_stream", "*", "lifecycle.yml"))
+	if err != nil {
+		return nil, fmt.Errorf("finding lifecycle paths failed: %w", err)
+	}
+	for _, lifecyclePath := range lifecyclePaths {
+		seen[filepath.Base(filepath.Dir(lifecyclePath))] = struct{}{}
+	}
```
mrodm, line 193: These files do not follow the same structure as the JSON files, I'm not sure this would work given this https://github.com/elastic/elastic-package/pull/3840/changes#r3783036713

internal/docs/readme_test.go
teresaromero: I've added some fixtures and updated the test cases

test/packages/other/readme_ilm/data_stream/ilm_ds/manifest.yml
mrodm, line 1: From CI, `fields` folder is required in the integration packages

```
Error: building package failed: invalid content found in built zip package: found 2 validation errors:
   1. expecting to find [fields] folder in folder [/Users/admin/builds/bk-agent-prod-orka-1786713270151226671/elastic/elastic-package/build/packages/readme_ilm-0.0.1.zip/data_stream/ilm_ds]
   2. expecting to find [fields] folder in folder [/Users/admin/builds/bk-agent-prod-orka-1786713270151226671/elastic/elastic-package/build/packages/readme_ilm-0.0.1.zip/data_stream/lifecycle_ds]
```

internal/docs/ilm.go
```diff
@@ -110,6 +107,9 @@ func renderILMPaths(packageRoot string, args []string) (string, error) {
 			return "", fmt.Errorf("finding ILM paths failed: %w", err)
 		}
 	}
+	if len(dataStreamNames) == 0 {
+		return "", nil
+	}
```
mrodm, line 115: According to the spec, input packages can define also `lifecycle.yml` 

https://github.com/elastic/package-spec/blob/4a1fe606db3b60978df8174fae478b44afc651ec/spec/input/spec.yml#L60-L65

IIUC the logic here, that would mean that this if would not allow to find and render the required sections for the input packages.

mrodm (commented):
It would be needed to check how this would behave with input packages too. According to the spec, those packages can define also a `lifecycle.yml` file at the root of the package.

And it would be interesting to test this branch with the packages from the integrations repository adding a comment like `test integrations`. It will help us to know how many packages would need to be updated with the next elastic-package release.

internal/docs/ilm.go
teresaromero: i've refactored the approach to add inputs too

teresaromero:
@kcreddy could you review again these changes; there was a bit more than what i expected for this "bugfix" :D

docs/howto/add_package_readme.md
```diff
@@ -124,6 +124,43 @@ List of placeholders that can be used in the Markdown templates:
...
+      ```
+      ### Data streams using ILM policies
+
+      #### metrics Policy
+      | Key | Value |
+      |---|---|
+      | policy.phases.delete.min_age | 30d |
+      | policy.phases.hot.actions.rollover.max_age | 30d |
+      ...
+      ```
```
kcreddy: A small example for `lifecycle.yml` will be nice here. Nit only.

kcreddy (commented):
Thanks! Nits only

kcreddy (approved):
Thanks! Nits only

docs/howto/add_package_readme.md
teresaromero: Added a rendered-output example for `lifecycle.yml` (DLM) next to the ILM one — solved at 680577ba.

internal/docs/ilm.go
```diff
@@ -95,62 +98,101 @@ func getILMPolicyFilePath(packageRoot, dataStreamName string) (string, error) {
...
+// renderILMPaths is the entry point for the {{ ilm }} template function.
+// args are data stream names and only apply to integration packages; passing
+// them skips the input-package path entirely. Bare invocation checks for a
+// root lifecycle.yml (input packages) first, then discovers data streams.
+func renderILMPaths(packageRoot string, args []string) (string, error) {
+	if len(args) > 0 {
+		return renderDataStreamILM(packageRoot, args)
+	}
+
+	result, err := renderInputPackageILM(packageRoot)
```
jillguyonnet: If I understand: `renderInputPackageILM` tries to read `lifecycle.yml` at the package root and returns `("", nil)` if the file doesn't exist, is that correct? If so, couple of questions:
1. Any reason it doesn't check `manifest.Type` to tell if the package has type input?
2. `renderILMPaths` returns early (below) if the result is non-empty: should we guard against the possibility of an integration package with a root `lifecycle.yml` (by accident or future spec evolution)? While improbably, in that case it looks like all per-data-stream ILM/lifecycle policies would be silently omitted from the rendered README with no error?

internal/docs/ilm.go
```diff
@@ -95,62 +98,101 @@ func getILMPolicyFilePath(packageRoot, dataStreamName string) (string, error) {
...
+			return "", nil
+		}
+		return "", fmt.Errorf("getting ILM policy map for root lifecycle.yml failed: %w", err)
+	}
+	sectionName := "lifecycle"
+	if manifest, mErr := packages.ReadPackageManifestFromPackageRoot(packageRoot); mErr == nil && manifest.Name != "" {
+		sectionName = manifest.Name
+	}
+	var out strings.Builder
+	out.WriteString("\n### Data streams using ILM policies\n")
```
jillguyonnet: Is this accurate for input packages? Should it be something like `### Lifecycle policy`?

jillguyonnet: It would look OK (I think) with the happy path:
```
### Lifecycle policy

#### <package-name> Policy
```

but if the fallback from line 140 kicks in it's not ideal 😅 :
```
### Lifecycle policy

#### lifecycle Policy
```

WDYT? We could propagate the error in line 140 rather than fall back to a placeholder string?

internal/docs/readme_test.go
```diff
@@ -470,10 +471,170 @@ Introduction to the package
...
+	root, err := os.OpenRoot(packageRoot)
+	require.NoError(t, err)
+	t.Cleanup(func() { root.Close() })
+
+	rendered, err := renderReadme(root, "README.md", packageRoot, templatePath, linksMap, urls)
+	require.NoError(t, err)
+	assert.Equal(t, expected, string(rendered))
+}
+
+func TestRenderILMPathsRootLifecycleDiscovered(t *testing.T) {
```
jillguyonnet, line 525: WDYT about adding a test that creates an integration-type manifest alongside a root `lifecycle.yml` and asserts the data-stream discovery path is still taken? Maybe it's not realistic but would guard against that unexpected config?

internal/docs/ilm.go
teresaromero: The file is not expected as per package-spec definition. However, i added a guard just in case during development this could happen. `build` or `check` commands will rise this as an error

internal/docs/ilm.go
teresaromero: Fixed: the heading is now `### Lifecycle policy` for input packages. The internal silent manifest-read and "lifecycle" fallback were removed — the package name is now passed in from `renderILMPaths` which reads the manifest (and propagates errors) before branching.

internal/docs/readme_test.go
teresaromero: Added a test, but this case is not real as having a lifecyle on root will be an error on validation

teresaromero:
test integrations

teresaromero:
> Created or updated PR in integrations repository to test this version. Check [elastic/integrations#20877](https://github.com/elastic/integrations/pull/20877)

**CI failures are expected and caused by this PR**

The 7 failing jobs in build #48434 (https://buildkite.com/elastic/integrations/builds/48434) are a direct consequence of the fix introduced in this PR (fix: prevent empty ILM heading when no policies are defined).

**Root cause**

Previously, renderILMPaths always emitted a ### Data streams using ILM policies heading even when no data streams had ILM policies, resulting in an empty section in the rendered README. This PR corrects that: renderDataStreamILM now returns "" early when the list of data streams is empty.

The stale README.md files in the affected packages were generated by the old elastic-package (with the orphan heading). The new elastic-package renders them without it, causing the lint check (README.md is outdated) to fail.

**Evidence**

Each failing job shows the same diff — two lines being removed:

```
@@ -492,4 +492,2 @@

-### Data streams using ILM policies
-
```

Example from Check integrations anthropic (https://buildkite.com/elastic/integrations/builds/48434#01a03421-6d8b-4209-a0a0-965b3e7b00c2) and Check integrations cursor (https://buildkite.com/elastic/integrations/builds/48434#01a03421-6db4-449e-9abc-a3b6986b967e).

**Affected packages**

These packages have {{ ilm }} in their README template but no actual ILM policies, so they previously rendered the empty heading:

- anthropic
- atlassian_cloud- cursor
- cyera
- elastic_package_registry
- hackerone                 
- kolide

**What needs to happen**           
                                                                                
A companion PR to elastic/integrations is needed to rebuild the README for each of these packages with the new elastic-package. The change per package is minimao empty lines (### Data streams using ILM policies+ blank line) from each README.md.

teresaromero:
@kcreddy @jillguyonnet please, re-review after addressing comments 👍🏻 thanks!!

kcreddy (approved):
Thanks for the fixes.

jillguyonnet (approved):
LGTM, thanks for the changes!

PR #3842, opened 2026-08-14
fix: exception field in validateMappingsNotInPreview silently suppresses sibling errors

mrodm:
## Summary

Spotted while reviewing #3839 (now merged): a pre-existing bug in
`validateMappingsNotInPreview` caused validation errors to be silently discarded
whenever an exception field appeared in the same flattened field map as an
undefined field.

## Root cause

Inside the `flattenFields` loop in `validateMappingsNotInPreview`
(`internal/fields/mappings.go`), the exception-field branch did `return nil`
instead of `continue`. Because Go map iteration order is non-deterministic, this
had two failure modes:

- **Exception field visited first** — the function returns immediately; the
  undefined sibling is never checked and no error is reported.
- **Exception field visited last** — errors already accumulated for the undefined
  sibling are discarded by the `return nil`.

In both orderings the function returns `nil`, silently hiding real validation
errors.

#3839 made this bug more reachable by adding a second call site for
`validateMappingsNotInPreview`: dynamic children under a `nested` ECS parent.
ECS nested fields (e.g. `email.attachments`) commonly have many children, some
of which may appear in `exceptionFields`, so the probability of hitting this
ordering-dependent path increased significantly.

### `internal/fields/mappings.go`
- `return nil` → `continue` in the exception-field branch of
  `validateMappingsNotInPreview` so each field in the flattened map is evaluated
  independently.

### `internal/fields/mappings_test.go`
- Added regression test: a `nested` parent with one exception-field child and one
  undefined sibling. Expects the undefined sibling to always be reported,
  regardless of Go's map iteration order. Confirmed to fail consistently (20/20
  runs) before the fix and pass consistently after.

## Proposed commit message

```
fix: exception field in validateMappingsNotInPreview silently suppresses sibling errors

`return nil` inside the exception-field branch of the flattenFields loop caused
the function to exit early regardless of map iteration order: either the
exception field was visited first (undefined siblings never checked) or last
(accumulated errors discarded). Changed to `continue` so each field is evaluated
independently.

Adds a regression test that places an exception field alongside an undefined
sibling under a dynamic nested parent (introduced by #3839), confirming the
undefined field is always reported.
```

## Testing

- Regression test runs 20/20 RED on the unfixed code, 20/20 GREEN after the fix
- Full unit suite passes: `go test ./internal/fields/...`

Spotted during review of #3839.

---

### DIFF
--- a/.github/workflows/run-updatecli.yml
+++ b/.github/workflows/run-updatecli.yml
@@ -35,7 +35,7 @@ jobs:
           echo "UPDATECLI_ACTION=apply" >> $GITHUB_ENV
 
       - name: Install Updatecli in the runner
-        uses: updatecli/updatecli-action@5dfc616e57b07d2302c3f37e5b72a4ead82dbbf8 #v3.4.0
+        uses: updatecli/updatecli-action@d422ba543720d01a96d4c3e3ab8544cc188cd711 #v3.5.0
 
       - name: Run updatecli
         # --experimental needed for commitusingapi option.
--- a/Makefile
+++ b/Makefile
@@ -85,10 +85,10 @@ test-stack-command-86:
 	./scripts/test-stack-command.sh 8.6.2
 
 test-stack-command-8x:
-	./scripts/test-stack-command.sh 8.19.20-291ed6ef-SNAPSHOT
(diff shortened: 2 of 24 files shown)


mrodm:
test integrations

teresaromero (approved):
🚀

PR #3860, opened 2026-08-20
modify pipeline-tag: fix to tag on_failure processors as well

chrisberkhout:
Proposed commit message:

```
modify pipeline-tag: fix to tag on_failure processors as well (#)

processPipeline only walked pipeline.Processors, so processors in the
pipeline's top-level on_failure block never received generated tags.
```

### DIFF
--- a/internal/modify/pipelinetag/pipelinetag.go
+++ b/internal/modify/pipelinetag/pipelinetag.go
@@ -61,13 +61,15 @@ func run(pkg *fleetpkg.Package) error {
 func processPipeline(pipeline *fleetpkg.Pipeline) error {
 	seen := map[string]*processorNode{}
 
-	for _, proc := range pipeline.Processors {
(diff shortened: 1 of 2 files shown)


teresaromero (commented):
can we add this case to the test fixture under `internal/modify/pipelinetag/testdata/default.yml` ?

chrisberkhout:
> can we add this case to the test fixture under `internal/modify/pipelinetag/testdata/default.yml` ?

There's already an untagged pipline-level `on_failure` processor in that fixture [here](https://github.com/elastic/elastic-package/blob/6371beeef0558dca8fe3ac2a1267677aafa79688/internal/modify/pipelinetag/testdata/default.yml#L16-L19). There's a new assertion in `pipelinetag_test.go` to make sure it gets tagged.

teresaromero (approved):

PR #3910, opened 2026-09-02
feat: add env var override for skipUploadPackageValidation Kibana flag

teresaromero:
## Summary

Adds `ELASTIC_PACKAGE_KIBANA_SKIP_UPLOAD_PACKAGE_VALIDATION=true` to force-enable `xpack.fleet.internal.skipUploadPackageValidation: true` in the generated Kibana config for `9.6.0-SNAPSHOT` onwards and backported branches detail below.

The flag is auto-injected for `>= 9.6.0-SNAPSHOT` (elastic/kibana#286094, merged on main). Once each lands, CI can activate the flag by setting the env var; the version ranges in the template ensure it is only injected into ongoing CI snapshot builds that actually contain the backport.

| Branch | Range | Kibana PR |
|--------|-------|-----------|
| 8.19.x | `[8.19.22-SNAPSHOT, 9.0.0-SNAPSHOT)` | elastic/kibana#287670 |
| 9.4.x  | `[9.4.7-SNAPSHOT, 9.5.0-SNAPSHOT)`   | elastic/kibana#287671 |
| 9.5.x  | `[9.5.4-SNAPSHOT, 9.6.0-SNAPSHOT)`   | elastic/kibana#287672 |

Lower bounds reflect the current CI snapshot for each branch (already-released builds are excluded). Upper bounds use SNAPSHOT forms so that prerelease versions of the next minor (e.g. `9.0.0-SNAPSHOT`, `9.5.0-SNAPSHOT`) are not inadvertently included. Setting the env var for a version outside these ranges has no effect, so it is safe to set in a shared CI environment.

## Changes

- `internal/stack/_static/kibana.yml.tmpl` — adds version-scoped conditions for the env var override: the flag is emitted when `$skipUpload` is true and the stack version falls within one of the three backport ranges above. Upper bounds are in SNAPSHOT form to exclude prerelease versions of adjacent minors.
- `internal/stack/resources.go` — exports `KibanaSkipUploadPackageValidationEnvVar` and passes its value as the `skip_upload_package_validation` template fact.
- `internal/stack/kibana_upload_flag_test.go` — covers three scenarios: version gate only (`>= 9.6.0-SNAPSHOT`), env var with a matching backport snapshot (enabled), and env var with a pre-backport or out-of-range version (disabled). Includes SNAPSHOT boundary cases for all range edges.
- `docs/howto/install_package.md` — documents the env var, the minimum snapshot versions per branch, and the manual `kibana.dev.yml` workaround for versions not yet covered.

## Related

- #3879 — initial version gate for 9.6.0-SNAPSHOT
- elastic/kibana#286094 — original PR
- elastic/kibana#287670 — backport to 8.19
- elastic/kibana#287671 — backport to 9.4
- elastic/kibana#287672 — backport to 9.5

### DIFF
--- a/internal/stack/_static/kibana.yml.tmpl
+++ b/internal/stack/_static/kibana.yml.tmpl
@@ -193,7 +193,14 @@ xpack.fleet.outputs:
 xpack.fleet.internal.registry.kibanaVersionCheckEnabled: false
 {{- end }}
 
-{{/* Present on 9.6.0-SNAPSHOT+ (elastic/kibana#286094). For backport branches (8.19, 9.4, 9.5) set ELASTIC_PACKAGE_KIBANA_SKIP_UPLOAD_PACKAGE_VALIDATION=true. */}}
-{{- if or (not (semverLessThan $version "9.6.0-SNAPSHOT")) (eq (fact "skip_upload_package_validation") "true") }}
+{{/* elastic/kibana#286094: merged in 9.6.0-SNAPSHOT; backported to 8.19.10-SNAPSHOT (#287670), 9.4.7-SNAPSHOT (#287671), 9.5.3-SNAPSHOT (#287672).
+     Backport branches also require ELASTIC_PACKAGE_KIBANA_SKIP_UPLOAD_PACKAGE_VALIDATION=true so CI controls when activation occurs. */}}
+{{- $skipUpload := eq (fact "skip_upload_package_validation") "true" }}
+{{- if or
+    (not (semverLessThan $version "9.6.0-SNAPSHOT"))
+    (and $skipUpload (not (semverLessThan $version "8.19.10-SNAPSHOT")) (semverLessThan $version "9.0.0"))
+    (and $skipUpload (not (semverLessThan $version "9.4.7-SNAPSHOT")) (semverLessThan $version "9.5.0"))
+    (and $skipUpload (not (semverLessThan $version "9.5.3-SNAPSHOT")) (semverLessThan $version "9.6.0-SNAPSHOT"))
+}}
 xpack.fleet.internal.skipUploadPackageValidation: true
 {{- end }}
--- a/internal/stack/_static/kibana.yml.tmpl
+++ b/internal/stack/_static/kibana.yml.tmpl
@@ -193,7 +193,7 @@ xpack.fleet.outputs:
 xpack.fleet.internal.registry.kibanaVersionCheckEnabled: false
 {{- end }}
 
-{{/* TODO: extend to 8.19, 9.4, and 9.5 once elastic/kibana#286094 is backported and released in those branches */}}
-{{- if not (semverLessThan $version "9.6.0-SNAPSHOT") }}
+{{/* Present on 9.6.0-SNAPSHOT+ (elastic/kibana#286094). For backport branches (8.19, 9.4, 9.5) set ELASTIC_PACKAGE_KIBANA_SKIP_UPLOAD_PACKAGE_VALIDATION=true. */}}
+{{- if or (not (semverLessThan $version "9.6.0-SNAPSHOT")) (eq (fact "skip_upload_package_validation") "true") }}
 xpack.fleet.internal.skipUploadPackageValidation: true
 {{- end }}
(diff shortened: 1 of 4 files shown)


internal/stack/resources.go
```diff
@@ -79,6 +80,13 @@ const (
 	configFleetAutoInstallContentPackagesEnabled = "stack.fleet_auto_install_content_packages_enabled"
 )
 
+var (
+	// KibanaSkipUploadPackageValidationEnvVar enables xpack.fleet.internal.skipUploadPackageValidation
+	// for Kibana versions that have the backport of elastic/kibana#286094 but are below 9.6.0-SNAPSHOT.
+	// Set to "true" in CI when targeting 8.19, 9.4, or 9.5 stacks with the backport applied.
+	KibanaSkipUploadPackageValidationEnvVar = environment.WithElasticPackagePrefix("KIBANA_SKIP_UPLOAD_PACKAGE_VALIDATION")
```
mrodm: It cannot be set just for specific stack versions in the integrations CI.

I would use this flag as a way to obtain if the stack version used in the test `stackVersion` is one that support that flag.

I guess it should be needed to know which 8.19.x-SNAPSHOT , 9.4.x-SNAPSHOT and 9.5.x-SNAPSHOT

internal/stack/_static/kibana.yml.tmpl
```diff
@@ -193,7 +193,7 @@ xpack.fleet.outputs:
 xpack.fleet.internal.registry.kibanaVersionCheckEnabled: false
 {{- end }}
 
-{{/* TODO: extend to 8.19, 9.4, and 9.5 once elastic/kibana#286094 is backported and released in those branches */}}
-{{- if not (semverLessThan $version "9.6.0-SNAPSHOT") }}
+{{/* Present on 9.6.0-SNAPSHOT+ (elastic/kibana#286094). For backport branches (8.19, 9.4, 9.5) set ELASTIC_PACKAGE_KIBANA_SKIP_UPLOAD_PACKAGE_VALIDATION=true. */}}
+{{- if or (not (semverLessThan $version "9.6.0-SNAPSHOT")) (eq (fact "skip_upload_package_validation") "true") }}
```
mrodm: If it is set like this, when the CI in the integrations adds the environment variable it will add for all the packages this Kibana setting, and there will be stack versions that do not support that.

internal/stack/kibana_upload_flag_test.go
```diff
@@ -15,75 +15,122 @@ import (
...
-				t.Fatalf("read kibana.yml: %v", err)
-			}
+	cases := []struct {
+		version string
+		want    bool
+	}{
+		// Entirely outside the backport branches — env var has no effect
+		{"8.18.0", false},
+		{"9.0.0", false},
+		{"9.3.0", false},
```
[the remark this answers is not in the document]
teresaromero: i dont think a new minor for 8.x will be out

internal/stack/_static/kibana.yml.tmpl
```diff
@@ -193,7 +193,14 @@ xpack.fleet.outputs:
...
 {{- end }}
 
-{{/* TODO: extend to 8.19, 9.4, and 9.5 once elastic/kibana#286094 is backported and released in those branches */}}
-{{- if not (semverLessThan $version "9.6.0-SNAPSHOT") }}
+{{/* elastic/kibana#286094: merged in 9.6.0-SNAPSHOT; backported to 8.19.10-SNAPSHOT (#287670), 9.4.7-SNAPSHOT (#287671), 9.5.3-SNAPSHOT (#287672).
+     Backport branches also require ELASTIC_PACKAGE_KIBANA_SKIP_UPLOAD_PACKAGE_VALIDATION=true so CI controls when activation occurs. */}}
+{{- $skipUpload := eq (fact "skip_upload_package_validation") "true" }}
+{{- if or
+    (not (semverLessThan $version "9.6.0-SNAPSHOT"))
+    (and $skipUpload (not (semverLessThan $version "8.19.10-SNAPSHOT")) (semverLessThan $version "9.0.0-SNAPSHOT"))
```
mrodm: Should this condition be using `8.19.22-SNAPSHOT` ?
8.19.10 was already published 


```suggestion
    (and $skipUpload (not (semverLessThan $version "8.19.22-SNAPSHOT")) (semverLessThan $version "9.0.0-SNAPSHOT"))
```

mrodm (approved):
LGTM, thanks!

PR #3918, opened 2026-09-03
Retry transient connection errors in HTTP retry policy

### LINKED ISSUE #3887: Kibana client never retries connection-level errors: `checkRetry` treats every `*url.Error` as unrecoverable
## Summary

The Kibana client has had HTTP retries since #1563 (Nov 2023): `internal/kibana/client.go` wraps the client with `retry.WrapHTTPClient` (`retryMax: 10`, backoff 1–5s, hashicorp/go-retryablehttp). Despite this, system tests still fail on single transient connection errors:

```
failed to create kibana policies: could not create test policy: could not create policy:
could not send request to Kibana API: Post "https://127.0.0.1:5601/api/fleet/agent_policies": EOF
```

The reason is a bug in the custom retry policy, `checkRetry` in `internal/retry/http.go`:

```go
var urlError *url.Error
if errors.As(err, &urlError) {
    // URL is invalid, not recoverable.
    return false, nil
}
```

Per the `net/http` documentation, **`http.Client.Do` wraps every returned error in `*url.Error`** — including transient transport errors such as `EOF`, `connection reset by peer`, and `broken pipe`. So this branch, intended to catch invalid-URL parse errors, matches *all* transport-level failures and short-circuits to "not recoverable". The `// Consider other errors as recoverable and retry.` fallthrough at the end of the error block is unreachable for anything `Client.Do` returns. (The `tls`/`x509` typed checks below it are likewise dead code — the `urlError` branch always matches first.)

The response-status path (429, 5xx) is unaffected and works.

Note the contrast with `retryablehttp.DefaultRetryPolicy`, which this function reimplements: the default policy inspects the `*url.Error`'s **inner** error for specific unrecoverable causes (redirect loops, unsupported scheme, certificate errors) and retries everything else. The custom version inverted that logic.

`internal/retry/http_test.go` (`TestUnrecoverableErrors`) doesn't catch this because it only exercises genuinely unrecoverable cases — invalid URL, redirect loop, unknown CA, invalid certificate — and never a transient connection error.

## Real-world impact

Auto-filed flaky-test issues across elastic/integrations. For the `network_traffic` package alone, 4 in July–August 2026:

- elastic/integrations#20793 — `POST /api/fleet/agent_policies: EOF` (http-basicauth)
- elastic/integrations#20223 — `POST /api/fleet/agent_policies: EOF` (http-post, LogsDB)
- elastic/integrations#20357 — `POST /api/fleet/package_policies: EOF` (redis-session)
- elastic/integrations#20317 — `POST /api/fleet/package_policies: EOF` (memcached-bin-tcp-single-load-store)

Log analysis of the #20793 run showed the classic stale keep-alive race (client reuses a pooled connection Kibana already closed → zero response bytes → `EOF`): no component crashed, ES logged zero errors, ~180 identical Fleet calls succeeded in the same run. Go's transport auto-replays such failures for GET/HEAD but never for POST (`net/http` `Request.isReplayable`), which is why these surface exclusively as POST EOFs — and the `checkRetry` bug means elastic-package's own retry layer, which should catch exactly this, never does.

## Proposed fix

1. **In `checkRetry`, stop treating `*url.Error` as terminal.** Match `retryablehttp.DefaultRetryPolicy` semantics: unwrap and check the inner error for genuinely unrecoverable causes (the existing `errTooManyRedirects`, `tls`/`x509` typed checks — which become reachable again — plus unsupported protocol scheme), and return `true` (retry) for the rest, which covers `EOF`/reset/refused. A single retry is mechanistically sufficient for the stale keep-alive race (the retry necessarily runs on a fresh connection); the existing `retryMax: 10` with 1–5s backoff is more than enough.
2. **Add a regression test**: a test server that accepts a connection and closes it without responding (or an `httptest` server closed between requests) must be retried.
3. *(Optional hardening)* Set a `Timeout` on the underlying `http.Client{}` (`client.go:226` — currently none), so a stalled Kibana handler can't hang a request until the server's 120s socket timeout.
4. *(Optional, strict idempotency)* The plumbing already exists: `kibana.Policy.ID` is marshaled into the create body (`json:"id,omitempty"`) and `CreatePolicy` already handles `ErrConflict` (409). Having the test runner set a generated ID would make retried creates fully idempotent. In practice the duplicate-policy risk in ephemeral test stacks is negligible.

## Impact of the fix

Removes the most common flaky-test signature (`Post .../agent_policies|package_policies: EOF`) for **all** integration packages. Remaining setup-failure classes (slow-response context deadlines, service container failures) are covered by a follow-up proposal to re-attempt setup-phase failures in the system test runner (see https://github.com/elastic/elastic-package/issues/3888)

kcreddy:
```
Retry transient connection errors in HTTP retry policy

The checkRetry function treated every *url.Error as an unrecoverable
invalid-URL error, which disabled retries for all connection-level
failures. Since http.Client.Do wraps every transport-level error in
*url.Error — including transient ones like EOF (from a keep-alive
connection the server already closed) or connection reset — this
effectively disabled retries for any connection error and made the
tls/x509 certificate checks below unreachable dead code.

Fix by inspecting the underlying error instead of the wrapper.
Keep genuinely unrecoverable causes non-retryable (unsupported
protocol scheme, certificate errors, redirect loops), and retry
everything else (EOF, connection reset, refused, etc.).

Fixes #3887

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
```

Closes #3887

### DIFF
--- a/internal/retry/http.go
+++ b/internal/retry/http.go
@@ -12,7 +12,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
-	"net/url"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
@@ -87,9 +87,13 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
 			return false, nil
 		}
 
-		var urlError *url.Error
-		if errors.As(err, &urlError) {
-			// URL is invalid, not recoverable.
+		// http.Client.Do wraps every error it returns in *url.Error, including
+		// transient transport failures such as EOF or connection reset that are
+		// worth retrying, so the wrapper itself says nothing about recoverability.
+		// Inspect the underlying error for the genuinely unrecoverable causes.
+
+		if strings.Contains(err.Error(), "unsupported protocol scheme") {
+			// Malformed request URL, not recoverable.
 			return false, nil
 		}
 
@@ -111,7 +115,9 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
 			return false, nil
 		}
 
-		// Consider other errors as recoverable and retry.
+		// Consider other errors, including connection-level failures such as
+		// EOF (e.g. a reused keep-alive connection closed by the server) or
+		// connection reset, as recoverable and retry.
 		return true, nil
 	}
 
--- a/internal/retry/http.go
+++ b/internal/retry/http.go
@@ -11,7 +11,9 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -92,9 +94,26 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
 		// worth retrying, so the wrapper itself says nothing about recoverability.
 		// Inspect the underlying error for the genuinely unrecoverable causes.
 
-		if strings.Contains(err.Error(), "unsupported protocol scheme") {
-			// Malformed request URL, not recoverable.
+		if errors.Is(err, http.ErrSchemeMismatch) {
+			// Server returned HTTP to an HTTPS client — a configuration error.
 			return false, nil
+		}
+
+		var urlErr *url.Error
+		if errors.As(err, &urlErr) {
+			// Go's stdlib does not export a typed error for unsupported protocol
+			// scheme, so match on the inner error string only (not the full
+			// url.Error string which embeds the URL and could produce false matches).
+			if strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
+				return false, nil
+			}
+
+			// Permanent DNS failure (NXDOMAIN). IsNotFound distinguishes it from
+			// transient resolver hiccups, which should still be retried.
+			var dnsErr *net.DNSError
+			if errors.As(urlErr.Err, &dnsErr) && dnsErr.IsNotFound {
+				return false, nil
+			}
 		}
 
 		var certVerificationError *tls.CertificateVerificationError
--- a/internal/retry/http.go
+++ b/internal/retry/http.go
@@ -104,7 +104,7 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
 			// Go's stdlib does not export a typed error for unsupported protocol
 			// scheme, so match on the inner error string only (not the full
 			// url.Error string which embeds the URL and could produce false matches).
-			if strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
+			if urlErr.Err != nil && strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
 				return false, nil
 			}
 
--- a/internal/retry/http_transport_errors_test.go
+++ b/internal/retry/http_transport_errors_test.go
@@ -9,6 +9,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -20,6 +21,23 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+// transientDNSError simulates a DNS failure that is NOT a permanent NXDOMAIN,
+// e.g. a transient resolver timeout.
+func transientDNSError(host string) error {
+	return &url.Error{Op: "Get", URL: "https://" + host + "/", Err: &net.DNSError{
+		Name:       host,
+		IsNotFound: false, // transient, not NXDOMAIN
+	}}
+}
+
+// permanentDNSError simulates a permanent NXDOMAIN failure.
+func permanentDNSError(host string) error {
+	return &url.Error{Op: "Get", URL: "https://" + host + "/", Err: &net.DNSError{
+		Name:       host,
+		IsNotFound: true,
+	}}
+}
 
 // TestCheckRetryTransientConnectionErrors feeds checkRetry the error shapes
 // http.Client.Do produces for transport-level failures, which are always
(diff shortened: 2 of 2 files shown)


internal/retry/http.go
```diff
@@ -87,9 +87,13 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
...
-			// URL is invalid, not recoverable.
+		// http.Client.Do wraps every error it returns in *url.Error, including
+		// transient transport failures such as EOF or connection reset that are
+		// worth retrying, so the wrapper itself says nothing about recoverability.
+		// Inspect the underlying error for the genuinely unrecoverable causes.
+
+		if strings.Contains(err.Error(), "unsupported protocol scheme") {
+			// Malformed request URL, not recoverable.
 			return false, nil
 		}
```
mrodm: Related also to the Copilot's message https://github.com/elastic/elastic-package/pull/3918/changes#r3923685512  

For `unsupported protocol scheme`: Go's stdlib doesn't export a typed error for this, so `strings.Contains` can't be avoided. However, it's safer to scope it to the inner error (urlErr.Err.Error()) rather than the full formatted string (err.Error()), which includes the URL and could cause false matches (although I think an unlikely edge case in practice) .

What about moving this to something like this ? 
```go
  var urlErr *url.Error
  if errors.As(err, &urlErr) && strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
      return false, nil
  }
```

I was thinking to try to not retry if there are other errors that are not necessary to be retried.

Copilot's message shows to do that via `strings.Contains` but maybe this could be a bit fragile.

Could we use some specific errors for that ? 

Claude suggested this:

```go
// Inner-error check for cases where http.Client wraps everything in *url.Error.
var urlErr *url.Error
if errors.As(err, &urlErr) {
    if strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
        return false, nil
    }
    var invalidAddrErr net.InvalidAddrError
    if errors.As(urlErr.Err, &invalidAddrErr) {
        return false, nil
    }
    var dnsErr *net.DNSError
    if errors.As(urlErr.Err, &dnsErr) && dnsErr.IsNotFound {
        return false, nil
    }
}
``` 

WDYT ? @kcreddy

mrodm: Related also to the Copilot's message https://github.com/elastic/elastic-package/pull/3918/changes#r3923685512

For `unsupported protocol scheme`: Go's stdlib doesn't export a typed error for this, so strings.Contains can't be avoided. However, it's safer to scope it to the inner error (`urlErr.Err.Error()`) rather than the full formatted string (`err.Error()`), which includes the URL and could cause false matches (although I think it is an unlikely edge case in practice).

What about moving this to something like this ?
```go
  var urlErr *url.Error
  if errors.As(err, &urlErr) && strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
      return false, nil
  }
```

I was thinking to try to not retry if there are other errors that are not necessary to be retried.

Copilot's message shows to do that via `strings.Contains` but maybe this could be a bit fragile.

Could we use some specific errors for that ?

Claude suggested this:

```go
// Inner-error check for cases where http.Client wraps everything in *url.Error.
var urlErr *url.Error
if errors.As(err, &urlErr) {
    // No exported type in Go's stdlib; string match on the inner error only
    // (not the full url.Error string which includes the URL).
    if strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
        return false, nil
    }

    // Permanent DNS failure (NXDOMAIN); IsNotFound distinguishes it from
    // transient DNS errors which should still be retried.
    var dnsErr *net.DNSError
    if errors.As(urlErr.Err, &dnsErr) && dnsErr.IsNotFound {
        return false, nil
    }
}
```

I think DNS errors could be valid to not be retried when testing with Serverless projects.

WDYT ? @kcreddy

mrodm: Should it be considered any other error there (wrapped  in url.Error) that should not be retried ?

mrodm (commented):
Thanks @kcreddy !

Added a comment about to try to be more specific in the `checkRetry` function if possible. Maybe some errors could be added as non-retriable.

internal/retry/http.go
kcreddy: Thanks @mrodm. 
Addressed all suggestions. Scoped the scheme match to the inner error, added `IsNotFound` for permanent DNS, and threw in `http.ErrSchemeMismatch` (server returns HTTP to HTTPS client) as another deterministic non-retryable.
[5fe7801](https://github.com/elastic/elastic-package/pull/3918/commits/5fe780196420dc5f20253e97fa37c5fc4af025bc)

internal/retry/http_transport_errors_test.go
```diff
@@ -0,0 +1,168 @@
...
+// Regression test for https://github.com/elastic/elastic-package/issues/3887.
+func TestCheckRetryTransientConnectionErrors(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+	}{
+		{"EOF (stale keep-alive close)", &url.Error{Op: "Post", URL: "https://127.0.0.1:5601/api/fleet/agent_policies", Err: io.EOF}},
+		{"unexpected EOF", &url.Error{Op: "Post", URL: "https://127.0.0.1:5601/api/fleet/package_policies", Err: io.ErrUnexpectedEOF}},
+		{"connection reset by peer", &url.Error{Op: "Get", URL: "https://127.0.0.1:5601/api/status", Err: syscall.ECONNRESET}},
+		{"connection refused", &url.Error{Op: "Get", URL: "https://127.0.0.1:5601/api/status", Err: syscall.ECONNREFUSED}},
```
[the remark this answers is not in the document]
mrodm: @kcreddy could you check this in case it is needed to use the suggestion from Copilot?

Thanks!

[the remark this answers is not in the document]
kcreddy: Agree. Updated by replacing `syscall` with `errors.New(...)` [3ed87d7](https://github.com/elastic/elastic-package/pull/3918/commits/3ed87d71097311dac265cb19dca88c2fa08b19c1)

internal/retry/http.go
```diff
@@ -87,12 +89,33 @@ func checkRetry(ctx context.Context, resp *http.Response, err error) (bool, erro
...
 
+		var urlErr *url.Error
+		if errors.As(err, &urlErr) {
+			// Go's stdlib does not export a typed error for unsupported protocol
+			// scheme, so match on the inner error string only (not the full
+			// url.Error string which embeds the URL and could produce false matches).
+			if strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
+				return false, nil
+			}
```
mrodm: As this method could be re-used in other internal packages, we could apply the fix that Copilot suggested here https://github.com/elastic/elastic-package/pull/3918/#pullrequestreview-5133850374


```suggestion
			if  urlErr.Err != nil && strings.Contains(urlErr.Err.Error(), "unsupported protocol scheme") {
				return false, nil
			}
```

Not sure how likely is to happen but it could be added as a safe guard.

kcreddy: Done. [ffa8c15](https://github.com/elastic/elastic-package/pull/3918/commits/ffa8c15c2e3898dd9d93cf4c0d881ae98e0af57b)

Actually not a big fan of Copilot reviews, it always tries to find a single-most actionable review item every time 🥲 and never seem to end

mrodm (approved):
Thanks!

Issue #3875, opened 2026-08-27
Dockerfile.terraform_deployer: google-cloud-sdk apt package removed, breaking Terraform-backed tests


teresaromero:
# Body

## Summary

The Docker image used by `elastic-package` to run Terraform deployers (`internal/servicedeployer/_static/Dockerfile.terraform_deployer`) pins `google-cloud-sdk=467.0.0-0`. Google silently dropped the `google-cloud-sdk` apt package name (renamed to `google-cloud-cli` in 2022) without a formal removal notice. Version `467.0.0-0` is no longer resolvable. Every integration that uses a Terraform-deployed test service is now broken in CI.

## Error

```
#7 [terraform 3/7] RUN ... apt-get install google-cloud-sdk=467.0.0-0 -yq ...
#7 5.046 Package google-cloud-sdk is not available, but is referred to by another package.
#7 5.046   google-cloud-cli
#7 5.052 E: Version '467.0.0-0' for 'google-cloud-sdk' was not found
failed to solve: process did not complete successfully: exit code: 100
```

## Root cause

`Dockerfile.terraform_deployer` lines 2–12:
```dockerfile
ENV GCLOUD_SDK_VERSION=467.0.0-0
...
RUN ... apt-get install google-cloud-sdk=${GCLOUD_SDK_VERSION} -yq ...
```

Google renamed the apt package from `google-cloud-sdk` to `google-cloud-cli` in March 2022 ([announcement](https://cloud.google.com/blog/products/application-development/redesigning-the-cloud-sdk-cli-for-easier-development)), promising backward compatibility *"for as long as needed"*. That promise was not kept — Google stopped publishing new versions under the old name without any formal removal notice. Community reports of `apt-get install google-cloud-sdk` failing date back to February 2024. The pinned version `467.0.0-0` worked as long as Docker's layer cache was valid; once a fresh `apt-get update` fetched the current repo state, the old name was gone.

## Proposed fix

In `internal/servicedeployer/_static/Dockerfile.terraform_deployer`:
- Change `google-cloud-sdk` → `google-cloud-cli`
- Update `GCLOUD_SDK_VERSION` to a currently available version (or drop the version pin)

## Impact

This is blocking CI across `elastic/integrations` on **main** and all open PRs that touch any of the 22+ affected packages. The breakage appears to have started around 2026-08-27.

**Affected packages** (all share `_dev/deploy/tf/` Terraform-deployed test infrastructure):
[...]

PR #3919, opened 2026-09-03
System test runner: re-attempt tests that fail during setup phase

### LINKED ISSUE #3888: System test runner: re-attempt tests that fail during the setup phase
## Summary

Follow-up to companion issue https://github.com/elastic/elastic-package/issues/3887. That issue fixes the retry policy in `internal/retry/http.go` so connection-close failures (`EOF`) on Fleet setup POSTs are retried. Two other classes of transient setup failures remain, and neither is fixable at the HTTP layer:

**Slow Kibana responses (request accepted, response too slow):**

- elastic/integrations#20617 — `GET /api/fleet/agents/<id>: context deadline exceeded` (http-get-2k-file, flow)
- elastic/integrations#20510 — `GET /api/fleet/agents/<id>: context deadline exceeded` (mongodb-inserts)

Re-sending the same GET immediately hits the same momentarily-slow Kibana; only a re-attempt after teardown, some tens of seconds later, finds a recovered stack.

**Service/agent container failures during setup:**

- elastic/integrations#20344 — `could not setup agent: service is unhealthy: container exited with code 143` (dns-tcp-axfr)
- elastic/integrations#20345 — `could not setup agent: service is unhealthy: could not inspect containers ("No such object")` (dns-udp-edns-ds)

Both from a single build — a container was SIGTERM'd during CI (exit 143 = 128+15) and had already been removed when inspected. No HTTP call failed at all.

Today a setup failure is immediately terminal: the test fails, an auto-filed flaky issue is created, and a human triages what turns out to be an infra transient. (Verified against main @ fedd1f1f: `internal/testrunner/runners/system/tester.go` has no re-attempt logic anywhere in the test flow.)

## Proposal

When a system test fails during the **setup phase** — policy creation, agent enrollment/assignment, service container startup/health — tear down that test's resources and re-attempt the **single failing test config** from scratch, once (configurable, default 1 re-attempt).

Scope constraints, deliberately strict:

1. **Setup-phase failures only.** Never re-attempt data-validation failures (document count/field mismatches). Those are real signal — a parsing bug, mapping problem, or genuine nondeterminism in the package — and auto-retrying them would silently mask defects. The setup/validation boundary in the system test runner is already clean, so this is enforceable.
2. **Per-test scope.** Only the failing test config is re-attempted; sibling tests and other data streams are untouched. Cost is ~30–60s per re-attempt, paid only when a failure occurs — negligible amortized across nightly runs. The boundary is already clean in the runner: setup lives in `prepareScenario`/`setupService` and validation in `validateTestScenario` (`internal/testrunner/runners/system/tester.go`), so "which phase failed" is directly determinable.
3. **Don't lose the signal.** A test that passes on re-attempt should be marked as flaky in the output (e.g., JUnit `<flakyFailure>` or a log annotation), so infra instability remains visible in aggregate even though it no longer fails builds or auto-files issues.

## Why both layers are worth having

| Failure class (examples above) | HTTP retry (companion issue https://github.com/elastic/elastic-package/issues/3887) | Setup re-attempt (this issue) |
|---|---|---|
| `POST .../agent_policies\|package_policies: EOF` | ✅ | ✅ (redundant backstop) |
| `GET /agents/<id>: context deadline exceeded` | ❌ | ✅ |
| Service container exited / removed during setup | ❌ | ✅ |

The HTTP retry is the surgical, near-free fix for the most common signature and should land first. This issue is the coarse safety net for everything else; together they cover all 8 flaky-test issues auto-filed against `network_traffic` in July–August 2026, and the same classes observed across other packages.

kcreddy:
```
System test runner: re-attempt tests that fail during the setup phase

When a system test fails during setup (policy creation, agent enrollment
and assignment, service container startup), the cause is usually a
transient environment issue such as a dropped connection to Kibana, a
container killed by the CI runner, or a slow Fleet response, not a
problem in the package under test. Today these failures are terminal:
the build fails, a flaky-test issue is auto-filed, and a human triages
what turns out to be infrastructure noise.

Add an opt-in setup re-attempt to the system test runner. When
prepareScenario fails with an environment error, the attempt is torn
down and the single failing test config is re-attempted from scratch,
with a fresh service run ID and config, up to the configured number of
times.

Scope, deliberately strict:

- Setup-phase environment errors only. Failures reported as
  ErrTestCaseFailed from prepareScenario (no documents within the
  wait-for-data timeout, assert.hit_count / min_count / fields_present
  not satisfied, no data streams discovered) are real test signal and
  are never re-attempted. Data validation in validateTestScenario never
  surfaces as a Go error, so it cannot trigger a re-attempt either.
- The one exception is the test service exiting with a non-zero code
  while waiting for documents. It is marked with errServiceExited so it
  is re-attempted as an environment failure, while still unwrapping to
  ErrTestCaseFailed so it keeps being reported as a test case failure
  (xUnit <failure>) with the same message as before. The
  docker_failing_test_service false-positive test guards this.
- Each attempt is fully torn down before the next one starts. No
  re-attempt happens if the context was cancelled or if the teardown of
  the failed attempt returned an error.
- Context cancellation during setup is captured in the result as on
  main, so the <error> entry is preserved and post-processing continues
  without re-attempting.
- The agent logs check after the run uses the start time of the last
  attempt, so logs from torn-down attempts are not scanned.
- The --setup, --no-provision and --tear-down workflows are unchanged.

Configuration:

- --setup-reattempts N on `elastic-package test system`, default 0
  (disabled), maximum 5.
- ELASTIC_PACKAGE_TEST_SETUP_REATTEMPTS env var as an alternative, so CI
  can enable it without changing scripts. The flag takes precedence.

Reporting:

- Tests passing only after a re-attempt are reported as a plain PASS in
  the human and JSON formats, keeping the result values standard for
  Buildkite and the flaky-test automation.
- The failures of the previous attempts are kept for aggregation in a
  new FlakyMsg field on TestResult, emitted as a <flakyFailure> element
  in xUnit and a flaky_details field in JSON.
- Tests exhausting all attempts are reported exactly as a single failed
  attempt is today.

Also adds unit tests for the re-attempt policy, the error
classification and the reporters, and a section in the system testing
HOWTO.

Closes #3888

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
```

Closes #3888

### DIFF
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -483,6 +483,7 @@ func getTestRunnerSystemCommand() *cobra.Command {
 	cmd.Flags().Bool(cobraext.TearDownFlagName, false, cobraext.TearDownFlagDescription)
 	cmd.Flags().Bool(cobraext.NoProvisionFlagName, false, cobraext.NoProvisionFlagDescription)
 	cmd.Flags().String(cobraext.AgentVersionFlagName, "", cobraext.AgentVersionFlagDescription)
+	cmd.Flags().Int(cobraext.SetupReattemptsFlagName, 1, cobraext.SetupReattemptsFlagDescription)
 
 	cmd.MarkFlagsMutuallyExclusive(cobraext.SetupFlagName, cobraext.TearDownFlagName, cobraext.NoProvisionFlagName)
 	cmd.MarkFlagsRequiredTogether(cobraext.ConfigFileFlagName, cobraext.SetupFlagName)
@@ -546,6 +547,11 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
 	deferCleanup, err := cmd.Flags().GetDuration(cobraext.DeferCleanupFlagName)
 	if err != nil {
 		return cobraext.FlagParsingError(err, cobraext.DeferCleanupFlagName)
+	}
+
+	setupReattempts, err := cmd.Flags().GetInt(cobraext.SetupReattemptsFlagName)
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.SetupReattemptsFlagName)
 	}
 
 	variantFlag, err := cmd.Flags().GetString(cobraext.VariantFlagName)
@@ -677,6 +683,7 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
 		FailOnMissingTests:     failOnMissing,
 		GenerateTestResult:     generateTestResult,
 		DeferCleanup:           deferCleanup,
+		SetupReattempts:        setupReattempts,
 		GlobalTestConfig:       globalTestConfig.System,
 		WithCoverage:           testCoverage,
 		CoverageType:           testCoverageFormat,
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -17,8 +17,8 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/elastic/elastic-package/internal/cobraext"
-	"github.com/elastic/elastic-package/internal/environment"
 	"github.com/elastic/elastic-package/internal/common"
+	"github.com/elastic/elastic-package/internal/environment"
 	"github.com/elastic/elastic-package/internal/files"
 	"github.com/elastic/elastic-package/internal/install"
 	"github.com/elastic/elastic-package/internal/logger"
--- a/cmd/testrunner.go
+++ b/cmd/testrunner.go
@@ -11,11 +11,13 @@ import (
 	"path/filepath"
 	"slices"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/elastic/elastic-package/internal/cobraext"
+	"github.com/elastic/elastic-package/internal/environment"
 	"github.com/elastic/elastic-package/internal/common"
 	"github.com/elastic/elastic-package/internal/files"
 	"github.com/elastic/elastic-package/internal/install"
@@ -483,7 +485,7 @@ func getTestRunnerSystemCommand() *cobra.Command {
 	cmd.Flags().Bool(cobraext.TearDownFlagName, false, cobraext.TearDownFlagDescription)
 	cmd.Flags().Bool(cobraext.NoProvisionFlagName, false, cobraext.NoProvisionFlagDescription)
 	cmd.Flags().String(cobraext.AgentVersionFlagName, "", cobraext.AgentVersionFlagDescription)
-	cmd.Flags().Int(cobraext.SetupReattemptsFlagName, 1, cobraext.SetupReattemptsFlagDescription)
+	cmd.Flags().Int(cobraext.SetupReattemptsFlagName, 0, cobraext.SetupReattemptsFlagDescription)
 
 	cmd.MarkFlagsMutuallyExclusive(cobraext.SetupFlagName, cobraext.TearDownFlagName, cobraext.NoProvisionFlagName)
 	cmd.MarkFlagsRequiredTogether(cobraext.ConfigFileFlagName, cobraext.SetupFlagName)
@@ -552,6 +554,19 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
 	setupReattempts, err := cmd.Flags().GetInt(cobraext.SetupReattemptsFlagName)
 	if err != nil {
 		return cobraext.FlagParsingError(err, cobraext.SetupReattemptsFlagName)
+	}
+	if !cmd.Flags().Changed(cobraext.SetupReattemptsFlagName) {
+		if v, ok := os.LookupEnv(environment.WithElasticPackagePrefix("TEST_SETUP_REATTEMPTS")); ok {
+			n, err := strconv.Atoi(v)
+			if err != nil {
+				return fmt.Errorf("invalid value for %s env var: %w", environment.WithElasticPackagePrefix("TEST_SETUP_REATTEMPTS"), err)
+			}
+			setupReattempts = n
+		}
+	}
+	const maxSetupReattempts = 5
+	if setupReattempts < 0 || setupReattempts > maxSetupReattempts {
+		return fmt.Errorf("--%s must be between 0 and %d, got %d", cobraext.SetupReattemptsFlagName, maxSetupReattempts, setupReattempts)
 	}
 
 	variantFlag, err := cmd.Flags().GetString(cobraext.VariantFlagName)
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -989,6 +989,28 @@ A sample workflow would look like:
 - Run `elastic-package stack up -d -v`
 - Navigate to the package folder in integrations and run `elastic-package test system -v`
 
+### Re-attempting tests that fail during the setup phase
+
+Failures during the setup phase of a system test (policy creation, agent
+enrollment and assignment, service container startup) are usually caused by
+transient issues in the environment, such as a dropped connection to Kibana or
+a container terminated by the CI runner, and not by problems in the package
+under test.
+
+To reduce the noise caused by these failures, when a system test fails during
+the setup phase, its resources are torn down and the failing test is
+re-attempted from scratch, once by default. The number of re-attempts can be
+adjusted with the `--setup-reattempts` flag, and re-attempts can be disabled
+with `--setup-reattempts=0`.
+
+Failures during data validation (missing documents, field or mapping problems,
+unexpected hit counts) are real test signal and are never re-attempted.
+
+A test that passes only after being re-attempted is still reported as passed,
+but it is annotated as flaky: the failures of the previous attempts are
+reported in a `<flakyFailure>` element in the xUnit report, so the instability
+of the environment remains visible in aggregate.
+
 ### Running system tests without cleanup (technical preview)
 
 By default, `elastic-package test system` command always performs these steps to run tests for a given package:
--- a/docs/howto/system_testing.md
+++ b/docs/howto/system_testing.md
@@ -999,17 +999,20 @@ under test.
 
 To reduce the noise caused by these failures, when a system test fails during
 the setup phase, its resources are torn down and the failing test is
-re-attempted from scratch, once by default. The number of re-attempts can be
-adjusted with the `--setup-reattempts` flag, and re-attempts can be disabled
-with `--setup-reattempts=0`.
+re-attempted from scratch. Setup re-attempts are disabled by default; enable
+them with `--setup-reattempts=N` (maximum 5). For example:
+
+```sh
+elastic-package test system -v --setup-reattempts 1
+```
 
 Failures during data validation (missing documents, field or mapping problems,
 unexpected hit counts) are real test signal and are never re-attempted.
 
-A test that passes only after being re-attempted is still reported as passed,
-but it is annotated as flaky: the failures of the previous attempts are
-reported in a `<flakyFailure>` element in the xUnit report, so the instability
-of the environment remains visible in aggregate.
+A test that passes only after being re-attempted is still reported as passed.
+The failures of the previous attempts are recorded in a `<flakyFailure>`
+element in the xUnit report, so the instability of the environment remains
+visible in aggregate even though the build does not fail.
 
 ### Running system tests without cleanup (technical preview)
 
--- a/internal/testrunner/reporters/formats/xunit.go
+++ b/internal/testrunner/reporters/formats/xunit.go
@@ -41,12 +41,19 @@ type testCase struct {
 	ClassName     string  `xml:"classname,attr"`
 	TimeInSeconds float64 `xml:"time,attr"`
 
-	Error   string   `xml:"error,omitempty"`
-	Failure string   `xml:"failure,omitempty"`
-	Skipped *skipped `xml:"skipped,omitempty"`
+	Error        string        `xml:"error,omitempty"`
+	Failure      string        `xml:"failure,omitempty"`
+	Skipped      *skipped      `xml:"skipped,omitempty"`
+	FlakyFailure *flakyFailure `xml:"flakyFailure,omitempty"`
 }
 
 type skipped struct {
+	Message string `xml:"message,attr"`
+}
+
+// flakyFailure reports a failure of a previous attempt of a test that
+// eventually passed, following the schema used by Maven Surefire.
+type flakyFailure struct {
 	Message string `xml:"message,attr"`
 }
 
@@ -102,6 +109,10 @@ func reportXUnitFormat(results []testrunner.TestResult) (string, error) {
 
 		if r.Skipped != nil {
 			c.Skipped = &skipped{r.Skipped.String()}
+		}
+
+		if r.FlakyMsg != "" {
+			c.FlakyFailure = &flakyFailure{Message: r.FlakyMsg}
 		}
 
 		numTests++
--- a/internal/testrunner/runners/system/reattempt_test.go
+++ b/internal/testrunner/runners/system/reattempt_test.go
@@ -1,0 +1,184 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package system
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/elastic/elastic-package/internal/testrunner"
+)
+
+// attemptOutcome describes what a single test attempt should produce in the
+// stub used by these tests.
+type attemptOutcome struct {
+	result testrunner.TestResult
+	runErr error
+	tdErr  error
+}
+
+func stubAttempts(t *testing.T, outcomes []attemptOutcome, calls *int) func() ([]testrunner.TestResult, error, error) {
+	t.Helper()
+	return func() ([]testrunner.TestResult, error, error) {
+		require.Less(t, *calls, len(outcomes), "more attempts than expected")
+		outcome := outcomes[*calls]
+		*calls++
+		return []testrunner.TestResult{outcome.result}, outcome.runErr, outcome.tdErr
+	}
+}
+
+func TestRunWithSetupReattempts(t *testing.T) {
+	setupErr := errSetupFailed{err: errors.New("service is unhealthy: container exited with code 143")}
+	setupResult := testrunner.TestResult{Name: "test", ErrorMsg: setupErr.Error()}
+	passResult := testrunner.TestResult{Name: "test"}
+	failedResult := testrunner.TestResult{Name: "test", FailureMsg: "could not find the expected hits"}
+
+	cases := []struct {
+		title           string
+		reattempts      int
+		outcomes        []attemptOutcome
+		expectedCalls   int
+		expectedErr     string
+		expectedFlaky   bool
+		expectedResults func(t *testing.T, results []testrunner.TestResult)
+	}{
+		{
+			title:         "pass on first attempt",
+			reattempts:    1,
+			outcomes:      []attemptOutcome{{result: passResult}},
+			expectedCalls: 1,
+		},
+		{
+			title:      "setup failure then pass is marked flaky",
+			reattempts: 1,
+			outcomes: []attemptOutcome{
+				{result: setupResult, runErr: setupErr},
+				{result: passResult},
+			},
+			expectedCalls: 2,
+			expectedFlaky: true,
+		},
+		{
+			title:      "setup failures until attempts are exhausted",
+			reattempts: 2,
+			outcomes: []attemptOutcome{
+				{result: setupResult, runErr: setupErr},
+				{result: setupResult, runErr: setupErr},
+				{result: setupResult, runErr: setupErr},
+			},
+			expectedCalls: 3,
+			expectedResults: func(t *testing.T, results []testrunner.TestResult) {
+				// The last failure is reported as an error entry, without
+				// aborting the run, and it is not marked as flaky.
+				require.Len(t, results, 1)
+				assert.NotEmpty(t, results[0].ErrorMsg)
+				assert.Empty(t, results[0].FlakyMsg)
+			},
+		},
+		{
+			title:      "validation failure is never re-attempted",
+			reattempts: 3,
+			outcomes: []attemptOutcome{
+				{result: failedResult},
+			},
+			expectedCalls: 1,
+		},
+		{
+			title:      "re-attempts disabled",
+			reattempts: 0,
+			outcomes: []attemptOutcome{
+				{result: setupResult, runErr: setupErr},
+			},
+			expectedCalls: 1,
+		},
+		{
+			title:      "hard errors are returned without re-attempt",
+			reattempts: 3,
+			outcomes: []attemptOutcome{
+				{result: passResult, runErr: errors.New("cannot load config")},
+			},
+			expectedCalls: 1,
+			expectedErr:   "cannot load config",
+		},
+		{
+			title:      "no re-attempt if teardown of the failed attempt failed",
+			reattempts: 3,
+			outcomes: []attemptOutcome{
+				{result: setupResult, runErr: setupErr, tdErr: errors.New("could not remove policy")},
+			},
+			expectedCalls: 1,
+			expectedErr:   "failed to tear down runner",
+		},
+		{
+			title:      "teardown failure after passing test is returned",
+			reattempts: 1,
+			outcomes: []attemptOutcome{
+				{result: passResult, tdErr: errors.New("could not remove policy")},
+			},
+			expectedCalls: 1,
+			expectedErr:   "failed to tear down runner",
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.title, func(t *testing.T) {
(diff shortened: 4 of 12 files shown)


internal/testrunner/reporters/formats/xunit.go
```diff
@@ -41,15 +41,22 @@ type testCase struct {
 	ClassName     string  `xml:"classname,attr"`
 	TimeInSeconds float64 `xml:"time,attr"`
 
-	Error   string   `xml:"error,omitempty"`
-	Failure string   `xml:"failure,omitempty"`
-	Skipped *skipped `xml:"skipped,omitempty"`
+	Error        string        `xml:"error,omitempty"`
+	Failure      string        `xml:"failure,omitempty"`
+	Skipped      *skipped      `xml:"skipped,omitempty"`
+	FlakyFailure *flakyFailure `xml:"flakyFailure,omitempty"`
```
mrodm, line 47: This new field is not added or taken into account in the json format.

Should it be added there for consistency ?

mrodm, line 47: As a note, this new field will not be used by the Buildkite pipeline (IIRC it is not an standard field from xUnit) nor by the automation in the integrations repository to create new GitHub issues with label `flaky-tests` (no information about this flakiness will be added there).

internal/testrunner/runners/system/tester.go
```diff
@@ -733,30 +740,90 @@ func (r *tester) run(ctx context.Context, stackConfig stack.Config) (results []t
...
 
-	partial, err := r.runTest(ctx, testConfig, stackConfig, svcInfo)
+		partial, err = r.runTest(ctx, testConfig, stackConfig, svcInfo)
 
-	skipDeferCleanup := len(partial) > 0 && partial[0].Skipped != nil
-	tdErr := r.tearDownTest(ctx, skipDeferCleanup)
-	if err != nil {
-		return partial, err
+		skipDeferCleanup := len(partial) > 0 && partial[0].Skipped != nil
+		return partial, err, r.tearDownTest(ctx, skipDeferCleanup)
```
mrodm, line 826: If the parameter `--defer-cleanup` is set in the CLI, it will block every re-attempt.

`skipDeferCleanup` is derived from `Skipped` status rather than from whether this is a re-attempt. Each cleanup between retry attempts waits the full `--defer-cleanup` duration (e.g. 5 minutes), making retries painfully slow in practice.

Would it be ok to wait in each re-attempt if the user sets that flag in the CLI?

```sh
elastic-package test system -v --defer-cleanup 5m
```

WDYT @teresaromero ? Keeping it as it is, it will let the developer to debug each re-attempt.

internal/testrunner/runners/system/tester.go
```diff
@@ -2120,7 +2187,16 @@ func (r *tester) runTest(ctx context.Context, config *testConfig, stackConfig st
...
+
+		// Test case failures (e.g. expected documents did not arrive) are
+		// real test signal, never re-attempted. Any other error during the
+		// preparation of the scenario is considered an environment issue and
+		// returned as a sentinel error so the caller can re-attempt the test.
+		var tcf testrunner.ErrTestCaseFailed
+		if errors.As(err, &tcf) {
+			return results, nil
+		}
+		return results, errSetupFailed{err: err}
```
mrodm: Related to Copilot's message https://github.com/elastic/elastic-package/pull/3919/changes#r3928052171

I wonder if we should manage here if there is any context error to not wrap it with `errSetupFailed`


```suggestion
		var tcf testrunner.ErrTestCaseFailed
		if errors.As(err, &tcf) {
			return results, nil
		}
		if ctx.Err() != nil {
			return results, ctx.Err()
		}
		return results, errSetupFailed{err: err}
```

Or would it be better to do it where Copilot's message points out ?

mrodm (commented):
Were you able to run a setup re-attempt locally? Maybe some way of forcing errors in the setup?
I was trying to do so, but not able to find a scenario where it happens. I'd like to try it locally if possible.

*EDIT*: It would also be interesting to add a comment `test integrations` here in this Pull Request to force testing all the packages in the integrations repository with these elastic-package changes.

internal/testrunner/runners/system/tester.go
kcreddy: Took your suggestion [a6c50f7](https://github.com/elastic/elastic-package/pull/3919/commits/a6c50f71fa46e3da7972b4177bb0ba00244371ca)

internal/testrunner/reporters/formats/xunit.go
kcreddy: Added in [a6c50f7](https://github.com/elastic/elastic-package/pull/3919/commits/a6c50f71fa46e3da7972b4177bb0ba00244371ca)

kcreddy:
test integrations

internal/testrunner/runners/system/tester.go
```diff
@@ -2120,7 +2187,19 @@ func (r *tester) runTest(ctx context.Context, config *testConfig, stackConfig st
...
 		results, _ := result.WithError(err)
-		return results, nil
+
+		// Test case failures (e.g. expected documents did not arrive) are
+		// real test signal, never re-attempted. Any other error during the
+		// preparation of the scenario is considered an environment issue and
+		// returned as a sentinel error so the caller can re-attempt the test.
+		var tcf testrunner.ErrTestCaseFailed
+		if errors.As(err, &tcf) {
+			return results, nil
```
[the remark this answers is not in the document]
kcreddy: This is valid, and worth fixing. Fixed in [dba1cb7](https://github.com/elastic/elastic-package/pull/3919/commits/dba1cb788a181be2e5e30ae43bb9fec6ad625a7e)

cmd/testrunner.go
```diff
@@ -483,6 +483,7 @@ func getTestRunnerSystemCommand() *cobra.Command {
 	cmd.Flags().Bool(cobraext.TearDownFlagName, false, cobraext.TearDownFlagDescription)
 	cmd.Flags().Bool(cobraext.NoProvisionFlagName, false, cobraext.NoProvisionFlagDescription)
 	cmd.Flags().String(cobraext.AgentVersionFlagName, "", cobraext.AgentVersionFlagDescription)
+	cmd.Flags().Int(cobraext.SetupReattemptsFlagName, 1, cobraext.SetupReattemptsFlagDescription)
```
teresaromero:
```suggestion
	cmd.Flags().Int(cobraext.SetupReattemptsFlagName, 0, cobraext.SetupReattemptsFlagDescription)
```

Can we change the default to 0 and be explicit on this setting where ever we are using the command?

cmd/testrunner.go
```diff
@@ -548,6 +549,11 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
 		return cobraext.FlagParsingError(err, cobraext.DeferCleanupFlagName)
 	}
 
+	setupReattempts, err := cmd.Flags().GetInt(cobraext.SetupReattemptsFlagName)
```
teresaromero, line 554: is there a max limit for reattempts?

teresaromero (commented):
I went through the code and left some comments. I am concern about having this activated by default; suggested to make it opt-in so we can be explicit about it. 

Related to the formats/outputs; based on the comment of Mario https://github.com/elastic/elastic-package/pull/3919/changes#r3959485113 ; is it necesary to show it as a different "success" ? if the goal is to retry on setup (because is flaly) i think the output should be FAIL/PASS after all , regardless the setup retries it got. this way we keep output standard for buildkite. wdyt?

cmd/testrunner.go
mrodm: If it is set to zero the default value to not change the current behavior, it will be good to support setting an environment variable too for this. That would help to update the CI in the integrations and elastic-package repository without having to update the CI scripts (adding new parameters). However this parameter is not global for the `test` command and it will be needed to split the commands in scripts.

So, I'd prefer if we do so, reading also an environment variable to allow setting the same value as the new parameter.

internal/testrunner/runners/system/tester.go
```diff
@@ -2120,7 +2187,18 @@ func (r *tester) runTest(ctx context.Context, config *testConfig, stackConfig st
...
 		results, _ := result.WithError(err)
-		return results, nil
+
+		// All errors from prepareScenario are environment issues and should be
+		// re-attempted. Note: ErrTestCaseFailed can reach here from
+		// verifyDataStream (service exited non-zero) or waitForDocs (no hits
+		// within the timeout), both of which are setup-phase events.
+		// validateTestScenario failures are never returned as Go errors — they
+		// are captured in result.FailureMsg via result.WithError — so there is
+		// no validation signal to guard against here.
```
mrodm: I'm not sure if this latest change would report the errors like `service exited non-zero` or `waitForDocs ` with no hits as test cases failed.

They should appear in the xUnit file as failures or errors. I'm not sure how they are reported, but they are.

Could you check that this is still working in this way ?

internal/testrunner/runners/system/tester.go
```diff
@@ -733,30 +740,90 @@ func (r *tester) run(ctx context.Context, stackConfig stack.Config) (results []t
...
+
+func (e errSetupFailed) Error() string {
+	return fmt.Sprintf("test setup failed: %v", e.err)
+}
+
+func (e errSetupFailed) Unwrap() error {
+	return e.err
+}
+
 func (r *tester) runTestPerVariant(ctx context.Context, stackConfig stack.Config, result *testrunner.ResultComposer, cfgFile, variantName string) ([]testrunner.TestResult, error) {
```
mrodm: Related to this comment https://github.com/elastic/elastic-package/pull/3919/#discussion_r3976869124, the `startingTime` is obtained before this function. 

That time is used to check the agent logs looking for any error or warning. If that startingTime is obtained before the retries, it could be checking other containres.

And this is also related to the other comment posted by Copilot https://github.com/elastic/elastic-package/pull/3919/#discussion_r3976869184, it would be needed to assess if this is true. If there is a re-attempt, elastic-package needs to ensure that the scenario (kibana, elasticsearch) is not dirty. Previous Elastic Agents should be unenrolled and removed, policies should be deleted, data streams should be deleted too, and so on.

internal/testrunner/runners/system/tester.go
```diff
@@ -733,30 +740,90 @@ func (r *tester) run(ctx context.Context, stackConfig stack.Config) (results []t
...
+	maxAttempts := 1 + reattempts
+	var setupFailures []string
+	for attemptNum := 1; ; attemptNum++ {
+		partial, err, tdErr := attempt()
+
+		var setupErr errSetupFailed
+		if errors.As(err, &setupErr) {
+			// Setup failures are already reported as part of the test results,
+			// they are not returned as hard errors to the caller.
+			err = nil
```
mrodm, line 846: If this error is overwritten with `nil`, and all the attempts finish with error. 

The condition in line 809, will not be true. I'm thinking for instance in those cases that the service cannot start due to some error. We want to report that as before in the xUnit so it gets shown in Buildkite and in Github issues.

Related also to this other comment https://github.com/elastic/elastic-package/pull/3919/changes#r3980680995

mrodm: It could be tested forcing an error in some Dockerfile in any of the test packages, or any other error that could lead to a failure here.

Then it should be reported as part of the xUnit (as it is done currently).

IIRC this is the command to create the xUnit files in system tests:
```
elastic-package test system -v --report-output file --report-format xUnit
```

internal/testrunner/runners/system/tester.go
mrodm: Moreover commit dba1cb7 removed the ErrTestCaseFailed guard, so waitForDocs failures, including    `assert.hit_count`, `assert.min_count` and `assert.fields_present`, are re-attempted. A test whose input genuinely produces no data now waits the 10 minute default timeout twice, plus a second setup. 

The PR description, the docs section, and the test comment at reattempt_test.go:86 all still claim these are never re-attempted. Could it be restored the guard for the `waitForDocs` error? In any case, it would be needed to update the description and docs independently of the solution chosen.

As it would happen in main, those `waitForDocs` errors should also be reported in the xUnit file.
It could be tested by changing some min_count or hit_count value in the test config of one package to force the failure.

internal/testrunner/runners/system/tester.go
```diff
@@ -2120,7 +2187,18 @@ func (r *tester) runTest(ctx context.Context, config *testConfig, stackConfig st
...
+
+		// All errors from prepareScenario are environment issues and should be
+		// re-attempted. Note: ErrTestCaseFailed can reach here from
+		// verifyDataStream (service exited non-zero) or waitForDocs (no hits
+		// within the timeout), both of which are setup-phase events.
+		// validateTestScenario failures are never returned as Go errors — they
+		// are captured in result.FailureMsg via result.WithError — so there is
+		// no validation signal to guard against here.
+		if ctx.Err() != nil {
+			return results, ctx.Err()
```
mrodm: IIRC on main a context cancellation during setup produced an `<error>` entry. Now `runTest` returns the context error as a hard error, and `testrunner.run` discards that tester's results


From main:

```
		// report all other errors as error entries in the xUnit file
		results, _ := result.WithError(err)
		return results, nil
```

It would be needed to validate if those errors are still reported in the xUnit file as `<error>` too.

mrodm (commented):
Added comments, my main concern is that these new re-attempts should not change the current behaviour for reporting failures and errors. And following that, it needs to be ensured that the cleaning (tearDownTest) runs between attempts and at the end of each test. Every re-attempt needs to be run after all the tearDown process is performed. 

**EDIT**: If new changes are pushed, it would be interesting to run the tests with all the packages integrations via posting the `test integrations` comment. Not for all commits, but we would need to validate from time to time that this is not breaking anything and that it is working as expected.

internal/testrunner/runners/system/tester.go
kcreddy: Thanks for the thorough look. This downstream result-loss risk are pre-existing behaviour in upstream/main and not introduced by this PR.

On main, `runTest` already returns `(results, nil)` for all setup errors (line return results, nil in the original prepareScenario error handler). As a result, `run()` already continues into the stack dump.

Our re-attempt loop reproduces that same behaviour: when all attempts are exhausted, `errSetupFailed` is cleared and we return `(partial, nil)`, so `run()` sees the same nil-error path it would have seen on main for a single failed attempt. Nothing is worse.

Agreed that fixing the result-loss in `run()` would be a good improvement, but it's a separate pre-existing issue. Happy to track it as a follow-up if that's useful.

internal/testrunner/reporters/formats/xunit.go
kcreddy: Revisited after Teresa's request to keep result as a plain PASS. JSON now carries the previous-attempt failures in a separate `flaky_details` field (omitted when empty).

internal/testrunner/runners/system/tester.go
kcreddy: Fixed in [45e8024](https://github.com/elastic/elastic-package/pull/3919/commits/45e802475ae9de4f2f2ff2d67c57080760f9d431)

internal/testrunner/runners/system/tester.go
kcreddy: Reverted this because propagating `ctx.Err()` as hard error made runner discard the results

internal/testrunner/runners/system/tester.go
kcreddy: Confirmed and fixed. `runTest` now handles the cancellation same way as main does. Also updated thread https://github.com/elastic/elastic-package/pull/3919/changes#r3989048115

cmd/testrunner.go
kcreddy: Done.

cmd/testrunner.go
kcreddy: Updated default as `0` and added env variable `TEST_SETUP_REATTEMPTS` that can be set inside integration pipeline to make re-attempts.

internal/testrunner/runners/system/tester.go
kcreddy: Restored the ErrTestCaseFailed guard, so waitForDocs failures (no hits, hit_count, min_count, fields_present, and undiscovered data streams) are not re-attempted and still surface as <failure> via FailureMsg, same as main.

kcreddy: For the service-exit case, the first version turned it into a plain error, which made it retryable but also changed its xUnit report from `<failure>` to `<error>`. The `docker_failing_test_service` false-positive test caught that. 

In [2d1eb0c](https://github.com/elastic/elastic-package/pull/3919/commits/2d1eb0c180620629e16558f5da1097f0b981a685), it is now wrapped in `errServiceExited`, which unwraps to `ErrTestCaseFailed`, so it is re-attempted but still reported as `<failure>` with the exact same message. `TestServiceExitedReportedAsFailure` guards this now.

Ran both locally against a 9.4.4 stack with `--report-format xUnit --report-output file`:
- docker_failing_test_service with defaults: one `<failure>`, both CI assertions pass.
- Same package with `--setup-reattempts 1`: attempt 1 fails, teardown runs, attempt 2 starts in a fresh namespace, fails again, single <failure> reported, no leftover containers.

internal/testrunner/runners/system/reattempt_test.go
```diff
@@ -0,0 +1,268 @@
...
+			expectedFlaky: true,
+		},
+		{
+			title:      "re-attempts disabled",
+			reattempts: 0,
+			outcomes: []attemptOutcome{
+				{result: setupResult, runErr: setupErr},
+			},
+			expectedCalls: 1,
+		},
```
mrodm, line 155: If there is an error with reattemps disabled, should there be an error ? 

Should it be added `expectedResults` field to check about `ErrorMsg` ? 
```golang
    expectedResults: func(t *testing.T, results []testrunner.TestResult) {
        // With re-attempts disabled a setup failure is still reported as
        // an error entry (same as the exhausted-attempts case), and is
        // never marked as flaky.
        require.Len(t, results, 1)
        assert.NotEmpty(t, results[0].ErrorMsg)
        assert.Empty(t, results[0].FlakyMsg)
    },
```

cmd/testrunner.go
```diff
@@ -548,6 +551,24 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return cobraext.FlagParsingError(err, cobraext.SetupReattemptsFlagName)
+	}
+	if !cmd.Flags().Changed(cobraext.SetupReattemptsFlagName) {
+		if v, ok := os.LookupEnv(environment.WithElasticPackagePrefix("TEST_SETUP_REATTEMPTS")); ok {
```
mrodm: Following how this is built in other parts of the codebase, the definition of the string for the environment variable should be placed at package level 

For instance here:
https://github.com/elastic/elastic-package/blob/8a4ba7438e77eb179a0c3122a0820ef78be2097e/internal/docs/links_map.go#L20

It could be something like:
```golang
var systemTestSetupReattemptsEnvVar = environment.WithElasticPackagePrefix("TEST_SETUP_REATTEMPTS")
```

cmd/testrunner.go
```diff
@@ -548,6 +551,24 @@ func testRunnerSystemCommandAction(cmd *cobra.Command, args []string) error {
...
+			if err != nil {
+				return fmt.Errorf("invalid value for %s env var: %w", environment.WithElasticPackagePrefix("TEST_SETUP_REATTEMPTS"), err)
+			}
+			setupReattempts = n
+		}
+	}
+	const maxSetupReattempts = 5
+	if setupReattempts < 0 || setupReattempts > maxSetupReattempts {
+		return fmt.Errorf("--%s must be between 0 and %d, got %d", cobraext.SetupReattemptsFlagName, maxSetupReattempts, setupReattempts)
+	}
```
mrodm, line 577: Should it be added here in this message that the error could come from the environment variable too ? Right now it just mention the CLI flag `--setup-reattemps`

docs/howto/system_testing.md
```diff
@@ -989,6 +989,31 @@ A sample workflow would look like:
...
+elastic-package test system -v --setup-reattempts 1
+```
+
+Failures during data validation (missing documents, field or mapping problems,
+unexpected hit counts) are real test signal and are never re-attempted.
+
+A test that passes only after being re-attempted is still reported as passed.
+The failures of the previous attempts are recorded in a `<flakyFailure>`
+element in the xUnit report, so the instability of the environment remains
+visible in aggregate even though the build does not fail.
```
mrodm: It is also reported in the JSON.

docs/howto/system_testing.md
```diff
@@ -989,6 +989,31 @@ A sample workflow would look like:
 - Run `elastic-package stack up -d -v`
 - Navigate to the package folder in integrations and run `elastic-package test system -v`
 
+### Re-attempting tests that fail during the setup phase
```
mrodm, line 992: These docs do not included information about the new environment variable added `ELASTIC_PACKAGE_TEST_SETUP_REATTEMPTS` to set the reattempts.

internal/testrunner/runners/system/reattempt_test.go
```diff
@@ -0,0 +1,268 @@
...
+		{
+			// validateTestScenario failures surface as FailureMsg in the result
+			// (nil runErr), not as errSetupFailed, so they are never re-attempted.
+			title:      "validateTestScenario failure (FailureMsg set, nil error) is never re-attempted",
+			reattempts: 3,
+			outcomes: []attemptOutcome{
+				{result: failedResult},
+			},
+			expectedCalls: 1,
+		},
```
mrodm, line 117: What about adding also the `expectedResults` to validate that there are failures or errors ?

```go
expectedResults: func(t *testing.T, results []testrunner.TestResult) {
    require.Len(t, results, 1)
    assert.Equal(t, "could not find the expected hits", results[0].FailureMsg)
    assert.Empty(t, results[0].ErrorMsg)
    assert.Empty(t, results[0].FlakyMsg)
},
```

internal/testrunner/runners/system/reattempt_test.go
```diff
@@ -0,0 +1,268 @@
...
+	failedResult := testrunner.TestResult{Name: "test", FailureMsg: "could not find the expected hits"}
+
+	cases := []struct {
+		title           string
+		reattempts      int
+		outcomes        []attemptOutcome
+		expectedCalls   int
+		expectedErr     string
+		expectedFlaky   bool
+		expectedResults func(t *testing.T, results []testrunner.TestResult)
```
mrodm, line 75: Maybe it is interesting to add the `expectedResults` field to all the cases, so it is explicit what results must be in each one.

WDYT ?

mrodm (commented):
Thanks @kcreddy for the changes!
I left some inline comments below. I'd also like to spend some time testing this locally before approving.

internal/testrunner/runners/system/reattempt_test.go
kcreddy: Agree. Applied the suggestion in [3a968b7](https://github.com/elastic/elastic-package/pull/3919/commits/3a968b7d1010b5c5692f6d23990662753e988474)

internal/testrunner/runners/system/reattempt_test.go
kcreddy: SGTM.

kcreddy:
> Thanks @kcreddy for the changes!
I left some inline comments below. I'd also like to spend some time testing this locally before approving.

@mrodm, No worries, I applied your latest suggestions. Feel free to test it out and let me know any feedback.

kcreddy:
https://github.com/elastic/elastic-package/pull/3919#issuecomment-5729784537 comment is stale. The CI is now green 🟢 after a retry. The earlier failure was an infra issue:
```xml
Error: could not retrieve package; API status code = 502; response body = <html>
  | <head><title>502 Bad Gateway</title></head>
  | <body>
  | <center><h1>502 Bad Gateway</h1></center>
  | <hr><center>nginx/1.29.7</center>
  | </body>
  | </html>
```

teresaromero (approved):
deferring final review to Mario.

From my pov looks good, only risk i see is the re-atempts can add more time to testing. Lets see how this works and if its adding the value we expect. System tests are critical.

mrodm:
Hi @kcreddy,

Sorry for the last response, I've been testing this branch locally, and I have some doubts about possible risks of being delayed Buildkite builds. Thinking mainly for the builds running in the integrations CI repository.

As a note, I've not being able to test locally issues with policy assignment or agent enrollment. Were you able to reproduce those doing some changes locally ?

This branch reports as errors issues in service containers start up process, or issues for the `waitForDocs`.

Thinking in some of the issues that have happened lately, I wonder if those should be avoided to be retried:
-  Docker image not found:
    - https://github.com/elastic/integrations/issues/21204
    - This would repeat the full setup and it will finish with the same error
- Issues with terraform container (missing package in apt):
    - https://github.com/elastic/elastic-package/issues/3875  
    - This would repeat the full setup and it will finish with the same error
- Some errors in the system test configuration related to assert conditions ?
    - `waitForDocs` has a maximum time of 10 minutes to wait for docs.
    - if the conditions are not met and finish with error that test, it will repeat again the full setup, is that right?
        - for instance (edge case), the service generates just 10 documents but there is a `assert.min_count` of 20 documents in the configuration.
        - in that edge-case, the CI step would wait 20 minutes to get the same error.

Could those errors be skipped in the reattempt process and be considered "final" errors? Does it make sense?
If possible, it would let other errors as policy creation, agent enrollment and assignment (as mentioned in the PR description) still be considered in this process. 

WDYT @kcreddy @teresaromero ? I'm trying to avoid some scenarios where the CI could last more time to get the same output (thinking about errors).

PR #3922, opened 2026-09-04
[CI] Reorder release step to run in parallel with tests

mrodm:
## Summary

The Buildkite release step was positioned at the end of the pipeline, after all test steps, and relied on `wait: continue_on_failure: true` barriers to ensure it would always be reached even when tests failed.

This PR moves the release step to the beginning of the pipeline so it runs in parallel with the test steps, making it independent of their outcome by design rather than by barrier configuration. As a side effect, running the release step in parallel with the tests reduces the overall build time.

## Proposed commit

```
[CI] Reorder release step

Previously, the release step was running at the end of the pipeline after
all test steps finished (with failure or success), relying on
`continue_on_failure: true` barriers to always reach it. This change
moves the release step to the beginning so it runs in parallel with
the test steps, independently of their outcome.
```

---

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -39,7 +39,7 @@ steps:
     agents:
       provider: "gcp"
     plugins:
-      # required to create Github releases
+      # required to create GitHub releases
       - elastic/vault-github-token#v0.1.0:
 
   - group: ":go: Run Unit tests"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -31,6 +31,17 @@ steps:
       cpu: "8"
       memory: "4G"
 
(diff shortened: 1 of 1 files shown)


teresaromero (commented):
why are we allowing a release if tests not succeed?

mrodm:
> why are we allowing a release if tests not succeed?

@teresaromero This was the previous way of working. I think as all the changes are merged via Pull Requests, that would mean that the CI (tests) have been running successfully.

Previously, all the tests were working but even if they fail, the release step is performed. And moreover, IIUC in golang just pushing a tag is considered a release, so not sure if not running the release step (that "just" generates the GitHub release) would make any difference.

teresaromero (approved):

PR #3923, opened 2026-09-04
lint(gocritic): enable exitAfterDefer check

mrodm:
## Summary

Enable the \`exitAfterDefer\` gocritic check, which detects \`defer\` calls that will never execute because the enclosing function exits via \`log.Fatalf\` (which calls \`os.Exit\` internally). This is part of the gradual tightening of the golangci-lint configuration tracked in #3443.

## Changes

- \`.golangci.yml\`: add \`exitAfterDefer\` to the enabled gocritic checks.
- \`internal/registry/gpgkey/fetch/main.go\`: refactor \`fetchKey\` and \`fingerprint\`
  to return errors instead of calling \`log.Fatalf\`, restoring idiomatic \`defer\`
  for response body cleanup.
- \`tools/readme/main.go\`: refactor \`generateReadme\` to return an error instead of
  calling \`log.Fatalf\`, using a named return with a custom defer to propagate
  \`Close\` errors (with path context) on the success path.

## Proposed commit

\`\`\`
lint(gocritic): enable exitAfterDefer check

Enable the exitAfterDefer gocritic check, which detects deferred calls
that will never execute because the enclosing function exits via
log.Fatalf (which calls os.Exit internally).

Fix the two flagged sites by refactoring fetchKey, fingerprint, and
generateReadme to return errors instead of calling log.Fatalf, restoring
idiomatic defer usage for resource cleanup.
\`\`\`

## Related issues

- Part of #3443 (Gradually stricter configuration for golangci-lint)

---

### DIFF
--- a/tools/readme/main.go
+++ b/tools/readme/main.go
@@ -99,8 +99,10 @@ func generateReadme(readmeTmpl *template.Template, cmdsDoc string) {
 	// process, so a deferred close would never run.
 	r := readmeVars{cmdsDoc}
 	if err := readmeTmpl.Execute(readme, r); err != nil {
-		readme.Close()
+		_ = readme.Close()
 		log.Fatalf("Writing README file %s failed: %v", readmePath, err)
 	}
-	readme.Close()
+	if err := readme.Close(); err != nil {
+		log.Fatalf("Closing README file %s failed: %v", readmePath, err)
+	}
 }
--- a/tools/readme/main.go
+++ b/tools/readme/main.go
@@ -95,8 +95,8 @@ func generateReadme(readmeTmpl *template.Template, cmdsDoc string) {
 	if err != nil {
 		log.Fatalf("Opening README file %s failed: %v", readmePath, err)
 	}
-	// Close readme explicitly instead of using defer: log.Fatalf exits the
-	// process, so a deferred close would never run.
+	// Close readme explicitly (no defer) because this function may call log.Fatalf,
+	// which exits the process and skips deferred calls.
 	r := readmeVars{cmdsDoc}
 	if err := readmeTmpl.Execute(readme, r); err != nil {
 		_ = readme.Close()
--- a/tools/readme/main.go
+++ b/tools/readme/main.go
@@ -95,10 +95,12 @@ func generateReadme(readmeTmpl *template.Template, cmdsDoc string) {
 	if err != nil {
 		log.Fatalf("Opening README file %s failed: %v", readmePath, err)
 	}
-	defer readme.Close()
-
+	// Close readme explicitly instead of using defer: log.Fatalf exits the
+	// process, so a deferred close would never run.
 	r := readmeVars{cmdsDoc}
 	if err := readmeTmpl.Execute(readme, r); err != nil {
+		readme.Close()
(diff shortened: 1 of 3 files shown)


teresaromero (approved):

efd6 (commented):
In the cases here all the close calls are releasing resources that do not need to be synced. The process termination will release those anyway.

mrodm:
> In the cases here all the close calls are releasing resources that do not need to be synced. The process termination will release those anyway.

@teresaromero @efd6 please, review again this PR. I did some changes to return errors instead of all the `log.Fatalf` calls. Now `log.Fatalf` is called once in the caller function with the error returned by the given function.

tools/readme/main.go
```diff
@@ -85,20 +87,25 @@ func loadReadmeTemplate() *template.Template {
...
 
-	readme, err := os.OpenFile(readmePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	readme, err := os.OpenFile(readmePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o644)
 	if err != nil {
-		log.Fatalf("Opening README file %s failed: %v", readmePath, err)
+		return fmt.Errorf("opening README file %s: %w", readmePath, err)
 	}
-	defer readme.Close()
+	defer func() {
+		if cerr := readme.Close(); cerr != nil && err == nil {
```
teresaromero: is this err here effective of being not nil? the err on line 107 is local scoped isnt it? that will trigger the defer but does it make any difference?

doubting here why we need to replace err with cerr and not just defer readme.Close

mrodm: Good catch on the shadowing. Changed `err :=` to `err =` so the named return variable is assigned directly in https://github.com/elastic/elastic-package/pull/3923/commits/5f29430774528ee1542da57ff2c320432a0a8f81, which makes the intent clearer and removes the scoping ambiguity. As it was previously, IIRC the `err` in the deferred function is the actual return variable, but I'd prefer to keep it without shadowing it so it is clear what value contains `err`.

To answer your other question: a plain `defer readme.Close()` would silently drop the close error. The named return + custom defer is what allows the close error to propagate to the caller.

teresaromero (approved):

tools/readme/main.go
efd6: Suggest something instead like
```
defer func() {
	serr := readme.Sync()
	if serr != nil {
		serr = fmt.Errorf("syncing README file %s: %w", readmePath, serr)
	}
	cerr := readme.Close()
	if cerr != nil {
		cerr = fmt.Errorf("closing README file %s: %w", readmePath, cerr)
	}
	err = errors.Join(err, serr, cerr)
}()
```
though the %w are likely not superfluous since the final error will be consumed by a function that has no conception of wrapped errors.

mrodm: I would omit for now the Sync call,but I'll keep your approach of using `errors.Join` instead. That will help to show the user both errors, thanks!

Pushed changed in https://github.com/elastic/elastic-package/pull/3923/commits/c523fab41eae44ebbfe3ea7ab4b18a9cd97d6e85

efd6 (approved):

